PR #16, opened 2025-04-11
[Xqci] Add R_RISCV_QC_* Relocation Tests

lenary:
These weren't added before because upstream LLVM didn't have a complete enough assembly implementation for Xqci. LLVM's implementation is now complete, so these tests can be landed.


quic-seaswara:
Can you rebase your commit with tip ?

quic-seaswara (approved):

PR #17, opened 2025-04-11
Change sort-section to permit space-delim

androm3da:
This clang command `clang -std=c99 -nostdinc -ffreestanding ... -Wl,--sort-section,alignment` from musl (https://github.com/quic/musl/blob/721ca8b60ec4180ff767c317f8869a5eca13fd92/configure#L566) fails like so:

	Fatal: cannot read file alignment
	Fatal: Linking had errors.

PR #19, opened 2025-04-11
Add delimiter to "cannot read" diagnostic

androm3da:
While working on #17, I got the error below:

	Fatal: cannot read file alignment

This error message suggests that there's a problem reading the alignment of something in a file.  But in fact, it's indicating that it couldn't open a file called "alignment."  This error message would be clearer if the delimiters indicated where the filename started/stopped.

### DIFF
--- a/include/eld/Diagnostics/DiagReaders.inc
+++ b/include/eld/Diagnostics/DiagReaders.inc
@@ -11,8 +11,8 @@
 //
 //===----------------------------------------------------------------------===//
 
-DIAG(fatal_cannot_read_input, DiagnosticEngine::Fatal, "cannot read file %0")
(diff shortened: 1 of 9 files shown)


quic-seaswara (commented):
LGTM with a test.

Could you please add a test here :- eld/test/Common/standalone/CommandLine

androm3da:
> LGTM with a test.
> 
> Could you please add a test here :- eld/test/Common/standalone/CommandLine

Will do

partaror (approved):
Looks good!

quic-seaswara:
There are some more tests failing, can you please look into and update the PR ?

quic-seaswara (approved):

PR #40, opened 2025-04-16
update README

quic-seaswara:
There were two things missing in the README file.

User need to enable projects llvm and clang, using system clang causes test failures.

Also include LLVM_ENABLE_ASSERTIONS to be turned OFF. A test fails when asserts are turned ON and needs to be investigated.

Change-Id: I412ccefa01cbe9d39b31e19622cc58f86a376c5a


apazos:
Thanks @quic-seaswara, that solves the issues we ran into locally.

lsahn-gh:
> User need to enable projects llvm and clang, using system clang causes test failures.

Interesting!

> Also include LLVM_ENABLE_ASSERTIONS to be turned OFF. A test fails when asserts are turned ON and needs to be investigated.

Thank you for the update

BR

quic-seaswara:
system clang may be older than what tests need. that is why tests may fail.

Issue #42, opened 2025-04-16
eld version output goes to stderr


quic-seaswara:
linker version output gets dumped to stderr, it needs to go to stdout.

**Reproducer**

`ld.eld --version > version.txt`

version.txt is empty and the version dump goes to stderr.

PR #43, opened 2025-04-16
eld version output should be emitted to stdout (#42)

quic-seaswara:
eld version was emitted to stderr, and zephyr builds assume that version output is emitted to stdout.

This is similar to how other linkers emit version output

Change-Id: I99397ede5087e2f3176fb25c8011cf6bc7f949dc


jonathonpenix:
Quick drive-by comment, we can easily handle this either way where Zephyr is concerned, so this isn't critical one way or another--I just was curious if there was a strong preference here given ld/lld's behavior. It also seems ex: GNU tools in general are somewhat opinionated here on using stdout (see ex: https://www.gnu.org/prep/standards/standards.html#g_t_002d_002dversion).

quic-seaswara:
Let us match ld/lld behavior to be consistent with other linkers.

partaror (approved):
Looks good!

PR #44, opened 2025-04-16
cmake - configure eld to create symlinks

quic-seaswara:
configure eld to create symlinks based on the flag ELD_CREATE_SYMLINKS.

Once the driver ld.eld is able to parse emulation options and pass the command line options to the right driver, we dont need target symlinks

Change-Id: Idf8e6ddca1cad1cdd56f9b69160bab81dd4bd6a3


partaror (changes requested):
Should we install ELD target symlinks in bin/ELD when `ELD_CREATE_SYMLINKS` is `Off`? If ELD does not have any symlinks anywhere, then it might be difficult for customers to use it even if they want to.

partaror (approved):

quic-seaswara:
Users should use -fuse-ld=eld

PR #46, opened 2025-04-17
include <windows.h> for case-sensitive targets

androm3da:
On a Windows system, the windows.h header file will normally appear to be available as Windows.h because its filesystems are case-insensitive. But when building on a linux host, for example, the header files from the cross-toolchain will be found instead as `windows.h`.


androm3da:
See also https://stackoverflow.com/a/15466951/489590

partaror (approved):
Looks good!

PR #49, opened 2025-04-18
[risc-v] Zero-page (offset from 0) relaxation

quic-akaryaki:
Absolute relocations accessing +-2KB around the zero address can use the hard-wired zero register as the base address instead of a value loaded with LUI. The LUI instruction can be deleted.

Zero-page relaxations are not enabled for exact zeroes because not yet assigned addresses may temporarily have the value of zero, which may incorrectly relax them.

### DIFF
--- a/lib/Target/RISCV/RISCVLDBackend.cpp
+++ b/lib/Target/RISCV/RISCVLDBackend.cpp
@@ -303,7 +303,7 @@ bool RISCVLDBackend::doRelaxationLui(Relocation *reloc, Relocator::DWord G) {
 
   /* Three types of relaxation can be applied here, in order of preference:
    * -- zero-page: LUI is deleted and the other instruction is converted to
-   *    x0-relative [ not implemented ];
+   *    x0-relative;
    * -- GP-relative, same as above but relative to GP, not available for PIC;
    * -- compressed LUI. */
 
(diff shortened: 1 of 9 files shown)


quic-seaswara (approved):

partaror (approved):
Looks good!

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -319,23 +319,35 @@ bool RISCVLDBackend::doRelaxationLui(Relocation *reloc, Relocator::DWord G) {
...
+                      llvm::isInt<12>(Value) && Value != 0;
+
   // HI will be deleted, LO will be converted to use GP as base.
   // GP must be available and relocation must fit in 12 bits relative to GP.
   // There is no GP for shared objects.
   bool canRelaxToGP =
       config().options().getRISCVRelax() &&
       config().options().getRISCVGPRelax() && !config().isCodeIndep() &&
-      G != 0 && fitsInGP(G, S + A, frag, reloc->targetSection(), SymbolSize);
+      G != 0 && fitsInGP(G, Value, frag, reloc->targetSection(), SymbolSize);
```
lenary: Aside: I think this needs a `Value != 0`, in the case where `GP` is within range of zero, and something doesn't yet have an address.

Maybe we should just exit early from the whole function if `Value == 0`, on the basis that we cannot be sure it has its final address?

quic-akaryaki: Good point. I added a check and a test. Can't exit early because we report "missed relaxation opportunities". Also should check S not Value.

lenary (approved):

quic-seaswara (approved):

lenary:
Test failures should be fixed by #90

quic-seaswara (approved):

lenary (approved):

quic-seaswara (approved):

PR #53, opened 2025-04-21
[LinkerScript] Add support for ALIGN_WITH_INPUT

### LINKED ISSUE #52: add support for parsing ALIGN_WITH_INPUT
There are several instances of ALIGN_WITH_INPUT usage which is seen in linker scripts that use bfd linker as the default.

Though this behavior is default with `eld` add support so that users dont need to modify linker scripts.

This allows easy transition to eld when used with GNU style linker scripts.

An example that has come across is when building `picolibc`

If the behavior does not match expectations, please reopen this bug with a test, will be glad to take a look.

quic-seaswara:
You can enforce that the difference between the VMA and LMA remains intact throughout this output section with the ALIGN_WITH_INPUT attribute.

This behavior is default and so there is nothing to do here.

Added a test that the behavior is consistent with expectations.

Also updated documentation.

Closes #52

Change-Id: Ic5c1e57542a9064b595fcffb63c781e2b71a4a4f

### DIFF
--- a/docs/userguide/documentation/linker_script.rst
+++ b/docs/userguide/documentation/linker_script.rst
@@ -307,3 +307,67 @@ NOCROSSREFS
      * A linker script can contain multiple NOCROSSREFS commands.
 
      * Each command is treated as an independent set of output sections that are checked for cross references.
+
+Output Section Description
+==========================
+
+A ``SECTIONS`` command can contain one or more output section descriptions.
+
+.. code-block:: plaintext
+
+    <section-name> [<virtual_addr>][(<type>)] :
+    [AT(<load_addr>)] [ALIGN(<section_align>) | ALIGN_WITH_INPUT]
+    [SUBALIGN(<subsection_align>)] [<constraint>]
+    {
+       ...
+       <output-section-command> <output-section-command>
+    }[><region>][AT><lma_region>][:<phdr>...][
+    =<fillexp>]
+
+Syntax
+------
+
+<section-name>
+    Specifies the name of the output section.
+
+<virtual_addr>
+    Specifies the virtual address of the output section (optional). The address value can be an expression (see Expressions).
+
+<type>
+    Specifies the section load property (optional).
+
+    - NOLOAD: Marks a section as not loadable.
+    - INFO: Parsed only; has no effect on linking.
+
+<load_addr>
+    Specifies the load address of the output section (optional). The address value can be specified as an expression (see Expressions).
+
+<section_align>
+    Specifies the section alignment of the output section (optional). The alignment value can be an expression (see Expressions).
+
+<subsection_align>
+    Specifies the subsection alignment of the output section (optional). The alignment value can be an expression (see Expressions).
+
+<constraint>
+    Specifies the access type of the input sections (optional).
+
+    - NOLOAD: All input sections are read-only.
+
+<output-section-command>
+    Specifies an output section command (see Output section commands). An output section description contains one or more output section commands.
+
+<region>
+    Specifies the region of the output section (optional). The region is expressed as a string. This option is parsed but has no effect on linking.
+
+<lma-region>
+    Specifies the load memory address (LMA) region of the output section (optional). The value can be an expression. This option is parsed, but it has no effect on linking.
+
+<fillexp>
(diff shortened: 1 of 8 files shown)


partaror (approved):
Looks good to me!

jonathonpenix:
I'm a bit fuzzy in the expected behavior here for GNU ld, but I don't think eld's default behavior aligns with `ALIGN_WITH_INPUT`. Small example based on [1] and [2]:
```
cat > test.c << '!'
__attribute__((section(".a")))
__attribute__((aligned(64)))
int foo(void) { return 1; }

__attribute__((section(".b")))
__attribute__((aligned(128)))
int bar(void) { return 2; }
!

cat > script.t << '!'
MEMORY {
  VMA_REGION (rwx) : ORIGIN = 0x1000, LENGTH = 0x1000
  LMA_REGION (rwx) : ORIGIN = 0x3fc0, LENGTH = 0x1000
}

SECTIONS {
  .a : ALIGN_WITH_INPUT {
    begin = .;
    *(.a)
  } > VMA_REGION AT > LMA_REGION
  .b : ALIGN_WITH_INPUT {
    *(.b)
    end = .;
  } > VMA_REGION AT > LMA_REGION
}
!

clang --target=riscv32-unknown-elf -Oz -c test.c
riscv32-unknown-elf-ld.bfd -T script.t test.o -o ld.a.out
riscv-link -T script.t test.o -o eld.a.out
```

Output snippets:
```
$ llvm-objdump --headers ld.a.out
...
  1 .a                00000004 00001000 00003fc0 TEXT
  2 .b                00000004 00001080 00004040 TEXT
...
$ llvm-objdump --headers eld.a.out
...
  1 .a                00000004 00001000 00003fc0 TEXT
  2 .b                00000004 00001080 00004000 TEXT
...
```

I think this breaks the use case described in [1] (essentially, a single memcpy from lma to vma) given that the lma/vma "views" are different sizes here due to the different different lma/vmas. I _think_ this is roughly what Zephyr is trying to do when it uses `ALIGN_WITH_INPUT`, not sure what picolibc is wants out of `ALIGN_WITH_INPUT`.

I think this tracks with the description in [3] as well ("...the difference between the VMA and LMA remains intact throughout this output section with the ALIGN_WITH_INPUT attribute"), but I might be missing/misunderstanding something.

Links:
[1] https://binutils.sourceware.narkive.com/MTsXxI3t/patch-ld-add-align-with-input-output-section-attribute
[2] https://archive.fosdem.org/2024/events/attachments/fosdem-2024-2340-linker-scripts-in-lld-and-how-they-compare-with-gnu-ld/slides/22752/Fosdem_linker_scripts_2024_uG7XLOI.pdf (see slide 30)
[3] https://sourceware.org/binutils/docs/ld/Forced-Output-Alignment.html

quic-seaswara:
You are right and thanks for the test. 

I was experimenting with using PHDR's and picolibc uses PHDRs. 

I will still consider this as default behavior, as you can see from the changes to the test, if the section is going to the same segment its being treated properly. This is what `picolibc` is using


```
cat > test.c << '!'
__attribute__((section(".a")))
__attribute__((aligned(64)))
int foo(void) { return 1; }

__attribute__((section(".b")))
__attribute__((aligned(128)))
int bar(void) { return 2; }
!

cat > script.t << '!'
PHDRS {
  A PT_LOAD;
}

MEMORY {
  VMA_REGION (rwx) : ORIGIN = 0x1000, LENGTH = 0x1000
  LMA_REGION (rwx) : ORIGIN = 0x3fc0, LENGTH = 0x1000
}

SECTIONS {
  .a : ALIGN_WITH_INPUT {
    begin = .;
    *(.a)
  } > VMA_REGION AT > LMA_REGION :A
  .b : ALIGN_WITH_INPUT {
    *(.b)
    end = .;
  } > VMA_REGION AT > LMA_REGION :A
}
!

```

I will look into this once I get the support for building picolibc is uploaded

jonathonpenix:
Sounds good, thanks!

I'll take a look again at what Zephyr is up to, but IIRC I think most of the upstream Zephyr stuff isn't using PHDRs and is along the lines of the memcpy example as above, so I think this will eventually come up there.

I went ahead and created https://github.com/qualcomm/eld/issues/56 to track this, please feel free to edit/close/etc. as appropriate!

PR #55, opened 2025-04-22
Add emulation option support for hexagon, arm and aarch64

### LINKED ISSUE #54: Add emulation option (`-m <emulation>`) support for hexagon, arm and aarch64
Currently, `-m <emulation>` option only works for RISCV. It should work for all the targets.

partaror:
This commit adds support for emulation option (`-m <emulation>`) for hexagon, arm, and aarch64 targets. It also modifies the LIT to use ld.eld instead of the target symlinks for running tests.

Closes #54

### DIFF
--- a/test/Hexagon/standalone/InvalidEmulation/InvalidEmulation.test
+++ b/test/Hexagon/standalone/InvalidEmulation/InvalidEmulation.test
@@ -1,3 +1,6 @@
+UNSUPPORTED: true
+# This test is temporarily marked unsupported because currently we do not do
+# consistency checks for multiple emulation options.
 #---InvalidEmulation.test--------------------- Executable --------------------#
 #BEGIN_COMMENT
 # This tests that the linker should reject invalid Emulations specified on the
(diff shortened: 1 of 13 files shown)


quic-seaswara (commented):
probably linkopts should add the appropriate emulation option for each target

partaror:
> probably linkopts should add the appropriate emulation option for each target

Currently, many tests do not use `%linkopts`. I wanted to preserve the original test behavior as much as I can -- we were adding `-march=<arch>` to `link` as well instead of `linkopts`. If we should use `linkopts` for storing the emulation option, then I would like to do that in a separate pull-request.

test/Hexagon/standalone/InvalidEmulation/InvalidEmulation.test
```diff
@@ -1,3 +1,6 @@
+UNSUPPORTED: true
```
quic-seaswara, line 1: Can you please create a issue for this ?

quic-seaswara (approved):

test/Hexagon/standalone/InvalidEmulation/InvalidEmulation.test
partaror: Done. I have created an issue to track this: #58

PR #61, opened 2025-04-23
assignments/defsym needs to be processed in link order

### LINKED ISSUE #60: eld does not handle assignments that refer to symbols that can be provided
It seems like eld does not handle assignments that refer to symbols in the linker script or symbols that can be defined by symbols in archive libraries.

This pattern is seen when building picolibc, and I am listing down the tests that will need to be supported in the linker to successfully link picolibc and run the unit tests.

** Symbols provided by linker script **

```
cat > 1.c << \!
int bar() { return 0; }
!

cat > script.t << \!
SECTIONS {
  .text : { *(.text*) }
  PROVIDE(foo = 0);
  PROVIDE(__text_align = ALIGNOF(.text));
  ASSERT(__text_align > foo, "good!")
}
!

clang -c -target riscv32 -ffunction-sections 1.c
ld.eld -march riscv32 1.o -T script.t

```

This test results in the linker reporting an error that __text_align symbol is undefined

```
Fatal: script.t: Error: undefined symbol '__text_align' referenced in expression
Error: Assertion failed ASSERT(__text_align(0x0) > foo(0x0), "good!")
Fatal: Linking had errors.
```

** Symbols referred by defsym expressions **

cat > 1.c << !
extern int foo();
int baz() {
   return foo();
}
!

cat > 2.c << \!
int bar() { return 0; }
!

cat > script.t << \!
foo = bar;
!

clang -c -target riscv32 1.c
clang -c -target riscv32 2.c
llvm-ar cr lib2.a 2.o
ld.eld -march riscv32 1.o --defsym foo=bar -t lib2.a
ld.eld -march riscv32 1.o -T script.t -t lib2.a
ld.eld -march riscv32 1.o lib2.a --defsym foo=bar

The defsym expressions are respected by the linker using link order. 

Only the last link line in the above example should fail.

This is also used by picolibc build process.

quic-seaswara:
It seems like assignments and defsym expressions need to be processed by link order, if the expression refers to a symbol that is not yet visited by the linker, the symbol will need to be processed as an undefined symbol. This means that any expression that refers to a symbol that can be PROVIDED by the linker script can fill in the symbol or resolve to that symbol.

This particular combination is exercised when building picolibc.

This bug allows picolibc to successfully build/run tests when tested on RISC-V.

Closes #60

Change-Id: I98d94d78cf415f28aacf04df0da10983642d9b41


lib/Script/Assignment.cpp
```diff
@@ -264,3 +261,9 @@ bool Assignment::isDot() const { return (Name.size() == 1 && Name[0] == '.'); }
 bool Assignment::hasDot() const {
   return isDot() || ExpressionToEvaluate->hasDot();
 }
+
+std::unordered_set<std::string> Assignment::getSymbolNames() const {
```
partaror, line 270: We have a very similar existing function `Assignment::getSymbolNames(std::unordered_set<std::string> &SymbolTokens)`. Can we use that function to implement this new function? This way we will not duplicate how we are computing the symbol names. This being said, it is only one line in both the functions -- `ExpressionToEvaluate->getSymbolNames(SymbolTokens);`

include/eld/SymbolResolver/NamePool.h
```diff
@@ -129,6 +129,11 @@ class NamePool {
     return nullptr;
   }
 
+  // Helper function to add an undefined symbol to the NamePool
+  void
+  addUndefinedELFSymbol(InputFile *I, std::string SymbolName,
```
partaror, line 134: `std::string` -> `const std::string &` or `llvm::StringRef`.

Also, should we rename the function to `addUndefinedSymbol`? We only support ELF target, and we typically do not use `ELF` for the other `NamePool` APIs.

partaror (commented):
Overall looks good. Some initial comments.

lib/Script/Assignment.cpp
quic-seaswara: Definitely, thanks for the catch

include/eld/SymbolResolver/NamePool.h
quic-seaswara: Sure

partaror:
We are getting an incorrect multiple definition error with this change.

Reproducible example:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
int foo() { return 1; }
int v = 11;
EOF

cat >script.t <<\EOF
SECTIONS {
  u = v;
  v = 13;
  w = v;
}
EOF

LDs=(ld.eld ld.lld arm-none-eabi-ld.bfd)
SFs=(eld lld bfd)

clang -o 1.o 1.c -c -ffunction-sections -fdata-sections

for i in "${!SFs[@]}"; do
done
```

include/eld/SymbolResolver/NamePool.h
quic-seaswara: Addressed the comments, thank  you!

lib/Script/Assignment.cpp
```diff
@@ -264,3 +256,19 @@ bool Assignment::isDot() const { return (Name.size() == 1 && Name[0] == '.'); }
 bool Assignment::hasDot() const {
   return isDot() || ExpressionToEvaluate->hasDot();
 }
+
+// Add undefined symbols for symbols referred by the assignment
+void Assignment::processAssignment(Module &CurModule, InputFile &I) {
```
partaror, line 261: nitpick: `processAssignment` name is vague and does not make it clear what is exactly happening in the function. Can we please name it something like: `addReferencedSymbols`?

lib/Script/Assignment.cpp
```diff
@@ -264,3 +256,19 @@ bool Assignment::isDot() const { return (Name.size() == 1 && Name[0] == '.'); }
...
+  CurModule.getScript().assignments().push_back(this);
+  if (isProvideOrProvideHidden())
+    return;
+  for (auto &Sym : getSymbolNames()) {
+    CurModule.getNamePool().addUndefinedELFSymbol(&I, Sym);
```
partaror, line 266: Is it right semantically to add a symbol as an undefined symbol even though the linker script contains the symbol definition?

For example:

```bash
u = 11;
v = u + a;
w = v + b;
```

We are adding `u` and `v` as an undefined symbols from this linker script, but this linker script contains the definitions of `u` and `v` as well. This is different from how we handle symbols from an object file.

lib/Script/Expression.cpp
```diff
@@ -687,6 +688,7 @@ void AlignExpr::getSymbols(std::vector<ResolveInfo *> &Symbols) {
 
 void AlignExpr::getSymbolNames(std::unordered_set<std::string> &SymbolTokens) {
   ExpressionToEvaluate.getSymbolNames(SymbolTokens);
+  AlignmentExpression.getSymbolNames(SymbolTokens);
```
partaror, line 691: This is an interesting find. We missed this important case until now.

include/eld/Diagnostics/DiagSymbolResolutions.inc
```diff
@@ -94,3 +94,5 @@ DIAG(error_patchable_script, DiagnosticEngine::Error,
      "Patchable symbol '%0' is overridden by linker script")
 DIAG(error_patchable_alias_not_found, DiagnosticEngine::Error,
      "Patchable alias '%0' was not found or was garbage collected")
+DIAG(warning_override_symbol, DiagnosticEngine::Warning,
+     "Symbol %0 defined in %1 overrides symbol in %2")
```
partaror, line 98: Can we please quote the symbol and input-file names in the diagnostics? It can make diagnostics easier to follow for uncommon symbol and input file names.

partaror (changes requested):
The changes in this pull-request breaks compatibility with the previous behavior:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
int foo = 11;
EOF

cat >script.t <<\EOF
PROVIDE(u = foo);
foo = 13 + u;
EOF

ld.eld -o a.eld.out 1.o -T script.t -Map a.eld.map.txt
```

Previously for this link, eld used to (correctly) report the error:

```
Fatal: script.t: Error: undefined symbol 'u' referenced in expression
Fatal: Linking had errors.
```

eld with this patch does not report any error for this link.

include/eld/Diagnostics/DiagSymbolResolutions.inc
quic-seaswara: We already do this this is the message 

Warning: Symbol v defined in s.t overrides symbol in 1.o

partaror:
> This is one of the behaviors that is needed by picolibc. This patch fixes this to reflect bfd behavior.

Oh, this is interesting. LLD errors out with the below message for this case:

```
ld.lld:  error: assignment to symbol foo does not converge
```

GNU ld links fine.

partaror:
This is another interesting (_weird_) case:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
int foo = 11;
int v = 1;
EOF

cat >script.t <<\EOF
PROVIDE(u = 13);
foo = u + v;
PROVIDE(u = 15);
EOF

LDs=(ld.eld ld.lld arm-none-eabi-ld.bfd)
SFs=(eld lld bfd)

for i in "${!SFs[@]}"; do
  ${LDs[$i]} -o a.${SFs[$i]}.out 1.o -T script.t -Map a.${SFs[$i]}.map.txt
done
```

Both lld and GNU ld selects `PROVIDE(u = 15);` for obtaining the value of `u`, however, eld selects the first `PROVIDE` statement: `PROVIDE(u = 13);`. If we are following GNU semantics, then should we select the second `PROVIDE` statement as well in this case?

quic-seaswara:
> This is another interesting (_weird_) case:
> 
> ```shell
> #!/usr/bin/env bash
> 
> cat >1.c <<\EOF
> int foo = 11;
> int v = 1;
> EOF
> 
> cat >script.t <<\EOF
> PROVIDE(u = 13);
> foo = u + v;
> PROVIDE(u = 15);
> EOF
> 
> clang -o 1.o 1.c -c -ffunction-sections
> LDs=(ld.eld ld.lld arm-none-eabi-ld.bfd)
> SFs=(eld lld bfd)
> 
> for i in "${!SFs[@]}"; do
>   ${LDs[$i]} -o a.${SFs[$i]}.out 1.o -T script.t -Map a.${SFs[$i]}.map.txt
> done
> ```
> 
> Both lld and GNU ld selects `PROVIDE(u = 15);` for obtaining the value of `u`, however, eld selects the first `PROVIDE` statement: `PROVIDE(u = 13);`. If we are following GNU semantics, then should we select the second `PROVIDE` statement as well in this case?

This is an old behavior, and requires a seperate issue. I believe we issue a warning, we should continue to do that as an improvement over GNU so that user knows and fix the error.

partaror:
> I believe we issue a warning, we should continue to do that as an improvement over GNU so that user knows and fix the error.

We do not report any warning :').

quic-seaswara:
> ```shell
> !/usr/bin/env bash
> 
> cat >1.c <<\EOF
> int foo = 11;
> int v = 1;
> EOF
> 
> cat >script.t <<\EOF
> PROVIDE(u = 13);
> foo = u + v;
> PROVIDE(u = 15);
> EOF
> 
> clang -o 1.o 1.c -c -ffunction-sections
> LDs=(ld.eld ld.lld arm-none-eabi-ld.bfd)
> SFs=(eld lld bfd)
> 
> for i in "${!SFs[@]}"; do
>   ${LDs[$i]} -o a.${SFs[$i]}.out 1.o -T script.t -Map a.${SFs[$i]}.map.txt
> done
> ```

```
$ ld.eld 1.o -T script.t  -Wall
Warning: Symbol foo defined in script.t overrides symbol in 1.o
Warning: script.t: Ignoring provide symbol 'u' redeclaration
```

lib/Script/Assignment.cpp
quic-seaswara: It needs to be `processAssignment` as the function just does not add the symbols referenced by expression.

lib/Script/Assignment.cpp
quic-seaswara: This is very interesting. The end behavior is the same. 

I would like to keep this to remain consistent instead of copying exactly bfd behavior. 

We can look into this case more as a usecase defined approach.

quic-seaswara:
I think I have addressed most of all comments.

There is definitely room for improvement in the documentation aspect. 

I will take this as the next step.

PR #65, opened 2025-04-28
Unconditionally create internal RISCV AttributeFragment

partaror:
Until now, we only created internal RISCV AttributeFragment if there
was an input file that contained .riscv.attributes section. This commit
changes this behavior by unconditionally creating the internal RISCV
AttributeFragment.

This change fixes the issue of the missing .riscv.attributes output
section when LTO inputs are used with linker scripts. The root cause
of the issue was that we were moving the content of the empty internal RISCV
AttributeSection into the output section due to the order of input sections
after section sorting and delayed creation of internal RISCV AttributeFragment.

### DIFF
--- a/lib/Object/ObjectLinker.cpp
+++ b/lib/Object/ObjectLinker.cpp
@@ -678,6 +678,7 @@ bool ObjectLinker::mayBeSortSections(std::vector<Section *> &Sections) {
                      ELFSection *B = llvm::dyn_cast<ELFSection>(BSection);
                      if (A == nullptr or B == nullptr)
                        return false;
+                     // FIXME: Redundant checks. All files have original input.
                      if (!A->originalInput())
                        return false;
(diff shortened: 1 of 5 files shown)


quic-seaswara (changes requested):
Its better to house an empty RISCVAttributeFragment which will later get filled.

Does this affect AArch64/ARM/Hexagon ?

partaror:
> Its better to house an empty RISCVAttributeFragment which will later get filled.

Yes, agreed. I have modified the PR to create an empty RISCV AttributeFragment instead of relying on the sort order.

> Does this affect AArch64/ARM/Hexagon ?

AttributeFragment is handled differently for AArch64/ARM/Hexagon. This issue does not seem to affect these targets.

quic-seaswara (approved):

PR #66, opened 2025-05-01
[LSParser] Ignore non-ascii unicode characters

### LINKED ISSUE #62: skip non ascii characters in linker scripts
I am creating this issue where the GNU linker skips non ascii characters in the linker script.

To support transition to the open source linker we want to make sure that we skip non ascii characters similar to the GNU linker.

You can use this simple example with all the attached linker script

```
cat > 1.c << \!
int foo() { return 0; }
!

clang -c -target riscv32 -c 1.c -ffunction-sections
ld.eld -march riscv32 1.o -T <script>

```

Use the attached linker scripts as part of this issue.

partaror:
This commit modifies linker script parser to ignore non-ascii unicode characters and report note diagnostics for the same.

Closes #62

### DIFF
--- a/include/eld/ScriptParser/ScriptLexer.h
+++ b/include/eld/ScriptParser/ScriptLexer.h
@@ -18,6 +18,7 @@
 #include "llvm/ADT/StringRef.h"
 #include "llvm/ADT/Twine.h"
 #include "llvm/Support/MemoryBufferRef.h"
+#include <optional>
 #include <vector>
 
 namespace eld {
@@ -39,6 +40,9 @@ public:
 
   // Set error as needed
   void setError(const llvm::Twine &Msg);
+
+  void setNote(const llvm::Twine &msg,
+               std::optional<llvm::StringRef> columnTok = std::nullopt) const;
 
   void lex();
 
@@ -75,16 +79,16 @@ public:
   bool consumeLabel(llvm::StringRef Tok);
 
   // Check if s encloses t
-  bool encloses(llvm::StringRef S, llvm::StringRef T);
+  bool encloses(llvm::StringRef S, llvm::StringRef T) const;
 
   // Get current location of token including filename
-  std::string getCurrentLocation();
+  std::string getCurrentLocation() const;
 
   // Unquote string if quoted
   llvm::StringRef unquote(llvm::StringRef S);
 
   // Get Line and column information
-  llvm::StringRef getLine();
+  llvm::StringRef getLine() const;
 
   /// Returns true if there are no reported errors that should
   /// end the link abruptly.
@@ -98,15 +102,23 @@ public:
 
 protected:
   // Get current memory buffer
-  llvm::MemoryBufferRef getCurrentMB();
+  llvm::MemoryBufferRef getCurrentMB() const;
 
-  size_t getColumnNumber();
+  size_t getColumnNumber() const;
+
+  size_t computeColumnWidth(llvm::StringRef s, llvm::StringRef e) const;
 
 private:
   // Handle expression splits.
   void maybeSplitExpr();
 
-  // Check if there is an error
+  llvm::StringRef noteAndSkipNonASCIIUnicodeChars(llvm::StringRef s) const;
+
+  bool isNonASCIIUnicode(char c) const { return c & 0x80; }
+
+  bool isFirstByteOfMultiByteUnicode(char c) const {
+    return (c & 0xc0) == 0xc0;
+  }
 
 protected:
   struct Buffer {
--- a/lib/ScriptParser/ScriptLexer.cpp
+++ b/lib/ScriptParser/ScriptLexer.cpp
@@ -54,7 +54,7 @@ using namespace eld;
 using namespace eld::v2;
 
 // Returns a whole line containing the current token.
-StringRef ScriptLexer::getLine() {
+StringRef ScriptLexer::getLine() const {
   StringRef S = getCurrentMB().getBuffer();
 
   size_t Pos = S.rfind('\n', PrevTok.data() - S.data());
@@ -64,11 +64,11 @@ StringRef ScriptLexer::getLine() {
 }
 
 // Returns 0-based column number of the current token.
-size_t ScriptLexer::getColumnNumber() {
-  return PrevTok.data() - getLine().data();
+size_t ScriptLexer::getColumnNumber() const {
+  return computeColumnWidth(getLine().data(), PrevTok.data());
 }
 
-std::string ScriptLexer::getCurrentLocation() {
+std::string ScriptLexer::getCurrentLocation() const {
   std::string Filename = std::string(getCurrentMB().getBufferIdentifier());
   return (Filename + ":" + Twine(PrevTokLine)).str();
 }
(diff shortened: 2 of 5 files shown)


lib/ScriptParser/ScriptLexer.cpp
```diff
@@ -318,3 +334,22 @@ void ScriptLexer::prev() {
     CurTok = {};
   }
 }
+
+llvm::StringRef
+ScriptLexer::warnAndSkipNonASCIIUnicodeChars(llvm::StringRef s) {
```
quic-seaswara: nit pick : since we are not warning and raising a note, the function may need to be changed accordingly ?

lib/ScriptParser/ScriptLexer.cpp
```diff
@@ -318,3 +334,22 @@ void ScriptLexer::prev() {
...
+  while (!s.empty() && isNonASCIIUnicode(s[0])) {
+    if (isFirstByteOfMultiByteUnicode(s[0]))
+      setNote("treating non-ascii unicode character as whitespace", s);
+    s = s.drop_front();
+  }
```
quic-seaswara, line 347: also function can be made const

test/Common/LinkerScriptParsing/Expressions/UnicodeCharacterWarning/UnicodeCharacterWarning.test
```diff
@@ -0,0 +1,12 @@
+#---UnicodeCharacterWarning.test--------------------- Executable------------------#
+#BEGIN_COMMENT
+# This test checks that linker script parser gives note diagnostics for
+# non-ascii unicode characters.
+#END_COMMENT
+RUN: %lsparserverifier %lsparserverifier_opts %p/Inputs/script.t 2>&1 | %filecheck %s
+CHECK: Note: {{.*}}script.t:7: treating non-ascii unicode character as whitespace
+CHECK: >>>    FOODATA_LMA_END    = FOODATA_LMA_START + 10;
+CHECK: >>>                    ^
+CHECK: Note: {{.*}}script.t:7: treating non-ascii unicode character as whitespace
```
quic-seaswara, line 10: column number missing ?

lib/ScriptParser/ScriptLexer.cpp
```diff
@@ -102,6 +102,21 @@ void ScriptLexer::setError(const Twine &Msg) {
...
+  std::string s = (getCurrentLocation() + ": " + msg).str();
+  if (PrevTok.size()) {
+    std::size_t columnNumber = 0;
+    if (columnTok)
+      columnNumber = computeColumnWidth(getLine().data(), columnTok->data());
+    else
+      columnNumber = getColumnNumber();
+    s += "\n>>> " + getLine().str() + "\n>>> " +
+         std::string(columnNumber, ' ') + "^";
+  }
```
quic-seaswara, line 116: An improvement here could be we can skip columns to the next ASCII ?

Example 

```

cat > s.t << \!
PHDRS
{
    RAM                                                                            PT_LOAD;
}


SECTIONS
{
  .foo : { *(.text.foo) } :RAM
}
!

cat > 1.c << \!
int foo() { return 0; }
!

clang -c -target riscv32 1.c
ld.eld -march riscv32  1.o -T s.t
```

quic-seaswara: git changes the non ascii character to a space, you can change it locally and see what I say, you will get atleast 100 note messages

include/eld/ScriptParser/ScriptLexer.h
```diff
@@ -102,11 +106,17 @@ class ScriptLexer {
...
 private:
   // Handle expression splits.
   void maybeSplitExpr();
 
-  // Check if there is an error
+  llvm::StringRef warnAndSkipNonASCIIUnicodeChars(llvm::StringRef s);
+
+  bool isNonASCIIUnicode(char c) { return c & 0x80; }
+
+  bool isFirstByteOfMultiByteUnicode(char c) { return (c & 0xc0) == 0xc0; }
```
quic-seaswara: some of these functions can be made const.

The functions that pass char can take input as const as well

include/eld/ScriptParser/ScriptLexer.h
```diff
@@ -40,6 +41,9 @@ class ScriptLexer {
   // Set error as needed
   void setError(const llvm::Twine &Msg);
 
+  void setNote(const llvm::Twine &msg,
+               std::optional<llvm::StringRef> columnTok = std::nullopt);
```
quic-seaswara: std::optional with default parameter ?

quic-seaswara (changes requested):

lib/ScriptParser/ScriptLexer.cpp
partaror: Thank you for pointing this out. I have fixed the function name.

test/Common/LinkerScriptParsing/Expressions/UnicodeCharacterWarning/UnicodeCharacterWarning.test
partaror: We do not print-out the column number, we instead use caret ('^') symbol to point to the column that triggered the diagnostic.

include/eld/ScriptParser/ScriptLexer.h
partaror: Yes, the default parameter is empty optional object. If `columnTok` is not passed explicitly, then the `PrevTok` is used for computing the column width.

include/eld/ScriptParser/ScriptLexer.h
partaror: Yes, thank you for pointing this out. I have made all such functions that are added by this API as `const`.

lib/ScriptParser/ScriptLexer.cpp
partaror: Oh, yes, seeing the same diagnostic a lot of times would not be very helpful. I have fixed the `noteAndSkipNonASCIIUnicodeChars` function such that the function only reports one diagnostic for one continuous block of non-ascii chars.

partaror:
Thank you @quic-seaswara for the detailed reviews!

lib/ScriptParser/ScriptLexer.cpp
partaror: I have made it `const` along with many other functions that were being used by the APIs added in this PR.

include/eld/ScriptParser/ScriptLexer.h
```diff
@@ -40,6 +41,9 @@ class ScriptLexer {
   // Set error as needed
   void setError(const llvm::Twine &Msg);
 
+  void setNote(const llvm::Twine &msg,
+               std::optional<llvm::StringRef> columnTok = std::nullopt) const;
+
```
quic-seaswara, line 46: Do you still need the default parameter here ?

partaror: The additional `columnTok` parameter is helpful. For example, if the line is: `SECTIONS <non-ascii-char>`, then if we do not explicitly specify the token where we found the non-ascii char, then we will need to use the previous token that was parsed to compute the column number. The previous token in this case as is `SECTIONS` and using it would result in incorrect column number.

Default parameter is helpful because for most situations we would not need to explicitly pass the token using which the column number is to be computed. So, when the default value (`std::nullopt`) is used, we take the `PrevTok` for computing the column number.

quic-seaswara (approved):

PR #68, opened 2025-05-05
Support 'archive: member' file-pattern in input section description

### LINKED ISSUE #67: linker script rule matches invalid sections
Linker script rule matches invalid sections with the below linker script

```bash
cat > 1.c << \!
int data = 10;
!

cat > script.t << \!
SECTIONS {
  .data : {
    *lib1.a: *(.*data*)
  }
}
!

clang -c -target hexagon  -c 1.c -g
llvm-ar cr lib1.a 1.o
ld.eld -mhexagonelf --whole-archive lib1.a -T script.t -Map x

```

The pattern lib1.a space translates to lib1.a:*(*) which made the linker match all the sections in the linker script rule.

partaror:
This commit adds support for 'archive: member' pattern in input section description. Generally, there should be no space between the 'archive:' and 'member'. However, it is a common mistake to put a space here. This commit makes this common use-case work as expected and reports a deprecated warning for the same.

Closes #67

### DIFF
--- a/lib/ScriptParser/ScriptLexer.cpp
+++ b/lib/ScriptParser/ScriptLexer.cpp
@@ -117,6 +117,14 @@ void ScriptLexer::setNote(const Twine &msg,
   ThisConfig.raise(Diag::note_linker_script) << s;
 }
 
+void ScriptLexer::setWarn(const Twine &Msg) {
+  std::string S = (getCurrentLocation() + ": " + Msg).str();
+  if (PrevTok.size())
+    S += "\n>>> " + getLine().str() + "\n>>> " +
+         std::string(getColumnNumber(), ' ') + "^";
+  ThisConfig.raise(Diag::warn_linker_script) << S;
+}
+
 void ScriptLexer::lex() {
   for (;;) {
     StringRef &S = CurBuf.S;
@@ -145,10 +153,9 @@ void ScriptLexer::lex() {
     if (S.starts_with("\"")) {
       size_t E = S.find("\"", 1);
       if (E == StringRef::npos) {
-        size_t Lineno =
-            StringRef(CurBuf.Begin, S.data() - CurBuf.Begin).count('\n');
+        size_t Lineno = computeLineNumber(S);
         ThisConfig.raise(Diag::error_linker_script)
-            << llvm::Twine(CurBuf.Filename + ":" + Twine(Lineno + 1) +
+            << llvm::Twine(CurBuf.Filename + ":" + Twine(Lineno) +
                            ": unclosed quote")
                    .str();
         return;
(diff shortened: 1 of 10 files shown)


lib/ScriptParser/ScriptLexer.cpp
```diff
@@ -314,7 +322,14 @@ StringRef ScriptLexer::unquote(StringRef S) {
 
 void ScriptLexer::prev() {
   if (!PrevTok.empty()) {
+    // FIXME: CurBuf.LineNumber needs to be updated!
```
quic-seaswara, line 340: Can you please create an issue ?

quic-seaswara (commented):
Thanks for working on this PR. Let us hold on merging this pull request. 

Discussing with the GNU community and checking the ambiguity would be very useful here.

partaror:
> Discussing with the GNU community and checking the ambiguity would be very useful here.

As per GNU, there is no ambiguity because GNU does not allow any whitespace between the '<archive>:<mem>' pattern.

From https://sourceware.org/binutils/docs/ld/Input-Section-Basics.html:

> You can also specify files within archives by writing a pattern matching the archive, a colon, then the pattern matching the file, with no whitespace around the colon.

quic-seaswara:
Can we convert this patch to emit a warning with -Wlinker-script ?

parth-07:
> Can we convert this patch to emit a warning with -Wlinker-script ?

I have modified the patch to emit a warning with `-Wlinker-script`.

quic-seaswara (approved):
Can you please add this example in the FAQ

```
cat > 1.c << \!
int data = 10;
!

cat > script.t << \!
SECTIONS {
  .data : {
     /* should show warning */
    *lib1.a: 1.o(.*data*)
     /* should not show warning */
    /tmp/lib1.a*: (.*data*)
    *lib1.a: (.*data*)
    /* Show a warning */
    *lib1.a: *(.*data*)
    /* show a warning */
    *lib1.a:
    *(.data*)
  }
}
!

clang -c   -c 1.c -g -target riscv32
llvm-ar cr lib1.a 1.o
riscv-link --whole-archive lib1.a -T script.t -Map x -Wlinker-script

```

parth-07:
> Can you please add this example in the FAQ

I have added the example as part of the FAQ.

One small caveat, we do not support new-line between 'archive:' and 'file' pattern, even with this patch. Therefore, I have modified the below excerpt from the example:

```
    /* show a warning */
    *lib1.a:
    *(.data*)
```

to:

```
    /* No warning. '*lib1.a:' and '*(.data*)' are considered as separate
       input section descriptions. */
    *lib1.a:
    *(.data*)
```

lib/ScriptParser/ScriptLexer.cpp
parth-07: I have created an issue: https://github.com/qualcomm/eld/issues/107

PR #70, opened 2025-05-06
[RISCV] Use New Nonstandard Relocation Macro

lenary:
This updates the relocation macros to match the refactor in llvm/llvm-project#138226. This should avoid breaking eld when other vendors upstream their own vendor relocations.

### DIFF
--- a/lib/Target/RISCV/RISCVRelocationInternal.h
+++ b/lib/Target/RISCV/RISCVRelocationInternal.h
@@ -67,10 +67,9 @@ enum : uint32_t {
   QUALCOMMVendorRelocationOffset =
       FirstQUALCOMMVendorRelocation - FirstNonstandardRelocation,
 
-#define ELF_RISCV_NONSTANDARD_RELOC(vendor_symbol, name, value)                \
-  name = value + vendor_symbol##VendorRelocationOffset,
+#define ELF_RISCV_NONSTANDARD_RELOC_QUALCOMM(name, value)                \
+  name = value + QUALCOMMVendorRelocationOffset,
 #include "llvm/BinaryFormat/ELFRelocs/RISCV_nonstandard.def"
-#undef ELF_RISCV_NONSTANDARD_RELOC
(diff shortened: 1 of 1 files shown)


quic-seaswara (approved):

lib/Target/RISCV/RISCVRelocationInternal.h
```diff
@@ -67,10 +67,9 @@ enum : uint32_t {
   QUALCOMMVendorRelocationOffset =
       FirstQUALCOMMVendorRelocation - FirstNonstandardRelocation,
 
-#define ELF_RISCV_NONSTANDARD_RELOC(vendor_symbol, name, value)                \
-  name = value + vendor_symbol##VendorRelocationOffset,
+#define ELF_RISCV_NONSTANDARD_RELOC_QUALCOMM(name, value)                \
+  name = value + QUALCOMMVendorRelocationOffset,
```
partaror, line 78: It seems that `RISCV_nonstandard.def` still expects `ELF_RISCV_NONSTANDARD_RELOC` macro to be defined. 

If we need both `ELF_RISCV_NONSTANDARD_RELOC` and `ELF_RISCV_NONSTANDARD_RELOC_QUALCOMM`, then
does the suggested change seems right?

```suggestion
#define ELF_RISCV_NONSTANDARD_RELOC(vendor_symbol, name, value)                \
  name = value + vendor_symbol##VendorRelocationOffset
  
#define ELF_RISCV_NONSTANDARD_RELOC_QUALCOMM(name, value)                \
  ELF_RISCV_NONSTANDARD_RELOC(QUALCOMM, name, value)
```

partaror: Please ignore my above comment, the linked llvm-project PR removes all usage of `ELF_RISCV_NONSTANDARD_RELOC` macros.

partaror (approved):
Looks good!

PR #89, opened 2025-05-08
[PluginAPI] findConfigFile() should return errors

### LINKED ISSUE #88: LinkerWrapper::findConfigFile does not return errors
despite returning `eld::Expected<std::string>` `findConfigFile()` never returns errors when the file is not found.

quic-areg:
Previously the API did not return errors when
the file passed is not found.

Fixes #88


test/Common/Plugin/FindConfigFile/findconfig.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include "OutputSectionIteratorPlugin.h"
+#include "PluginVersion.h"
+#include <string>
+#include <iostream>
+
+#ifdef _WIN32
+#define DLL_EXPORT __declspec(dllexport)
+#else
+#define DLL_EXPORT
+#endif
```
quic-seaswara: This is already part of LinkerPlugin.h

test/Common/Plugin/FindConfigFile/findconfig.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+using namespace eld::plugin;
+
+class DLL_EXPORT findconfig : public OutputSectionIteratorPlugin {
```
quic-seaswara: can you use LLVM coding conventions here, might want to call it as FindConfig

test/Common/Plugin/FindConfigFile/findconfig.cpp
```diff
@@ -0,0 +1,61 @@
...
+public:
+  findconfig() : OutputSectionIteratorPlugin("findconfig") {}
+
+  // Perform any initialization here
+  void Init(std::string Options) override {
+    if (getLinker()->getState() != LinkerWrapper::AfterLayout)
+     return;
+    auto E = getLinker()->findConfigFile(Options);
+    ELDEXP_REPORT_AND_RETURN_VOID_IF_ERROR(getLinker(), E);
+    std::cerr << "found config file " << Options << std::endl;
```
quic-seaswara, line 20: can use Linker diag

test/Common/Plugin/FindConfigFile/FindConfigFile.test
```diff
@@ -0,0 +1,6 @@
+# tests that LinkerWrapper::findConfigFile() properly returns errors
```
quic-seaswara, line 3: it will be useful to use the test template since all the new tests are already using the template.

test/Common/Plugin/SearchDiagnosticsPlugin/SearchDiagnosticsPlugin.cpp
```diff
@@ -15,8 +15,6 @@ class DLL_A_EXPORT SearchDiagnosticsPlugin : public eld::plugin::LinkerPlugin {
   void Init(const std::string &Options) override {
     eld::Expected<std::string> expConfigFile =
         getLinker()->findConfigFile(PluginName + ".config");
-    ELDEXP_REPORT_AND_RETURN_VOID_IF_ERROR(
-        getLinker(), expConfigFile);
```
quic-seaswara, line 19: expConfigFile becomes an unused variable, may be remove the variable.

quic-seaswara (changes requested):

test/Common/Plugin/FindConfigFile/findconfig.cpp
quic-areg: Done

test/Common/Plugin/FindConfigFile/findconfig.cpp
quic-areg: Done. Any plugin test will not follow LLVM conventions due to `RegisterAll()`, `Cleanup()` etc

test/Common/Plugin/FindConfigFile/findconfig.cpp
quic-areg: I think this is fine for a test.

quic-seaswara:
Unfortunately we cannot change the plugin API's but let us try to make the code as closer to LLVM coding conventions.

test/Common/Plugin/FindConfigFile/FindConfigFile.test
quic-areg: Done, but I do not think it is useful

`---FindConfigFile.test--------------------- Executable`

Repeating the filename. We already know what file we have opened. The test does not actually generate an executable. There is no appropriate entry in `ExampleOfMyLitTest.test` for this test. 

```
#BEGIN_COMMENT
# tests that LinkerWrapper::findConfigFile() properly returns errors
#END_COMMENT
```

My comment was already denoted by the `#`. Here we use **BEGIN**_COMMENT, to denote the start of a comment, but **START**_TEST to denote the start of a test.

```
#START_TEST
RUN: %not %link %linkopts -L %p/Inputs -T %p/Inputs/script.t -o %t.out 2>&1 | %filecheck %s
#END_TEST
```

start/end test are already denoted by `RUN` lines.

```
CHECK: found config file file-that-exists
CHECK: findconfig:Error: plugin configuration file not found: '/foo/bar/file-that-does-not-exist'
```

Why is there no label for `CHECK` lines? `#START_CHECK` `END_CHECK`?

test/Common/Plugin/SearchDiagnosticsPlugin/SearchDiagnosticsPlugin.cpp
quic-areg: Done

partaror (approved):
LGTM.

PR #92, opened 2025-05-13
Fix dot counter handling for non-alloc sections

### LINKED ISSUE #91: Non-alloc sections that contains symbol assignments are missing from the output image
Non-alloc sections that contains symbols assignments in the linker script are missing from the output image.

Reproducible example:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
int foo() { return 1; }
int bar() { return 3; }
int baz() { return 5; }
EOF

cat >script.t <<\EOF
SECTIONS {
  .foo : { *(*foo*) }
  debug_info : { 
    start = .;
    *(.debug_info*) 
    end = .;
  }
  TEXT : { *(*text*) }
} 
EOF

clang -o 1.o 1.c -c -ffunction-sections -g
LDs=(ld.eld ld.lld)
SFs=(eld lld)

for i in "${!SFs[@]}"; do
done
```

`llvm-readelf -sS 1.eld.out` does not contain `debug_info` output section.

partaror:
Until now, we were not properly handling dot counter in non-alloc sections and this was leading to incorrect symbol assignment evaluations in non-alloc sections and non-alloc sections size incorrectly being set to 0. This commit fixes the dot counter handling in non-alloc sections. Now dot counter is moved around as usual in non-alloc sections but is restored to its original value after performing symbol evaluations and updating the output section size.

Closes #91

### DIFF
--- a/test/Common/standalone/linkerscript/NonAllocSectionSymbolAssignment/Inputs/script.t
+++ b/test/Common/standalone/linkerscript/NonAllocSectionSymbolAssignment/Inputs/script.t
@@ -1,0 +1,9 @@
+SECTIONS {
+  .foo : { *(*foo*) }
+  debug_info : {
+    start = .;
+    *(.debug_info*)
+    end = .;
+  }
+  TEXT : { *(*text*) }
+}
--- a/lib/Target/GNULDBackend.cpp
+++ b/lib/Target/GNULDBackend.cpp
@@ -1904,6 +1904,7 @@ void GNULDBackend::evaluateAssignments(OutputSectionEntry *out,
 
   // Initial dot symbol value.
   dotSymbol->setValue(OutSection->addr());
+  LDSymbol::ValueType InitialDotValue = OutSection->addr();
 
   if (m_Module.getPrinter()->traceAssignments())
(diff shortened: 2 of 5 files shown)


test/Common/standalone/linkerscript/NonAllocSectionSymbolAssignment/Inputs/script.t
```diff
@@ -0,0 +1,9 @@
+SECTIONS {
+  .foo : { *(*foo*) }
+  debug_info : {
+    start = .;
+    *(.debug_info*)
+    end = .;
+  }
```
quic-seaswara, line 7: Can you add another test to test assignments before and after the section 

SECTIONS {

  .foo : { *(*foo) }
  
  . =  . + 100; 
 my_start = .;
 .debug_info : {
    start = .;
   *(.debug_info*)
   end = . ;
 }
 myend = .;
 mytext : { *(.text) }
}

partaror: I have added this test.

PR #95, opened 2025-05-16
Make dynamic string table simpler to use

### LINKED ISSUE #94: Dynamic string table is difficult to use because string offsets cannot be computed before layout
Dynamic string table (`ELFFileFormat::f_pDynStrTab`) is difficult to use because the string offsets of the strings in the dynamic string table cannot be computed before the layout. The dynamic string table is currently roughly used as follows:

1) **Before layout: ** Compute the size required for the dynamic string table by iterating over all the dynamic symbols, sonames, DT_NEEDED entries, and so on.
2) **After layout: ** Again traverse over all the dynamic symbols, so names, DT_NEEDED entries, and this time copy the string into the dynamic string table and carefully assign the string offsets where the strings are referred.

This is cumbersome, error-prone and duplication of work that can be avoided. It is easy to have an inconsistency in iterating over dynamic elements in the *Before layout* stage or in the *After layout* stage. It makes things significantly complicated for dynamic sections such as `.gnu.version_d` and `.gnu.version_r` where certain offsets inside these sections needs to store offset into the dynamic string table.

partaror:
Currently, dynamic string table is used as follows:

1) Before layout: Compute the size of the dynamic string table by
   traversing over dynamic symbols, sonames, DT_NEEDED entries and so
   on.
2) During output image writing: Again traverse over dynamic symbols,
   sonames, DT_NEEDED entries, ..., This time copy the string into the
   dynamic string table and set the string offset in the dynamic-related
   section that was referring the string.

This separation of computing size, offsets and actually writing makes the dynamic string table use cumbersome and error-prone. Especially for sections such as `.gnu.version_r` and `.gnu.version_d`.

With this commit, the dynamic string table can be used as:

```cpp
std::size_t SONameOffset = FileFormat->addStringToDynStrTab(
    llvm::cast<ELFDynObjectFile>(lib)->getSOName());
auto DTEntry = dynamic()->reserveNeedEntry();
DTEntry->setValue(llvm::ELF::DT_NEEDED, SONameOffset);
```

There is no need to separate size and offset computation anymore.

Closes #94

### DIFF
--- a/include/eld/Target/ELFFileFormat.h
+++ b/include/eld/Target/ELFFileFormat.h
@@ -61,6 +61,46 @@ public:
 
   std::vector<ELFSection *> &getSections() { return f_pSections; }
 
+  std::size_t addStringToDynStrTab(const std::string &S) {
+    return DynamicStringTableContents.addString(S);
+  }
+
+  std::size_t getDynStrTabSize() const {
+    return DynamicStringTableContents.size();
+  }
+
+  std::optional<std::size_t> getOffsetInDynStrTab(const std::string &S) const {
+    return DynamicStringTableContents.getOffset(S);
+  }
+
+  const std::string &getDynStrTabContents() const {
+    return DynamicStringTableContents.Strings;
+  }
+
+private:
+  struct StringTableContents {
+    std::string Strings = std::string(1, '\0');
+    std::unordered_map<std::string, std::size_t> StringOffsets;
+
+    std::size_t addString(const std::string &S) {
+      auto iter = StringOffsets.find(S);
+      if (iter != StringOffsets.end())
+        return iter->second;
+      std::size_t Offset = Strings.size();
+      Strings += S + std::string(1, '\0');
+      return StringOffsets[S] = Offset;
+    }
+
+    std::size_t size() const { return Strings.size(); }
+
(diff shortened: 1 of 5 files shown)


include/eld/Target/ELFFileFormat.h
```diff
@@ -61,6 +61,46 @@ class ELFFileFormat : public LDFileFormat {
...
 
+  std::size_t addStringToDynStrTab(const std::string &S) {
+    return DynamicStringTableContents.addString(S);
+  }
+
+  std::size_t getDynStrTabSize() const {
+    return DynamicStringTableContents.size();
+  }
+
+  std::optional<std::size_t> getOffsetInDynStrTab(const std::string &S) const {
```
quic-seaswara, line 72: why std::optional

quic-seaswara (approved):

include/eld/Target/ELFFileFormat.h
parth-07: This API is used during image writing step. It expects the string has already been inserted in the dynamic string table. If this is not true, then this API returns an empty optional object (and the user of this API can choose to error/assert as appropriate). It does not add the string in the dynamic string table as the counterpart function `addStringToDynStrTab`.

quic-seaswara (approved):

PR #103, opened 2025-05-22
ThinLTO cache commit

quic-akaryaki:
As a result of upstream LLVM PR #136121, users of CacheFileStream must call its commit() function before deleting the stream.

### DIFF
--- a/lib/Object/ObjectLinker.cpp
+++ b/lib/Object/ObjectLinker.cpp
@@ -2980,6 +2980,8 @@ bool ObjectLinker::doLto(llvm::lto::LTO &LTO) {
       report_fatal_error(Twine("unable to add memory buffer: ") +
                          toString(S.takeError()));
     *(*S)->OS << MB->getBuffer();
+    if (Error Err = (*S)->commit())
+      report_fatal_error(std::move(Err));
   };
 
   llvm::FileCache ThinLTOCache;
--- a/test/Common/LTO/ThinLTOCaching/ThinLTOCaching.test
+++ b/test/Common/LTO/ThinLTOCaching/ThinLTOCaching.test
@@ -1,0 +1,16 @@
+## A basic test for thin LTO cache
+
+RUN: %clang %clangopts -flto=thin -O2 -c %p/Inputs/1.c -o %t.1.bc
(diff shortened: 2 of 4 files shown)


lib/Object/ObjectLinker.cpp
```diff
@@ -2980,6 +2980,8 @@ bool ObjectLinker::doLto(llvm::lto::LTO &LTO) {
       report_fatal_error(Twine("unable to add memory buffer: ") +
                          toString(S.takeError()));
     *(*S)->OS << MB->getBuffer();
+    if (Error Err = (*S)->commit())
+      report_fatal_error(std::move(Err));
```
parth-07, line 2984: We typically use `report_fatal_error` through `ASSERT`. Does the `(*S)->commit()` failing always signify a toolchain issue? Can a user input make `(*S)->commit()` return an error?

test/Common/LTO/ThinLTOCaching/ThinLTOCaching.test
```diff
@@ -0,0 +1,13 @@
+## A basic test for thin LTO cache
+
+RUN: %clang %clangopts -flto=thin -O2 -c %p/Inputs/1.c -o %t.1.bc
+RUN: %clang %clangopts -flto=thin -O2 -c %p/Inputs/2.c -o %t.2.bc
+
+## Test --save-temps with thin LTO cache
+RUN: rm -rf cache
```
parth-07: Can we use absolute path for the cache-file here? Would %t1.cache work?

lib/Object/ObjectLinker.cpp
quic-akaryaki: The purpose of this change is to adjust to the API changes introduced by PR #136121. `report_fatal_error` is used in the LLVM code, and was used in case of errors during commit before #136121. The errors are more likely not due to user input but system error such as full disks and permissions,

test/Common/LTO/ThinLTOCaching/ThinLTOCaching.test
quic-akaryaki: Done

quic-seaswara (approved):

PR #111, opened 2025-05-27
Add `plugin::OutputSection::recomputeSize`

### LINKED ISSUE #110: `plugin::OutputSection::recomputeSize` function would be helpful
Recomputing output section size is a recurring need in many plugins. It would be beneficial for the plugin framework to provide this functionality.

parth-07:
This commit adds `plugin::OutputSection::recomputeSize` API. The recurring need for this functionality in plugins is the main motivation for adding this API.

Closes #110


include/eld/Plugin/PluginManager.h
```diff
@@ -20,6 +20,19 @@ class Plugin;
...
+    Init,
+    VisitSections,
+    VisitSymbols,
+    RuleMatching,
+    SectionMerging,
+    PerformingLayout,
+    WritingOutput,
+    Destroy
+  };
+
```
quic-seaswara: why another link state, we already have a state in LinkerWrapper ? Can we extend that ?

parth-07: I fear that we cannot use the already existing link state without breaking backward compatibility in some silent ways. The `LinkerWrapper::State` consist of `Initializing`, `BeforeLayout`, `CreatingSections`, `CreatingSegments` and `AfterLayout`.

Many plugin framework APIs use `LinkerWrapper::State`. They may all need to be modified. We also cannot change the value of `Module::LinkerWrapper::State` member because plugin types other than `LinkerPlugin` are not aware of these new link states, and many plugins have code that checks for link states -- all of these checks may need to be updated if we modify `LinkerWrapper::State`.

Also, `LinkerWrapper::State::BeforeLayout` can encompass all of `Init`, `VisitSections`, `VisitSymbol`, `RuleMatching` and `SectionMerging`.

PR #112, opened 2025-05-27
Add support for `--require-defined` option

### LINKED ISSUE #72: `--require-defined` command-line option can be helpful
`--require-defined` option can be helpful for our users and help us increasing GNU compatibility of our tools.
From GNU LD documentation:

> --require-defined=symbol
> 
> Require that symbol is defined in the output file. This option is the same as option --undefined except that if symbol is not defined in the output file then the linker will issue an error and exit.

MSMazaya:
Implement the `--require-defined` command-line option for improved GNU LD compatibility. Error if the specified symbol is not defined in the output.

Closes #72

### DIFF
--- a/include/eld/Config/GeneralOptions.h
+++ b/include/eld/Config/GeneralOptions.h
@@ -974,6 +974,13 @@ public:
 
   bool isReproduceOnFail() const { return RecordInputFilesOnFail; }
 
+  // --require-defined support
+  void addRequireDefinedSymbol(const std::string &S) { RequireDefinedSymbols.push_back(S); }
+
+  const std::vector<std::string> &getRequireDefinedSymbols() const {
+    return RequireDefinedSymbols;
+  }
+
   // -- enable relaxation on hexagon ----
   void enableRelaxation() { BRelaxation = true; }
 
@@ -1244,6 +1251,7 @@ private:
   bool ProgressBar = false;                  // Show progressbar.
   bool RecordInputFiles = false;             // --reproduce
   bool RecordInputFilesOnFail = false;       // --reproduce-on-fail
+  std::vector<std::string> RequireDefinedSymbols; // --require-defined
   // FIXME: Change the name to CompressReproduceTar
   bool CompressTar = false;         // --reproduce-compressed
   bool DisplaySummary = false;      // display linker run summary
--- a/include/eld/Diagnostics/DiagSymbolResolutions.inc
+++ b/include/eld/Diagnostics/DiagSymbolResolutions.inc
@@ -33,6 +33,8 @@ DIAG(symbol_undefined_by_user, DiagnosticEngine::Error,
      "Symbol %0 defined in %1 is undefined")
 DIAG(symbol_not_found, DiagnosticEngine::Fatal,
      "Malformed image: Symbol `%0' not found")
+DIAG(missing_required_symbol, DiagnosticEngine::Error,
+     "Required symbol `%0' not defined")
 DIAG(dynlist_symbol_undefined_by_user, DiagnosticEngine::Error,
      "Symbol %0 defined in Dynamic List  is undefined")
 DIAG(tls_non_tls_symbol_mismatch, DiagnosticEngine::Fatal, "%0")
--- a/lib/Core/Linker.cpp
+++ b/lib/Core/Linker.cpp
@@ -429,6 +429,18 @@ bool Linker::resolve() {
     return false;
   }
 
+  // Making sure that all of the required symbols from --require-defined are defined.
+  std::vector<std::string> UndefRequiredSymbols;
+  for (const std::string &S : ThisConfig->options().getRequireDefinedSymbols())
+    if (!ThisModule->getNamePool().isSymbolPresent(S))
+      UndefRequiredSymbols.push_back(S);
+  if (!UndefRequiredSymbols.empty()) {
+    for (const std::string &UndefSymbol : UndefRequiredSymbols)
+      ThisConfig->raise(Diag::missing_required_symbol) << UndefSymbol;
+    ThisModule->setFailure(true);
+    return false;
+  }
+
   {
     LinkerProgress->incrementAndDisplayProgress();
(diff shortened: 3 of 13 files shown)


include/eld/Diagnostics/DiagSymbolResolutions.inc
```diff
@@ -33,6 +33,8 @@ DIAG(symbol_undefined_by_user, DiagnosticEngine::Error,
      "Symbol %0 defined in %1 is undefined")
 DIAG(symbol_not_found, DiagnosticEngine::Fatal,
      "Malformed image: Symbol `%0' not found")
+DIAG(missing_required_symbol, DiagnosticEngine::Fatal,
```
quic-seaswara: Can we use DiagnosticEngine::Error instead of fatal  ?

lib/Core/Linker.cpp
```diff
@@ -429,6 +429,29 @@ bool Linker::resolve() {
...
+  std::vector<std::string> UndefRequiredSymbols;
+  for (const std::string &S : ThisConfig->options().getRequireDefinedSymbols()) {
+    bool found = false;
+    for (auto &G : ThisModule->getNamePool().getGlobals()) {
+      eld::ResolveInfo *SymInfo = G.getValue();
+      if (SymInfo->name() == S && !SymInfo->isUndef()) {
+        found = true;
+        break;
+      }
+    }
```
quic-seaswara: You can do a namepool lookup if the required symbol is found instead of iterating through all the global symbols.

quic-seaswara (commented):
Need tests 

* static linking 
* shared library
* what happens if you say require-defined and the symbol is an archive 
* what happens if you say require-defined and the symbol gets garbage collected ?

include/eld/Config/GeneralOptions.h
```diff
@@ -962,6 +962,13 @@ class GeneralOptions {
 
   bool isReproduceOnFail() const { return RecordInputFilesOnFail; }
 
+  // --require-defined support
+  void addRequireDefinedSymbol(const std::string & S) { RequireDefinedSymbols.push_back(S); }
```
quic-areg: clang-format `const std::string &S`

lib/Core/Linker.cpp
```diff
@@ -429,6 +429,21 @@ bool Linker::resolve() {
     return false;
   }
 
+  // Making sure that all of the required symbols from --require-defined are defined.
+  std::vector<std::string> UndefRequiredSymbols;
+  for (const std::string &S : ThisConfig->options().getRequireDefinedSymbols()) {
```
quic-areg: you can omit braces here 

https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements

lib/Core/Linker.cpp
```diff
@@ -429,6 +429,21 @@ bool Linker::resolve() {
...
+    if (!ThisModule->getNamePool().findSymbol(S)) {
+      UndefRequiredSymbols.push_back(S);
+    }
+  }
+  if (!UndefRequiredSymbols.empty()) {
+    for (const std::string &UndefSymbol : UndefRequiredSymbols) {
```
quic-areg: and here

MSMazaya: Removed.

MSMazaya:
Sure. The following is for shared library symbols:

ld:
```
$ clang -c -fpic -o 1.o ./Inputs/1.c 
$ ld -shared -o 1.so  1.o 
$ ld --require-defined foo -o 1.out 1.so
ld: warning: cannot find entry symbol _start; not setting start address
$ echo $?
0
```

eld:
```
$ clang -target arm -mfloat-abi=soft -c -fpic -o 1.o ./Inputs/1.c 
$ ld.eld  -shared -o 1.so  1.o
$ ld.eld --require-defined foo -o 1.out 1.so
Error: Required symbol `foo' not defined
Fatal: Linking had errors.
```

And this one for archived symbols:

ld:
```
$ clang -c -o 1.o ./Inputs/1.c
$ ar rcs 1.a 1.o 
$ ld -require-defined foo -o 1.out 1.a
ld: warning: cannot find entry symbol _start; defaulting to 0000000000401000
$ echo $?
0
```

eld:
```
$ clang -target arm -mfloat-abi=soft -c -o 1.o ./Inputs/1.c
$ ar rcs 1.a 1.o
$ ld.eld --require-defined foo -o 1.out 1.a
Error: Required symbol `foo' not defined
Fatal: Linking had errors.
```

With `Inputs/1.c` from this PR.

quic-seaswara:
Thank you!

MSMazaya:
One thing I have noticed too is that, it seems like unused symbols are not being garbage collected? With the same `Inputs/1.c` and `script.t` from this PR, I get the following result:

ld:
```
$ clang -c -o 1.o ./Inputs/1.c
$ ld -T script.t --gc-sections -o 1.out 1.o
$ objdump -t 1.out

1.out:     file format elf64-x86-64

SYMBOL TABLE:
no symbols

```

Whereas in eld:
```
$ clang -target arm -mfloat-abi=soft -c -o 1.o ./Inputs/1.c
$ ld.eld -T script.t --gc-sections -o 1.out 1.o
$ objdump -t 1.out

1.out:     file format elf32-little

SYMBOL TABLE:
00000000 l    d  .text	00000000 .text
00000000 l    d  .ARM.attributes	00000000 .ARM.attributes
00000000 l    d  .comment	00000000 .comment
00000008 l    d  .ARM.exidx	00000000 .ARM.exidx
00000000 l    df *ABS*	00000000 1.c
00000000 l       .text	00000000 $a.0
00000000 g     F .text	00000008 foo
00000000 g       *ABS*	00000000 __RWPI_BASE__

```

Seems like foo is still there:
```
00000000 g     F .text	00000008 foo
```

quic-seaswara:
without an entry symbol specified, we choose the first executable section as a default entry section. 

We prefer not to change this behavior, but it is worth noting the differences in our eld documentation.

PR #119, opened 2025-06-03
Do not check PHDR errors for partial link

### LINKED ISSUE #118: eld errors out when doing partial link if the linker script contains `PHDRS`
eld errors out when doing partial link if the linker script contains `PHDRS`. 

Reproducible example:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
int foo() { return 1; }

int var = 3;
EOF

cat >script.t <<\EOF
PHDRS {
  A PT_LOAD;
  B PT_LOAD;
}

SECTIONS {
  .text : { *(.text.*) } :B
  .data : { *(.data.*) }
}
EOF

clang -o 1.o 1.c -target riscv64 -c -ffunction-sections -fdata-sections
ld.eld -m elf64lriscv -o 1.out 1.o -T script.t -r
```

Output:
```
Error: Loadable section .text not in any load segment
Error: Loadable section .data not in any load segment
Fatal: Linking had errors.
```

Instead of erroring out, the linker should simply ignore `PHDRS` when doing partial link.

parth-07:
This commit fixes the incorrect PHDR-related linker error when doing partial link. The linker should ignore the PHDRS specified in the linker script when doing partial link.

Closes #118

### DIFF
--- a/lib/Target/GNULDBackend.cpp
+++ b/lib/Target/GNULDBackend.cpp
@@ -3142,7 +3142,7 @@ bool GNULDBackend::layout() {
     doPostLayout();
   }
 
-  if (m_Module.getScript().phdrsSpecified())
+  if (m_Module.getScript().phdrsSpecified() && !config().isLinkPartial())
     checkForLinkerScriptPhdrErrors();
 
   if (!config().getDiagEngine()->diagnose()) {
--- a/test/Common/standalone/linkerscript/PartialLinkWithPHDRS/Inputs/1.c
+++ b/test/Common/standalone/linkerscript/PartialLinkWithPHDRS/Inputs/1.c
@@ -1,0 +1,3 @@
+int foo() { return 1; }
+
+int var = 3;
--- a/test/Common/standalone/linkerscript/PartialLinkWithPHDRS/Inputs/script.t
+++ b/test/Common/standalone/linkerscript/PartialLinkWithPHDRS/Inputs/script.t
@@ -1,0 +1,9 @@
+PHDRS {
+  A PT_LOAD;
+  B PT_LOAD;
+}
+
+SECTIONS {
+  .text : { *(.text.*) } :B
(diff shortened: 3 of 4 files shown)


lib/Target/GNULDBackend.cpp
```diff
@@ -3152,7 +3152,7 @@ bool GNULDBackend::layout() {
     doPostLayout();
   }
 
-  if (m_Module.getScript().phdrsSpecified())
+  if (m_Module.getScript().phdrsSpecified() && !config().isLinkPartial())
```
quic-seaswara, line 3145: We should also raise a warning that PHDR's are used for partial linking with -Wlinker-script. 

Why Is it even valid to use a linker script with PHDR's and partial linking ?

parth-07: > We should also raise a warning that PHDR's are used for partial linking with -Wlinker-script.

Yes, I plan to do this in a separate patch.

> Why Is it even valid to use a linker script with PHDR's and partial linking ?

I am not sure about this. I think it is for convenience purposes. We can choose to error out here but both GNU ld and lld do not report any error for this. Should I raise an issue in LLVM for erroring out when `PHDRS` is specified with partial link?

quic-seaswara: Sure.

quic-seaswara: MEMORY/PHDRS/Section address/AT all would be warnings as well.

Dot assignments outside the section would be a warning IMO as well.

parth-07: I created an LLD issue for PHDR warning this morning: https://github.com/llvm/llvm-project/issues/142926. I will create issues for MEMORY/section address/AT/... as well depending upon their response for the PHDR warning issue.

quic-seaswara (approved):

PR #133, opened 2025-06-09
Add symbols referred by shared libs to dynamic symbol table

### LINKED ISSUE #106: Missing symbol in the dynamic symbol table
The symbols that are used in shared libraries and defined in regular object files should be emitted to the dynamic symbol table, but we
are not always emitting such symbols to the dynamic symbol table.

Reproducible example:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
int bar();
int foo() { return bar(); }
int bar() { return 3; }
EOF

cat >2.c <<\EOF
int bar() { return 5; }
EOF

cat >main.c <<\EOF
int foo();
int bar();

int baz() { return 7; }
int main() {
  return foo() + bar() + baz();
}
EOF

riscv64-unknown-linux-gnu-gcc -o 1.o 1.c -c -ffunction-sections -fPIC
riscv64-unknown-linux-gnu-gcc -o lib1.so 1.o -shared
riscv64-unknown-linux-gnu-gcc -o 2.o 2.c -c -ffunction-sections
riscv64-unknown-linux-gnu-gcc -o main.o main.c -c -ffunction-sections

LDs=("ld.eld -melf64lriscv" ld.lld riscv64-unknown-linux-gnu-ld)
SFs=("eld" lld bfd)

for i in "${!SFs[@]}"; do
  ${LDs[$i]} -o a.${SFs[$i]}.out main.o lib1.so 2.o
done
```

`llvm-readelf --dyn-syms a.eld.out` does not contain `bar`. bfd and lld artifacts does contain `bar` in the dynamic symbol table.

Interestingly enough, if we change the order of input files to `main.o 2.o lib1.so`, then eld emits `bar` to the dynamic symbol table as well. 
However, this behavior should not be input-order dependent.

parth-07:
This commit fixes the missing symbol from dynamic symbol table issue when a symbol was referred by dynamic symbol table but was resolved to a symbol by a reloctable object that was present after the shared library in the link command.

The issue is fixed by properly propagating exportToDyn property across symbols during symbol resolution.

Closes #106


test/Common/standalone/SharedLibSymRefExport/SharedLibSymRefExport.test
```diff
@@ -0,0 +1,21 @@
...
+RUN: %clang %clangopts -o %t1.1.o %p/Inputs/1.c -c -fPIC
+RUN: %link %linkopts -o %t1.lib1.so %t1.1.o -shared
+RUN: %clang %clangopts -o %t1.2.o %p/Inputs/2.c -c
+RUN: %clang %clangopts -o %t1.3.o %p/Inputs/3.c -c -fPIC
+RUN: %link %linkopts -o %t1.lib3.so %t1.3.o -shared
+RUN: %clang %clangopts -o %t1.main.o %p/Inputs/main.c -c
+RUN: %link %linkopts -o %t1.main.1.out -dy %t1.main.o %t1.lib1.so %t1.2.o
+RUN: %readelf --dyn-symbols %t1.main.1.out | %filecheck %s
+RUN: %link %linkopts -o %t1.main.2.out -dy %t1.main.o %t1.lib3.so %t1.2.o
+RUN: %readelf --dyn-symbols %t1.main.2.out | %filecheck %s
```
quic-seaswara, line 17: Does LTO preserve the symbol that needs to be exported ? Can you pls check ?

lib/SymbolResolver/NamePool.cpp
```diff
@@ -369,6 +369,8 @@ ResolveInfo NamePool::createInputSymbolRI(
   RI.setInBitCode(IF.isBitcode());
   if (IsPatchable)
     RI.setPatchable();
+  if (IsDyn && RI.canBePreemptible())
+    RI.setExportToDyn();
```
quic-seaswara, line 373: what is done for weak symbols and undefined weak symbols ?

test/Common/standalone/SharedLibSymRefExport/SharedLibSymRefExport.test
parth-07: Yes, LTO seems to preserve the symbol that needs to be exported. I have added a test with LTO as well.

lib/SymbolResolver/NamePool.cpp
parth-07: The behavior will be the same as strong symbol in the shared library in this case. If the symbol library contains a weak or an undefined weak symbol, and one of the object files also defines the same symbol, then the dynamic symbol table will contain the symbol. I have modified the test to check for the weak symbol case as well.

quic-seaswara (approved):

PR #135, opened 2025-06-09
Move LinkerWrapper::State to Module

### LINKED ISSUE #134: `LinkerWrapper::State` should not be exposed to the users
`LinkerWrapper::State` needs to be modified whenever a new state is created. Any modification to `LinkerWrapper::State` may result in ABI-incompatibility and all the plugins must be recompiled. It would be helpful to hide `LinkerWrapper::State` and expose `LinkerWrapper` member functions to expose the link state to the plugins. An additional benefit of this approach is that it would force users to specify the exact link state instead of using less-than/greater-than comparison of link states that can be error prone.

parth-07:
This commit moves LinkerWrapper::State to Module::LinkState and adds LinkerWrapper member fucntions for the plugins to query the link state. The main motivation for this change is to increase the ABI-compatibility robustness of the plugin framework. Previously, Any modification to LinkerWrapper::State could result in ABI-incompatibility. Now link states can be added without affecting the ABI-compatibility.

An added additional benefit is that the new design forces users to specify the exact link state that they need instead of using the error-prone less-than/greater-than link state comparisons.

Closes #134

### DIFF
--- a/include/eld/Core/Module.h
+++ b/include/eld/Core/Module.h
@@ -138,6 +138,33 @@ public:
 
   typedef std::pair<uint64_t, uint64_t> DynamicListStartEndIndexPair;
 
+  /// Stages of the link process.
+  /// What actions a plugin can perform depends on the
+  /// link state. For example:
+  /// - Plugins can change the output section of an input
+  ///   section only in BeforeLayout link state.
+  /// - Plugins can move chunks from one output section to
+  ///   another only in CreatingSections link state.
+  /// - Plugins can only compute the output image layout checkum using
+  ///   `LinkerWrapper::getImageLayoutChecksum` only in AfterLayout linker
+  ///   state.
+  /// - Plugins can reassign virtual addresses using
+  ///   `LinkerWrapper::reassignVirtualAddresses()` only in CreatingSegments
+  ///   link state.
+  ///
+  /// Plugin authors should ensure that the action being performed by the
+  /// plugin is meaningful in the link state in which it is executed.
+  /// Executing invalid actions for a link state can result in undefined
+  /// behavior.
+  enum LinkState : uint8_t {
+    Unknown,
+    Initializing,
+    BeforeLayout,
+    CreatingSections,
+    CreatingSegments,
+    AfterLayout,
+  };
+
 public:
   explicit Module(LinkerScript &CurScript, LinkerConfig &Config,
                   LayoutInfo *LayoutInfo);
@@ -418,9 +445,9 @@ public:
   /// On updating the state, this function also checks linker invariants for
   /// the state. It returns true if all the invariants are true; Otherwise it
   /// returns false.
-  bool setState(plugin::LinkerWrapper::State S);
+  bool setLinkState(LinkState S);
 
-  plugin::LinkerWrapper::State getState() const { return State; }
+  LinkState getState() const { return State; }
 
   llvm::StringRef getStateStr() const;
 
@@ -607,8 +634,16 @@ public:
     return VersionScripts;
   }
 
-  bool isBeforeLayoutState() const {
-    return getState() == plugin::LinkerWrapper::State::BeforeLayout;
+  bool isLinkStateBeforeLayout() const {
+    return getState() == Module::LinkState::BeforeLayout;
+  }
+
+  bool isLinkStateCreatingSections() const {
+    return getState() == Module::LinkState::CreatingSections;
+  }
+
+  bool isLinkStateAfterLayout() const {
+    return getState() == Module::LinkState::AfterLayout;
   }
 
   void setFragmentPaddingValue(Fragment *F, uint64_t V);
@@ -632,6 +667,8 @@ private:
 
   // Read one plugin config file
   bool readOnePluginConfig(llvm::StringRef Cfg, bool IsDefaultConfig);
+
+
 
 private:
   LinkerScript &UserLinkerScript;
@@ -664,7 +701,7 @@ private:
   LayoutInfo *ThisLayoutInfo = nullptr;
   bool Failure = false;
   bool UsesLto = false;
-  plugin::LinkerWrapper::State State = plugin::LinkerWrapper::Unknown;
+  LinkState State = LinkState::Unknown;
   ReplaceFragsVectorT ReplaceFrags;
   PluginDataMapT PluginDataMap;
   eld::OutputTarWriter *OutputTar = nullptr;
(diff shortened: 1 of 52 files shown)


include/eld/Core/Module.h
```diff
@@ -418,9 +446,9 @@ class Module {
   /// On updating the state, this function also checks linker invariants for
   /// the state. It returns true if all the invariants are true; Otherwise it
   /// returns false.
-  bool setState(plugin::LinkerWrapper::State S);
+  bool setState(LinkState S);
```
quic-seaswara: setLinkState

include/eld/PluginAPI/LinkerWrapper.h
```diff
@@ -808,6 +779,16 @@ class DLL_A_EXPORT LinkerWrapper {
   eld::Expected<bool> doesRuleMatchWithSection(const LinkerScriptRule &R,
                                                const Section &S,
                                                bool doNotUseRMName = false);
+  bool isBeforeLayoutLinkState() const;
```
quic-seaswara: Probably isLinkStateBeforeLayout() sounds better ?

include/eld/PluginAPI/LinkerWrapper.h
```diff
@@ -828,8 +809,11 @@ class DLL_A_EXPORT LinkerWrapper {
 private:
   eld::Module &m_Module;
   eld::Plugin *m_Plugin;
+
   eld::DiagnosticEngine *m_DiagEngine;
   friend class LayoutWrapper;
+  friend struct OutputSection;
+  friend struct Segment;
```
quic-seaswara: why introducing new friends ?

lib/Core/Module.cpp
```diff
@@ -854,10 +854,10 @@ bool Module::verifyInvariantsForCreatingSectionsState() const {
   return false;
 }
 
-bool Module::setState(plugin::LinkerWrapper::State S) {
+bool Module::setState(LinkState S) {
   bool Verification = true;
-  if (State == plugin::LinkerWrapper::State::BeforeLayout &&
-      S == plugin::LinkerWrapper::State::CreatingSections)
+  if (State == LinkState::BeforeLayout &&
+      S == LinkState::CreatingSections)
```
quic-seaswara: convert this to functions as well ?

lib/Core/Module.cpp
```diff
@@ -792,7 +792,7 @@ bool Module::resetSymbol(ResolveInfo *R, Fragment *F) {
 
 uint64_t Module::getImageLayoutChecksum() const {
   uint64_t Hash = 0;
-  if (State != plugin::LinkerWrapper::AfterLayout)
+  if (State != LinkState::AfterLayout)
```
quic-seaswara: convert to functions

lib/LinkerWrapper/LinkerWrapper.cpp
```diff
@@ -318,7 +319,7 @@ eld::Expected<void> LinkerWrapper::resetOffset(OutputSection O) {
 
 eld::Expected<std::pair<OutputSection, LinkerScriptRule>>
 LinkerWrapper::getOutputSectionAndRule(Section S) {
-  if (getState() < State::CreatingSections)
+  if (getLinkState() < Module::LinkState::CreatingSections)
     RETURN_INVALID_LINK_STATE_ERR(
```
quic-seaswara: Same here, no less than or greater than conditions.

include/eld/Core/Module.h
parth-07: Done.

include/eld/PluginAPI/LinkerWrapper.h
parth-07: We are using `LinkerWrapper::getLinkState` function in these friend classes, this function is marked private now.

include/eld/PluginAPI/LinkerWrapper.h
parth-07: I have modified all such functions to follow the isLinkState<StateName> pattern.

lib/Core/Module.cpp
parth-07: Is there any issue on directly referring to enums in the linker core? 

We cannot directly use a function for `S == LinkState::CreatingSections` expression because the link state has not been changed at the time we are doing this comparison (later in this function, we set the link state to `S`).

lib/Core/Module.cpp
parth-07: Is there any benefit for changing all enum comparisons to functions inside the linker core codebase? This patch only adds 'isLinkState<StateName>' functions for the LinkerWrapper. We cannot access those functions in the linker core.

lib/LinkerWrapper/LinkerWrapper.cpp
parth-07: In `LinkerWrapper` functions, using comparison like this is very helpful, less verbose and less error-prone to future modifications to the link states. For example, if a function works for the link states `CreatingSections`, `CreatingSegments` and the `AfterLayout`, then the current link state check-expression would look like this:

(1)
```cpp
if (getState() >= Module::LinkState::CreatingSections) {
  ...
}
```

If we change it to use link state query functions,  then the expression would transform to:

(2)
```cpp
if (getLinker()->isLinkStateCreatingSections() || getLinker()->isLinkStateCreatingSegments() 
      || getLinker()->isLinkStateAfterLayout()) {
  ...
} 
```

(2) is significantly more verbose than (1), and less robust as well, for example, if we later add a new link state `S` between `CreatingSegments` and `AfterLayout`, then (1) will work as it is but (2) would need to updated for the new link state as well.

include/eld/PluginAPI/LinkerWrapper.h
parth-07: I have removed the friends as per our discussion.

lib/Core/Module.cpp
parth-07: I have removed `State == LinkState::Beforelayout` as I realized that this check was redundant. We cannot convert `S == LinkState::CreatingSections` because `S` is an argument instead of the current link state.

lib/LinkerWrapper/LinkerWrapper.cpp
parth-07: I have removed all less-than / greater-than conditions involving link state.

lib/Core/Module.cpp
parth-07: Done. I have changed this to use `Module::isLinkStateAfterLayout`

quic-seaswara (approved):

PR #137, opened 2025-06-09
Annotate autoloaded plugins in the text map-file

### LINKED ISSUE #136: Map-file should contain an annotation for autoloaded plugins
Map-file should contain an annotation for autoloaded plugins so that it is easy to recognize which plugins were loaded by the linker and which were specified by the user.

parth-07:
This commit adds an annotation for autoloaded plugins in the text map-file. The main motivation for this map-file addition is to be able to easily tell which plugins were loaded automatically by the linker and which were loaded by the user.

This commit also adds a dummy HelloWorldPlugin. This serves two purpose: it provides a skeleton linker plugin for the users to start with and to provide a plugin for testing the autoload annotation in the map-file.

Closes #136

### DIFF
--- a/include/eld/Script/Plugin.h
+++ b/include/eld/Script/Plugin.h
@@ -26,7 +26,8 @@ class Module;
 class Plugin {
 public:
   explicit Plugin(plugin::Plugin::Type T, std::string Name, std::string R,
-                  std::string O, bool Stats, Module &Module);
+                  std::string O, bool Stats, bool DefaultPlugin,
+                  Module &Module);
 
   ~Plugin();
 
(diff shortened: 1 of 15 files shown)


include/eld/Script/Plugin.h
```diff
@@ -213,6 +214,8 @@ class Plugin {
   /// Calls ActBeforeRuleMatching hook handler.
   void callActBeforeRuleMatchingHook();
 
+  bool isDefaultPlugin() const { return IsDefaultPlugin; }
+
```
quic-seaswara, line 218: Do we need to change the major version ? 

This breaks backwards compatibility.

quic-seaswara (approved):

PR #139, opened 2025-06-10
[RISCV] Handle VENDOR Relocations with emit-relocs

lenary:
Before this patch, the output of `--emit-relocs` and partial linking
when using `R_RISCV_VENDOR` relocations was corrupted. This is because
we use an internal relocation numbering scheme in `handleRelocation`
when parsing the relocation, and there was no logic to map this back to
the public relocation number.

To map the internal relocation numbers back to their external relocation
numbers, we add `GNULDBackend::remapInternalRelocationType` with a
default implementation that returns the input value. If this changes
the type of the relocation, a verbose log message is printed.

On RISC-V this hook:
- Returns standard (<= 255) codes as is
- Maps QUALCOMM internal relocations back to `R_RISCV_CUSTOM<N>`
- Maps Relaxation internal relocations to `R_RISCV_NONE`


lib/Object/ObjectLinker.cpp
```diff
@@ -2168,6 +2168,8 @@ bool ObjectLinker::relocation(bool EmitRelocs) {
         R->setAddend(Relocation->symOffset(*ThisModule));
       }
     }
+
+    R->setType(ThisBackend.remapInternalRelocationType(R->type()));
```
quic-seaswara: What is done with partial linking ?

include/eld/Target/GNULDBackend.h
```diff
@@ -701,6 +701,14 @@ class GNULDBackend {
 
   virtual bool shouldIgnoreRelocSync(Relocation *Reloc) const { return false; }
 
+  // Sometimes an internal value is stored in Relocation::Type (RISC-V
+  // Vendor/Internal relocations), This maps the internal relocation type back
+  // to the external relocation type it came from, for emit-relocs.
+  virtual Relocation::Type
+  remapInternalRelocationType(Relocation::Type t) const {
```
quic-seaswara: getRemapInternalRelocationType

lib/Object/ObjectLinker.cpp
quic-seaswara: It might make sense to change line 2158 to ThisBackend.getRelocationType(Relocation->type())

quic-seaswara: Do we want to add a verbose message to use it for diagnostics for tracking ?

lib/Object/ObjectLinker.cpp
```diff
@@ -2154,8 +2154,18 @@ bool ObjectLinker::relocation(bool EmitRelocs) {
       RelocCount[OutputSect] = 0;
       MaxSectSize[OutputSect] = OutputSect->size();
     }
+
+    Relocation::Type ExtType = ThisBackend.getRemappedInternalRelocationType(Relocation->type());
+
+    if (ExtType != Relocation->type() && ThisModule->getPrinter()->isVerbose()) {
+      ThisConfig.raise(Diag::remapped_internal_reloc)
+        << ThisBackend.getRelocator()->getName(Relocation->type())
+        << ExtType
```
quic-seaswara: Can this be a string instead of a number ? makes it easy to figure out to see to what reloc it was changed to

include/eld/Diagnostics/DiagRelocations.inc
```diff
@@ -70,3 +70,5 @@ DIAG(unsupported_reloc, DiagnosticEngine::Fatal,
 DIAG(resolve_undef_weak_guard, DiagnosticEngine::Warning,
      "Undefined weak symbol %0 referred from %1 is resolved to %2")
 DIAG(error_relocations_plugin, DiagnosticEngine::Error, "unable to apply relocations")
+DIAG(remapped_internal_reloc, DiagnosticEngine::Verbose,
```
quic-seaswara: verbose_remapped_internal_reloc

quic-seaswara (changes requested):
Can we also add a test for partial linking (-r)

lib/Object/ObjectLinker.cpp
lenary: I originally wanted to do this, but `getName` on the custom relocs doesn't give the `R_RISCV_CUSTOM<N>` name. I will try to make it do so but we might end up with a bunch of boilerplate code that doesn't do much.

lib/Object/ObjectLinker.cpp
lenary: Added.

include/eld/Target/GNULDBackend.h
lenary: Done

include/eld/Diagnostics/DiagRelocations.inc
lenary: Done

lib/Object/ObjectLinker.cpp
lenary: Done

quic-akaryaki (approved):
LG

PR #140, opened 2025-06-10
[RISCV] Implement Relaxations for QC.E.J/QC.E.JAL

lenary:
These relaxations are like call relaxations, so they are done in the same phase as call relaxations.

The Relaxations are:
- QC.E.J to C.J or JAL X0
- QC.E.JAL to C.JAL or JAL X1

This change also adds a command-line switch for turning off all the xqci relaxations.

These relaxations are described in http://github.com/quic/riscv-elf-psabi-quic-extensions/releases/latest

This is the first commit to implement #147 so includes adding the flag as well.

### DIFF
--- a/lib/Target/RISCV/RISCVLDBackend.cpp
+++ b/lib/Target/RISCV/RISCVLDBackend.cpp
@@ -299,6 +299,75 @@ bool RISCVLDBackend::doRelaxationCall(Relocation *reloc, bool DoCompressed) {
   return true;
 }
 
+bool RISCVLDBackend::doRelaxationQCCall(Relocation *reloc, bool DoCompressed) {
+  // This function performs the relaxation to replace: QC.E.JAL or QC.E.J with
+  // one of JAL, C.J, or C.JAL.
+
+  Fragment *frag = reloc->targetRef()->frag();
+  RegionFragmentEx *region = llvm::dyn_cast<RegionFragmentEx>(frag);
+  if (!region)
+    return true;
+  uint64_t offset = reloc->targetRef()->offset();
+
+  // extract instruction
+  uint64_t qc_e_jump = reloc->target() & 0xffffffffffff;
+  bool isTailCall = (qc_e_jump & 0xf1f07f) == 0x00401f;
+
+  Relocator::DWord S = getSymbolValuePLT(*reloc);
+  Relocator::DWord A = reloc->addend();
+  Relocator::DWord P = reloc->place(m_Module);
+  Relocator::DWord X = S + A - P;
+
+  bool canRelaxXqci =
+      config().targets().is32Bits() && config().options().getRISCVRelaxXqci();
+  bool canRelax =
+      config().options().getRISCVRelax() && canRelaxXqci && llvm::isInt<21>(X);
+  bool canCompress = DoCompressed && llvm::isInt<12>(X);
+
+  if (!canRelax) {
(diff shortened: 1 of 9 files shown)


test/RISCV/standalone/Relaxation/QC_E_CALL_TO_JAL/QC_E_CALL_TO_JAL.test
```diff
@@ -0,0 +1,116 @@
+#----------AUIPC_JALR_TO_JAL.test----------------- Executable------------------#
```
quic-seaswara: nit : change test name

test/RISCV/standalone/Relaxation/QC_E_CALL_TO_JAL/QC_E_CALL_TO_JAL_PLT.test
```diff
@@ -0,0 +1,13 @@
+#----------AUIPC_JALR_TO_JAL_PLT.test-------------------------------#
+#BEGIN_COMMENT
+# Check that offset to PLT is used when relaxing.
+#END_COMMENT
+#--------------------------------------------------------------------
+RUN: %clang %clangopts -c %p/Inputs/shared.s -o %t.shared.o -menable-experimental-extensions -march=rv32gc_xqcilb0p2
+RUN: %link %linkopts --relax-xqci -shared %t.shared.o -o %t.so --section-start=.text=0x1000 --section-start=.plt=0x20000
```
quic-seaswara, line 8: Do we support building shared objects with xqci ?

test/RISCV/standalone/Relaxation/QC_E_CALL_TO_JAL/QC_E_CALL_TO_JAL_PLT.test
```diff
@@ -0,0 +1,13 @@
+#----------AUIPC_JALR_TO_JAL_PLT.test-------------------------------#
```
quic-seaswara: nit : test name change

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -759,6 +829,11 @@ void RISCVLDBackend::mayBeRelax(int relaxation_pass, bool &pFinished) {
               pFinished = false;
           break;
         }
+        case ELF::riscv::internal::R_RISCV_QC_E_JUMP_PLT: {
+          if (nextRelax && relaxation_pass == RELAXATION_CALL)
```
quic-seaswara, line 832: RELAXATION_QC_CALL ?

test/RISCV/standalone/Relaxation/QC_E_CALL_TO_JAL/QC_E_CALL_TO_JAL_PLT.test
lenary: xqci only adds static relocations, and the call-like jumps (QC.E.J and QC.E.JAL) have a relocation that can introduce a PLT if required, so I don't see why this shouldn't be compatible with shared objects?

lib/Target/RISCV/RISCVLDBackend.cpp
lenary: I was originally going to use a different phase, but I decided to instead use the same phase as CALLs, as the commit message notes. I'll remove the enum value I added.

quic-seaswara (approved):

PR #154, opened 2025-06-18
Return true when symbol table section is missing

bhoumik03:
This change ensures the linker does not fail when the symbol table section is absent, which is valid for some ELF inputs.


parth-07 (commented):
As this is a non-x86 change, can we please add the test to Common/standalone directory?

test/Common/standalone/MissingSymTab/MissingSymTab.test
```diff
@@ -0,0 +1,21 @@
+#----x86_64_MissingSymTab.test-----------#
+
```
partaror, line 2: Please remove redundant new lines in the test everywhere.

test/Common/standalone/MissingSymTab/MissingSymTab.test
```diff
@@ -0,0 +1,21 @@
+#----x86_64_MissingSymTab.test-----------#
```
partaror: Remove x86_64 from the test name.

test/Common/standalone/MissingSymTab/MissingSymTab.test
```diff
@@ -0,0 +1,21 @@
...
+# Test ELD handling of object files with missing symbol table sections.
+# The readSymbols() function should return true when the symbol table is missing.
+END_COMMENT
+
+#START_TEST
+
+RUN: %clang %clangopts -c %p/Inputs/1.c -o %t.1.o -ffunction-sections
+RUN: %strip %t.1.o
+
+RUN: %readelf -S %t.1.o | %filecheck %s --check-prefix=NO-SYMTAB
```
partaror: Please change `NO-SYMTAB` to `SYMTAB` for brevity.

test/Common/standalone/MissingSymTab/MissingSymTab.test
```diff
@@ -0,0 +1,16 @@
...
+END_COMMENT
+
+#START_TEST
+RUN: %clang %clangopts -c %p/Inputs/1.c -o %t.1.o -ffunction-sections
+RUN: %strip %t.1.o
+RUN: %readelf -S %t.1.o | %filecheck %s --check-prefix=SYMTAB
+RUN: %link %linkopts -o %t.1.eld.out %t.1.o
+
+SYMTAB-NOT: .symtab
+LINK-SUCCESS: 0
```
quic-seaswara, line 15: What is testing for LINK-SUCCESS ?

lib/Readers/ELFReader.cpp
```diff
@@ -295,7 +295,7 @@ template <class ELFT> eld::Expected<bool> ELFReader<ELFT>::readSymbols() {
            ? llvm::ELF::SHT_DYNSYM
            : llvm::ELF::SHT_SYMTAB));
   if (!symTabSec)
-    return false;
+    return true;
```
quic-seaswara, line 298: Does it behave the same way if dynsym is not present ?

bhoumik03: It is okay to return true. lld and bfd does not error when a shared library does not contain a dynamic symbol table and the link does not require anything from the shared library.

PR #158, opened 2025-06-18
[ci] cancel current running actions for PR

quic-seaswara:
automatically cancel current running github actions when a new commit is uploaded

https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow


.github/workflows/ci.yml
```diff
@@ -1,6 +1,16 @@
...
+on:
+  pull_request:
+    paths-ignore:
+      - '**.md'
+      - 'docs/**'
+
+# This is what will cancel the workflow
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
navaneethshan: why this is needed? Changing false to true in line 15 will do the same.

navaneethshan: * line 20 below.

navaneethshan (approved):

PR #161, opened 2025-06-18
[ci] improve ci workflow

quic-seaswara:
improve ci workflow by skipping files that dont require a build

https://github.com/qualcomm/eld/pull/158 was creating emails that show that github eld actions fail, and this would take care of not having those emails sent.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,25 +1,25 @@
 name: CI
 
-# dont start the job if there are changes only to .md files or docs
 on:
   push:
-    paths-ignore:
-      - '**.md'
-      - 'docs/**'
   pull_request:
-    paths-ignore:
-      - '**.md'
-      - 'docs/**'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
 
 jobs:
   build:
     runs-on: self-hosted
-    concurrency:
-      group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,25 +1,25 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
 
 jobs:
   build:
     runs-on: self-hosted
-    concurrency:
-      group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
-      cancel-in-progress: true
```
navaneethshan, line 19: Didn't this work?

.github/workflows/ci.yml
```diff
@@ -1,25 +1,25 @@
...
     permissions:
       contents: write
       statuses: write
+    if: github.event_name == 'push' || github.event_name == 'pull_request'
```
navaneethshan, line 17: Bringing this filter here will still start the workflow, but job will not run, is that what you need? may be to satisfy the merge condition?

quic-seaswara: Yes

.github/workflows/ci.yml
quic-seaswara: Yes I moved it it up to make the build conditions easy to control

quic-seaswara:
This avoids the mails from ci system that no jobs need to run. It shows up as a failure.

navaneethshan (approved):

navaneethshan (approved):

PR #170, opened 2025-06-20
ci: improve cache key granularity and restore strategy

navaneethshan:
- Use PR number and run ID in cache key to avoid collisions
- Add restore-keys fallback to reuse cache from same PR or base branch
- Enable cancel-in-progress to reduce redundant builds


parth-07:
I am a little confused about the cache-usage. Can you please answer the below doubt to help me better understand the cache usage?

> Use PR number and run ID in cache key to avoid collisions
> Add restore-keys fallback to reuse cache from same PR or base branch

If the builder builds PR 12 and then builds PR 18, then will the cache generated during PR 12 be usable for PR 18 build?

quic-seaswara:
@navaneethshan can you please update on this ?

.github/workflows/ci.yml
```diff
@@ -96,7 +98,7 @@ jobs:
       uses: actions/cache@v4
       with:
         path: /local/mnt/workspace/ccache/${{ env.BASE_BRANCH_NAME }}
-        key: ${{ runner.os }}-ccache-${{ github.sha }}
+        key: ${{ runner.os }}-ccache-${{ github.event.pull_request.number }}-${{ github.run_id }}
```
quic-seaswara, line 101: I am trying to understand this change.

The key to ccache is the pull request number, independent pull requests will not get the previously compiled ccache.

navaneethshan: While saving the cache, saving it with more granular info, because I did see the conflict if we save with generic name. ( like ${{ runner.os }}-ccache)
But while retriving, we use fall-backs upto "${{ runner.os }}-ccache-"  (line 30), so the cache created for a PR can be used for any other PR's build.

PR #181, opened 2025-06-23
Fix `getAllOutputSections()` crash when used in `ActBeforePerformingLayout` hook

quic-rb10:
This commit aims to fix the crash that occurs when `getAllOutputSection()` is called in the `ActBeforePerformingLayout` hook. The crash occurs due to an errornous check in `getAllOutputSections()` where it currently emits an error diagnostic when called before `AfterLayout` state of the linker.


lib/LinkerWrapper/LinkerWrapper.cpp
```diff
@@ -970,8 +970,9 @@ bool LinkerWrapper::isVerbose() const {
 
 eld::Expected<std::vector<plugin::OutputSection>>
 LinkerWrapper::getAllOutputSections() const {
-  if (getState() < State::AfterLayout)
-    RETURN_INVALID_LINK_STATE_ERR("CreatingSegments, AfterLayout");
+  if (getState() < State::CreatingSections)
+    RETURN_INVALID_LINK_STATE_ERR(
+        "CreatingSegments, BeforeLayout, AfterLayout, CreatingSections");
```
partaror: Is `BeforeLayout` right here? It happens _before_ `CreatingSections`.

quic-rb10: Yes `beforelayout` shouldn't be here have removed it.

partaror:
Can you please add the `Closes #<issue-number>` with the PR description and the commit message?

quic-seaswara (approved):

PR #182, opened 2025-06-23
Fix DT_NEEDED entry for shared libs without SO name

### LINKED ISSUE #180: `DT_NEEDED` should have the shared library path if the library has no SONAME.
`DT_NEEDED` field should have the shared library path, as is passed to the linker, if the library does not have SONAME. 

Reproducible example:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
int foo() { return 1; }
EOF

cat >main.c <<\EOF
int foo();

int main() { return foo(); }
EOF

mkdir -p a
cd a
clang -target riscv64 -o 1.o ../1.c -c -fPIC
ld.eld -m elf64lriscv -o lib1.so 1.o -shared

cd ..

clang -target riscv64 -o main.o main.c -c
ld.eld -melf64lriscv -o main.eld.out main.o a/lib1.so
llvm-readelf -d main.eld.out
# has  (NEEDED)    Shared library: [lib1.so]
# should have: (NEEDED)   Shared library: [a/lib1.so]
```

### LINKED ISSUE #184: GNU-incompatibility in DT_NEEDED entry for `:<lib>` namespec library
We have a GNU-incompatibility in DT_NEEDED entry for `:<lib>` namespec library.

Reproducible example:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
int foo() { return 1; }
EOF

cat >main.c <<\EOF
int foo();

int main() { return foo(); }
EOF

mkdir -p a
cd a
clang -target riscv64 -o 1.o ../1.c -c -fPIC
ld.eld -m elf64lriscv -o lib1.so 1.o -shared
cd ..

clang -target riscv64 -o main.o main.c -c
ld.eld -melf64lriscv -o main.eld.out main.o -L $(pwd) -l:a/lib1.so
riscv64-unknown-linux-gnu-ld -o main.bfd.out main.o -L $(pwd) -l:a/lib1.so

# Emits NEEDED lib1.so
llvm-readelf -d main.eld.out 

# Emits NEEDED a/lib1.so
llvm-readelf -d main.bfd.out
```

partaror:
This commit fixes DT_NEEDED entry value for shared libs without SO name. We were emitting the shared library basename as DT_NEEDED entry when SO name was not available. We should instead emit the shared library path that is passed to the linker.

Closes #180
Closes #184

### DIFF
--- a/lib/Input/ELFDynObjectFile.cpp
+++ b/lib/Input/ELFDynObjectFile.cpp
@@ -20,3 +20,13 @@ bool ELFDynObjectFile::isELFNeeded() {
     return true;
   return isUsed();
 }
+
+std::string ELFDynObjectFile::getFallbackSOName() const {
+  Input *I = getInput();
+  if (!getInput()->isNamespec())
+    return I->getResolvedPath().native();
+  const auto &filename = I->getFileName();
+  if (filename[0] == ':')
+    return filename.substr(1);
+  return I->getResolvedPath().filename().native();
+}
--- a/include/eld/Input/ELFDynObjectFile.h
+++ b/include/eld/Input/ELFDynObjectFile.h
@@ -33,6 +33,8 @@ public:
 
   virtual ~ELFDynObjectFile() {}
 
+  std::string getFallbackSOName() const;
+
 private:
   std::vector<ELFSection *> Sections;
 };
--- a/lib/Readers/DynamicELFReader.cpp
+++ b/lib/Readers/DynamicELFReader.cpp
@@ -116,9 +116,7 @@ eld::Expected<bool> DynamicELFReader<ELFT>::readDynamic() {
 
   // if there is no SONAME in .dynamic, then set it from input path
   if (!hasSOName)
-    dynObjFile->setSOName(
-        this->m_InputFile.getInput()->getResolvedPath().filename().native());
-
+    dynObjFile->setSOName(dynObjFile->getFallbackSOName());
   return true;
 }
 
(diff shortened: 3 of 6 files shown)


lib/Input/ELFDynObjectFile.cpp
```diff
@@ -20,3 +20,15 @@ bool ELFDynObjectFile::isELFNeeded() {
     return true;
   return isUsed();
 }
+
+std::string ELFDynObjectFile::getFallbackSOName() const {
+  if (getInput()->getInputType() == Input::InputType::Namespec) {
```
quic-seaswara: add a function isNameSpec, instead of this comparison.

Also use early returns.

lib/Input/ELFDynObjectFile.cpp
```diff
@@ -20,3 +20,15 @@ bool ELFDynObjectFile::isELFNeeded() {
...
+    Input *I = getInput();
+    const auto &filename = I->getFileName();
+    // FIXME: Should filename.empty() be an assert instead?
+    if (!filename.empty() && filename[0] == ':')
```
quic-seaswara: can filename be empty ?

quic-seaswara (changes requested):

lib/Input/ELFDynObjectFile.cpp
partaror: Done.

lib/Input/ELFDynObjectFile.cpp
partaror: It seems that filename should never be empty. I have remove the `!filename.empty()` check.

quic-seaswara (approved):

PR #185, opened 2025-06-23
add multiple tls segments support

### LINKED ISSUE #175: Relocation overflow errors when first PT_TLS segment is empty
**Description**
eld is producing (I think) incorrect relocation overflow errors in the reproducer below.

I think the important condition is whether the first `PT_TLS` segment (`tls_init` below) is empty--in the example below, eld does **not** error if any of the below happen:
- The `tls_init` PHDR is omitted (and ex: `.tdata` is commented out)
- The empty `tls_init PT_TLS;` is moved after `tls PT_TLS;`
- There is data in `tls_init` (ex: creating a variable that ends up in `.tdata`)

This was seen when building picolibc for aarch64 (not sure if this can happen with other targets).

**Reproducer**
```
cat > test.c << '!'
__thread int a;

  return a;
}
!

cat > script.t << '!'
PHDRS {
  flash PT_LOAD;
  ram_init PT_LOAD;
  ram PT_LOAD;
  tls_init PT_TLS;
  tls PT_TLS;
}

MEMORY {
  flash (rx!w) : ORIGIN = 0x40000000, LENGTH = 0x100
  ram (w!rx)   : ORIGIN = 0x40000200, LENGTH = 0x100
}

SECTIONS {
  .text : { *(.text) } >flash AT>flash :flash
  .tdata : { *(.tdata) } >ram AT>flash :tls_init :ram_init
  .tbss (NOLOAD) : { *(.tbss) } >ram AT>ram :tls :ram
}
!

clang --target=aarch64-none-elf -c test.c
# ld and lld seem ok with this, though lld's output looks a bit odd to me
ld.lld -m aarch64elf test.o -T script.t
aarch64-linux-gnu-ld.bfd -m aarch64elf test.o -T script.t

# eld errors: 
# Error: Relocation overflow when applying relocation `R_AARCH64_TLSLE_ADD_TPREL_HI12' for symbol `a' referred from test.o[.text] symbol defined in test.o[.tbss]
# Error: Relocation overflow when applying relocation `R_AARCH64_TLSLE_ADD_TPREL_LO12_NC' for symbol `a' referred from test.o[.text] symbol defined in test.o[.tbss]
ld.eld -m aarch64elf test.o -T script.t
```

quic-seaswara:
Though its not common with standard loaders that support multiple TLS segments, picolibc introduces multiple TLS segments. This is supported on GNU linker and lld, so support this with eld as well.

Resolves #175


include/eld/Diagnostics/DiagLDScript.inc
```diff
@@ -125,4 +125,6 @@ DIAG(error_non_power_of_2_value_to_align_output_section, DiagnosticEngine::Error
      "be 0 or a power of 2")
 DIAG(warn_linker_script, DiagnosticEngine::Warning, "%0")
 DIAG(warn_memory_region_has_zero_size, DiagnosticEngine::Warning,
-     "%0: Memory region '%1' has zero size")
\ No newline at end of file
+     "%0: Memory region '%1' has zero size")
+DIAG(error_empty_pt_tls_segment, DiagnosticEngine::Error,
+      " TLS segments are empty")
```
partaror:
```suggestion
      "TLS segments are empty")
```

lib/Target/GNULDBackend.cpp
```diff
@@ -649,11 +649,24 @@ uint64_t GNULDBackend::finalizeTLSSymbol(LDSymbol *pSymbol) {
...
-  if (!tls_seg) {
+  std::vector<ELFSegment *> tls_segs =
+      elfSegmentTable().getSegments(llvm::ELF::PT_TLS);
+  if (!tls_segs.size()) {
     config().raise(Diag::no_pt_tls_segment);
     return false;
   }
+  ELFSegment *tls_seg = nullptr;
+  for (auto &seg : tls_segs) {
+    if (seg->size()) {
```
partaror, line 660: Can multiple TLS segments have non-zero size?

test/Common/standalone/linkerscript/TLS/MultipleTLSSegments/MultipleTLSSegments.test
```diff
@@ -0,0 +1,10 @@
+UNSUPPORTED:arm
+#---MultipleTLSSegments.test--------------------- Executable,LS------------------#
+#BEGIN_COMMENT
+# This tests that the linker can handle multiple TLS segments.
```
partaror, line 4: Should the test check for multiple TLS segments in the output? Currently, we are checking just for `.tbss` section.

test/Common/standalone/linkerscript/TLS/MultipleTLSSegments/MultipleTLSSegments.test
```diff
@@ -0,0 +1,10 @@
+UNSUPPORTED:arm
```
partaror, line 1: Why is the test unsupported for ARM?

lib/Target/GNULDBackend.cpp
quic-seaswara: Yes, you can have two TLS segments and not have any content associated with it.

test/Common/standalone/linkerscript/TLS/MultipleTLSSegments/MultipleTLSSegments.test
quic-seaswara: Yes, thanks for the suggestion

test/Common/standalone/linkerscript/TLS/MultipleTLSSegments/MultipleTLSSegments.test
quic-seaswara: ARM fails with a undefined symbol required for TLS support

PR #195, opened 2025-06-28
Fix x86_64 default image base

bhoumik03:
Change the default image base from 0x0 to 0x400000 for x86_64 as x86_64 cannot load executables at address 0x0.
Add test case to verify the default image base is 0x400000 now.

### DIFF
--- a/lib/Target/x86_64/x86_64StandaloneInfo.h
+++ b/lib/Target/x86_64/x86_64StandaloneInfo.h
@@ -19,7 +19,7 @@ public:
 
   uint64_t startAddr(bool linkerScriptHasSectionsCmd, bool isDynExec,
                      bool loadPhdr) const override {
-    return 0;
+    return 0x400000;
   }
   void initializeAttributes(InputBuilder &pBuilder) override {
     pBuilder.makeBStatic();
--- a/test/x86_64/linux/DefaultImageBase/DefaultImageBase.test
+++ b/test/x86_64/linux/DefaultImageBase/DefaultImageBase.test
@@ -1,0 +1,18 @@
+#---DefaultImageBase.test-----Executable--------#
+
+BEGIN_COMMENT
+# Test x86_64 default image address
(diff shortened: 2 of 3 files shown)


lib/Target/x86_64/x86_64StandaloneInfo.h
```diff
@@ -19,7 +19,7 @@ class x86_64StandaloneInfo : public x86_64Info {
 
   uint64_t startAddr(bool linkerScriptHasSectionsCmd, bool isDynExec,
                      bool loadPhdr) const override {
-    return 0;
+    return 0x400000;
```
partaror: Can we modify this function such that it only returns `0x400000` for executables, i.e, when `m_Config.codeGenType() == LinkerConfig::Exec`. We can continue returning `0` for other codegen types.

parth-07 (approved):
LGTM.

quic-seaswara (approved):

test/x86_64/linux/DefaultImageBase/DefaultImageBase.test
```diff
@@ -0,0 +1,18 @@
...
+# without requiring manual --image-base=0x400000 flag.
+#END_COMMENT
+
+#START_TEST
+RUN: %clang %clangopts -c %p/Inputs/1.c -o %t.1.o
+RUN: %link %linkopts -o %t.1.out %t.1.o
+RUN: %t.1.out; echo $? > %t.code
+RUN: %filecheck --input-file=%t.code %s --check-prefix=EXITCODE
+RUN: readelf -l %t.1.out | grep LOAD | head -1 > %t.load
+RUN: %filecheck --input-file=%t.load %s --check-prefix=IMAGEBASE
```
parth-07:
```suggestion
RUN: %readelf -l %t.1.out | %grep LOAD -m 1 | %filecheck %s --check-prefix=IMAGEBASE
```

PR #196, opened 2025-06-28
 Add R_X86_64_PLT32 static relocation support

bhoumik03:
Implement R_X86_64_PLT32 relocation handling for static linking using PC-relative calculations.
Add test case to verify correct working of the implementation.

### DIFF
--- a/lib/Target/x86_64/x86_64Relocator.cpp
+++ b/lib/Target/x86_64/x86_64Relocator.cpp
@@ -346,6 +346,28 @@ Relocator::Result eld::relocPCREL(Relocation &pReloc, x86_64Relocator &pParent,
   return applyRel(pReloc, Result, pRelocDesc, DiagEngine, options);
 }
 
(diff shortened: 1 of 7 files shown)


lib/Target/x86_64/x86_64Relocator.cpp
```diff
@@ -346,6 +346,28 @@ Relocator::Result eld::relocPCREL(Relocation &pReloc, x86_64Relocator &pParent,
...
+
+  FragmentRef *target_fragref = pReloc.targetRef();
+  Fragment *target_frag = target_fragref->frag();
+  ELFSection *target_sect = target_frag->getOutputELFSection();
+
+  // Static Linking : PLT32 behaves as PC32
+  Relocator::Address Result = S + A - P;
+
+  if (!target_sect->isAlloc()) {
+    return applyRel(pReloc, Result, pRelocDesc, DiagEngine, options);
```
partaror: What is the difference between `applyRel` inside and outside this if-condition?

parth-07 (changes requested):
Can you please rebase this patch and add a test?

PR #197, opened 2025-06-28
Fix x86_64 GOT implementation

bhoumik03:
Correct GOT entry size from 32-bit to 64-bit; update alignment and size parameters.
Replace union pattern with direct storage

### DIFF
--- a/lib/Target/x86_64/x86_64GOT.h
+++ b/lib/Target/x86_64/x86_64GOT.h
@@ -21,13 +21,13 @@ public:
   // Going to be used by GOTPLT0
   x86_64GOT(GOTType T, ELFSection *O, ResolveInfo *R, uint32_t Align,
             uint32_t Size)
-      : GOT(T, O, R, Align, Size), Value{0, 0, 0, 0} {
+      : GOT(T, O, R, Align, Size), Value(0) {
     if (O)
       O->addFragmentAndUpdateSize(this);
   }
 
   // Helper constructor for GOT.
-  x86_64GOT(GOTType T, ELFSection *O, ResolveInfo *R) : GOT(T, O, R, 4, 4) {
+  x86_64GOT(GOTType T, ELFSection *O, ResolveInfo *R) : GOT(T, O, R, 8, 8) {
     if (O)
       O->addFragmentAndUpdateSize(this);
   }
@@ -39,22 +39,17 @@ public:
   virtual x86_64GOT *getNext() { return nullptr; }
 
   virtual llvm::ArrayRef<uint8_t> getContent() const override {
-    // Convert uint32_t to ArrayRef.
-    typedef union {
-      uint32_t a;
-      uint8_t b[4];
-    } C;
-    C Content;
-    Content.a = 0;
+    Value = 0;
     // If the GOT contents needs to reflect a symbol value, then we use the
     // symbol value.
     if (getValueType() == GOT::SymbolValue)
-      Content.a = symInfo()->outSymbol()->value();
+      Value = symInfo()->outSymbol()->value();
     if (getValueType() == GOT::TLSStaticSymbolValue)
-      Content.a =
+      Value =
           symInfo()->outSymbol()->value() - GNULDBackend::getTLSTemplateSize();
-    std::memcpy((void *)Value, (void *)&Content.a, sizeof(Value));
-    return llvm::ArrayRef(Value);
+
+    return llvm::ArrayRef(reinterpret_cast<const uint8_t *>(&Value),
+                          sizeof(Value));
   }
 
   static x86_64GOT *Create(ELFSection *O, ResolveInfo *R) {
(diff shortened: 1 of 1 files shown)


lib/Target/x86_64/x86_64GOT.h
```diff
@@ -21,13 +21,13 @@ class x86_64GOT : public GOT {
...
             uint32_t Size)
-      : GOT(T, O, R, Align, Size), Value{0, 0, 0, 0} {
+      : GOT(T, O, R, Align, Size), Value(0) {
     if (O)
       O->addFragmentAndUpdateSize(this);
   }
 
   // Helper constructor for GOT.
-  x86_64GOT(GOTType T, ELFSection *O, ResolveInfo *R) : GOT(T, O, R, 4, 4) {
+  x86_64GOT(GOTType T, ELFSection *O, ResolveInfo *R) : GOT(T, O, R, 8, 8) {
```
partaror: Nitpick: Please annotate the `8, 8` arguments with the parameter names as: `/*Align=*/ 8, /*Size=*/ 8`

lib/Target/x86_64/x86_64GOT.h
```diff
@@ -39,36 +39,31 @@ class x86_64GOT : public GOT {
...
 private:
-  uint8_t Value[4];
+  mutable uint64_t Value;
 };
 
 class x86_64GOTPLT0 : public x86_64GOT {
 public:
   x86_64GOTPLT0(ELFSection *O, ResolveInfo *R)
-      : x86_64GOT(GOT::GOTPLT0, O, R, 4, 16) {}
+      : x86_64GOT(GOT::GOTPLT0, O, R, 8, 16) {}
```
partaror: Please annotate `8, 16` with the parameter names here.

lib/Target/x86_64/x86_64GOT.h
```diff
@@ -39,36 +39,31 @@ class x86_64GOT : public GOT {
...
+                          sizeof(Value));
   }
 
   static x86_64GOT *Create(ELFSection *O, ResolveInfo *R) {
     return make<x86_64GOT>(GOT::Regular, O, R);
   }
 
 private:
-  uint8_t Value[4];
+  mutable uint64_t Value;
```
partaror, line 61: Does `Value` needs to be a member variable? Can it be a local variable in the `getContent` member function?

partaror (changes requested):
@bhoumik03 We need a test for this patch.

test/x86_64/linux/relocGOTPCREL/Inputs/relocGOTPCREL_test.c
```diff
@@ -0,0 +1,20 @@
+// Test R_X86_64_GOTPCREL relocation
+int global_var = 42;
+
+void _start() {
+  int result;
+
+  asm volatile("movq global_var@GOTPCREL(%%rip), %%rax\n"
```
parth-07, line 7: Why do we need `volatile` here?

test/x86_64/linux/relocGOTPCREL/relocGOTPCREL.test
```diff
@@ -0,0 +1,14 @@
...
+
+BEGIN_COMMENT
+# Test R_X86_64_GOTPCREL relocation type support.
+#END_COMMENT
+
+#START_TEST
+RUN: %clang %clangopts -c %p/Inputs/relocGOTPCREL_test.c -o %t.relocGOTPCREL_test.o
+RUN: %link %linkopts -o %t.relocGOTPCREL_test.out %t.relocGOTPCREL_test.o
+RUN: %t.relocGOTPCREL_test.out; echo $? > %t.code
+RUN: cat %t.code
```
parth-07: Can this be removed?

lib/Target/x86_64/x86_64Relocator.cpp
```diff
@@ -187,6 +190,17 @@ void x86_64Relocator::scanGlobalReloc(InputFile &pInputFile, Relocation &pReloc,
     }
     return;
   }
+  case llvm::ELF::R_X86_64_GOTPCREL:
+  case llvm::ELF::R_X86_64_GOTPCRELX:
+  case llvm::ELF::R_X86_64_REX_GOTPCRELX: {
+    std::lock_guard<std::mutex> relocGuard(m_RelocMutex);
```
parth-07: Can we move the mutex inside the if condition?

parth-07 (commented):
Looks good, apart from the few nitpick comments.

test/x86_64/linux/relocGOTPCREL/Inputs/relocGOTPCREL_test.c
bhoumik03: My understanding is that volatile prevents the compiler from optimizing away the inline assembly. Might not be the case here, but I think it is safer to have it generally. Please let me know if this reasoning is incorrect or if there's a better approach.

partaror (approved):

PR #200, opened 2025-07-01
Improve mapping utility functions

quic-rb10:
This patch aims to fix misleading parameter names in two mapping utility `LinkerConfig::getHashFromFile()` and `LinkerConfig::getFileFromHash` as well as remove an extra const attached to the return type of these functions which could have inhibited some optimizations.


parth-07 (approved):
Looks good!

Issue #147, opened 2025-06-17
Implement Relaxations from RISC-V ELF psABI Qualcomm Extensions


lenary:
The release at https://github.com/quic/riscv-elf-psabi-quic-extensions/releases/tag/v0.2 describes several relaxations for the Xqci instructions/relocations, which we want to implement in `eld`.

The intention is to implement them behind a `--relax-xqci` flag, so that standard implementations are not affected. This will work like `--no-relax-c`, but be disabled-by-default.

The relaxations we intend to implement, in order and grouped together, are:
- [x] "Extended to Standard Function/Tail Call" and "Extended to Compressed Function/Tail Call" 
    (`QC.E.JAL` -> `C.JAL` or `JAL ra`; `QC.E.J` -> `C.J` or `JAL x0`) (#140)
- [x] "Extended to Standard Address", "Extended to Compressed LUI Relaxation", and "Extended to Standard Global-pointer Address Relaxation" 
    (`QC.E.LI` -> `C.LUI`, `QC.LI`, or `ADDI` (gp-relative); `QC.LI` -> `C.LUI`) (#194)
- [x] "Standard to Extended Function/Tail Call Relaxation" 
    (`AUIPC; JALR ra` -> `QC.E.JAL`; `AUIPC; JALR x0` -> `QC.E.J`) (#201)
- [ ] ~"Standard to Extended address relaxation", "Standard to Extended global-pointer address relaxation", and "Zero-page address relaxation"
    (`LUI; ADDI` -> `QC.LI`, `QC.E.LI`, or `QC.E.ADDI` (gp-relative))~
- [ ] ~"Standard to Extended Zero-page Load/Store Relaxation" and "Standard to Extended Global-pointer relaxation"
    (`LUI; L{W,H,HU,B,BU}` -> `QC.E.L{W,H,HU,B,BU}` (zero- or gp-relative); `LUI; S{W,H,B}` -> `QC.E.S{W,H,B}`)~

~For the two-instruction sequences starting `LUI`, we will likely also end up relaxing the `AUIPC` equivalent, but this depends on being able to tell whether the user wants the resulting image to be position-independent or not.~

lenary:
I believe this is now complete. We released a new version of the spec removing the relaxations of hi/lo paired relocations: https://github.com/quic/riscv-elf-psabi-quic-extensions/releases/tag/v0.3

I will file a new issue if we find future work.

PR #201, opened 2025-07-02
[RISCV] Implement Xqci Call Relaxations

lenary:
These relaxations are implemented as part of
RISCVLDBackend::doRelaxationCall as they apply to the same relocation/instructions.

The relaxations are:
- `AUIPC; JALR ra`   -> `QC.E.JAL`
- `AUIPC; JALR zero` -> `QC.E.J`

These respect the existing command-line switches which prevent these being applied when relaxations (or xqci relaxations) are disabled. This is particularly important in this case as standard instructions are replaced by custom Xqci instructions, which will fail to execute on most RISC-V implementations.

These relaxations are described in
http://github.com/quic/riscv-elf-psabi-quic-extensions/releases/latest

This continues the implementation of #147.

I also took the time to refactor `doRelaxationCall` to make the priority between relaxations clearer, and the verbose "relaxation missed" messages be emitted later, once we're surer the relaxation has been missed.

### DIFF
--- a/lib/Target/RISCV/RISCVLDBackend.cpp
+++ b/lib/Target/RISCV/RISCVLDBackend.cpp
@@ -222,11 +222,23 @@ void RISCVLDBackend::reportMissedRelaxation(StringRef Name,
   recordRelaxationStats(Section, 0, NumBytes);
 }
 
-bool RISCVLDBackend::doRelaxationCall(Relocation *reloc, bool DoCompressed) {
+bool RISCVLDBackend::doRelaxationCall(Relocation *reloc) {
+  /* Three similar relaxations can be applied here, in order of preference:
+   * -- auipc;jalr -> c.j/c.jal (saves 6 bytes)
+   * -- auipc;jalr -> jal (saves 4 bytes)
+   * -- auipc;jalr -> qc.e.j/qc.e.jal (saves 2 bytes) (Xqci only)
+   */
+
   Fragment *frag = reloc->targetRef()->frag();
   RegionFragmentEx *region = llvm::dyn_cast<RegionFragmentEx>(frag);
   if (!region)
     return true;
+
+  Relocator::DWord S = getSymbolValuePLT(*reloc);
+  Relocator::DWord A = reloc->addend();
+  Relocator::DWord P = reloc->place(m_Module);
+  Relocator::DWord X = S + A - P;
+
   uint64_t offset = reloc->targetRef()->offset();
 
   // extract the next instruction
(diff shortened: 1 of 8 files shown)


lenary:
This PR has bugs. Working on them.

lenary:
Trying something now, where I undo unnecessary changes to try to work out what's causing the failures/bugs I'm seeing.

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -238,68 +250,88 @@ bool RISCVLDBackend::doRelaxationCall(Relocation *reloc, bool DoCompressed) {
...
+  // For `JAL`, the offset needs to be 21 bits.
+  bool canRelaxJal = config().options().getRISCVRelax() && llvm::isInt<21>(X);
+
+  // For `QC.E.J` and `QC.E.JAL`, we have to have xqci relaxations enabled,
+  // and be on 32-bit, and the destination must be either `x0` or `ra`.
+  bool canRelaxXqci = config().options().getRISCVRelax() &&
+                      config().targets().is32Bits() &&
+                      config().options().getRISCVRelaxXqci();
+  bool canRelaxQcEJ =
+      canRelaxXqci && (rd == 0 || rd == 1) && isInt<32>(X) && S != 0;
```
quic-seaswara: Why S != 0 ? I thought this could happen only with GP relocations ? 

Do you have a test that covers where the relaxation goes wrong ?

I would recommend adding a test 

Eventually we want to get the zero page relaxation page enabled by default, but if we dont have a test it might end up breaking ?

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -238,68 +250,88 @@ bool RISCVLDBackend::doRelaxationCall(Relocation *reloc, bool DoCompressed) {
...
-    }
+    region->replaceInstruction(offset, reloc, jal, 4 /* Replace bytes */);
+    reloc->setTargetData(jal);
+    reloc->setType(llvm::ELF::R_RISCV_JAL);
+    // Delete the next instruction
+    relaxDeleteBytes("RISCV_CALL", *region, offset + 4, 4,
+                     reloc->symInfo()->name());
+
+    // Report missed relaxation as we could still do a `C.J`/`C.JAL`
     reportMissedRelaxation("RISCV_CALL_C", *region, offset, 2,
```
quic-seaswara, line 307: Why are we reporting a missed relaxation here, previously it would do after returning from this function

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -238,68 +250,88 @@ bool RISCVLDBackend::doRelaxationCall(Relocation *reloc, bool DoCompressed) {
...
+    uint64_t qc_e_j = (rd == 1) ? 0xc01ful : 0x401ful;
+    const char *msg =
+        (rd == 1) ? "R_RISCV_CALL_QC_E_JAL" : "R_RISCV_CALL_QC_E_J";
+
+    region->replaceInstruction(offset, reloc, qc_e_j, 6);
+    reloc->setTargetData(qc_e_j);
+    reloc->setType(ELF::riscv::internal::R_RISCV_QC_E_CALL_PLT);
+    relaxDeleteBytes(msg, *region, offset + 6, 2, reloc->symInfo()->name());
+
+    reportMissedRelaxation(msg, *region, offset, 4, reloc->symInfo()->name());
```
quic-seaswara, line 323: Same here could not follow why reportMissedRelaxation

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -238,68 +250,88 @@ bool RISCVLDBackend::doRelaxationCall(Relocation *reloc, bool DoCompressed) {
...
+
+    return true;
+  }
+
+  reportMissedRelaxation("RISCV_CALL", *region, offset, 6,
+                         reloc->symInfo()->name());
```
quic-seaswara, line 329: It may be useful to report missed relaxation from the input file that caused the missed relaxation opportunity ? 

It may be a follow up patch

lib/Target/RISCV/RISCVLDBackend.cpp
lenary: `S != 0` effectively is a check as to whether `S` is defined right now. This is the linker script definition issue we've covered before. We are inconsistent with when we check this `S != 0`, making assumptions that GP is not near zero, and others, which is why we sometimes have the check and sometimes don't. This probably needs a general refactor.

I'm not sure it's actually necessary for this relaxation, I can try a run today without this or the `isInt<32>(X)` check (which should also be always true). On RV32, all symbols are ±32-bits from all others, regardless of when they are resolved.

lib/Target/RISCV/RISCVLDBackend.cpp
lenary: The entire function has a slightly different structure than it had before, for clarity.

This report missedrelaxation is because we did relax the call to a JAL, but we didn't relax it to a C_JAL, which would have been even smaller.

lib/Target/RISCV/RISCVLDBackend.cpp
lenary: Again, this is similar. We report a missed relaxation because we turned 8 bytes into 6 (QC_E_JAL), but we didn't manage to turn the 8 bytes into a 2-byte C_JAL, so there were 4 more bytes on the table.

lib/Target/RISCV/RISCVLDBackend.cpp
lenary: The file is included in the message. See below: `VERBOSE_NOXQCI: RISCV_CALL : Cannot relax 6 bytes for symbol 'f' in section .text+0x1000e0 file {{.*}}.o`

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -238,68 +250,87 @@ bool RISCVLDBackend::doRelaxationCall(Relocation *reloc, bool DoCompressed) {
...
-  region->replaceInstruction(offset, reloc, instr, 4 /* Replace bytes */);
-  // Replace the reloc to R_RISCV_JAL
-  reloc->setType(llvm::ELF::R_RISCV_JAL);
-  reloc->setTargetData(instr);
-  // Delete the next instruction
-  relaxDeleteBytes("RISCV_CALL", *region, offset + 4, 4,
-                   reloc->symInfo()->name());
-  return true;
+  if (canRelaxQcEJ) {
+    uint64_t qc_e_j = (rd == 1) ? 0xc01ful : 0x401ful;
```
quic-akaryaki, line 314: Nit: may be easier to read without `ul`?

test/RISCV/standalone/Relaxation/HI20_LO12_I_TO_GP/HI20_LO12_I_TO_GP.test
```diff
@@ -16,10 +16,8 @@ CHECK: lw      a{{.*}}, 0(gp)
 RUN: %filecheck %s --input-file=%t.1.map --check-prefix=MAP
 MAP: # LinkStats Begin
 MAP: # RelaxationBytesDeleted : {{18|16}}
-MAP-NOT: # RelaxationBytesMissed
```
quic-akaryaki, line 19: What was wrong about this one?

quic-akaryaki (approved):

lib/Target/RISCV/RISCVLDBackend.cpp
lenary: I wanted to be clear that these are 64-bit numbers, I wasn't sure a better way to do so. I do agree it's a little hard to see where the hex ends and the suffix begins.

test/RISCV/standalone/Relaxation/HI20_LO12_I_TO_GP/HI20_LO12_I_TO_GP.test
lenary: `RelaxationBytesMissed` is not emitted only if the value is zero. Now the value is either 0 or 2 (iirc), so it's hard to write a match line for it, but we cannot match the absence of the line any more

lib/Target/RISCV/RISCVLDBackend.cpp
lenary: I re-ran all our internal test suites without the `isInt<32>(X) && S != 0` checks, and they all came back without any failures.

lenary:
From my PoV, this is ready to land, and I would like it to be backported to llvm 21. I think I have addressed the core of all of the comments so far.

quic-seaswara (approved):

PR #203, opened 2025-07-02
Fix crash on parsing ':ALIGN' in linker script

### LINKED ISSUE #202: Linker crashes when trying to parse ':ALIGN'
If there is no space between ":" and ALIGN, linker will misparse the following script:

```bash
SECTIONS
{
  S :ALIGN(4096) {}
}
```

Moreover, when recovering from this error, it will enter infinite recursion and overflow the stack.

Thank you @quic-akaryaki for reporting this issue.

parth-07:
The linker was crashing due to stack overflow when parsing ':ALIGN' in an output section description. This commit fixes the linker script parser so that the crash does not happen.

The root cause of the stack overflow is how we parse expressions (readExpr) in linker script and the behavior of ScriptLexer::expect(...) utility. ScriptLexer::expect does not do anything if errors have already been encountered during linker script parsing. In particular, it never increments the current token position in the script file, even if the current token is the same as the expected token. This causes an infinite call cycle on parsing an expression such as '(4096)' when an error has already been encountered.

readExpr() calls readPrimary()
readPrimary() calls readParenExpr()

readParenExpr():

  expect("("); // no-op, current token still points to '('
  Expression *E = readExpr(); // The cycle continues...

Closes #202

### DIFF
--- a/lib/ScriptParser/ScriptParser.cpp
+++ b/lib/ScriptParser/ScriptParser.cpp
@@ -155,6 +155,17 @@ void ScriptParser::readEntry() {
 }
 
 eld::Expression *ScriptParser::readExpr() {
+  if (atEOF()) {
+    Module &Module = ThisScriptFile.module();
+    GNULDBackend &Backend = ThisScriptFile.backend();
+    // We do not return nullptr here because the returned expression is
+    // dereferenced at many places. We can add a null-pointer check everywhere,
+    // but that would impose issues if we want to extend the parser to continue
+    // parsing despite errors (the way we do with --no-inhibit-exec for overall
+    // linking). Adding checks everywhere would also violate the parser design
+    // to be able to continue parsing even after errors have occurred.
+    return make<NullExpression>(Module, Backend);
+  }
   // Our lexer is context-aware. Set the in-expression bit so that
   // they apply different tokenization rules.
   enum LexState Orig = LexState;
--- a/include/eld/Diagnostics/DiagLDScript.inc
+++ b/include/eld/Diagnostics/DiagLDScript.inc
@@ -128,3 +128,5 @@ DIAG(warn_memory_region_has_zero_size, DiagnosticEngine::Warning,
      "%0: Memory region '%1' has zero size")
 DIAG(error_empty_pt_tls_segment, DiagnosticEngine::Error,
       "TLS segments are empty")
+DIAG(internal_error_null_expression, DiagnosticEngine::InternalError,
+     "Trying to evaluate null expression")
--- a/include/eld/Script/Expression.h
+++ b/include/eld/Script/Expression.h
@@ -78,7 +78,8 @@ public:
     LOGICAL_AND,
     LOGICAL_OR,
     ORIGIN,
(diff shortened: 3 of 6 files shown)


lib/ScriptParser/ScriptParser.cpp
```diff
@@ -155,6 +155,17 @@ void ScriptParser::readEntry() {
...
+  if (atEOF()) {
+    Module &Module = ThisScriptFile.module();
+    GNULDBackend &Backend = ThisScriptFile.backend();
+    // We do not return nullptr here because the returned expression is
+    // dereferenced at many places. We can add a null-pointer check everywhere,
+    // but that would impose issues if we want to extend the parser to continue
+    // parsing despite errors (the way we do with --no-inhibit-exec for overall
+    // linking). Adding checks everywhere would also violate the parser design
+    // to be able to continue parsing even after errors have occurred.
+    return make<Integer>(Module, Backend, "0", 0);
```
quic-seaswara: If we dont need Module and Backend, why create local variables ? 

It may be useful to patch the sites that end up calling expression with a nullptr, and continue ? 

If we still plan on going this route, I would prefer to return a sentinel expression, probably creating a null expression class.

quic-seaswara (changes requested):

lib/ScriptParser/ScriptParser.cpp
parth-07: > If we dont need Module and Backend, why create local variables ?

Module and Backend local variables are used in the `make<Integer>(...)` expression.

> It may be useful to patch the sites that end up calling expression with a nullptr, and continue ?

I do not prefer this because it would be too much patching, and missing a check at any place would result in a null pointer dereference UB and can cause issues with the long-term plan of continuing parsing despite errors.

> If we still plan on going this route, I would prefer to return a sentinel expression, probably creating a null expression class.

I do not think we should create a new class if there is only one use-case for it. I think returning a `0` expression is fine here because the link will terminate after linker script parsing due to errors encountered during the parsing. If `--no-inhibit-exec` is used and we want to continue parsing despite the linker script errors, then using the expression value of `0` seems right as well. Please let me know your thoughts on this.

quic-seaswara: I am worried that the value returned will be interpreted and later used in a wrong way.

quic-seaswara: let us add a sentinel entry and assert if the expression gets evaluated

parth-07: We are only returning this sentinel 0 when the linker script contains error. And the link ends if the linker script parsing emits any error. We should not be continuing link if the linker script parsing emits error, and if we do for the `--no-inhibit-exec` case, then the value of `0` seems right, I think.

quic-seaswara: Ok

parth-07: As we discussed, I have added `NullExpression` subtype to use as a sentinel expression object for invalid expressions.

PR #206, opened 2025-07-03
Fix reproduce crash when input is passed through linkerscript

### LINKED ISSUE #205: `-reproduce` option fails to work when file is passed through linkerscript.
This test fails when the user extracts the reproduce tar ball and runs the response file:

```
cat > 1.c << \!
int foo() { return 0; }
!

cat > script.t << \!
INPUT(1.o)
!

hexagon-clang -c 1.c
hexagon-link script.t --reproduce a.tar
```

There are other kinds of commands that appears to also not work, for example

GROUP
AS_NEEDED

quic-rb10:
This patch aims to fix the reproduce crash when the input file is passed through the linkerscript.

Resolved #205

### DIFF
--- a/include/eld/Script/ScriptFile.h
+++ b/include/eld/Script/ScriptFile.h
@@ -95,6 +95,8 @@ public:
 
   size_t size() const { return LinkerScriptCommandQueue.size(); }
 
+  InputToken *findResolvedFilename(InputToken * Input);
+
   std::string findIncludeFile(const std::string &Filename, bool &Result,
                               bool State);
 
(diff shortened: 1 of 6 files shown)


lib/ScriptParser/ScriptParser.cpp
```diff
@@ -614,9 +614,12 @@ void ScriptParser::addFile(StringRef Name) {
   if (Name.consume_front("-l"))
     InputStrTok = ThisScriptFile.createNameSpecToken(Name.str(),
                                                      ThisScriptFile.asNeeded());
-  else
-    InputStrTok =
-        ThisScriptFile.createFileToken(Name.str(), ThisScriptFile.asNeeded());
+  else {
+    std::string ResolvedFilename =
+        ThisScriptFile.findResolvedFilename(Name.str());
+    InputStrTok = ThisScriptFile.createFileToken(ResolvedFilename,
```
quic-seaswara: Is there a way this can be moved to ObjectLinker::normalize ?

PR #211, opened 2025-07-07
Add both ternary branch expressions symbols into symbol resolver

parth-07:
This commit fixes undefined symbol reference error due to ternary branch expression symbols not getting inserted into symbol resolver. For the below ternary expression:

  u = condition ? expr1 : expr2;

We were only adding symbols from expr2 to the symbol resolver because condition expression value is always 0, as it is not evaluated yet.

This commit modifies the behavior such that the symbols from both the expr1 and expr2 gets added to the symbol resolver. With this change, our behavior is similar to the lld's behavior. Ideally, we should only add symbols from the ternary branch expression that gets selected by the condition expression. However, we need to add linker script symbols way before we can evaluate the linker script expressions.


parth-07:
Depends upon #210

PR #212, opened 2025-07-07
Initialize Expression::result as an empty optional object

### LINKED ISSUE #204: `Expression::Result` should be empty by default
`std::optional<uint64_t> Expression::Result` should be empty by default. This will make sure that the value is never used before it is properly set, and the `Expression::Result` can be used to identify if the expression was ever (successfully) evaluated before, which is helpful for supporting lazy evaluation.

parth-07:
This commit changes Expression::result initialization value from 0 to std::nullopt. Now, Expression::result is an empty object by default. The main motivation for this change is to make it possible to determine whether a linker script expression has been evaluated and to find cases where expression values were (incorrectly) used before evaluating the expression.

Closes #204

### DIFF
--- a/lib/Script/Expression.cpp
+++ b/lib/Script/Expression.cpp
@@ -15,13 +15,14 @@
 #include "eld/Target/ELFSegment.h"
 #include "eld/Target/GNULDBackend.h"
 #include "llvm/Support/MathExtras.h"
+#include <optional>
 
 using namespace eld;
 
 Expression::Expression(std::string Name, Type Type, Module &Module,
                        GNULDBackend &Backend, uint64_t Value)
     : Name(Name), ThisType(Type), ThisModule(Module), ThisBackend(Backend),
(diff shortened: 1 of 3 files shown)


parth-07:
Depends upon #210 and #211. These issues were due to (incorrectly) using expression result values before even the expression was evaluated.

lib/Script/Expression.cpp
```diff
@@ -33,6 +34,12 @@ uint64_t Expression::result() const {
   return *MResult;
 }
 
+uint64_t Expression::resultOrZero() const {
+  if (MResult)
+    return MResult.value();
+  return 0;
+}
+
```
quic-seaswara, line 42: Can we add a function to say if the expression result is available ?

parth-07: Done.

quic-seaswara (approved):

PR #216, opened 2025-07-07
improve ci workflow by speeding up checkouts

quic-seaswara:
In Git, fetch-depth refers to how many commits deep Git should go when fetching or cloning a repository. It’s used to create a shallow clone, which limits the history you download—super handy when you don’t need the full commit log.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,7 @@ jobs:
       if: steps.cache.outputs.cache-hit != 'true'
       uses: actions/checkout@v4
       with:
+        fetch-depth: 2
         repository: llvm/llvm-project
         path: pr-${{ env.PR_NUMBER }}/llvm-project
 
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -31,13 +31,15 @@ jobs:
       if: steps.cache.outputs.cache-hit != 'true'
       uses: actions/checkout@v4
       with:
+        fetch-depth: 2
```
parth-07, line 34: I think the default `fetch-depth` for `actions/checkout@v4` is 1.

From https://github.com/actions/checkout

> \# Number of commits to fetch. 0 indicates all history for all branches and tags.
> \# Default: 1
> fetch-depth: ''

quic-seaswara:
I will abandon this change as it is not creating any speed up.

@navaneethshan will push a fix to make sure we are able to cache previous built artifacts

.github/workflows/ci.yml
quic-seaswara: Yeah, thanks for pointing it out. 

I noticed it after pushing this PR.

PR #228, opened 2025-07-10
Emit empty segment warning only if linker-script warnings are enabled

### LINKED ISSUE #227: Empty segment warning should only be emitted when linker-script warnings are enabled.
Empty segment warning should only be emitted when linker-script warnings are enabled.

parth-07:
Emit empty segment warning only if linker-script warnings are enabled

With this change, the empty segment warning is only emitted when
linker-script warning category is enabled.

Closes #227


quic-seaswara (approved):

quic-seaswara:
Follow up : move the verifySegments to GNULDBackend::postLayout after checking for overlaps

PR #233, opened 2025-07-11
Enable reproduce feature for findConfigFile + getFileContents use idiom

### LINKED ISSUE #232: File passed to `LinkerWrapper::findConfigFile` should be stored as part of the reproducer.
File passed to `LinkerWrapper::findConfigFile` should be stored as part of the reproducer.

parth-07:
This commit enables reproduce feature for findConfigFile +
getFileContents use idiom. More concretely:

```cpp
  std::string C;
  auto expConfigPath = getLinker()->findConfigFile(filename);
  if (expConfigPath)
    C = getLinker()->getFileContents(expConfigPath.value());
```

This commit enables reproduce for such usage by adding an additional
'StringMap' in the mapping file.

Closes https://github.com/qualcomm/eld/issues/232


lib/LinkerWrapper/LinkerWrapper.cpp
```diff
@@ -647,11 +647,17 @@ std::string LinkerWrapper::getFileContents(std::string FileName) {
...
   const eld::sys::fs::Path *P = Directories.findFile(
       "plugin configuration INI file", FileName, getPlugin()->getPluginName());
-  if (P)
-    return P->getFullPath();
+  if (P) {
+    std::string FullPath = P->getFullPath();
+    if (m_Module.getOutputTarWriter() && llvm::sys::fs::exists(FullPath))
+      m_Module.getOutputTarWriter()->createAndAddConfigFile(FileName, FullPath);
+    return FullPath;
+  }
```
quic-seaswara, line 660: We just need to store a mapping in the mapping.ini file, no need to write this to the tarball.

Add a test to make sure that the file is not in the tarball but in mapping.ini file, if someone wants someone who calls findConfigFile but not read the file using getFileContents

Add a note in the documentation saying that users have to use findConfigFile and getFileContents together.

parth-07: Thank you for the suggestions. I have applied them.

quic-seaswara (approved):

PR #237, opened 2025-07-14
fix segment alignment when building for omagic

### LINKED ISSUE #192: eld produces different segment p_aligns under --omagic/-N than lld/ld
**Description**
As far as I can tell,  `--omagic`/`-N` in ld/lld ends up meaning that the p_align for (some? most? all?) segments is the max of their output sections rather than (I assume?) page aligned--see the reproducer below where eld produces segments with p_align 0x1000 as compared to ld/lld producing segments with p_align 0x2/0x8.

The test below is using PHDRs, but this behavior seemed consistent when using one/both/none of MEMORY and PHDRs.

**Reproducer**
```
cat > test.c << '!'
int data = 1;
long long int bss;

int main(int argc, char *argv[]) { return 0; }
!

cat > script.t << '!'
PHDRS {
  text PT_LOAD;
  data PT_LOAD;
}

SECTIONS {
  .text : { *(.text) } :text
  .data : { *(.data) } :data
  .bss : { *(.bss) } :data
}
!

clang --target=riscv32-unknown-elf -c test.c

#  LOAD           0x001000 0x00000000 0x00000000 0x00020 0x00020 R E 0x1000
#  LOAD           0x001020 0x00000020 0x00000020 0x00004 0x00010 RW  0x1000
ld.eld -m elf32lriscv test.o -N -T script.t

#   LOAD           0x000094 0x00000000 0x00000000 0x00020 0x00020 R E 0x2
#   LOAD           0x0000b8 0x00000020 0x00000020 0x00004 0x00010 RW  0x8
ld.lld test.o -N -T script.t

#   LOAD           0x000094 0x00000000 0x00000000 0x00020 0x00020 RWE 0x2
#   LOAD           0x0000b8 0x00000020 0x00000020 0x00004 0x00010 RW  0x8
riscv32-unknown-elf-ld.bfd test.o -N -T script.t
```

quic-seaswara:
omagic should set the segment alignment for loadable segments to be the maximum section alignment of the sections that are contained in the load segment

Resolves #192


partaror (approved):
LGTM.

PR #240, opened 2025-07-16
Fix mapping file feature for `findConfigFile` and `getFileContents` idiom

parth-07:
This commit fixes mapping file feature for findConfigFile and getFileContents idiom. There were two issues:

- A common map was being used to store two kinds of mappings:
  - original filename -> hashed filename in the reproducer
  - string key -> string value

  Here string key -> string value represents mappings specified in
  `[StringMap]` section of the mapping ini file.

  We must not use a common map for storing these two kinds of mappings
  because they can conflict with each other. A common example that
  demonstrates the conflict is when findConfigFile and getFileContents
  are called with absolute paths.

- getFileContents should call getHashFromFile instead of getFileFromHash. 
  Previously, it worked fine because Plugin option was getting used in findConfigFile called, 
  and we transform the plugin option based on the mapping file as well.


include/eld/Config/LinkerConfig.h
```diff
@@ -85,6 +85,21 @@ class LinkerConfig {
     std::optional<bool> EnableCommandLineWarnings;
   };
 
+  struct MappingFileInfo {
+    std::unordered_map<std::string, std::string> StringMapKeyToValue;
+
+    void addStringMapEntry(llvm::StringRef Key, llvm::StringRef Value) {
+      StringMapKeyToValue[Key.str()] = Value.str();
```
quic-seaswara, line 92: Should this assert if there was already a key for the mapping ?

include/eld/Config/LinkerConfig.h
```diff
@@ -85,6 +85,21 @@ class LinkerConfig {
...
+    }
+
+    llvm::StringRef getStringMapValueFromKey(llvm::StringRef Key) const  {
+      auto iter = StringMapKeyToValue.find(Key.str());
+      if (iter != StringMapKeyToValue.end())
+        return iter->second;
+      return "";
```
quic-seaswara, line 99: Can this happen ?

lib/LinkerWrapper/LinkerWrapper.cpp
```diff
@@ -630,7 +630,7 @@ plugin::LinkerScript LinkerWrapper::getLinkerScript() {
 
 std::string LinkerWrapper::getFileContents(std::string FileName) {
   if (m_Module.getConfig().options().hasMappingFile())
-    FileName = m_Module.getConfig().getFileFromHash(FileName);
+    FileName = m_Module.getConfig().getHashFromFile(FileName);
```
quic-seaswara, line 633: Can we add a test to capture if getFileContents() is called using an absolute path ?

test/Common/Plugin/ReproducerWithFindConfigFileAbsolutePath/ReproducerWithFindConfigFileAbsolutePath.cpp
```diff
@@ -0,0 +1,41 @@
...
+  void ActBeforeRuleMatching() override {
+    auto E = getLinker()->findConfigFile(MyConfigFile);
+    ELDEXP_REPORT_AND_RETURN_VOID_IF_ERROR(getLinker(), E);
+    std::cout << "Found config file " << E.value() << "\n";
+
+    std::string C = getLinker()->getFileContents(E.value());
+    std::cout << "Contents of config file: " << C << "\n";
+  }
+
+  std::string MyConfigFile;
```
quic-seaswara, line 38: private ?

include/eld/Config/LinkerConfig.h
parth-07: Yes, it can happen. A plugin can be designed in such a way that the argument of `findConfigFile(...)` is different for the original link and the reproducer link. I do not think we can prevent all the cases. By returning an empty string, we are leaving up to the caller of `getStringMapValueFromKey` to decide what happens if a mapping value is not present when we expected it to be present.

include/eld/Config/LinkerConfig.h
parth-07: I don't think so. Currently, we add a mapping for each call to `LinkerWrapper::findConfigFile`. There can be multiple calls to this API with the same argument, however, they should all have the same value. We are assuming that they should all have the same value. If the assumption is wrong, then we may need to improve the mapping file.

lib/LinkerWrapper/LinkerWrapper.cpp
parth-07: It is called using absolute file in the test added in this PR: test/Common/Plugin/ReproducerWithFindConfigFileAbsolutePath/ReproducerWithFindConfigFile.test

test/Common/Plugin/ReproducerWithFindConfigFileAbsolutePath/ReproducerWithFindConfigFileAbsolutePath.cpp
parth-07: It is just a test :). Should we be careful about access specifiers for the test artifacts?

quic-seaswara (approved):

PR #241, opened 2025-07-16
Fix linker crash when invalid section is used with 'ALIGNOF' linker script command.

### LINKED ISSUE #239: Linker crashes when invalid section is used with 'ALIGNOF' linker script command
Linker crashes when invalid section name is used with 'ALIGNOF' linker script command. The crash occurs with both the old and the new linker script parsers.

Reproducible example:

```
#!/usr/bin/env bash

cat >1.c <<\EOF
int foo()

{ return 1; }
EOF

cat >script.t <<\EOF
SECTIONS {
TEXT :

{ *(*text*) }
. = ALIGN(ALIGNOF(NON_EXISTENT_SECTION));
DATA :

{ *(*data*) }
}
EOF

clang -target hexagon -o 1.o 1.c -c -ffunction-sections
ld.eld -mhexagonelf -o 1.out 1.o -T script.t
```

Ris-Bali:
This patch fixes the linker scrash that occurs when an invalid section is used with the 'ALIGNOF' linker script command.

Resolves #239

### DIFF
--- a/lib/Script/Expression.cpp
+++ b/lib/Script/Expression.cpp
@@ -726,7 +726,10 @@ eld::Expected<uint64_t> AlignOf::evalImpl() {
   // sections are known only after all the assignments are complete
   if (ThisSection == nullptr)
     ThisSection = ThisModule.getScript().sectionMap().find(Name);
-  assert(ThisSection != nullptr);
+  if (!ThisSection) {
+    return std::make_unique<plugin::DiagnosticEntry>(
+        Diag::error_sect_invalid, std::vector<std::string>{Name});
+  }
   // evaluate sub expressions
   return ThisSection->getAddrAlign();
 }
--- a/include/eld/Diagnostics/DiagLDScript.inc
+++ b/include/eld/Diagnostics/DiagLDScript.inc
@@ -29,6 +29,8 @@ DIAG(promoting_bss_to_progbits, DiagnosticEngine::Note,
      "Section %0(%1) is before %2(%3), Promoting %0 to PROGBITS")
 DIAG(undefined_symbol_in_linker_script, DiagnosticEngine::Fatal,
      "%0: Error: undefined symbol '%1' referenced in expression")
+DIAG(error_sect_invalid, DiagnosticEngine::Error,
+     "%0: Invalid section '%1'")
 DIAG(warn_section_no_segment, DiagnosticEngine::Warning,
      "Section %0 does not have segment assignment in linker script.")
 DIAG(fatal_segment_not_defined_ldscript, DiagnosticEngine::Fatal,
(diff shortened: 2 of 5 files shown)


include/eld/Diagnostics/DiagReaders.inc
```diff
@@ -25,6 +25,8 @@ DIAG(user_syntax_error, DiagnosticEngine::Fatal,
      "Syntax error found at line %1 while parsing file `%0'")
 DIAG(idx_sect_not_found, DiagnosticEngine::Fatal,
      "Section index `%0` out of bounds in file `%1`")
+DIAG(sect_invalid, DiagnosticEngine::Error,
```
parth-07: Prefix the diagnostic id with the diagnostic severity.

include/eld/Diagnostics/DiagReaders.inc
```diff
@@ -25,6 +25,8 @@ DIAG(user_syntax_error, DiagnosticEngine::Fatal,
...
+     "Section %0 is invalid")
```
parth-07: `"Invalid section '%0'"` is more concise.

lib/Script/Expression.cpp
```diff
@@ -726,7 +726,11 @@ eld::Expected<uint64_t> AlignOf::evalImpl() {
   // sections are known only after all the assignments are complete
   if (ThisSection == nullptr)
     ThisSection = ThisModule.getScript().sectionMap().find(Name);
-  assert(ThisSection != nullptr);
+  if (!ThisSection) {
+    ThisModule.getConfig().raise(Diag::sect_invalid) << Name;
```
parth-07: Return the error as `DiagnosticEntry` instead, for this function return value of `0` does not represent failure.

parth-07 (changes requested):

include/eld/Diagnostics/DiagReaders.inc
```diff
@@ -25,6 +25,8 @@ DIAG(user_syntax_error, DiagnosticEngine::Fatal,
      "Syntax error found at line %1 while parsing file `%0'")
 DIAG(idx_sect_not_found, DiagnosticEngine::Fatal,
      "Section index `%0` out of bounds in file `%1`")
+DIAG(error_sect_invalid, DiagnosticEngine::Error,
+     "Invalid Section %0")
```
parth-07: Can you please enclose the section name with quotes (`'%0'`)?

lib/Script/Expression.cpp
```diff
@@ -726,7 +726,11 @@ eld::Expected<uint64_t> AlignOf::evalImpl() {
   // sections are known only after all the assignments are complete
   if (ThisSection == nullptr)
     ThisSection = ThisModule.getScript().sectionMap().find(Name);
-  assert(ThisSection != nullptr);
+  if (!ThisSection) {
+    ThisModule.getConfig().raise(Diag::error_sect_invalid) << Name;
+    ThisModule.setFailure(true);
```
parth-07: Do we need this here?

lib/Script/Expression.cpp
Ris-Bali: Have made these changes

quic-seaswara (approved):

PR #243, opened 2025-07-16
Fix parsing of ':ALIGN(...)' in linker script

### LINKED ISSUE #242: `:ALIGN(...)` in linker script should not give error
Currently, `:ALIGN(...)` in linker script gives error because ':ALIGN' is parsed as ':ALIGN' instead of ':' and 'ALIGN'. The parsing should be fixed here for GNU compatibility.

parth-07:
This commit improves GNU compatibility by fixing the parsing of ':ALIGN(...)' in linker script. Previously, ':ALIGN' was parsed as one token causing confusing error. Now it gets correctly parsed as two tokens: ':' and 'ALIGN'.

Closes #242


lib/ScriptParser/ScriptParser.cpp
```diff
@@ -758,7 +757,7 @@ OutputSectDesc::Prolog ScriptParser::readOutputSectDescPrologue() {
       Prologue.PluginCmd->setHasOutputSection();
   }
 
-  expect(":");
+  expect(LexState::Expr, ":");
```
quic-seaswara, line 760: Why are we using a Expr state to read a : ?

parth-07: Because `:` is a token separator for `Expr` lex state, so this lex state correctly parses things such as `:ALIGN(...)`.

PR #244, opened 2025-07-16
Support PIE with TLS

### LINKED ISSUE #236: eld errors on R_RISCV_TPREL_* relocations with -pie
**Description**
eld produces errors on `R_RISCV_TPREL_HI20` and `R_RISCV_TPREL_LO12_I` relocations it encounters when built with `-pie`. I don't think this is right--in the example below, (IIUC) the compiler is ok to use local-exec/the TPREL relocations and local-exec should be fine/correct since we're referencing a definition in the executable from the executable.

FWIW, I don't think this is limited to just -static -pie, that's just where we're seeing this come up--as best I can tell, `R_RISCV_TPREL_HI20`/`R_RISCV_TPREL_LO12_I` can be valid in non-static code.

**Example**
```
cat > main.c << '!'
__thread int b = 2;                                                                                         
  return b;
}
!

clang --target=riscv32-unknown-elf -fPIE -ftls-model=global-dynamic -c main.c

# Using linux variant since the baremetal one doesn't like -pie, but I think the idea is the same for purposes here
riscv32-unknown-linux-gnu-ld.bfd -pie -static --no-dynamic-linker main.o
ld.lld -pie -static --no-dynamic-linker main.o
ld.eld -m elf32lriscv -static -pie --no-dynamic-linker main.o
# Fatal: Current link configuration does not support relocation type `R_RISCV_TPREL_HI20' for symbol `b' referred from main.o[.text] , recompile with -fPIC
# Fatal: Current link configuration does not support relocation type `R_RISCV_TPREL_LO12_I' for symbol `b' referred from main.o[.text] , recompile with -fPIC
```

quic-seaswara:
local exec relocations are allowed when building for PIE.

Resolves #236

### DIFF
--- a/test/Common/standalone/TLS/TLSPIE/Inputs/tls.c
+++ b/test/Common/standalone/TLS/TLSPIE/Inputs/tls.c
@@ -1,0 +1,12 @@
+int j;
+__thread int a;
+__thread int b __attribute((tls_model ("local-exec")));
+__thread int c __attribute((tls_model ("initial-exec")));
+__thread int d __attribute((tls_model ("local-dynamic")));
+__thread int e __attribute((tls_model ("global-dynamic")));
+
+int
+main (void)
+{
+  return a + b + c + d + e + j;
+}
--- a/lib/Target/AArch64/AArch64Relocator.cpp
+++ b/lib/Target/AArch64/AArch64Relocator.cpp
@@ -128,14 +128,12 @@ bool AArch64Relocator::isInvalidReloc(Relocation &pReloc) const {
   if (!config().isCodeIndep())
     return false;
   switch (pReloc.type()) {
-  case llvm::ELF::R_AARCH64_ABS32:
   case llvm::ELF::R_AARCH64_ABS16:
-  case llvm::ELF::R_AARCH64_TLSIE_LD64_GOTTPREL_LO12_NC:
-  case llvm::ELF::R_AARCH64_TLSLE_LDST64_TPREL_LO12_NC:
+    return true;
   case llvm::ELF::R_AARCH64_TLSLE_ADD_TPREL_LO12_NC:
   case llvm::ELF::R_AARCH64_TLSLE_ADD_TPREL_LO12:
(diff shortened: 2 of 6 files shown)


test/Common/standalone/TLS/TLSPIE/Inputs/tls.c
```diff
@@ -0,0 +1,12 @@
+int j;
+__thread int a;
+__thread int b __attribute((tls_model ("local-exec")));
+__thread int c __attribute((tls_model ("initial-exec")));
+__thread int d __attribute((tls_model ("local-dynamic")));
+__thread int e __attribute((tls_model ("global-dynamic")));
```
jonathonpenix, line 6: Not necessarily a blocker, but is the point here to generate the various relocations for local-exec/initial-exec/local-dynamic/global-dynamic and check that all of them are accepted under -pie?

If so, FWIW, I think clang might be foiling this--I think these all turn into local-exec accesses (all `R_RISCV_TPREL*` for riscv, for example). See https://godbolt.org/z/Een5Gn9Y5 as an example.

I'm assuming this is essentially https://github.com/llvm/llvm-project/issues/101913 --not sure what the best way to work around this is in the short term besides directly writing the asm.

quic-seaswara: There are two things here, I prefer 'C' code more than asm to test multiple targets. The Common test runs for all targets and makes sure that all the TLS relocs are appropriately handled without producing a invalid relocation error for all the TLS models. 

The test may be eventually used to check for TLS relaxations when implemented.

jonathonpenix: Right, I don't have an issue with C code in principle, but due to the clang bug/current behavior above, I just wanted to point out that I think _only_ local-exec TLS relocations are being generated by the compiler and tested here.

So I think this might not be testing what we're hoping

quic-seaswara: Yes, I agree. 

The goal is that local-exec model gets caught by linker equally across all targets.

PR #263, opened 2025-07-22
Fix reproduce crash when input is passed through linkerscript

rb1005:
This patch aims to fix the reproduce crash when the input file is passed through the linkerscript

Resolves: #205


parth-07 (commented):
Looks good! Can you please add a test for the GROUP command as well?

quic-seaswara (approved):

PR #271, opened 2025-07-23
Infer OS/ABI from input files

### LINKED ISSUE #165: Revisit how the ELF OSABI field is set
**Description**
Currently (for aarch64 targets at least), eld will set the ELF OSABI flag to ELFOSABI_LINUX/GNU if given a linux-y emulation or triple (`-m aarch64linux`, for example). Other linkers don't seem to do that--I'm not sure if lld will set ELFOSABI_LINUX in general, and my understanding from the links below is that ld looks for GNU-specific features.

The OSABI flag being set to ELFOSABI_LINUX in this way has caused some surprises in the past, so it might be worth changing this.

Checking for GNU-specific features or looking for ELFOSABI_LINUX/GNU in the "incoming" object files might be options worth considering (and both are discussed/mentioned in the links below).

A few maybe-helpful links:
- https://github.com/llvm/llvm-project/issues/59199 (related discussion in lld)
- https://github.com/llvm/llvm-project/issues/81740 (another related discussion in lld)
- https://github.com/llvm/llvm-project/issues/40541 (semi-related discussion in objcopy)

**Reproducer**
```
cat > empty.c << '!'
  return 0;
}
!

clang --target=aarch64-none-elf empty.c -c
ld.eld -m aarch64linux empty.o
# Shows: OS/ABI: UNIX - GNU
readelf -h a.out
```

quic-seaswara:
Infer the OS/ABI from input files instead from the emulation specified.

Resolves #165


include/eld/Diagnostics/DiagReaders.inc
```diff
@@ -96,3 +96,5 @@ DIAG(error_patch_dynamic_input, DiagnosticEngine::Error,
 DIAG(error_patch_invalid_symbol, DiagnosticEngine::Error,
      "Symbol '%0' from file '%1' cannot be patched. Only global non-common "
      "symbols can be patched")
+DIAG(warning_abi_differences, DiagnosticEngine::Warning,
+     "%0 has OS/ABI set to %1, previously seen %2")
```
partaror: Can we please enclose the input file in quotes? It helps when the input filename contains non-trivial characters such as whitespace.

lib/Config/LinkerConfig.cpp
```diff
@@ -241,6 +241,13 @@ bool LinkerConfig::setWarningOption(llvm::StringRef WarnOption) {
     setShowWholeArchiveWarning(true);
     return true;
   }
+  // -Wosabi and -Wno-osabi
+  if (WarnOpt == "osabi") {
+    setShowOSABIWarning(true);
+  }
+  if (WarnOpt == "no-osabi") {
+    setShowOSABIWarning(true);
```
partaror:
```suggestion
    setShowOSABIWarning(false);
```

lib/Readers/ELFReader.cpp
```diff
@@ -421,6 +429,10 @@ eld::Expected<bool> ELFReader<ELFT>::isCompatible() const {
         Diag::invalid_elf_class,
         {inputFile.getInput()->decoratedPath(), config.targets().getArch()}));
 
+  eld::Expected<bool> expCheckOSAbi = checkOSABI();
+  if (config.showOSABIWarnings() && !expCheckOSAbi)
+    config.getDiagEngine()->raiseDiagEntry(std::move(expCheckOSAbi.error()));
```
partaror: So far, we only use `eld::Expected` for errors. Can we continue reserving `eld::Expected` just for error? `.error()` call for warning is also misleading. What do you think about this?

lib/Readers/ELFReader.cpp
```diff
@@ -400,6 +400,14 @@ template <class ELFT> bool ELFReader<ELFT>::checkClass() const {
   return elfHeader.e_ident[llvm::ELF::EI_CLASS] == backend.getInfo().ELFClass();
 }
 
+template <class ELFT> eld::Expected<bool> ELFReader<ELFT>::checkOSABI() const {
```
partaror: Can we return `eld::Expected<void>` here?

test/AArch64/standalone/OSABI/OSABI.test
```diff
@@ -0,0 +1,17 @@
...
+#END_COMMENT
+RUN: %clang --target=aarch64-none-elf  -c %p/Inputs/1.c  -o %t1.o
+RUN: %link -maarch64linux %t1.o  -o %t.out.1
+RUN: %readelf -h %t.out.1 | %filecheck %s
+RUN: %clang --target=aarch64-none-elf  -c %p/Inputs/1.c -flto  -o %t1.bc.o
+RUN: %link -maarch64linux %t1.o  -o %t.out.2
+RUN: %readelf -h %t.out.2 | %filecheck %s
+RUN: %mkdir -p %t && %touch %t/empty.o
+RUN: %link -maarch64linux %t/empty.o -r -o %t.r.o
+RUN: %not %link -maarch64linux %t1.o %t.r.o -Wosabi -Werror -o %t2.out
```
partaror: Can we please also add the error message as part of the test as well?

include/eld/Diagnostics/DiagReaders.inc
quic-seaswara: Sure

lib/Config/LinkerConfig.cpp
quic-seaswara: Thanks for the catch!

lib/Readers/ELFReader.cpp
quic-seaswara: Ok sure. I will need to move raising the diagnostics to checkOSABI. 

We should probably fix this as its inconvenient to use. 

I would expect callees to return expected and its upto the caller to report it if its warning or error.

lib/Readers/ELFReader.cpp
quic-seaswara: Sure

test/AArch64/standalone/OSABI/OSABI.test
quic-seaswara: This is a good suggestion, will do

include/eld/Diagnostics/DiagReaders.inc
```diff
@@ -96,3 +96,5 @@ DIAG(error_patch_dynamic_input, DiagnosticEngine::Error,
 DIAG(error_patch_invalid_symbol, DiagnosticEngine::Error,
      "Symbol '%0' from file '%1' cannot be patched. Only global non-common "
      "symbols can be patched")
+DIAG(warning_abi_differences, DiagnosticEngine::Warning,
+     "'%0' has OS/ABI set to %1, previously seen %2")
```
jonathonpenix, line 100: Would it be worth making it explicit that we're ignoring this new value and we'll be using %2 instead?

jonathonpenix (commented):
As a broader policy question: IIUC we're taking the first OSABI we see from the input files. I think lld does something a bit different (first non-NONE OSABI with some specific overrides based on the emulation, see https://github.com/qualcomm/eld/issues/165#issuecomment-2989555405).

Would it be worth doing something like that here (first non-NONE OSABI)? My understanding is that object files with a mix of NONE and LINUX would be more common than all LINUX, for example.

include/eld/Diagnostics/DiagReaders.inc
quic-seaswara: bfd appears to take the lowest ABI.

**bfd**

```
$llvm-readelf -h r.o | grep ABI
  OS/ABI:                            Standalone App
$llvm-readelf -h 1.o | grep ABI
  OS/ABI:                            UNIX - System V
$aarch64-linux-gnu-ld.bfd -maarch64linux 1.o r.o -o r
$llvm-readelf -h r | grep ABI
  OS/ABI:                            UNIX - System V
$aarch64-linux-gnu-ld.bfd  1.o r.o -o r
  OS/ABI:                            UNIX - System V
```

**lld**

```
$ ld.lld 1.o r.o -o r
$ llvm-readelf -h r | grep ABI
  OS/ABI:                            Standalone App
$ ld.lld -maarch64linux 1.o r.o -o r
$ llvm-readelf -h r | grep ABI
  OS/ABI:                            UNIX - System V
```


It looks like the ABI setting is implementation defined.

docs/userguide/documentation/linker_diagnostics.rst
```diff
@@ -95,6 +95,8 @@ General Warning Flags
      - Warns when ``--whole-archive`` is used inappropriately or causes symbol bloat.
    * - ``-Wzero-sized-sections``
      - Flags sections that are defined but have zero size, which may indicate a script or input error.
+   * - ``-Wosabi``
```
parth-07, line 98: We also need to add this to `GNULinkerOptions.td`

docs/userguide/documentation/linker_diagnostics.rst
```diff
@@ -95,6 +95,8 @@ General Warning Flags
...
+     - Generates a warning when an input file's OS/ABI value differs from those encountered in previously processed files by eld.
```
parth-07:
```suggestion
     - Generates a warning when an input file's OS/ABI value differs from those encountered in previously processed files.
```
We can remove 'by eld' here.

lib/Target/TargetInfo.cpp
```diff
@@ -138,3 +142,16 @@ uint8_t TargetInfo::ELFClass() const {
...
+
+bool TargetInfo::setOSABI(const InputFile &In, const uint8_t ABI) {
+  if (!OSAbi)
+    OSAbi = ABI;
+  if ((OSAbi != ABI) && m_Config.showOSABIWarnings()) {
+    m_Config.raise(Diag::warning_abi_differences)
+        << In.getInput()->decoratedPath()
+        << std::string(llvm::ELF::convertOSABIToName(ABI))
+        << std::string(llvm::ELF::convertOSABIToName(*OSAbi));
+    return false;
```
parth-07: Why do we return false for a warning?

lib/Readers/ELFReader.cpp
```diff
@@ -421,6 +429,9 @@ eld::Expected<bool> ELFReader<ELFT>::isCompatible() const {
         Diag::invalid_elf_class,
         {inputFile.getInput()->decoratedPath(), config.targets().getArch()}));
 
+  if (!checkOSABI())
+    return false;
```
parth-07: Do we want to fail the link for incompatible ABIs?

include/eld/Diagnostics/DiagReaders.inc
parth-07: Perhaps an another issue is to determine which ABIs are compatible with each other and which are not. In general, taking the ABI with the highest requirements seems right as long as all the other ABIs are the subset of the ABI with the highest requirement.

jonathonpenix: > bfd appears to take the lowest ABI

bfd looks for certain features being present AFAICT:

```
cat > foo.c << '!'
void foo(void);
void my_foo(void) { }
void (*resolve_foo(void)) (void) { return my_foo; }
!

cat > bar.c << '!'
int bar(void) { return 2; }
!

clang --target=aarch64-linux-gnu -c foo.c bar.c -c
aarch64-linux-gnu-ld.bfd foo.o bar.o
# Gives: OS/ABI:                            UNIX - GNU
```

In the output you showed above, I don't think either of the input .o files will have GNU set (so lld won't mark the output as GNU) and there aren't any GNU features, so bfd doesn't set it.

> It looks like the ABI setting is implementation defined.

Yeah, I don't think there's a standardized way of doing this (and honestly I'm not sure how much it matters). IMO, lld's behavior seems like an ok starting point in that I think it does something reasonable enough and I don't think is particularly complex/intrusive.

docs/userguide/documentation/linker_diagnostics.rst
quic-seaswara: good catch

docs/userguide/documentation/linker_diagnostics.rst
quic-seaswara: Sure

lib/Target/TargetInfo.cpp
quic-seaswara: Yeah I agree

lib/Readers/ELFReader.cpp
quic-seaswara: will change it to a void function

quic-seaswara:
Addressed all comments.

lib/Target/TargetInfo.cpp
```diff
@@ -138,3 +137,19 @@ uint8_t TargetInfo::ELFClass() const {
 TargetInfo::TargetRelocationType TargetInfo::getTargetRelocationType() const {
   return relocType;
 }
+
+void TargetInfo::setOSABI(const InputFile &In, const uint8_t ABI) {
+  if (!OSAbi) {
+    OSAbi = ABI;
+    return;
+  }
+  if ((OSAbi != ABI) && m_Config.showOSABIWarnings()) {
```
jonathonpenix, line 146: Should the warning be conditional on `!isOSABINone(ABI)` too?

Ex: if we process a.o (GNU) then b.o(NONE) I think we're warning here (and I'm not sure if we should?)

quic-seaswara: No, it would be useful to see all object files that have different OS/ABI values. This way if there is a way to cleanup or fix the build, users can.

PR #272, opened 2025-07-24
Fix evaluation of expressions containing forward references

### LINKED ISSUE #169: eld erroring on forward reference in PROVIDE with ternary operator
**Description**
eld is producing errors about undefined symbols in the reproducer below. eld accepts it if either of the `PROVIDE` or ternary operator are omitted and lld and ld both seem to accept the reproducer so I'm assuming this is valid?

This was reduced from one of picolibc's linker scripts (rv64 with TLS enabled).

**Reproducer**
```
cat > main.c << '!'
  return 0;
}
!

cat > script.t << '!'
SECTIONS {
  __tls_size = __tls_tail_extra_size ? 0x10 : 0x20;
  PROVIDE( __tls_tail_extra_size = 0x15 );

  .text : { *(.text) }
}
!

clang --target=riscv32-unknown-elf main.c -c
# errors: Fatal: script.t: Error: undefined symbol '__tls_tail_extra_size' referenced in expression
ld.eld -m elf32lriscv -T script.t main.o
# lld and ld both accept it and set __tls_size to 0x10
ld.lld -T script.t main.o
riscv32-unknown-elf-ld.bfd -T script.t main.o
```

parth-07:
This commit fixes evaluation of expressions containing forward
references. At a high-level, eld evaluates assignments in sequence
and consequently the expressions containing forward references needs to
be reevaluated when the forward reference value is computed for the correct
computation results. For example:

```
  u = v1;    // Assignment 1
  v1 = v2;   // Assignment 2
  v2 = 0xa;  // Assignment 3
```

eld evaluates the assignments in order, that is, the evaluation happens
as: [Assignment 1, Assignment 2, Assignment 3]. If we follow this
evaluation order, the symbols `u` and `v1` will have incorrect values
because the value of `v2` is unknown when the assignments of `u` and
`v1` are evaluated.

This commit fixes evaluation of expressions containing forward
references by making the below two changes:

1) Assignments which cannot be completely evaluated during the
   sequential evaluation of expressions are marked as pending
   assignments. After the layout is done, but before the relaxation,
   these pending assignments are recursively reevaluated until there
   is no improvement in pending assignments or a MaxIteration limit is
   reached.

2) During a layout iteration, assignments may get evaluated multiple times
   as layout needs to reset itself based on a few conditions
   (new segment needs to be created, and so on...). It is important to reset
   the assignments and the symbol values if a layout gets reset. Let's
   see why it is important with the help of an example:

```
   SECTIONS {
     FOO : {
       u = v; // Assignment 1
       v = 0xa; // Assignment 2
       *(.text.foo)
     }
     BAR : {
       v = 0xb; // Assignment 3
     }
     v = 0xc; // Assignment 4
  }
```

  The sequential assignment evaluation order is: [A1, A2, A3, A4]. When
  A1 is evaluated, `v` is not defined, hence we mark A1 as pending
  assignment. However, if the layout gets reset after evaluating A2,
  then A1 will be evaluated again, but this time, `v` is defined (from
  assignment 2 evaluation) and thus A1 can be completely evaluated.
  This is wrong because A1 should get the value from the assignment 4
  instead of assignment 2!

  We fix this issue by resetting the assignments and the symbol values
  whenever a layout is reset.

  The same issue happens when the layout needs to be recomputed after a
  relaxation pass. And the same solution of resetting the assignments and
  the symbol values works for this case as well.

This commit also adds a new trace category 'pending-assignments' for
tracing pending assignment evaluation.

Closes https://github.com/qualcomm/eld/issues/169

### DIFF
--- a/include/eld/Diagnostics/DiagVerbose.inc
+++ b/include/eld/Diagnostics/DiagVerbose.inc
@@ -163,3 +163,9 @@ DIAG(verbose_linker_script_output_file, DiagnosticEngine::Verbose,
      "Using output file name '%0' specified in the linker script '%1'")
 DIAG(verbose_rule_matching_cache_func_hash, DiagnosticEngine::Verbose,
      "Rule-matching cache-functionailty hash: %0")
+DIAG(verbose_performing_layout_iteration, DiagnosticEngine::Verbose,
+     "Performing layout iteration %0")
+DIAG(verbose_eval_pending_assignments, DiagnosticEngine::Verbose,
+     "Evaluating pending assignments")
+DIAG(verbose_reset_tracked_assignment, DiagnosticEngine::Verbose,
+     "Resetting tracked assignments")
--- a/include/eld/Diagnostics/DiagnosticPrinter.h
+++ b/include/eld/Diagnostics/DiagnosticPrinter.h
@@ -47,6 +47,7 @@ public:
     TraceMergeStrings = 0x8000,
     TraceLinkerScript = 0x10000,
     TraceSymDef = 0x100000,
+    TracePendingAssignments = 0x200000
   };
 
   enum Verbose : uint16_t { None = 0x0, Default = 0x1 };
@@ -67,6 +68,8 @@ public:
   bool traceTrampolines() { return Trace & TraceTrampolines; }
 
   bool traceAssignments() { return Trace & TraceAssignments; }
+
+  bool tracePendingAssignments() { return Trace & TracePendingAssignments; }
 
   bool traceFiles() { return Trace & TraceFiles; }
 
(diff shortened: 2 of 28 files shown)


parth-07:
Thank you for the suggestions and insights.

> -- trace=pending-assignments

I have added the pending-assignments trace.

> linker map file should show what expressions were marked pending

I will update map-file in a follow-up patch. I am concerned that this patch is already doing a lot of things.

> Document the behavior and how users can inspect behavior

I will add the documentation along with the map-file patch.

> Pending assignments that evaluate LOADADDR etc should be marked as always pending

I will make this change in a follow-up patch. I do not want to add too much functionality change in a single PR as that can make it difficult to do testing.

> Every expression that was evaluated pending remains to be pending in every iteration of program layout (need to see test)
> Test cases should use both PHDRs and no PHDRs

I will add more tests for these as part of this pull-request by tomorrow.

parth-07:
This was resolved with a better design/framework as part of this epic: https://github.com/qualcomm/eld/issues/429

PR #291, opened 2025-08-02
Add a plugin API to add files to reproduce tarball

rb1005:
This patch adds a plugin API that would enable plugin users to add plugin generated files to the reproduce tarball. This would improve the debugging workflow and help users identify bugs.


include/eld/PluginAPI/LinkerWrapper.h
```diff
@@ -176,6 +176,8 @@ class DLL_A_EXPORT LinkerWrapper {
   /// \em BeforeLayout and beyond.
   eld::Expected<Symbol> getSymbol(const std::string &Sym) const;
 
+  void createAndAddFileToReproduceTar(std::string FileName);
```
quic-seaswara: - why createAnd ? 

- Just say addFileToReproduceTar

- The function needs to return eld::Expected  (for example if the user tries to run with no reproduce switch)

- const std::string &FileName 

- add Doxygen comments

include/eld/Support/MappingFile.h
```diff
@@ -20,6 +20,7 @@ class MappingFile {
     ObjectFile,
     LinkerScript,
     Plugin,
+    PluginGenerated,
```
quic-seaswara: Can we use Plugin itself ?

lib/LinkerWrapper/LinkerWrapper.cpp
```diff
@@ -1041,6 +1042,13 @@ LinkerWrapper::getUnbalancedChunkAdds() const {
   return unbalancedChunkAdds;
 }
 
+void LinkerWrapper::createAndAddFileToReproduceTar(std::string FileName) {
+  if (!llvm::sys::fs::exists(FileName))
+    return;
+  m_Module.getOutputTarWriter()->createAndAddPluginGeneratedFile(FileName,
+                                                                 FileName);
```
quic-seaswara: same here, dont add create

quic-seaswara (changes requested):

include/eld/PluginAPI/LinkerWrapper.h
Ris-Bali: Have made these changes

include/eld/Support/MappingFile.h
Ris-Bali: I have added "Miscellaneous" for this case as discussed

lib/LinkerWrapper/LinkerWrapper.cpp
```diff
@@ -1041,6 +1042,21 @@ LinkerWrapper::getUnbalancedChunkAdds() const {
   return unbalancedChunkAdds;
 }
 
+eld::Expected<void>
+LinkerWrapper::addFileToReproduceTar(std::string &FileName) {
+  if (!m_Module.getOutputTarWriter())
+    return std::make_unique<eld::plugin::DiagnosticEntry>(
+        Diag::error_reproduce_flag_not_used);
```
quic-seaswara, line 1066: Can you add this test ?

lib/LinkerWrapper/LinkerWrapper.cpp
```diff
@@ -1041,6 +1042,21 @@ LinkerWrapper::getUnbalancedChunkAdds() const {
...
+  if (!llvm::sys::fs::exists(FileName))
+    return std::make_unique<eld::plugin::DiagnosticEntry>(
+        Diag::error_plugin_file_does_not_exist,
+        std::vector<std::string>{FileName});
+  if (m_Module.getConfig().options().hasMappingFile())
+    FileName = m_Module.getConfig().getHashFromFile(FileName);
```
quic-seaswara: Why are we using Mapping file here ?

lib/Support/OutputTarWriter.cpp
```diff
@@ -218,6 +218,12 @@ void OutputTarWriter::createAndAddConfigFile(std::string Filename,
                    /*isLTO*/ false);
 }
 
+void OutputTarWriter::addPluginGeneratedFile(std::string Filename,
+                                             std::string ResolvedPath) {
```
quic-seaswara: Do you need two parameters ?

Ris-Bali: No I have removed the extra parameter

lib/LinkerWrapper/LinkerWrapper.cpp
Ris-Bali: I have added this in "test/Common/Plugin/AddPluginFileToReproduceTar/AddPluginToReproduceFlagMissing.test" in this patch.

lib/LinkerWrapper/LinkerWrapper.cpp
Ris-Bali: It may be useful if reproduce is used again in the tar. Its an edge case but for most cases this won't be used.

quic-seaswara: Let us not have this corner case. Running reproducer from a reproducer is something we can warn or error out.

include/eld/PluginAPI/LinkerWrapper.h
```diff
@@ -462,6 +471,8 @@ class DLL_A_EXPORT LinkerWrapper {
   /// be the file path returned by the `LinkerWrapper::findConfigFile`.
   std::string getFileContents(std::string FileName);
 
+  std::optional<eld::plugin::MemoryBuffer> getBuffer(std::string FileName);
```
quic-seaswara: const function

quic-seaswara: const std::string &FileName

include/eld/Support/OutputTarWriter.h
```diff
@@ -76,6 +76,8 @@ class OutputTarWriter {
 
   void createAndAddScriptFile(std::string filename, std::string resolvedPath);
 
+  void addPluginGeneratedFile(std::string filename);
```
quic-seaswara: const std::string &filename

include/eld/PluginAPI/LinkerWrapper.h
Ris-Bali: Can't do a const string since FileName is being modified in getBuffer().

lib/LinkerWrapper/LinkerWrapper.cpp
```diff
@@ -645,6 +646,23 @@ std::string LinkerWrapper::getFileContents(std::string FileName) {
   return buf.getContents().str();
 }
 
+std::optional<eld::plugin::MemoryBuffer>
+LinkerWrapper::getBuffer(std::string FileName) const {
```
quic-seaswara, line 650: const std::string &Filename

quic-seaswara (approved):

PR #293, opened 2025-08-04
Skip invalid ASCII chars when parsing linker script

### LINKED ISSUE #292: unknown directive when parsing linker script
[script.t.gz](https://github.com/user-attachments/files/21566525/script.t.gz)

```
cat > 1.c << \!
int foo() { return 0; }

clang -c -target hexagon 1.c
gunzip script.t.gz
ld.eld -T script.t 1.o

```

**eld**

```
ld.eld 1.o -T script.t 
Error: script.t:16: unknown directive:
>>> }
>>>  ^
Fatal: Linker script script.t has errors, Please check
```

**ld.bfd**

```
$ riscv-ld 1.o -T script.t
riscv-ld:script.t:16: ignoring invalid character `\000' in script
```

parth-07:
This commit fixes unknown linker script directive error for ascii non-printable characters by skipping invalid chars when parsing linker scripts.

Resolves #292


lib/ScriptParser/ScriptLexer.cpp
```diff
@@ -349,6 +352,29 @@ ScriptLexer::noteAndSkipNonASCIIUnicodeChars(llvm::StringRef s) const {
   return s;
 }
 
+llvm::StringRef
+ScriptLexer::noteAndSkipInvalidASCIIChar(llvm::StringRef s) const {
+  if (s.empty() || !isASCIIControlChar(s[0]))
+    return s;
+  setNote("Ignoring invalid ACII character '" + convertToOct(s[0]) + "'", s);
```
quic-seaswara: nit : ASCII character

lib/ScriptParser/ScriptLexer.cpp
```diff
@@ -349,6 +352,29 @@ ScriptLexer::noteAndSkipNonASCIIUnicodeChars(llvm::StringRef s) const {
...
+  s = s.drop_front();
+  return s;
+}
+
+bool ScriptLexer::isASCIIControlChar(char c) const {
+  int ASCIIControlCharEnd = 0x1F;
+  int ASCIIControlCharStart = 0x00;
+  if (c >= ASCIIControlCharStart && c <= ASCIIControlCharEnd)
```
quic-seaswara: Can we use https://en.cppreference.com/w/c/string/byte/isprint ?

lib/ScriptParser/ScriptLexer.cpp
```diff
@@ -234,6 +234,9 @@ StringRef ScriptLexer::skipSpace(StringRef S) {
     StringRef Saved = S;
     S = noteAndSkipNonASCIIUnicodeChars(S);
     S = S.ltrim();
+    // No note will be emitted for whitespace characters as they are already
+    // skipped as part of 'S = S.ltrim()' above.
+    S = noteAndSkipInvalidASCIIChar(S);
```
quic-seaswara: Can we simply skip any characters that we dont recognize instead of this Invalid ASCII character ?

if (!isprint(ch))
continue;

lib/ScriptParser/ScriptLexer.cpp
parth-07: Thank you for mentioning this. I did not know such as API existed. I have simplified the code by using this API.

lib/ScriptParser/ScriptLexer.cpp
parth-07: We can just skip the character. This function skips the character and emit a note for the same. We decided to keep the note diagnostic for diagnostic compatibility with GNU ld.

lib/ScriptParser/ScriptLexer.cpp
```diff
@@ -349,6 +351,21 @@ ScriptLexer::noteAndSkipNonASCIIUnicodeChars(llvm::StringRef s) const {
...
+llvm::StringRef
+ScriptLexer::noteAndSkipInvalidASCIIChar(llvm::StringRef s) const {
+  if (s.empty() || llvm::isPrint(s[0]) || isNonASCIIUnicode(s[0]))
+    return s;
+  setNote("Ignoring invalid ASCII character '" + convertToOct(s[0]) + "'", s);
+  s = s.drop_front();
+  return s;
+}
+
+std::string ScriptLexer::convertToOct(char c) const {
```
quic-seaswara: unsigned char c ?

lib/ScriptParser/ScriptLexer.cpp
```diff
@@ -349,6 +351,21 @@ ScriptLexer::noteAndSkipNonASCIIUnicodeChars(llvm::StringRef s) const {
...
+  if (s.empty() || llvm::isPrint(s[0]) || isNonASCIIUnicode(s[0]))
+    return s;
+  setNote("Ignoring invalid ASCII character '" + convertToOct(s[0]) + "'");
+  s = s.drop_front();
+  return s;
+}
+
+std::string ScriptLexer::convertToOct(char c) const {
+  char buf[10];
+  sprintf(buf, "\\%03o", static_cast<unsigned char>(c));
```
quic-seaswara: Can we use the below instead of 'C' ?

```
std::ostringstream oss;
oss << "\\" << std::oct << std::setw(3) << std::setfill('0') << static_cast<int>(static_cast<unsigned char>(c));
return oss.str();
```

Also why octal ?

parth-07: I do not mind using the ostringstream here but it seems a lot more complex than `sprintf`. What do you think about this?

I am using octal because GNU ld uses octal for escape codes in the diagnostics. I suppose we can use hex value as well.

lib/ScriptParser/ScriptLexer.cpp
parth-07: This API is being called with `char` value and the parser tokens are char as well. I think it is better to have `static_cast` inside the function call as opposed to outside. Please let me know your thoughts on this.

lib/ScriptParser/ScriptLexer.cpp
quic-seaswara: hex values are much better IMO easy to debug. 

You might get a different result if the character value goes above 127. I think it would need to be unsigned int.

Prefer c++ over C :) Since this happens only if you have a invalid character, I think its ok to use C++ :)

parth-07: Thank you for the suggestions. I have applied them.

quic-seaswara (approved):

PR #300, opened 2025-08-05
RISC-V: cleaning up applying relocations

quic-akaryaki:
The main point of this PR is to separate address computation and instruction encoding steps.

### DIFF
--- a/lib/Target/RISCV/RISCVHelper.h
+++ b/lib/Target/RISCV/RISCVHelper.h
@@ -210,13 +210,11 @@ template <typename T> T getPCRELLO(T Value, T PC) {
 /// The following instruction types are supported
 /// I, S, SB, U, UJ, RVC, RVC(LUI), RVC(B), RVC(J)
 ///------------------------------------------
-template <typename T> T encodeI(T Result) {
-  Result &= 0xFFF;
-  Result <<= 20;
-  return Result;
+inline int32_t encodeI(int32_t Result) {
+  return (Result & 0xfff) << 20;
 }
 
-template <typename T> T encodeS(T &Result) {
+inline int32_t encodeS(int32_t Result) {
   uint32_t Imm11_5 = extractBits(Result, 11, 5) << 25;
   uint32_t Imm4_0 = extractBits(Result, 4, 0) << 7;
   Result = Imm11_5 | Imm4_0;
@@ -241,9 +239,8 @@ template <typename T> T encodeUJ(T Result) {
   return Result;
 }
 
-template <typename T> T encodeU(T Result) {
-  Result &= 0xFFFFF000;
-  return Result;
+inline int32_t encodeU_HI20(int32_t Result) {
+  return (Result + 0x800) & 0xFFFFF000;
 }
 
 template <typename T> T encodeU_ABS20(T Result) {
--- a/lib/Target/RISCV/RISCVRelocator.cpp
+++ b/lib/Target/RISCV/RISCVRelocator.cpp
@@ -891,8 +891,8 @@ RISCVRelocator::Result applyHI(Relocation &pReloc, RISCVLDBackend &Backend,
     // b. If the relocation overflows PCREL
     //             and
     // c. if the relocation would fit within LUI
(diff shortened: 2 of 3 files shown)


lib/Target/RISCV/RISCVHelper.h
```diff
@@ -210,13 +210,15 @@ template <typename T> T getPCRELLO(T Value, T PC) {
...
 /// I, S, SB, U, UJ, RVC, RVC(LUI), RVC(B), RVC(J)
 ///------------------------------------------
-template <typename T> T encodeI(T Result) {
-  Result &= 0xFFF;
-  Result <<= 20;
-  return Result;
+inline int32_t encodeI(int32_t Result) {
+  int32_t Hi20 = (Result + 0x800) & 0xfffff000;
+  Result -= Hi20;
+  return (Result & 0xfff) << 20;
```
lenary: This seems a really complex way of calculating this - I think it should just be `return (Result & 0xfff) << 20;`

lib/Target/RISCV/RISCVHelper.h
```diff
@@ -210,13 +210,15 @@ template <typename T> T getPCRELLO(T Value, T PC) {
...
 }
 
-template <typename T> T encodeS(T &Result) {
+inline int32_t encodeS(int32_t Result) {
+  int32_t Hi20 = (Result + 0x800) & 0xfffff000;
+  Result -= Hi20;
   uint32_t Imm11_5 = extractBits(Result, 11, 5) << 25;
   uint32_t Imm4_0 = extractBits(Result, 4, 0) << 7;
   Result = Imm11_5 | Imm4_0;
```
lenary: Again, this seems more complex than necessary. You should just be able to extract the bits, from `Result` and insert them into the right place of the result

lib/Target/RISCV/RISCVRelocator.cpp
```diff
@@ -868,47 +880,33 @@ RISCVRelocator::Result applyRel(Relocation &pReloc, RISCVLDBackend &Backend,
...
-RISCVRelocator::Result applyHI(Relocation &pReloc, RISCVLDBackend &Backend,
-                               RelocationDescription &pRelocDesc) {
-  if (RelocDescs.count(pReloc.type()) == 0)
-    return RISCVRelocator::Unsupport;
-
-  int64_t S = Backend.getSymbolValuePLT(pReloc);
-  int64_t A = pReloc.addend();
-  int64_t Result = S + A + 0x800;
-
+  // TODO: This should be moved to the separate relaxation/transformation pass.
```
lenary, line 885: Agreed 👍

lib/Target/RISCV/RISCVHelper.h
quic-akaryaki: I think it can be `((Result + 0x800) & 0xfff) << 20`, but I also think it's less clear because `(Result + 0x800) & 0xfff` essentially has the sign bit in the middle of the number. `Result - Hi20` is a normal signed number, which will later be truncated.

lenary: Maybe I'm missing a `llvm::SignExtend32<12>(Result)` in my expression (you'll need LLVM's MathExtras.h)? I don't like the random adding of `0x800` when dealing with the lower part of a hi20/lo12 pair, which is really accounting for the sign overflow in the high part.

quic-akaryaki: You are right, I missed a negation so 0x800 gets cancelled but it's not obvious.
I wrote the code to follow what CPU does to compute the effective address lo12 = EA - hi20.
And the "random" 0x800 is still there in the hi20 computation.

lib/Target/RISCV/RISCVRelocator.cpp
quic-akaryaki: I simplified it in the latest forced push.

quic-akaryaki:
Sorry for spamming, I added another NFC here. It is in the the final form now (subject to testing etc).

lenary (approved):
I think this LGTM, but it needs more eyes.

PR #304, opened 2025-08-06
Use Expression::resetOrZero for printing map-file

parth-07:
This commit replaces the use of `Expression::result` with `Expression::resultOrZero` in `TextLayoutPrinter` and `LayoutWrapper`. This is because map-file can be printed (or layout information can be requested) even when the link has failed before expression evaluation.

### DIFF
--- a/lib/LinkerWrapper/LayoutWrapper.cpp
+++ b/lib/LinkerWrapper/LayoutWrapper.cpp
@@ -51,7 +51,7 @@ LayoutWrapper::getPaddings(eld::plugin::OutputSection &Section) {
     if (!(p.endOffset - p.startOffset))
       continue;
     recordPadding(paddings, p.startOffset, p.endOffset - p.startOffset,
-                  p.Exp ? p.Exp->result() : 0, /*isAlignment*/ false);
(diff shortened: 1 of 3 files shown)


lib/LinkerWrapper/LayoutWrapper.cpp
```diff
@@ -83,7 +83,7 @@ LayoutWrapper::getPaddings(eld::plugin::OutputSection &Section) {
                NextRuleWithContent ? NextRuleWithContent->getFirstFrag()
                                    : nullptr)) {
         recordPadding(paddings, p.startOffset, p.endOffset - p.startOffset,
-                      p.Exp ? p.Exp->result() : 0, /*isAlignment*/ false);
+                      p.Exp ? p.Exp->resultOrZero() : 0, /*isAlignment*/ false);
```
quic-seaswara, line 86: Do we have a case when expression is not evaluated ?

quic-seaswara: @parth-07  please add a test post merge

quic-seaswara (approved):

PR #305, opened 2025-08-06
[ARM][AArch64] Fix computation of local-exec TLS relocations

### LINKED ISSUE #302: Incorrect offsets for TLS accesses on AArch64
picolibc's TLS tests (https://github.com/picolibc/picolibc/blob/main/test/tls.c) are currently failing when built with eld. I think this is due to incorrect offsets being used when accessing TLS variables where other TLS variables with sufficiently large alignments are present.

It looks like eld is always using a 16-byte offset (for the TCB I guess) rather than also accounting for possible alignment padding which is what lld/bfd/picolibc all seem to be doing or expecting. I'm not sure what the official source is, but a few possibly-relevant snippets from codebases about this:
* lld: https://github.com/llvm/llvm-project/blob/4c2d56318fec16d1d5241cd91d38b59cb65bce83/lld/ELF/InputSection.cpp#L726-L729
* picolibc: https://github.com/picolibc/picolibc/blob/c623d2b6001f41e2d5bfcd91943ff1ead960f3d5/picolibc.ld.in#L271C6-L271C74 and https://github.com/picolibc/picolibc/blob/c623d2b6001f41e2d5bfcd91943ff1ead960f3d5/picocrt/machine/aarch64/crt0.c#L60

**Reproducer**
This isn't exactly a full running/working reproducer, but is hopefully sufficient. picolibc's test is a full running example too
```
cat > test.c << '!'
__thread int a = 12;
__attribute__((aligned(128)))
__thread int b = 13;

int foo(void) { return a; }
!

clang --target=aarch64-none-elf -c test.c
ld.eld -maarch64linux test.o -o eld.a.out
ld.lld test.o -o lld.a.out
ld.bfd test.o -o bfd.a.out
```

**Output**
```
eld:
  400120: d53bd048      mrs     x8, TPIDR_EL0
  400124: 91400108      add     x8, x8, #0x0, lsl #12   // =0x0
  400128: 91004108      add     x8, x8, #0x10

lld/bfd (output snippet here is effectively the same between the two linkers, just different addresses):
  2101c8: d53bd048      mrs     x8, TPIDR_EL0
  2101cc: 91400108      add     x8, x8, #0x0, lsl #12   // =0x0
  2101d0: 91020108      add     x8, x8, #0x80
```

parth-07:
This commit fixes computation of local-exec TLS relocations for
ARM and AArch64. Previously, we were not using the correct start offset
in static thread local storage block for computing the relocation.

Lets understand this with the help of an example

```

             <--------- Static thread local storage block  ------------------>
| --------- | ----------------- | --------------------- | ----------------------- |
     TCB    TP  Reserved space      Alignment padding      Thread variables

```
ARM/AArch64 local-exec relocation resolves to:
  SymbolValueInTLSSegment + Reserved Space + Alignment padding

Previously, we were not accounting for the alignment padding. But why is
this alignment padding required? It is required so that the symbol
alignments assumed at the compile/link time is satisfied at runtime as
well. The exact condition that should be satisfied is:

&nbsp;&nbsp; VMA(TP) + Reserved space + Alignment padding congruent to TLSSegment[vaddr] (modulo TLSSegment[align])

This condition becomes more clear with the help of an example:

&nbsp;&nbsp;  StartVMAForStaticThreadVariables = VMA(TP) + Reserved space + Alignment padding

&nbsp;&nbsp;  StartVMAForStaticThreadVariables should be congruent to TLSSegment[vaddr] (modulo TLSSegment[align])

Resolves #302

### DIFF
--- a/lib/Target/AArch64/AArch64Relocator.cpp
+++ b/lib/Target/AArch64/AArch64Relocator.cpp
@@ -1031,7 +1031,14 @@ Relocator::Result tls_gottprel_lo(Relocation &pReloc,
 // R_AARCH64_TLSLE_ADD_TPREL_LO12
 // R_AARCH64_TLSLE_ADD_TPREL_LO12_NC : TPREL(S+A)
 Relocator::Result tls_tprel(Relocation &pReloc, AArch64Relocator &pParent) {
-  Relocator::DWord X = pParent.getSymValue(&pReloc) + 0x10;
+  const auto &OptTLSBlockVarOffset = pParent.getStaticTLSBlockVarOffset();
+  if (!OptTLSBlockVarOffset.has_value()) {
+    pParent.config().raise(Diag::no_pt_tls_segment);
+    return Relocator::Result::BadReloc;
+  }
+
+  Relocator::DWord X =
+      pParent.getSymValue(&pReloc) + OptTLSBlockVarOffset.value();
 
   if (pReloc.type() == llvm::ELF::R_AARCH64_TLSLE_ADD_TPREL_HI12) {
     if (X >= 0x1000000)
@@ -1142,4 +1149,17 @@ Relocator::Result copyInstruction(Relocation &pReloc,
   std::memcpy((void *)&insn, data, AArch64InsnHelpers::InsnSize);
   pReloc.target() = insn;
   return Relocator::OK;
+}
+
+void AArch64Relocator::computeTLSOffsets() {
+  const uint32_t WordSize = 0x8;
+  auto OptFirstTLSSegVirtualAddr = getFirstTLSSegmentVirtualAddr();
+  auto OptMaxTLSSegAlignment = getMaxTLSSegmentAlign();
+  if (!OptFirstTLSSegVirtualAddr || !OptMaxTLSSegAlignment)
+    return;
+  auto FirstTLSSegVirtualAddr = OptFirstTLSSegVirtualAddr.value();
+  auto MaxTLSSegAlignment = OptMaxTLSSegAlignment.value();
+  StaticTLSBlockVarOffset = 2 * WordSize;
+  *StaticTLSBlockVarOffset +=
+      ((FirstTLSSegVirtualAddr - 2 * WordSize) & (MaxTLSSegAlignment - 1));
 }
--- a/lib/Target/AArch64/AArch64Relocator.h
+++ b/lib/Target/AArch64/AArch64Relocator.h
@@ -75,6 +75,12 @@ public:
   void partialScanRelocation(Relocation &pReloc,
                              const ELFSection &pSection) override;
 
+  void computeTLSOffsets() override;
+
+  std::optional<uint64_t> getStaticTLSBlockVarOffset() const {
+    return StaticTLSBlockVarOffset;
+  }
+
 private:
   bool isInvalidReloc(Relocation &pType) const;
   void scanLocalReloc(InputFile &pInput, Relocation &pReloc,
@@ -86,6 +92,11 @@ private:
 
 private:
   AArch64GNUInfoLDBackend &m_Target;
+  /// The static TLS block contains an optional gap at the beginning,
+  /// that is followed by an optional alignment padding. The TLS variables
+  /// are stored after the alignment padding. This member stores the
+  /// offset in the static TLS block from where the variables start.
+  std::optional<uint64_t> StaticTLSBlockVarOffset;
 };
 
 } // namespace eld
--- a/lib/Target/ARM/ARMRelocator.cpp
+++ b/lib/Target/ARM/ARMRelocator.cpp
@@ -1590,9 +1590,15 @@ Relocator::Result tls_ldo32(Relocation &pReloc, ARMRelocator &pParent) {
   return Relocator::OK;
 }
 
-// R_ARM_TLS_LE32: S + A - tp
+// R_ARM_TLS_LE32: S + A + 2 * WordSize +
 Relocator::Result tls_le32(Relocation &pReloc, ARMRelocator &pParent) {
-  pReloc.target() += pParent.getSymValue(&pReloc) + pReloc.addend() + 0x8;
+  const auto &OptTLSBlockVarOffset = pParent.getStaticTLSBlockVarOffset();
+  if (!OptTLSBlockVarOffset.has_value()) {
+    pParent.config().raise(Diag::no_pt_tls_segment);
+    return Relocator::Result::BadReloc;
+  }
+  pReloc.target() += pParent.getSymValue(&pReloc) + pReloc.addend() +
+                     OptTLSBlockVarOffset.value();
   return Relocator::OK;
 }
 
@@ -1628,4 +1634,17 @@ Relocator::Result relocLDR12(Relocation &pReloc, ARMRelocator &pParent) {
   Relocator::DWord X = (S + A) - pReloc.place(pParent.module());
   pReloc.target() = pReloc.target() | (X & 0xfff);
   return Relocator::OK;
+}
+
+void ARMRelocator::computeTLSOffsets() {
+  const uint32_t WordSize = 0x4;
+  auto OptFirstTLSSegVirtualAddr = getFirstTLSSegmentVirtualAddr();
+  auto OptMaxTLSSegAlignment = getMaxTLSSegmentAlign();
+  if (!OptFirstTLSSegVirtualAddr || !OptMaxTLSSegAlignment)
+    return;
+  auto FirstTLSSegVirtualAddr = OptFirstTLSSegVirtualAddr.value();
+  auto MaxTLSSegAlignment = OptMaxTLSSegAlignment.value();
+  StaticTLSBlockVarOffset = 2 * WordSize;
+  *StaticTLSBlockVarOffset +=
+      ((FirstTLSSegVirtualAddr - 2 * WordSize) & (MaxTLSSegAlignment - 1));
 }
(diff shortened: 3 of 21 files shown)


lib/Target/AArch64/AArch64Relocator.cpp
```diff
@@ -1031,7 +1031,21 @@ Relocator::Result tls_gottprel_lo(Relocation &pReloc,
...
+  auto OptMaxTLSSegAlignment = pParent.getMaxTLSSegmentAlign();
+  if (!OptFirstTLSSegVirtualAddr.has_value() ||
+      !OptMaxTLSSegAlignment.has_value()) {
+    pParent.config().raise(Diag::no_pt_tls_segment);
+    return Relocator::Result::BadReloc;
+  }
+
+  auto FirstTLSSegVirtualAddr = OptFirstTLSSegVirtualAddr.value();
+  auto MaxTLSSegAlignment = OptMaxTLSSegAlignment.value();
+  X += ((FirstTLSSegVirtualAddr - 2 * WordSize) & (MaxTLSSegAlignment - 1));
```
quic-seaswara: This can go to AArch64Relocator, less duplication of code

lib/Target/ARM/ARMRelocator.cpp
```diff
@@ -1590,9 +1590,21 @@ Relocator::Result tls_ldo32(Relocation &pReloc, ARMRelocator &pParent) {
...
+  auto OptMaxTLSSegAlignment = pParent.getMaxTLSSegmentAlign();
+  if (!OptFirstTLSSegVirtualAddr.has_value() ||
+      !OptMaxTLSSegAlignment.has_value()) {
+    pParent.config().raise(Diag::no_pt_tls_segment);
+    return Relocator::Result::BadReloc;
+  }
+  auto FirstTLSSegVirtualAddr = OptFirstTLSSegVirtualAddr.value();
+  auto MaxTLSSegAlignment = OptMaxTLSSegAlignment.value();
+  pReloc.target() +=
+      ((FirstTLSSegVirtualAddr - 2 * WordSize) & (MaxTLSSegAlignment - 1));
```
quic-seaswara: This also affects other models IE, LD and GD. 

It makes sense to add a function getTLSOffset() in AArch64Relocator.

lib/Target/Relocator.cpp
```diff
@@ -294,3 +295,22 @@ Relocation::Address Relocator::getSymValue(const Relocation *R) {
     return getTarget().finalizeTLSSymbol(R->symInfo()->outSymbol());
   return R->symValue(m_Module);
 }
+
+std::optional<uint64_t> Relocator::getFirstTLSSegmentVirtualAddr() const {
+  const auto &ELFSegTable = getTarget().elfSegmentTable();
+  const auto &TLSSegments = ELFSegTable.getSegments(llvm::ELF::PT_TLS);
```
quic-seaswara, line 301: Can we cache the pointer to the PT_TLS segment in GNULDBackend instead of getting the segments all the time.

quic-seaswara (commented):
Please add tests

lib/Target/Relocator.cpp
parth-07: We cannot cache PT_TLS segment in GNULDBackend as there can be multiple relevant PT_TLS segment, but we can cache the FirstTLSSegmentVirtualAddress and MaxTLSSegmentAlignment. I am changing the code to cache these two values.

quic-seaswara: You can store the cache the TLS segments as a vector in GNULDBackend.

lib/Target/ARM/ARMRelocator.cpp
parth-07: It seems that the other models except local-exec does not access static thread local storage block.

lib/Target/Relocator.cpp
quic-seaswara: Thanks!

lib/Target/AArch64/AArch64Relocator.cpp
parth-07: I have modified the code such that we do not compute TLS offset for each relocation.

lib/Target/Relocator.cpp
parth-07: We now compute TLS offset only once, thus, we do not make so many calls to `ELFSegTable.getSegments(llvm::ELF::PT_TLS)`. I have avoided caching TLS segment variables because I believe the gain is not too much now, and I was worried that we may accidentally use stale value of cached TLS segments at someplace.

lib/Target/AArch64/AArch64Relocator.h
```diff
@@ -75,6 +75,12 @@ class AArch64Relocator : public Relocator {
   void partialScanRelocation(Relocation &pReloc,
                              const ELFSection &pSection) override;
 
+  void computeTLSOffsets() override;
+
+  std::optional<uint64_t> getStaticTLSBlcokVarOffset() const {
```
quic-seaswara: nit: getStaticTLSBlock

lib/Target/Relocator.cpp
```diff
@@ -294,3 +295,22 @@ Relocation::Address Relocator::getSymValue(const Relocation *R) {
...
+  if (TLSSegments.empty())
+    return std::nullopt;
+  return TLSSegments[0]->vaddr();
```
quic-seaswara, line 304: This function should live in GNULDBackend

lib/Target/Relocator.cpp
```diff
@@ -294,3 +295,22 @@ Relocation::Address Relocator::getSymValue(const Relocation *R) {
...
+std::optional<uint64_t> Relocator::getMaxTLSSegmentAlign() const {
+  const auto &ELFSegTable = getTarget().elfSegmentTable();
+  const auto &TLSSegments = ELFSegTable.getSegments(llvm::ELF::PT_TLS);
+  if (TLSSegments.empty())
+    return std::nullopt;
+  uint64_t TLSAlign = 0;
+  for (const auto &S : TLSSegments)
+    TLSAlign = std::max(TLSAlign, S->align());
+  return TLSAlign;
+}
```
quic-seaswara, line 316: This function too should live in GNULDBackend

test/AArch64/standalone/TLSLocalExec/TLSLocalExec.test
```diff
@@ -0,0 +1,15 @@
+#---TLSLocalExec.test--------------------------- PartialLink -----------------#
+#BEGIN_COMMENT
+# This tests checks that eld correctly computes TLS local-exec
+# relocation computation when the TLS segment alignment is greater
+# than 16.
+#END_COMMENT
+#START_TEST
+RUN: %clang %clangopts -o %t1.1.o %p/Inputs/1.c -c
+RUN: %link %linkopts -o %t1.1.out %t1.1.o --defsym __aeabi_read_tp=0
```
quic-seaswara: Need test with linker script too

test/ARM/standalone/TLSLocalExec/TLSLocalExec.test
```diff
@@ -0,0 +1,15 @@
+#---TLSLocalExec.test--------------------------- PartialLink -----------------#
+#BEGIN_COMMENT
+# This tests checks that eld correctly computes TLS local-exec
+# relocation computation when the TLS segment alignment is greater
+# than 16.
+#END_COMMENT
+#START_TEST
+RUN: %clang %clangopts -o %t1.1.o %p/Inputs/1.c -c
+RUN: %link %linkopts -o %t1.1.out %t1.1.o --defsym __aeabi_read_tp=0
```
quic-seaswara, line 9: Same here need test with linker script too

quic-seaswara: add tests like the below using explicit linker script ALIGN's /SUBALIGN that change the alignment of the section.

I have a thought that the computeTLSOffsets is happening too early

quic-seaswara (changes requested):

quic-akaryaki:
Is this problem specific to aarch64?

quic-seaswara:
> Is this problem specific to aarch64?

ARM/AArch64

lib/Target/AArch64/AArch64Relocator.h
parth-07: I do not want to name it getStaticTLSBlock because that would be misleading. Static TLS block contains 2 reserved word bytes, padding alignment, and then TLS data. I want this variable to indicate that it contains the offset to the TLS data/variables inside the static TLS block.

lib/Target/Relocator.cpp
parth-07: This function is currently not required in GNULDBackend. GNULDBackend is more accessible than Relocator. I want to keep the functionality close / restricted to where it is required. Can we please move it to GNULDBackend later when it seems that a similar feature is required at other places too?

lib/Target/Relocator.cpp
parth-07: Can we please move it to GNULDBackend later when it seems that a similar feature is required at other places too?

lib/Target/AArch64/AArch64Relocator.h
quic-seaswara: I meant there is a spell check error here 

getStaticTLSBlcokVarOffset -> getStaticBlockVarOffset

lib/Target/Relocator.cpp
quic-seaswara: This should really live in GNULDBackend. We can move it later too.

test/ARM/standalone/TLSLocalExec/TLSLocalExec.test
parth-07: I have add more tests.

quic-seaswara (approved):

PR #307, opened 2025-08-06
[RISCV] Add support for TLSDESC

quic-akaryaki:


lib/Target/GNULDBackend.cpp
```diff
@@ -3938,6 +3938,10 @@ bool GNULDBackend::relax() {
         if (hasError)
           m_Module.setFailure(true);
       }
+      if (LinkerConfig::Object != config().codeGenType()) {
+        if (!setupProgramHdrs())
+          m_Module.setFailure(true);
```
partaror, line 4089: Should we break from the loop as well if setupProgramHdrs fails?

lib/Target/GNULDBackend.cpp
```diff
@@ -3938,6 +3938,10 @@ bool GNULDBackend::relax() {
         if (hasError)
           m_Module.setFailure(true);
       }
+      if (LinkerConfig::Object != config().codeGenType()) {
+        if (!setupProgramHdrs())
```
partaror: Does it makes sense to call this function once the layout is finalized?

quic-seaswara: Good observation, Parth. Why are we doing this before layout is complete ?

lib/Target/GNULDBackend.cpp
quic-akaryaki: I guess.

lib/Target/GNULDBackend.cpp
quic-akaryaki: We are doing this because we need to know values of symbols during relaxations.

quic-akaryaki:
Looks like the build failed because upstream LLVM does not emit R_RISCV_ALIGN.

lib/Target/GNULDBackend.cpp
quic-akaryaki: For some reason, TLS symbol values are computed using TLS segment addresses, which are not valid before this function is called.

quic-seaswara: ah ok. Now I understand.

All TLS symbols are accessed TP relative. This is why we calculate TLS symbol values relative to PT_TLS.

quic-seaswara: Thanks for the explanation

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -1428,16 +1428,18 @@ bool RISCVLDBackend::handlePendingRelocations(ELFSection *section) {
...
                           std::get<3>(r), std::get<4>(r), /*pLastVisit*/ true))
       return false;
 
   // Sort the relocation table, in offset order, since the pending relocations
   // that got added at end of the relocation table may not be in offset order.
+  // Another reason to sort relocations is to make sure R_RISCV_RELAX are
+  // adjacent to the relocation they affect. The psABI is not clear if
+  // R_RISCV_RELAX must be adjacent, but using `.reloc` in assembly may generate
+  // those that are not. Note that because of this, this function must not have
+  // earlier non-error exists.
```
quic-seaswara: nit : exits

test/AArch64/standalone/BSS_none_bss/BSS_none_bss.yaml
```diff
@@ -1,3 +1,11 @@
+# RUN: %yaml2obj %s -o %t.o
+# RUN: %link %linkopts -march aarch64 -T %p/t.t %t.o -o %t.out
+# RUN: llvm-objdump -s %t.out | %filecheck %s
```
quic-seaswara: %objdump

lib/Target/RISCV/RISCVRelocator.cpp
```diff
@@ -446,6 +450,54 @@ void RISCVRelocator::scanRelocation(Relocation &pReloc, eld::IRBuilder &pLinker,
     if (!section->isAlloc())
       return;
 
+    ELFObjectFile *Obj = llvm::dyn_cast<ELFObjectFile>(&pInputFile);
+
+    // Common relocation processing for both local and global symbols.
+    switch (pReloc.type()) {
```
quic-seaswara, line 458: probably making this to be a function and calling this from local and global reloc would make it easy since its a common pattern that we follow across targets.

lib/Target/RISCV/RISCVRelocator.cpp
```diff
@@ -446,6 +450,54 @@ void RISCVRelocator::scanRelocation(Relocation &pReloc, eld::IRBuilder &pLinker,
...
+    case llvm::ELF::R_RISCV_TLSDESC_HI20:
+    case llvm::ELF::R_RISCV_TLSDESC_LOAD_LO12:
+    case llvm::ELF::R_RISCV_TLSDESC_ADD_LO12:
+    case llvm::ELF::R_RISCV_TLSDESC_CALL: {
+      std::lock_guard<std::mutex> relocGuard(m_RelocMutex);
+      if (config().isBuildingExecutable()) {
+        // Non-preemptible symbols in executables will be optimized or relaxed,
+        // no GOT needed.
+        if (!m_Target.isSymbolPreemptible(*rsym))
+          return;
```
quic-seaswara, line 468: It looks like we dont do anything for hidden symbols / protected symbols that may get resolved as HIDDEN/PROTECTED across modules ?

cat > 1.c << \!
__thread int tls_var = 0;
int foo() { return tls_var; }
!

cat > 2.c << \!
extern __thread int tls_var;
int bar() { return tls_var; }
!

clang -c -target riscv32 1.c -fPIC -mtls-dialect=desc -fvisibility=hidden
clang -c -target riscv32 2.c -fPIC -mtls-dialect=desc 
riscv-ld 1.o 2.o

test/lld/ELF/riscv-tlsdesc-relax.test
```diff
@@ -0,0 +1,178 @@
+REQUIRES: riscv32 || riscv64
+
+RUN: %llvm-mc -filetype=obj --defsym PAD=0 -mattr=+c,+relax %p/Inputs/riscv-tlsdesc-relax/a.s -o %t.a.o
+RUN: %llvm-mc -filetype=obj --defsym PAD=5000 -mattr=+c,+relax %p/Inputs/riscv-tlsdesc-relax/a.s -o %t.aa.o
+RUN: %llvm-mc -filetype=obj -mattr=+c,+relax %p/Inputs/riscv-tlsdesc-relax/c.s -o %t.c.o
+RUN: %link %linkopts -shared %t.c.o -o %t.c.so
```
quic-seaswara, line 6: May be useful to add tests using 'C' code to make it easy to debug ?

lib/Target/RISCV/RISCVRelocator.cpp
quic-akaryaki: Actually I wanted to merge local/global functions because there is a lot duplication between them. We have had bugs in one of the branches because it's written a little different than the other.

lib/Target/RISCV/RISCVRelocator.cpp
quic-akaryaki: They should not work differently because the code does not check for it. If there is a bug in isSymbolPreemptible(), then the symbol will be wrongly preemptible or not but TLS DESC code will still work correctly.

test/lld/ELF/riscv-tlsdesc-relax.test
quic-akaryaki: The test check for particular instruction at particular offsets which C compiler will not guarantee. Also, some tests have interleave patterns that compiler probably can never generate.

lib/Target/RISCV/RISCVLDBackend.cpp
quic-akaryaki: Sure

test/AArch64/standalone/BSS_none_bss/BSS_none_bss.yaml
quic-akaryaki: Sure (it was the original code).

lib/Target/RISCV/RISCVRelocator.cpp
quic-seaswara: Sure

quic-seaswara:
I would like to see tests using 'C' code as well. You can still keep the asm test.

Please add tests that include 

* Hidden visibility (build shared objects)
* Test cases with partial linking
* Test cases with PIE 
* Test cases with mno-relax

```
cat > 1.c << \!
__attribute__((visibility("hidden"))) __thread int tls_var = 10;
int foo() { return tls_var; }
!

cat > 2.c << \!
extern __thread int tls_var;
int bar() { return tls_var; }
!

clang -c -target riscv32 1.c -fPIC -mtls-dialect=desc
clang -c -target riscv32 2.c -fPIC
# case 1
ld.eld -march riscv32 1.o 2.o -shared -o 1.so
llvm-readelf --dyn-syms 1.so | grep tls_var && echo BADIMAGE
llvm-readelf -r 1.so | grep tls_var && echo BADIMAGE
# case 2 (partial link)
ld.eld -march riscv32 1.o 2.o -r -o r.o
ld.eld -march riscv32 r.o -shared -o 1.so
llvm-readelf --dyn-syms 1.so | grep tls_var && echo BADIMAGE
llvm-readelf -r 1.so | grep tls_var && echo BADIMAGE
# case 3 (no relax)
clang -c -target riscv32 1.c -fPIC -mtls-dialect=desc -mno-relax
clang -c -target riscv32 2.c -fPIC -mno-relax
ld.eld -march riscv32 1.o 2.o -shared -o 1.so
llvm-readelf --dyn-syms 1.so | grep tls_var && echo BADIMAGE
llvm-readelf -r 1.so | grep tls_var && echo BADIMAGE
```

quic-akaryaki:
> I would like to see tests using 'C' code as well. You can still keep the asm test.
> 
> I would like to see more detailed testing mentioned below :
> 
> * Basic TLSDESC tests without relaxation (Executable, Shared library, PIE, static linking, **symbol visibility** tests)
>   
>   * Probably we want to have a flag --no-tlsdesc-relax
> * TLSDESC tests with relaxation for the cases above
> * Please add what relaxations is supported right now, and what is punted for later
>   
>   * It is not clear from the commit message on what TLS relaxationare included in this patch
> * Any addition to map file for TLS relaxations will be useful
> * Test cases with mno-relax as needed
> * Documentation should follow as a seperate subtask (to know what the linker did for our users)
> 
> Please create subtasks as needed and totally I am ok to see the limited support that we add for 21.0 but my focus is to build this feature with more clarity.
> 
> ```
> cat > 1.c << \!
> __attribute__((visibility("hidden"))) __thread int tls_var = 10;
> int foo() { return tls_var; }
> !
> 
> cat > 2.c << \!
> extern __thread int tls_var;
> int bar() { return tls_var; }
> !
> 
> clang -c -target riscv32 1.c -fPIC -mtls-dialect=desc
> clang -c -target riscv32 2.c -fPIC
> # case 1
> ld.eld -march riscv32 1.o 2.o -shared -o 1.so
> llvm-readelf --dyn-syms 1.so | grep tls_var && echo BADIMAGE
> llvm-readelf -r 1.so | grep tls_var && echo BADIMAGE
> # case 2 (partial link)
> ld.eld -march riscv32 1.o 2.o -r -o r.o
> ld.eld -march riscv32 r.o -shared -o 1.so
> llvm-readelf --dyn-syms 1.so | grep tls_var && echo BADIMAGE
> llvm-readelf -r 1.so | grep tls_var && echo BADIMAGE
> # case 3 (no relax)
> clang -c -target riscv32 1.c -fPIC -mtls-dialect=desc -mno-relax
> clang -c -target riscv32 2.c -fPIC -mno-relax
> ld.eld -march riscv32 1.o 2.o -shared -o 1.so
> llvm-readelf --dyn-syms 1.so | grep tls_var && echo BADIMAGE
> llvm-readelf -r 1.so | grep tls_var && echo BADIMAGE
> ```

Just to clarify, this patch has full support for TLSDESC, including all 4 types of static relocations, linker optimizations (replacing instructions with nops) and relaxations (removing instructions). Of course, it may have bugs, which is the case with existing code, which could be fixed before the 21 releases. Since new requirements come 1 day before the 21.x feature deadline, it won't go into the release. 

Also, please clarify:
-- if --no-tlsdesc-relax is needed
-- which addition to map file for TLS relaxations will be useful

Thank you

lib/Target/RISCV/RISCVRelocator.cpp
quic-akaryaki: I have added a C test for hidden.

quic-seaswara (commented):
Great!. Let us get it merged. 

@lenary if you can go through it and have your comments, it would be great!

lib/LinkerWrapper/RISCVLinkDriver.cpp
```diff
@@ -135,6 +135,10 @@ RISCVLinkDriver::parseOptions(ArrayRef<const char *> Args,
       ArgList.hasFlag(OPT_RISCVLinkOptTable::riscv_relax_xqci,
                       OPT_RISCVLinkOptTable::no_riscv_relax_xqci, false));
 
+  // --no-relax-tlsdesc
+  if (ArgList.hasArg(OPT_RISCVLinkOptTable::no_riscv_tlsdesc))
+    Config.options().setRISCVRelaxTLSDESC(false);
```
lenary, line 140: Why do we have the option? The relaxation uses Base ISA instructions, both in the original sequence and the relaxed sequence, so when would this relaxation not be valid (when relaxations are enabled)

quic-seaswara: It is just a safety option to disable if required.

quic-seaswara (approved):

PR #312, opened 2025-08-08
Fix PROVIDE feature for output section prologue expressions

### LINKED ISSUE #309: `PROVIDE` does not provide symbols referenced in output section prologue
`PROVIDE` is not providing symbols referenced in output section prologue.

Reproducible example:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
int foo() { return 11; }
EOF

cat >script.t <<\EOF
PHDRS {
  A PT_LOAD;
}

SECTIONS {
  PROVIDE(u = 0x1000);
  .foo (u) : { *(.text.foo) } :A
}
EOF

LDs=("ld.eld -m elf64lriscv")
SFs=(eld)

clang -o 1.o -target riscv64 1.c -c 
for i in "${!SFs[@]}"; do
  # Reports:
  # Fatal: script.t: Error: undefined symbol 'u' referenced in expression
  ${LDs[$i]} -o 1.${SFs[$i]}.out 1.o -T script.t 
done
```

parth-07:
This commit fixes PROVIDE feature for output section prologue expressions. PROVIDE was not providing undefined reference symbols referred in output section prologue expressions such VMA, LMA, Align and SUBALIGN.

Fixes #309

### DIFF
--- a/lib/Script/OutputSectDesc.cpp
+++ b/lib/Script/OutputSectDesc.cpp
@@ -26,21 +26,21 @@ using namespace eld;
 //===----------------------------------------------------------------------===//
 OutputSectDesc::OutputSectDesc(const std::string &PName)
     : ScriptCommand(ScriptCommand::OUTPUT_SECT_DESC), Name(PName),
-      OutpuSectDescProlog() {
-  OutpuSectDescProlog.OutputSectionVMA = nullptr;
-  OutpuSectDescProlog.ThisType = OutputSectDesc::DEFAULT_TYPE;
-  OutpuSectDescProlog.SectionFlag = OutputSectDesc::DEFAULT_PERMISSIONS;
-  OutpuSectDescProlog.OutputSectionLMA = nullptr;
-  OutpuSectDescProlog.Alignment = nullptr;
-  OutpuSectDescProlog.OutputSectionSubaAlign = nullptr;
-  OutpuSectDescProlog.SectionConstraint = OutputSectDesc::NO_CONSTRAINT;
-  OutpuSectDescProlog.ThisPlugin = nullptr;
-  OutpuSectDescProlog.PluginCmd = nullptr;
+      OutputSectDescProlog() {
+  OutputSectDescProlog.OutputSectionVMA = nullptr;
+  OutputSectDescProlog.ThisType = OutputSectDesc::DEFAULT_TYPE;
+  OutputSectDescProlog.SectionFlag = OutputSectDesc::DEFAULT_PERMISSIONS;
+  OutputSectDescProlog.OutputSectionLMA = nullptr;
+  OutputSectDescProlog.Alignment = nullptr;
+  OutputSectDescProlog.OutputSectionSubaAlign = nullptr;
+  OutputSectDescProlog.SectionConstraint = OutputSectDesc::NO_CONSTRAINT;
+  OutputSectDescProlog.ThisPlugin = nullptr;
+  OutputSectDescProlog.PluginCmd = nullptr;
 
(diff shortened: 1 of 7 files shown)


lib/Script/OutputSectDesc.cpp
```diff
@@ -317,16 +316,32 @@ eld::Expected<void> OutputSectDesc::activate(Module &CurModule) {
...
+  InputFile *IF = getInputFileInContext();
+  if (OutputSectDescProlog.hasVMA())
+    OutputSectDescProlog.vma().addRefSymbolsAsUndefSymbolToNP(IF, NP);
+  if (OutputSectDescProlog.hasLMA())
+    OutputSectDescProlog.lma().addRefSymbolsAsUndefSymbolToNP(IF, NP);
+  if (OutputSectDescProlog.hasAlign())
+    OutputSectDescProlog.align().addRefSymbolsAsUndefSymbolToNP(IF, NP);
+  if (OutputSectDescProlog.hasSubAlign())
+    OutputSectDescProlog.subAlign().addRefSymbolsAsUndefSymbolToNP(IF, NP);
+
```
quic-seaswara: Current ObjectLinker goes through all assignments and adds all the symbols.

Can we go over the linker script and fetch symbols from all the linker scripts.

I am not sure what happens with symbol names in 

* MEMORY (LENGTH/ORIGIN) etc  
* PHDRS for flags.

quic-seaswara (changes requested):

lib/Script/OutputSectDesc.cpp
parth-07: > Can we go over the linker script and fetch symbols from all the linker scripts.

We go over all the nodes in the linker script AST and call `activate` member function of that node. I think we should add these symbols to the name pool as part of the activate function. I have added the code for adding output section description prologue symbols to the name pool in `OutputSectDesc::activate`. I think we should use `activate` functions to also add symbols from the assignments for consistency. The same method can be used for adding symbols from all other linker script nodes. Please let me know your thoughts on this.

quic-seaswara: I see, This may require a large refactor. We can punt this for a later window of time.

parth-07:
> Please make the changes to fix the spelling error as NFC and a seperate commit

I have moved the spelling changes to a separate NFC commit.

lib/Script/OutputSectDesc.cpp
parth-07: Okay, sounds good!

lib/Script/OutputSectDesc.cpp
```diff
@@ -282,7 +281,7 @@ eld::Expected<void> OutputSectDesc::activate(Module &CurModule) {
     // By default assign LMA region = VMA Region when the output
     // section does not have a LMA region specified and there is no
     // LMA override
-    if (!OutpuSectDescProlog.hasLMA() && !OutpuSectDescEpilog.hasLMARegion())
+    if (!OutputSectDescProlog.hasLMA() && !OutpuSectDescEpilog.hasLMARegion())
```
quic-seaswara: Can we fix the epilog spelling error as well ? Since you are already changing prolog ?

quic-seaswara (commented):
Please document this behavior

lib/Script/OutputSectDesc.cpp
parth-07: Done. I have fixed the spelling of OutpuSectDescEpilog as well.

quic-seaswara (approved):

PR #313, opened 2025-08-08
Fix assertion failure due to invalid glob pattern in linker script

### LINKED ISSUE #284: eld asserts-out if an invalid glob pattern is specified in the linker script
eld asserts-out if an invalid glob pattern is specified in the linker script. We should error-out instead.

Reproducer:

```bash
#!/usr/bin/env bash

cat >s.t <<\EOF
SECTIONS
{
  MYSEC :
  {
    *([)
  } 
}
EOF

cat >1.c <<\EOF
int foo() { return 1; }
EOF

clang -target riscv64 -o 1.o 1.c -c -ffunction-sections
ld.eld -m elf64lriscv -o 1.eld.out 1.o -T s.t
# LLVM ERROR: Assertion `0` failed in {{.*}}WildcardPattern.cpp line # 84: Pattern cannot be created for [
# Fatal: !!!UNEXPECTED LINKER BEHAVIOR!!!

```

parth-07:
This commit fixes linker assertion failure due to invalid glob pattern in linker script. With this commit, the linker errors out instead of crashing due to assertion failure.

Resolves #284

### DIFF
--- a/lib/Script/WildcardPattern.cpp
+++ b/lib/Script/WildcardPattern.cpp
@@ -14,10 +14,13 @@
 //===- WildcardPattern.cpp ------------------------------------------------===//
 //===----------------------------------------------------------------------===//
 #include "eld/Script/WildcardPattern.h"
+#include "eld/Diagnostics/DiagnosticEngine.h"
 #include "eld/Script/ExcludeFiles.h"
 #include "eld/Support/MsgHandling.h"
+#include "eld/PluginAPI/DiagnosticEntry.h"
 #include "llvm/ADT/Hashing.h"
 #include "llvm/Support/raw_ostream.h"
+#include <optional>
 #include <string>
 
 using namespace eld;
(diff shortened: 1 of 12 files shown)


lib/Script/WildcardPattern.cpp
```diff
@@ -86,3 +89,17 @@ void WildcardPattern::createGlobPattern(llvm::StringRef Pattern) {
   if (!hasGlob())
     setHash(llvm::hash_combine(Pattern));
 }
+
+WildcardPattern *WildcardPattern::create(StrToken *S, SortPolicy PPolicy,
+                                         ExcludeFiles *PExcludeFileList) {
+  // A single backslash is an invalid glob pattern. It is supported for
+  // GNU-compatibility.
+  if (S->name() != "\\") {
+    auto E = llvm::GlobPattern::create(llvm::StringRef(S->name()));
```
quic-seaswara, line 98: I would prefer llvm::GlobPattern have this check right ?

What if the pattern contains something like this : *(text\)

parth-07: `if (S->name() != "\\")` is an old check that is being duplicated here. It is also present at line 83 of this file (function: `WildcardPattern::createGlobPattern`).

> I would prefer llvm::GlobPattern have this check right ?

llvm::GlobPattern correctly returns invalid glob for this. But we want to accept this glob pattern for GNU compatibility.

> What if the pattern contains something like this : *(text)

*(text) is a valid pattern. It should work as expected, right?

PR #351, opened 2025-08-25
Nightly musl test suite validation for x86_64 backend

bhoumik03:
Add Nightly ELD + musl Test Suite

**Stacked on #197 **

Adds nightly GitHub Actions workflow to test ELD linker against musl libc-test suite.

### DIFF
--- a/.github/workflows/nightly-musl-builder.yml
+++ b/.github/workflows/nightly-musl-builder.yml
@@ -1,0 +1,110 @@
+name: Nightly builder for testing eld against musl-test-suite for x86_64
+
+on: 
+  schedule:
+      - cron: '0 2 * * *' # 2:00 AM
+  pull_request:
+    paths:
+      - '.github/workflows/nightly-musl-builder.yml'
+
+jobs:
+  test-eld-musl-x86_64:
+    runs-on: ubuntu-latest
+    
+    steps:
+      - name: Set up Clang 20
+        uses: egor-tensin/setup-clang@v1
+        with:
+          version: "20"
+          platform: x64
+
+      - name: Install system dependencies
+        run: |
+          sudo apt update
+          sudo apt install -y libc++-dev libc++abi-dev ccache libclang-rt-20-dev cmake ninja-build make
+          
+      - name: Verify clang installation
+        run: |
+          clang --version
+          clang++ --version
+
+      - name: Checkout LLVM Project
+        uses: actions/checkout@v4
+        with:
+          repository: llvm/llvm-project
+          path: llvm-project
+          ref: main
+
+      - name: Checkout ELD
+        uses: actions/checkout@v4
+        with:
+          path: llvm-project/llvm/tools/eld
+
+      - name: Configure ELD Build
+        run: |
+          mkdir -p obj
+          cd obj
+          cmake -G Ninja \
+            -DCMAKE_BUILD_TYPE=Release \
+            -DCMAKE_INSTALL_PREFIX=$PWD/../install \
+            -DLLVM_ENABLE_PROJECTS="llvm;clang;compiler-rt" \
+            -DCMAKE_C_COMPILER=`which clang` \
+            -DCMAKE_CXX_COMPILER=`which clang++` \
+            -DCMAKE_CXX_FLAGS="-stdlib=libc++" \
+            -DLLVM_TARGETS_TO_BUILD="X86" \
+            -DELD_TARGETS_TO_BUILD='x86_64' \
+            ../llvm-project/llvm
+
+      - name: Build ELD
+        run: |
+          cd obj 
+          ninja -j$(nproc) ld.eld FileCheck
+          ninja install
+
+      - name: Setup ELD Environment
+        run: |
+          echo "PATH=$PWD/install/bin:$PATH" >> $GITHUB_ENV
+          echo "LD_LIBRARY_PATH=$PWD/obj/lib:$LD_LIBRARY_PATH" >> $GITHUB_ENV
+
+
+      - name: Download and Build musl
+        run: |
+          git clone --depth 1 https://github.com/bminor/musl.git
+          cd musl
+          CC=clang CXX=clang++ ./configure --prefix=$PWD/../musl-install
+          make -j$(nproc)
+          make install
+
+      - name: Fix musl-clang Wrapper for ELD
+        run: |
+          sed -i 's/exec \$(\$cc -print-prog-name=ld)/exec ld.$LD/' musl-install/bin/ld.musl-clang
+          
+      - name: Configure build environment and toolchain
+        run: |
+          # Set environment variables for subsequent steps
+          echo "LD=eld" >> $GITHUB_ENV
+          echo "PATH=$PWD/musl-install/bin:$PWD/install/bin:$PATH" >> $GITHUB_ENV
+          echo "CC=$PWD/musl-install/bin/musl-clang" >> $GITHUB_ENV
+          echo "CFLAGS=-Wno-unused-command-line-argument" >> $GITHUB_ENV
+
+      - name: Download and Setup libc-test
+        run: |
+          git clone --depth 1 https://repo.or.cz/libc-test.git
+          cd libc-test
+          cp config.mak.def config.mak
+          sed -i 's/LDFLAGS += -g/LDFLAGS += -g -static/' config.mak
+          sed -i '1s/$/ -Wno-parentheses/' config.mak
+
+      - name: Build and test libc-test with ELD + musl
+        run: | 
+          cd libc-test
+          make -j$(nproc) 2>&1 | tee build.log
+
+      - name: Validate test results with FileCheck
+        run: |  
+          cd libc-test       
+          FILECHECK="$PWD/../obj/bin/FileCheck"
+          $FILECHECK --check-prefixes=CHECK ../llvm-project/llvm/tools/eld/test/musl/x86_64/static/test-failures.txt < src/REPORT
+          $FILECHECK --check-prefixes=COUNT ../llvm-project/llvm/tools/eld/test/musl/x86_64/static/test-failures.txt < src/REPORT
+
+          echo "Test results validated"
--- a/test/musl/x86_64/static/test-failures.txt
+++ b/test/musl/x86_64/static/test-failures.txt
@@ -1,0 +1,20 @@
+CHECK-DAG: FAIL src/api/main.exe
+CHECK-DAG: FAIL src/functional/dlopen.exe
+CHECK-DAG: FAIL src/functional/strptime-static.exe
+CHECK-DAG: FAIL src/functional/strptime.exe
+CHECK-DAG: FAIL src/functional/tgmath-static.exe
+CHECK-DAG: FAIL src/functional/tgmath.exe
+CHECK-DAG: FAIL src/functional/tls_align-static.exe
+CHECK-DAG: FAIL src/functional/tls_align.exe
+CHECK-DAG: FAIL src/functional/tls_align_dlopen.exe
+CHECK-DAG: FAIL src/functional/tls_init-static.exe
+CHECK-DAG: FAIL src/functional/tls_init.exe
+CHECK-DAG: FAIL src/functional/tls_init_dlopen.exe
+CHECK-DAG: FAIL src/functional/tls_local_exec-static.exe
+CHECK-DAG: FAIL src/functional/tls_local_exec.exe
+CHECK-DAG: FAIL src/math/fmal.exe
+CHECK-DAG: FAIL src/math/powf.exe
+CHECK-DAG: FAIL src/regression/tls_get_new-dtv.exe
+
+COUNT-COUNT-17: {{^FAIL }}
+COUNT-NOT: {{^FAIL }}


test/x86_64/linux/relocGOTPCREL/relocGOTPCREL.test
```diff
@@ -0,0 +1,13 @@
+#--relocGOTPCRELStatic.test----------Executable--------#
+
```
quic-seaswara: remove empty line

test/x86_64/linux/relocGOTPCREL/relocGOTPCREL.test
```diff
@@ -0,0 +1,13 @@
...
+BEGIN_COMMENT
+# Test R_X86_64_GOTPCREL relocation type support.
+#END_COMMENT
+
+#START_TEST
+RUN: %clang %clangopts -c %p/Inputs/relocGOTPCREL_test.c -o %t.relocGOTPCREL_test.o
+RUN: %link %linkopts -o %t.relocGOTPCREL_test.out %t.relocGOTPCREL_test.o
+RUN: %t.relocGOTPCREL_test.out; echo $? > %t.code
```
quic-seaswara: We had a similar comment in the previous pull requests too, change this to FileCheck

lib/Target/x86_64/x86_64Relocator.cpp
```diff
@@ -187,6 +190,17 @@ void x86_64Relocator::scanGlobalReloc(InputFile &pInputFile, Relocation &pReloc,
...
     return;
   }
+  case llvm::ELF::R_X86_64_GOTPCREL:
+  case llvm::ELF::R_X86_64_GOTPCRELX:
+  case llvm::ELF::R_X86_64_REX_GOTPCRELX: {
+    if (!(rsym->reserved() & ReserveGOT)) {
+      std::lock_guard<std::mutex> relocGuard(m_RelocMutex);
+      x86_64GOT *gotEntry =
+          m_Target.createGOT(GOT::GOTType::Regular, Obj, rsym);
+      gotEntry->setValueType(GOT::SymbolValue);
```
quic-seaswara, line 203: if we are building static executables only

lib/Target/x86_64/x86_64GOT.h
```diff
@@ -39,36 +40,31 @@ class x86_64GOT : public GOT {
...
+                          sizeof(Value));
   }
 
   static x86_64GOT *Create(ELFSection *O, ResolveInfo *R) {
     return make<x86_64GOT>(GOT::Regular, O, R);
   }
 
 private:
-  uint8_t Value[4];
+  mutable uint64_t Value;
```
quic-seaswara, line 61: Do we use this Value outside getContent()

bhoumik03: Value member variable is used only within getContent(), but since getContent() returns ArrayRef<uint_8>, we can't make it a local variable.
We could change return type to uint64 _t, but then that would require modifying the base class as well.
So having Value as member variable prevents this dangling pointer issue.

.github/workflows/nightly-musl-builder.yml
```diff
@@ -0,0 +1,153 @@
...
+
+      - name: Build and test libc-test with ELD + musl
+        run: |          
+          cd libc-test
+          make -j$(nproc) 2>&1 | tee build.log
+    
+      - name: Check for Unexpected Test Failures
+        run: |  
+          chmod +x .github/scripts/check-test-results.sh
+          .github/scripts/check-test-results.sh libc-test
```
quic-seaswara: You can use LLVM FileCheck to test the results.

lib/Target/x86_64/x86_64GOT.h
quic-seaswara: Can we simply make the Value[4] to Value[8] ?

lib/Target/x86_64/x86_64RelocationFunctions.h
```diff
@@ -120,7 +123,30 @@ struct RelocationDescription x86RelocDesc[] = {
...
+    {/*.func = */ none, /*.type = */ 31, /*.forceVerify = */ false},
+    {/*.func = */ none, /*.type = */ 32, /*.forceVerify = */ false},
+    {/*.func = */ none, /*.type = */ 33, /*.forceVerify = */ false},
+    {/*.func = */ none, /*.type = */ 34, /*.forceVerify = */ false},
+    {/*.func = */ none, /*.type = */ 35, /*.forceVerify = */ false},
+    {/*.func = */ none, /*.type = */ 36, /*.forceVerify = */ false},
+    {/*.func = */ none, /*.type = */ 37, /*.forceVerify = */ false},
+    {/*.func = */ none, /*.type = */ 38, /*.forceVerify = */ false},
+    {/*.func = */ none, /*.type = */ 39, /*.forceVerify = */ false},
+    {/*.func = */ none, /*.type = */ 40, /*.forceVerify = */ false},
```
quic-seaswara, line 142: What are the relocation types ?

.github/workflows/nightly-musl-builder.yml
```diff
@@ -0,0 +1,153 @@
...
+      #     chmod +x bin/ld.eld
+      #     chmod +x lib/libLW.so.4
+          
+      #     # Set library path
+      #     export LD_LIBRARY_PATH=$PWD/lib:$LD_LIBRARY_PATH
+      #     echo "LD_LIBRARY_PATH=$PWD/lib:$LD_LIBRARY_PATH" >> $GITHUB_ENV
+          
+      #     echo "Download complete:"
+      #     ls -lh bin/ld.eld
+      #     ls -lh lib/libLW.so.4
```
quic-seaswara: Can you remove these commented lines ?

.github/workflows/nightly-musl-builder.yml
```diff
@@ -0,0 +1,153 @@
+name: Nightly builder for testing eld against musl-test-suite for x86
```
quic-seaswara: x86_64

.github/workflows/nightly-musl-builder.yml
```diff
@@ -0,0 +1,153 @@
...
+          path: llvm-project/llvm/tools/eld
+
+      - name: Configure ELD Build
+        run: |
+          mkdir -p obj
+          cd obj
+          
+          cmake -G Ninja \
+            -DCMAKE_BUILD_TYPE=Release \
+            -DLLVM_ENABLE_PROJECTS="llvm;clang;clang-tools-extra" \
```
quic-seaswara: why clang-tools-extra ?

.github/workflows/nightly-musl-builder.yml
```diff
@@ -0,0 +1,153 @@
...
+            -DLLVM_ENABLE_SPHINX=OFF \
+            ../llvm-project/llvm
+
+      - name: Build ELD
+        run: |
+          cd obj 
+          ninja -j$(nproc) ld.eld
+          echo "ELD build completed"
+          echo "ELD binary location:"
+          ls -la bin/ld.eld
```
quic-seaswara: Do you need this ?

.github/workflows/nightly-musl-builder.yml
```diff
@@ -0,0 +1,153 @@
...
+
+      - name: Setup ELD Environment
+        run: |          
+          mkdir -p bin lib
+          cp obj/bin/ld.eld bin/
+          cp obj/lib/libLW.so.4 lib/
+          chmod +x bin/ld.eld
+          echo "LD_LIBRARY_PATH=$PWD/lib:$LD_LIBRARY_PATH" >> $GITHUB_ENV
```
quic-seaswara: Please do ninja install

.github/workflows/nightly-musl-builder.yml
```diff
@@ -0,0 +1,153 @@
...
+          echo "CC=$PWD/musl-install/bin/musl-clang" >> $GITHUB_ENV
+          echo "CFLAGS=-Wno-unused-command-line-argument" >> $GITHUB_ENV
+          echo "CXXFLAGS=-stdlib=libc++" >> $GITHUB_ENV
+
+      - name: Download and Setup libc-test
+        run: |
+          git clone --depth 1 https://repo.or.cz/libc-test.git
+          cd libc-test
+          cp config.mak.def config.mak
+          sed -i 's/LDFLAGS += -g/LDFLAGS += -g -Wl,--image-base=0x400000 -static/' config.mak
```
quic-seaswara: Can we change the linker to set the image base as 40000 ?

quic-seaswara (changes requested):

lib/Target/x86_64/x86_64GOT.h
parth-07: I had requested Bhoumik to change the Value[4] to an uint64_t integer instead. We can change the Value[4] to Value[8] as well. However, I believe that the variable usage and the arithmetic operations are much simpler to perform/understand when the value is `uint64_t`. Please let us know your thoughts on this.

.github/workflows/nightly-musl-builder.yml
parth-07: @bhoumik03 Can we remove the `--image-base` altogether now? I think you fixed the default image base for x86-64 sometime back, right?

lib/Target/x86_64/x86_64RelocationFunctions.h
bhoumik03: Relocations numbered 25-37 are not supported as per this PR, but we need this padding for 41 (R_X86_64_GOTPCRELX) and 42 (R_X86_64_GOTPCRELX) as we need index-to-type correspondence. Will add types as and when that numbered relocation support is added. 
Moreover, there are no relocations corresponding to 38-40 and 46-49.

Please let me know if I should add the types for 25-37 as part of this PR itself.

.github/workflows/nightly-musl-builder.yml
bhoumik03: Yes, thank for pointing out, we don't need to add this flag.

.github/workflows/nightly-musl-builder.yml
```diff
@@ -0,0 +1,120 @@
...
+            -DELD_TARGETS_TO_BUILD='x86_64' \
+            ../llvm-project/llvm
+
+      - name: Build ELD
+        run: |
+          set -e
+          cd obj 
+          ninja -j$(nproc) ld.eld FileCheck
+          # Verify build succeeded
+          test -f bin/ld.eld || { echo "Build failed: ld.eld not found"; exit 1; }
```
quic-seaswara: Remove this, if eld failed the previous step would fail and stop the build

.github/workflows/nightly-musl-builder.yml
```diff
@@ -0,0 +1,120 @@
...
+          test -f bin/FileCheck || { echo "Build failed: FileCheck not found"; exit 1; }
+
+      - name: Setup ELD Environment
+        run: |
+          set -e        
+          echo "PATH=$PWD/obj/bin:$PATH" >> $GITHUB_ENV
+          echo "LD_LIBRARY_PATH=$PWD/obj/lib:$LD_LIBRARY_PATH" >> $GITHUB_ENV
```
quic-seaswara, line 67: can we install all the artifacts and set the directory to the install area and run tests ?

.github/workflows/nightly-musl-builder.yml
```diff
@@ -0,0 +1,120 @@
...
+          set -e         
+          cd libc-test
+          make -j$(nproc) 2>&1 | tee build.log
+
+      - name: Validate test results with FileCheck
+        run: |
+          set -e
+          cd libc-test       
+          FILECHECK="$PWD/../obj/bin/FileCheck"
+          $FILECHECK ../llvm-project/llvm/tools/eld/.github/scripts/expected-failures.txt < src/REPORT
```
quic-seaswara: why the .txt file is in .github/scripts/, can we move it to eld-home/test/musl/x86_64/static/test-failures.txt

.github/workflows/nightly-musl-builder.yml
```diff
@@ -0,0 +1,120 @@
...
+        run: |
+          set -e
+          git clone --depth 1 https://repo.or.cz/libc-test.git
+          cd libc-test
+          cp config.mak.def config.mak
+          sed -i 's/LDFLAGS += -g/LDFLAGS += -g -static/' config.mak
+
+      - name: Build and test libc-test with ELD + musl
+        run: | 
+          set -e
```
quic-seaswara: Do you need this ?

.github/workflows/nightly-musl-builder.yml
```diff
@@ -0,0 +1,120 @@
...
+          git clone --depth 1 https://repo.or.cz/libc-test.git
+          cd libc-test
+          cp config.mak.def config.mak
+          sed -i 's/LDFLAGS += -g/LDFLAGS += -g -static/' config.mak
+
+      - name: Build and test libc-test with ELD + musl
+        run: | 
+          set -e         
+          cd libc-test
+          make -j$(nproc) 2>&1 | tee build.log
```
quic-seaswara, line 101: Is there a way to print the summary ?

.github/workflows/nightly-musl-builder.yml
```diff
@@ -0,0 +1,120 @@
...
+          path: llvm-project/llvm/tools/eld
+
+      - name: Configure ELD Build
+        run: |
+          set -e 
+          mkdir -p obj
+          cd obj
+          cmake -G Ninja \
+            -DCMAKE_BUILD_TYPE=Release \
+            -DLLVM_ENABLE_PROJECTS="llvm;clang" \
```
quic-seaswara: need to enable compiler builtins as well

.github/workflows/nightly-musl-builder.yml
```diff
@@ -0,0 +1,120 @@
...
+          sed -i 's/exec \$(\$cc -print-prog-name=ld)/exec ld.$LD/' musl-install/bin/ld.musl-clang
+
+      - name: Configure build environment and toolchain
+        run: |
+          set -e
+          # Set environment variables for subsequent steps
+          echo "LD=eld" >> $GITHUB_ENV
+          echo "PATH=$PWD/musl-install/bin:$PWD/obj/bin:$PATH" >> $GITHUB_ENV
+          echo "CC=$PWD/musl-install/bin/musl-clang" >> $GITHUB_ENV
+          echo "CFLAGS=-Wno-unused-command-line-argument" >> $GITHUB_ENV
```
quic-seaswara, line 88: export LDFLAGS="-fuse-ld=eld -static"

.github/workflows/nightly-musl-builder.yml
```diff
@@ -0,0 +1,120 @@
...
+          git clone --depth 1 https://github.com/bminor/musl.git
+          cd musl
+          CC=clang CXX=clang++ ./configure --prefix=$PWD/../musl-install
+          make -j$(nproc)
+          make install
+          
+      - name: Fix musl-clang Wrapper for ELD
+        run: |
+          set -e
+          sed -i 's/exec \$(\$cc -print-prog-name=ld)/exec ld.$LD/' musl-install/bin/ld.musl-clang
```
quic-seaswara, line 80: No need to make this change if you use -fuse-ld=eld in the below comment

.github/workflows/nightly-musl-builder.yml
```diff
@@ -0,0 +1,120 @@
...
+          echo "CFLAGS=-Wno-unused-command-line-argument" >> $GITHUB_ENV
+          echo "CXXFLAGS=-stdlib=libc++" >> $GITHUB_ENV
+
+      - name: Download and Setup libc-test
+        run: |
+          set -e
+          git clone --depth 1 https://repo.or.cz/libc-test.git
+          cd libc-test
+          cp config.mak.def config.mak
+          sed -i 's/LDFLAGS += -g/LDFLAGS += -g -static/' config.mak
```
quic-seaswara, line 95: remove usage of sed throughout the workflow

.github/workflows/nightly-musl-builder.yml
```diff
@@ -0,0 +1,120 @@
...
+          cd obj
+          cmake -G Ninja \
+            -DCMAKE_BUILD_TYPE=Release \
+            -DLLVM_ENABLE_PROJECTS="llvm;clang" \
+            -DCMAKE_C_COMPILER=`which clang` \
+            -DCMAKE_CXX_COMPILER=`which clang++` \
+            -DCMAKE_CXX_FLAGS="-stdlib=libc++" \
+            -DLLVM_TARGETS_TO_BUILD="X86" \
+            -DLLVM_CCACHE_BUILD=ON \
+            -DLLVM_CCACHE_DIR=$PWD/ccache/ \
```
quic-seaswara: Do you need caching if this is just nightly builds ?

.github/workflows/nightly-musl-builder.yml
```diff
@@ -0,0 +1,120 @@
...
+
+on: 
+  schedule:
+      - cron: '0 2 * * *' # 2:00 AM
+  pull_request:
+    paths:
+      - '.github/workflows/nightly-musl-builder.yml'
+
+jobs:
+  test-eld-external:
```
quic-seaswara: test-eld-musl-x86_64:

.github/workflows/nightly-musl-builder.yml
```diff
@@ -0,0 +1,120 @@
...
+          cd libc-test
+          make -j$(nproc) 2>&1 | tee build.log
+
+      - name: Validate test results with FileCheck
+        run: |
+          set -e
+          cd libc-test       
+          FILECHECK="$PWD/../obj/bin/FileCheck"
+          $FILECHECK ../llvm-project/llvm/tools/eld/.github/scripts/expected-failures.txt < src/REPORT    
+          echo "Test results validated"
```
quic-seaswara, line 110: Remove this line ?

.github/scripts/expected-failures.txt
```diff
@@ -0,0 +1,18 @@
+CHECK: FAIL src/api/main.exe
+CHECK: FAIL src/functional/dlopen.exe
+CHECK: FAIL src/functional/strptime-static.exe
+CHECK: FAIL src/functional/strptime.exe
```
quic-seaswara: CHECK-DAG

.github/scripts/expected-failures.txt
```diff
@@ -0,0 +1,18 @@
...
+CHECK: FAIL src/functional/tls_align_dlopen.exe
+CHECK: FAIL src/functional/tls_init-static.exe
+CHECK: FAIL src/functional/tls_init.exe
+CHECK: FAIL src/functional/tls_init_dlopen.exe
+CHECK: FAIL src/functional/tls_local_exec-static.exe
+CHECK: FAIL src/functional/tls_local_exec.exe
+CHECK: FAIL src/math/fmal.exe
+CHECK: FAIL src/math/powf.exe
+CHECK: FAIL src/regression/tls_get_new-dtv.exe
+CHECK-NOT: FAIL
```
quic-seaswara: Can you check of failures CHECK-COUNT : 17 ?

.github/workflows/nightly-musl-builder.yml
bhoumik03: Yes, to ensure immediate failure

.github/scripts/expected-failures.txt
bhoumik03: CHECK-COUNT-17 could miss when specific tests change (one passes, another fails)

.github/workflows/nightly-musl-builder.yml
bhoumik03: Not needed, removed it

.github/workflows/nightly-musl-builder.yml
bhoumik03: Added printing of scr/REPORT which lists the failures and also tells if it was a segmentation fault or build error

.github/workflows/nightly-musl-builder.yml
quic-seaswara: I could not follow

.github/workflows/nightly-musl-builder.yml
quic-seaswara: no need of doing cat src/REPORT, you can pass this to filecheck directly

.github/workflows/nightly-musl-builder.yml
```diff
@@ -0,0 +1,115 @@
...
+          git clone --depth 1 https://repo.or.cz/libc-test.git
+          cd libc-test
+          cp config.mak.def config.mak
+
+      - name: Build and test libc-test with ELD + musl
+        run: | 
+          set -e         
+          cd libc-test
+          make -j$(nproc) 2>&1 | tee build.log
+          echo "Test Summary"
```
quic-seaswara: Remove Test summary

test/musl/x86_64/static/test-failures.txt
```diff
@@ -0,0 +1,18 @@
...
+CHECK-DAG: FAIL src/functional/tls_align_dlopen.exe
+CHECK-DAG: FAIL src/functional/tls_init-static.exe
+CHECK-DAG: FAIL src/functional/tls_init.exe
+CHECK-DAG: FAIL src/functional/tls_init_dlopen.exe
+CHECK-DAG: FAIL src/functional/tls_local_exec-static.exe
+CHECK-DAG: FAIL src/functional/tls_local_exec.exe
+CHECK-DAG: FAIL src/math/fmal.exe
+CHECK-DAG: FAIL src/math/powf.exe
+CHECK-DAG: FAIL src/regression/tls_get_new-dtv.exe
+CHECK-NOT: FAIL
```
quic-seaswara: CHECK-COUNT: 17

dont add CHECK-NOT:FAIL

.github/workflows/nightly-musl-builder.yml
bhoumik03: can you please illustrate? are you referring to compiler-rt?

.github/workflows/nightly-musl-builder.yml
bhoumik03: Using export LDPFLAGS does not fix the issue. It leads to usage of the default eld rather that the musl wrapper. For musl wrapper to be used, I had to use sed to directly modify the wrapper.

.github/workflows/nightly-musl-builder.yml
```diff
@@ -0,0 +1,121 @@
...
+          
+      - name: Configure build environment and toolchain
+        run: |
+          set -e
+          # Set environment variables for subsequent steps
+          echo "LD=eld" >> $GITHUB_ENV
+          echo "PATH=$PWD/musl-install/bin:$PWD/install/bin:$PATH" >> $GITHUB_ENV
+          echo "CC=$PWD/musl-install/bin/musl-clang" >> $GITHUB_ENV
+          echo "CFLAGS=-Wno-unused-command-line-argument" >> $GITHUB_ENV
+          echo "CXXFLAGS=-stdlib=libc++" >> $GITHUB_ENV
```
parth-07: Do we need to set -stdlib=libc++ here?

.github/workflows/nightly-musl-builder.yml
parth-07: The other mechanism would be to store the config.mak file with the source code. Should we prefer that over sed? Hopefully, this is a temporary measure, as we aim to completely support x86-64 in the near future.

.github/workflows/nightly-musl-builder.yml
bhoumik03: I have build eld with ```-DCMAKE_CXX_FLAGS="-stdlib=libc++"``` in sync with other workflows.
So to keep consistency while testing we need this flag.

.github/workflows/nightly-musl-builder.yml
```diff
@@ -0,0 +1,121 @@
+name: Nightly builder for testing eld against musl-test-suite for x86_64
+
+on: 
+  schedule:
+      - cron: '0 2 * * *' # 2:00 AM
+  pull_request:
```
quic-seaswara, line 6: Why also on pull request ?

.github/workflows/nightly-musl-builder.yml
quic-seaswara: @bhoumik03 pls create an issue for this

.github/workflows/nightly-musl-builder.yml
quic-seaswara: What is immediate failure, we should not have set -e IMHO

quic-seaswara (changes requested):

.github/workflows/nightly-musl-builder.yml
bhoumik03: Without set -e the remaining commands of that step would still execute even if the earlier fails. 
So that would be helpful in case of failure. But yes, it is not essential and can be removed

.github/workflows/nightly-musl-builder.yml
bhoumik03: It would run only when some change is made to this yml, for verification.

.github/workflows/nightly-musl-builder.yml
bhoumik03: This was only for printing the results, not for filecheck

.github/workflows/nightly-musl-builder.yml
bhoumik03: ok

test/musl/x86_64/static/test-failures.txt
```diff
@@ -0,0 +1,21 @@
...
+CHECK-DAG: FAIL src/functional/tls_align.exe
+CHECK-DAG: FAIL src/functional/tls_align_dlopen.exe
+CHECK-DAG: FAIL src/functional/tls_init-static.exe
+CHECK-DAG: FAIL src/functional/tls_init.exe
+CHECK-DAG: FAIL src/functional/tls_init_dlopen.exe
+CHECK-DAG: FAIL src/functional/tls_local_exec-static.exe
+CHECK-DAG: FAIL src/functional/tls_local_exec.exe
+CHECK-DAG: FAIL src/math/fmal.exe
+CHECK-DAG: FAIL src/math/powf.exe
+CHECK-DAG: FAIL src/musl/pleval-static.exe
```
bhoumik03: This is a compilation error

src/musl/pleval.c:53:9: error: chained comparison 'X <= Y >= Z' does not behave the same as a mathematical expression [-Wparentheses]

test/musl/x86_64/static/test-failures.txt
bhoumik03: both the approaches have the following issue - if there is a new unexpected failure in the middle of the input, it would go undetected as the CHECK commands start scanning after where the previous CHECK command scanned.

To fix this, I am using --check-prefix flag which ensures that the count check happens from the start of the input.

parth-07 (approved):
Looks good!

test/musl/x86_64/static/test-failures.txt
quic-seaswara: Are you using -Werror -Wall ?

bhoumik03: musl test-suite config.mak adds these flags.
Added -Wno-parentheses to disable the warning that was causing the compilation failure.

PR #358, opened 2025-08-26
Fix fragment padding size computation when output section is unaligned

### LINKED ISSUE #346: map file shows improper section addresses
Below test shows incorrect addresses in the map file :-

```
cat > script.t << '!'
SECTIONS {
  text (0x14) : {
    *(.text*)
  }
}
!

cat > test.c << '!'
int foo() { return 1; }
int bar() { return 0; }
!

clang -falign-functions=64 --target=riscv32-unknown-elf -c test.c -ffunction-sections
ld.eld test.o -T script.t -Map map.txt -o eld.out
```

**Map file snippet**

```
text    0x14    0x7e # Offset: 0x1014, LMA: 0x40, Alignment: 0x40, Flags: SHF_ALLOC|SHF_EXECINSTR, Type: SHT_PROGBITS
# RelaxationBytesDeleted : 36
*(.text*) #Rule 1, script.t [3:0]
.text   0x14    0x0     test.o  #SHT_PROGBITS,SHF_ALLOC|SHF_EXECINSTR,2
.text.foo       0x14    0x3e    test.o  #SHT_PROGBITS,SHF_ALLOC|SHF_EXECINSTR,64
        0x40            $x
        0x40            foo
PADDING_ALIGNMENT       0x52    0x2     0x0
.text.bar       0x54    0x3e    test.o  #SHT_PROGBITS,SHF_ALLOC|SHF_EXECINSTR,64
        0x80            $x
        0x80            bar
*(text) #Rule 2, Internal-LinkerScript (Implicit rule inserted by Linker) [0:0]
```

parth-07:
This commit fixes fragment padding size computation when output section address is unaligned. The fragment padding size computation (incorrectly) assumed that the fragment offset is congruent to the output section address (modulo fragment alignment). This assumption is violated when the output section address is explicitly specified in the linker script.

This change is causing UnalignedCrash.test to fail because the test expects the link to fail but now the link is passing. The link passing seems correct as both bfd/lld pass as well for this test case.

Resolves #346

### DIFF
--- a/lib/Fragment/Fragment.cpp
+++ b/lib/Fragment/Fragment.cpp
@@ -13,6 +13,7 @@
 #include "eld/Fragment/Fragment.h"
 #include "eld/Core/Module.h"
 #include "eld/Diagnostics/DiagnosticEngine.h"
+#include "eld/Object/OutputSectionEntry.h"
 #include "eld/Object/SectionMap.h"
 #include "eld/Readers/ELFSection.h"
 #include "eld/Support/MsgHandling.h"
@@ -107,7 +108,14 @@ Fragment *Fragment::getNextNode() {
 size_t Fragment::paddingSize() const {
   if (!hasOffset())
     return 0;
-  return llvm::offsetToAlignment(UnalignedOffset, llvm::Align(Alignment));
+  // FIXME: Will a fragment ever not have an owning section / output section?
+  // We should have linker invariants such as 'All fragments must always have a
+  // fragment' so that we can write codes in a uniform way without having too
+  // many if-checks.
+  ELFSection *OutSect = (OwningSection ? getOutputELFSection() : nullptr);
+  size_t OutSectAddr = (OutSect ? OutSect->addr() : 0);
+  return llvm::offsetToAlignment(UnalignedOffset + OutSectAddr,
+                                 llvm::Align(Alignment));
 }
 
 void Fragment::setOffset(uint32_t POffset) {
--- a/test/Common/standalone/FragPaddingWithUnalignedOutSect/Inputs/script2.t
+++ b/test/Common/standalone/FragPaddingWithUnalignedOutSect/Inputs/script2.t
@@ -1,0 +1,7 @@
+SECTIONS {
+  .text (0x14) : {
+    . = . + 0x2;
+    *(.text.foo)
+    *(*text*)
+  }
+}
--- a/test/Common/standalone/FragPaddingWithUnalignedOutSect/FragPaddingWithUnalignedOutSect.test
+++ b/test/Common/standalone/FragPaddingWithUnalignedOutSect/FragPaddingWithUnalignedOutSect.test
@@ -1,0 +1,100 @@
+#---FragPaddingWithUnalignedOutSect.test------------- Executable------------------#
+#BEGIN_COMMENT
+# This tests verifies that the fragment padding computation is correct
+# when the output section address is unaligned.
+#END_COMMENT
+#BEGIN_TEST
+RUN: %clang %clangopts -o %t1.1.o %p/Inputs/1.c -c -ffunction-sections
+RUN: %link -MapStyle txt %linkopts -o %t1.1.1.out %t1.1.o -T %p/Inputs/script1.t -Map %t1.1.1.map.txt
+RUN: %filecheck %s --check-prefix=CHECK1_1 < %t1.1.1.map.txt
+RUN: %clang %clangopts -o %t1.2.o %p/Inputs/2.c -c -ffunction-sections
+RUN: %link -MapStyle txt %linkopts -o %t1.2.1.out %t1.2.o -T %p/Inputs/script1.t -Map %t1.2.1.map.txt
+RUN: %filecheck %s --check-prefix=CHECK2_1 < %t1.2.1.map.txt
+
+RUN: %link -MapStyle txt %linkopts -o %t1.1.2.out %t1.1.o -T %p/Inputs/script2.t -Map %t1.1.2.map.txt
+RUN: %filecheck %s --check-prefix=CHECK1_2 < %t1.1.2.map.txt
+RUN: %link -MapStyle txt %linkopts -o %t1.2.2.out %t1.2.o -T %p/Inputs/script2.t -Map %t1.2.2.map.txt
+RUN: %filecheck %s --check-prefix=CHECK2_2 < %t1.2.2.map.txt
+
+RUN: %clang %clangopts -o %t1.3.o %p/Inputs/3.c -c -ffunction-sections
+RUN: %link -MapStyle txt %linkopts -o %t1.3.thread1.out %t1.3.o -T %p/Inputs/script.thread1.t -Map %t1.3.thread1.map.txt
+RUN: %filecheck %s --check-prefix=CHECK3_THREAD1 < %t1.3.thread1.map.txt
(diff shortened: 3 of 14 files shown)


lib/Fragment/Fragment.cpp
```diff
@@ -107,7 +108,14 @@ Fragment *Fragment::getNextNode() {
...
   if (!hasOffset())
     return 0;
-  return llvm::offsetToAlignment(UnalignedOffset, llvm::Align(Alignment));
+  // FIXME: Will a fragment ever not have an owning section / output section?
+  // We should have linker invariants such as 'All fragments must always have a
+  // fragment' so that we can write codes in a uniform way without having too
+  // many if-checks.
+  ELFSection *OutSect = (OwningSection ? getOutputELFSection() : nullptr);
+  size_t OutSectAddr = (OutSect ? OutSect->addr() : 0);
+  return llvm::offsetToAlignment(UnalignedOffset + OutSectAddr,
```
quic-seaswara, line 117: This should probably use dotSymbol->value() ?

test/UnitTests/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ set(LLVM_LINK_COMPONENTS
     LTO
     MC
     Object
+    Option
```
quic-seaswara, line 10: Why this change ?

quic-seaswara (changes requested):

test/UnitTests/CMakeLists.txt
parth-07: I accidentally pushed this change. Thank you for catching this. I am removing this change.

lib/Fragment/Fragment.cpp
parth-07: We cannot use dotSymbol here because the dotSymbol value keeps changing as we perform the layout. If we use `dotSymbol` here, then the `Fragment::getOffset()` will give incorrect results when used outside the layout-iteration context.

test/UnitTests/CMakeLists.txt
parth-07: I have removed this change.

lib/Fragment/Fragment.cpp
quic-seaswara: Outside the layout-iteration context, did not think about it!.

We should remove the dependency of the output section and use the owning section address ( all fragments are relative to the owning section right ? ) 

Can you add a test that includes debug sections and check for offset alignments ?

lib/Fragment/Fragment.cpp
```diff
@@ -107,7 +108,14 @@ Fragment *Fragment::getNextNode() {
 size_t Fragment::paddingSize() const {
   if (!hasOffset())
     return 0;
-  return llvm::offsetToAlignment(UnalignedOffset, llvm::Align(Alignment));
+  // FIXME: Will a fragment ever not have an owning section / output section?
+  // We should have linker invariants such as 'All fragments must always have a
+  // fragment' so that we can write codes in a uniform way without having too
+  // many if-checks.
```
quic-seaswara, line 114: All fragments definitely have a owning section, cant think of any that does not have one.

test/Common/standalone/FragPaddingWithUnalignedOutSect/Inputs/script2.t
```diff
@@ -0,0 +1,7 @@
+SECTIONS {
+  .text (0x14) : {
+    . = . + 0x2;
```
quic-seaswara, line 3: There are users such as picolibc that do something like this :

```
        .tbss_space : {
                . = ADDR(.tbss);
                . = . + SIZEOF(.tbss);
        } >ram AT>ram :ram
```

I am pretty sure we dont have a test, but add this test to make sure that we dont regress this.

quic-seaswara (changes requested):

lib/Fragment/Fragment.cpp
parth-07: > We should remove the dependency of the output section and use the owning section address ( all fragments are relative to the owning section right ? )

I think with the current design we will need to keep output section dependency. If I understand correctly, then the owning section will refer to the original input section -- we only assign address to the output section, rule sections and fragments.

> Can you add a test that includes debug sections and check for offset alignments ?

Thank you for mentioning debug sections. We often miss testing for debug section offsets. I will add a test for this.

quic-seaswara: we can assign the address to the section as well, and build this feature over the owning section address or punt it after this patch, whatever works is fine. But i feel that we can build over the input section address assignment.

lib/Fragment/Fragment.cpp
parth-07: One of the test did fail with a crash due to output section being a nullptr when I did not have these ternary-condition checks. I feel that we need to make the linker invariants more clear and add a mechanism to enforce them. For example, we can use a sentinel `NullOutputSection` in the case when an input section does not have an output section due to an error but we want to continue the link because the user has specified `--no-inhibit-exec`. Please share your thoughts on this.

quic-seaswara: Which test is that ? May be that was called before output sections were set ? 

I think a std::optional would work here instead of NullOutputSection ?

parth-07: > Which test is that ?

I do not recall the name. I will check and update this thread.

> I think a std::optional would work here instead of NullOutputSection ?

I think std::optional will work for most cases, but `--no-inhibit-exec` can create interesting cases where we want to continue link despite the errors. Or in the cases, where we are printing map-file when the link has failed before the link was able to assign an output section to all the input sections.

parth-07: Yes, `std::optional` will work. But then we still need to check if the optional has value or not. We cannot simply call `OutSect->addr()`, we instead need to do `OutSect.has_value() ? (*OutSect).addr() : 0`. It is the same as the pointer case.

quic-seaswara: Good suggestion!

test/Common/standalone/FragPaddingWithUnalignedOutSect/Inputs/script2.t
parth-07: I tried adding a test for this, but it seems that we have an another bug that causes incorrect `.tbss_space` layout if `.tbss` has an explicit address and a memory region. I have created an issue for this: #367.

I have added tests for checking layout of `.tdata` and `.tbss` fragments when the output section address is unaligned.

lib/Fragment/Fragment.cpp
parth-07: > But i feel that we can build over the input section address assignment.

Yes, we can build over the input section address assignment, but I believe that should be handled as an independent refactoring effort to improve the design, as that increases the code changes and testing scope significantly.

parth-07: > Can you add a test that includes debug sections and check for offset alignments ?

I could not add test for debug sections because there is no way to assign explicit address/offset to debug sections and because the debug section alignment is always 1. To test this change, we need to make output section address/offset incongruent to the fragment address/offset (modulo fragment alignment).

quic-seaswara: You can include a debug section in a alloc section, or use SUBALIGN

parth-07: Thank you for the suggestions. I have added a test for the debug section as well.

quic-seaswara (approved):

PR #364, opened 2025-08-28
Fix Relaxation/UnalignedCrash/UnalignedCrash.test

### LINKED ISSUE #363: Link command in`Relaxation/UnalignedCrash-QTOOL-105850/UnalignedCrash.test` should not fail
The test `test/RISCV/standalone/Relaxation/UnalignedCrash-QTOOL-105850/UnalignedCrash.test` checks that the link command  fails with the `Insufficient number of spare bytes for alignment relaxation` error message. However, the link command for this test should actually pass. It is causing insufficient bytes for alignment relaxation error message due to incorrect address assignment of fragment when output section address is not congruent to fragment offset (module fragment alignment).

parth-07:
This commit fixes the test
RISCV/standalone/Relaxation/UnalignedCrash/UnalignedCrash.test. The test checks that the link command fails with the insufficient bytes for alignment relaxation error. However, for this test, the link should pass without any errors. It was currently erroring out because of the incorrect fragment address. This commit fixes the test so that the link should fail with the insufficient bytes for alignment relaxation error without relying on the incorrect fragment address.

Resolves #363

### DIFF
--- a/test/RISCV/standalone/Relaxation/InsufficientAlignmentRelaxation/Inputs/1.s
+++ b/test/RISCV/standalone/Relaxation/InsufficientAlignmentRelaxation/Inputs/1.s
@@ -1,10 +1,9 @@
         .text
         .globl  foo
         .type   foo, @function
-        call norelax
-        call norelax
-        call norelax
-        call norelax
+        .option rvc
+        j 1f
+        .option norvc
         .balign 32
 foo:    nop
         // Add a relaxation relocation as RISC-V ABI demands that R_ALIGN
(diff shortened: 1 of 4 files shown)


parth-07:
LLVM mc used to generate incorrect (error-prone?) alignment padding when `rvc` and `norvc` code were mixed in the same section. This was fixed here: https://github.com/llvm/llvm-project/pull/150816. The test case present in this PR should not result in 'Insufficient bytes for alignment relaxation' with the LLVM fix.

PR #372, opened 2025-08-28
[patch 1/7] Fix only TBSS sections causing empty segment

### LINKED ISSUE #353: empty segment created when the image has only tbss
An empty segment is created when the image has only tbss.

**Test**
```

cat > 1.c << \!
__thread int b = 0;
int data = 100;
int foo() { return 0; }
int main() { return 0; }
!

cat > script.t << \!
MEMORY
{
  RAM (rw) : ORIGIN = 10000, LENGTH = 5000
}

SECTIONS {
 .foo : { *(.text.foo) } > RAM
 .empty : {} > RAM
 tbss : { *(.tbss*) } > RAM
 .empty : {} > RAM
 .main : { *(.text.main) } > RAM
 output_data : { BYTE(0x0) *(.data*) } > RAM
 /DISCARD/ : { *(.riscv.attributes) }
}
!

clang -c -target riscv32 -c  1.c -ffunction-sections -fdata-sections
ld.eld 1.o -T script.t
llvm-readelf -l -W a.out
```

**Bug**

```
$ llvm-readelf -l -W a.out

Elf file type is EXEC (Executable file)
Entry point 0x0
There are 5 program headers, starting at offset 52

Program Headers:
  Type           Offset   VirtAddr   PhysAddr   FileSiz MemSiz  Flg Align
  LOAD           0x000710 0x00002710 0x00002710 0x00012 0x00012 R E 0x1000
  LOAD           0x000724 0x00002724 0x00002724 0x00000 0x00000 RW  0x1000
  LOAD           0x000724 0x00002724 0x00002724 0x00016 0x00016 R E 0x1000
  LOAD           0x00073c 0x0000273c 0x0000273c 0x00008 0x00008 RW  0x1000
  TLS            0x000724 0x00002724 0x00002724 0x00000 0x00004 R   0x4

 Section to Segment mapping:
  Segment Sections...
   00     .foo
   01
   02     .main
   03     output_data
   04     tbss
   None   .comment .shstrtab .symtab .strtab
```

quic-seaswara:
TBSS sections are special, they do not consume any virtual memory size at load time, they incur a cost only when setting up size of the initial TLS image by either libc or libpthread.

Creating a segment when a TBSS section is encountered causes linker to create an empty segment.

Resolves #353

### DIFF
--- a/lib/Target/CreateProgramHeaders.hpp
+++ b/lib/Target/CreateProgramHeaders.hpp
@@ -247,8 +247,6 @@ bool GNULDBackend::createProgramHdrs() {
       continue;
     }
 
-    bool isPrevTBSS = prev && prev->isTBSS();
-
     bool isPrevRelRO =
         config().options().hasRelro() && prev && isRelROSection(prev);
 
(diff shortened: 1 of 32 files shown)


quic-seaswara:
The patch will be followed up with more test case including tests that can run. 

`createScriptProgramHdrs` also needs a change to support and will be a follow up.

Steven6798 (approved):

quic-areg (approved):

PR #373, opened 2025-08-30
Fix incorrect handling of negative VMA/LMA

Ris-Bali:
Currently eld does not emit any diagnostic when
negative VMA/LMA is specified in the linkerscript. This patch adds an error diagnostic to handle
such cases.


lib/Target/CreateProgramHeaders.hpp
```diff
@@ -250,7 +250,12 @@ bool GNULDBackend::createProgramHdrs() {
     bool doAlign = true;
     // If the output section specified a VMA value.
     if ((*out)->prolog().hasVMA()) {
-      (*out)->prolog().vma().evaluateAndRaiseError();
+      auto vmaExpr = (*out)->prolog().vma().evaluateAndRaiseError();
+      if (vmaExpr.has_value() && (int64_t)vmaExpr.value() < 0 &&
+          (int64_t)vmaExpr.value() > -0x100000) {
```
parth-07: Why do we have `-0x100000` magic constant here? Why is the first comparison with 0 not enough?

test/Common/standalone/NegativeVMA_LMA/NegativeVMA_LMA.test
```diff
@@ -0,0 +1,11 @@
+#--NegativeVMA_LMA.test------------------ Executable----------------#
+#BEGIN_COMMENT
+# Verify that the linker errors out when the VMA/LMA
+# is negative.
+#END_COMMENT
+RUN: %clang %clangopts -c %p/Inputs/1.c -o %t1.1.o
+RUN: %not %link %linkopts %t1.1.o -T %p/Inputs/script.t -o %t2_vma.out 2>&1 | %filecheck %s
+RUN: %not %link %linkopts %t1.1.o -T %p/Inputs/script_1.t -o %t2_lma.out 2>&1 | %filecheck %s --check-prefix=LMA
+
+CHECK: VMA is negative
```
parth-07, line 10: Please add the VMA/LMA value in the error message as well.

include/eld/Diagnostics/DiagLDScript.inc
```diff
@@ -75,6 +75,10 @@ DIAG(fatal_modulo_by_zero, DiagnosticEngine::Fatal,
      "%0: modulo by zero in expression %1")
 DIAG(error_experimental_not_supported, DiagnosticEngine::Error,
      "Linker has only experimental support for handling %0")
+DIAG(error_negative_vma, DiagnosticEngine::Fatal,
```
parth-07, line 78: Change the diagnostic severity to error here.

lib/Target/CreateProgramHeaders.hpp
Ris-Bali: If the address is very large casting it to int64_t gives a negative number. Eg: `HugeVMA.test` comparing only with zero will cause this test to fail.

parth-07: How did you come up with the 0x100000 constant?

Ris-Bali: Okay I think we should only check the address when its literal depending on the sign. Have removed the constant

PR #377, opened 2025-09-02
Fix LIT test failures in windows.

quic-shaksuma:
This commit fixes some LIT tests and marks some unsupported that are not supported on windows.

### DIFF
--- a/test/Common/Plugin/LinkerPluginVersion/LinkerPluginVersion.test
+++ b/test/Common/Plugin/LinkerPluginVersion/LinkerPluginVersion.test
@@ -11,27 +11,27 @@ RUN: %clang %clangopts -o %t1.1.o %p/Inputs/1.c -c
 RUN: not %link %linkopts --plugin-config %p/Inputs/PluginConfigNoVersion.yaml %t1.1.o --trace=plugin -o %t.out 2>&1 | %filecheck --check-prefix=NO_VERSION %s
 NO_VERSION: Plugin {{.*}} in library {{.*}} does not export getPluginAPIVersion function
 
-RUN: ELD_TEST_LINKER_PLUGIN_VERSION_MAJOR=4 ELD_TEST_LINKER_PLUGIN_VERSION_MINOR=0 \
+RUN: env ELD_TEST_LINKER_PLUGIN_VERSION_MAJOR=4 ELD_TEST_LINKER_PLUGIN_VERSION_MINOR=0 \
 RUN:   %link %linkopts --plugin-config %p/Inputs/PluginConfig.yaml %t1.1.o --trace=plugin -o %t.out 2>&1 | %filecheck --check-prefix=CURRENT %s
 CURRENT: Plugin version 4.0 in library
 CURRENT: Calling LinkerPluginVersion::Init plugin hook
 
-RUN: ELD_TEST_LINKER_PLUGIN_VERSION_MAJOR=4 ELD_TEST_LINKER_PLUGIN_VERSION_MINOR=0 \
+RUN: env ELD_TEST_LINKER_PLUGIN_VERSION_MAJOR=4 ELD_TEST_LINKER_PLUGIN_VERSION_MINOR=0 \
 RUN:   %link %linkopts --plugin-config %p/Inputs/PluginConfig.yaml %t1.1.o --trace=plugin -o %t.out 2>&1 | %filecheck --check-prefix=PAST_MINOR %s
 PAST_MINOR: Plugin version 4.0 in library
 PAST_MINOR: Calling LinkerPluginVersion::Init plugin hook
 
-RUN: ELD_TEST_LINKER_PLUGIN_VERSION_MAJOR=4 ELD_TEST_LINKER_PLUGIN_VERSION_MINOR=999 \
+RUN: env ELD_TEST_LINKER_PLUGIN_VERSION_MAJOR=4 ELD_TEST_LINKER_PLUGIN_VERSION_MINOR=999 \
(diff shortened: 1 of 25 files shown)


test/Common/Plugin/LinkerPluginVersion/LinkerPluginVersion.test
```diff
@@ -11,27 +11,27 @@ RUN: %clang %clangopts -o %t1.1.o %p/Inputs/1.c -c
...
 CURRENT: Calling LinkerPluginVersion::Init plugin hook
 
-RUN: ELD_TEST_LINKER_PLUGIN_VERSION_MAJOR=4 ELD_TEST_LINKER_PLUGIN_VERSION_MINOR=0 \
+RUN: env ELD_TEST_LINKER_PLUGIN_VERSION_MAJOR=4 ELD_TEST_LINKER_PLUGIN_VERSION_MINOR=0 \
 RUN:   %link %linkopts --plugin-config %p/Inputs/PluginConfig.yaml %t1.1.o --trace=plugin -o %t.out 2>&1 | %filecheck --check-prefix=PAST_MINOR %s
 PAST_MINOR: Plugin version 4.0 in library
 PAST_MINOR: Calling LinkerPluginVersion::Init plugin hook
 
-RUN: ELD_TEST_LINKER_PLUGIN_VERSION_MAJOR=4 ELD_TEST_LINKER_PLUGIN_VERSION_MINOR=999 \
+RUN: env ELD_TEST_LINKER_PLUGIN_VERSION_MAJOR=4 ELD_TEST_LINKER_PLUGIN_VERSION_MINOR=999 \
```
parth-07, line 24: Why do we need `env` here? I think we had this test previously passing on Windows without the `env`, right?

test/Common/standalone/BinaryFormatInputFileReproduce/BinaryFormatInputFileReproduce.test
```diff
@@ -1,4 +1,4 @@
-UNSUPPORTED: reproduce_fail
+#UNSUPPORTED: windows, reproduce_fail
```
parth-07, line 1: Why is this test unsupported for windows? Can you please add a comment regarding this?

test/Common/Plugin/TarWriterTests/TarWriterUnwritablePathTest/TarWriterUnwritablePathTest.test
```diff
@@ -1,4 +1,4 @@
-UNSUPPORTED: windows
+#UNSUPPORTED: windows
```
parth-07, line 1: Why do we need `#` before UNSUPPORTED? Previously, the UNSUPPORTED directive was working fine on Windows as well, right?

test/Common/standalone/CommandLine/Reproduce/Reproduce.test
```diff
@@ -1,4 +1,4 @@
-UNSUPPORTED: reproduce_fail
+#UNSUPPORTED: windows, reproduce_fail
```
parth-07, line 1: Please add a comment regarding why this test is unsupported on Windows.

parth-07 (changes requested):
Can you please add a comment briefly describing why the test is unsupported for windows in the test itself for the tests that previously were not unsupported for windows?

quic-shaksuma:
> Can you please add a comment briefly describing why the test is unsupported for windows in the test itself for the tests that previously were not unsupported for windows?

These tests are marked unsupported temporarily to unblock the downstream builders. These tests seem to fail on those builder due variation in behavior of tools like rm, tar, etc. available currently on those builders.
These tests will be re-visited later to re-enable them. A ticket is created for that.

test/Common/standalone/CommandLine/Reproduce/Reproduce.test
quic-shaksuma: > Please add a comment regarding why this test is unsupported on Windows.

to unblock downstream builder. tar.exe used there doesn't behave as expected.

test/Common/Plugin/TarWriterTests/TarWriterUnwritablePathTest/TarWriterUnwritablePathTest.test
quic-shaksuma: > Why do we need `#` before UNSUPPORTED? Previously, the UNSUPPORTED directive was working fine on Windows as well, right?

It was showing up in failed list, also, '#' matches the format we use in all tests

test/Common/Plugin/LinkerPluginVersion/LinkerPluginVersion.test
quic-shaksuma: > Why do we need `env` here? I think we had this test previously passing on Windows without the `env`, right?

using 'env' is the proper way to set vars like ELD_TEST_LINKER_PLUGIN_VERSION_MAJOR. This was failing on the downstream builder.

test/Common/standalone/BinaryFormatInputFileReproduce/BinaryFormatInputFileReproduce.test
quic-shaksuma: same as other unsupported tests

test/Common/Plugin/LinkerPluginVersion/LinkerPluginVersion.test
parth-07: Oh, okay. Thank you for the info.

parth-07 (approved):
Thank you for the explanations. Looks good to me.

PR #381, opened 2025-09-03
Fix DiagnosticEngine deadlock that can happen when the link crashes

### LINKED ISSUE #379: The linker goes into a dead-lock if it crashes when emitting a diagnostic
The linker goes into a dead-lock if it crashes when emitting a diagnostic. Let's see why it happens:

<pre>
DiagEngine->raise(...)
  - Mutex->lock()
  - ...
  - _crash happens_
  - SignalHandler is called

SignalHandler()
  - Calls DiagEngine->raise(...) to emit 'Unexpected linker behavior diagnostic'. But Mutex was never unlocked!!!
</pre>

Reproducer:

```bash
cat >PluginDiagCrash.cpp <<\EOF
#include "Defines.h"
#include "LinkerPlugin.h"
#include "PluginBase.h"
#include "PluginVersion.h"
#include <iostream>

class DLL_A_EXPORT PluginDiagCrash : public eld::plugin::LinkerPlugin {
public:
  PluginDiagCrash() : eld::plugin::LinkerPlugin("PluginDiagCrash") {}
  void Init(const std::string &options) override {
    auto diagID = getLinker()->getNoteDiagID("A note diagnostic: %0 %1");
    getLinker()->reportDiag(diagID, "A", "Long", "List", "Of", "Too", "Many",
                            "Arguments", "That", "Exceeds", "The", "Max",
                            "Number", "Of", "Supported", "Arguments");
  };
};

ELD_REGISTER_PLUGIN(PluginDiagCrash)
EOF

cat >PluginConfig.yaml <<\EOF
GlobalPlugins:
  - Type: LinkerPlugin
    Name: PluginDiagCrash
    Library: PluginDiagCrash
EOF

# Build the plugin

# Linker goes into deadlock
ld.eld -o /dev/null /dev/null --plugin-config PluginConfig.yaml
```

parth-07:
This commit fixes DiagnosticEngine deadlock that happens when a signal is delivered to a thread that is currently emitting a diagnostic. Let's see why the deadlock happens:

Thread T1 emitting a diagnostic -> DiagnosticEngine.Mutex.Lock() -> crash / assert happens -> SignalHandler() -> Emits 'Unexpected linker behavior' diagnostic but the DiagnosticEngine.Mutex lock was never freed.

This commit fixes this deadlock by:

1) Aborting the diagnostic-in-flight in-case the same thread that is
   emitting a diagnostic receives the signal. This is the most common
   case because program errors are thread-directed signals.
2) Checking if the diagnostic engine is usable. And if so, use the
   diagnostic engine; otherwise, manually emit error to the error
   stream.

Resolves #379


parth-07:
>Please run a windows build, and make sure it is green

I have made sure the functionality works on Windows.

lib/Diagnostics/DiagnosticEngine.cpp
```diff
@@ -180,6 +182,14 @@ DiagnosticEngine::DiagIDType DiagnosticEngine::getBaseDiagID(DiagIDType Id) {
   return Id;
 }
 
+bool DiagnosticEngine::isUsable() {
+  std::unique_lock<std::timed_mutex> Lock(Mutex, std::defer_lock_t{});
+  std::chrono::seconds WaitThreshold(2);
```
quic-seaswara, line 187: Can you add a comment here it is 2 seconds, for people reading the code, it might be easier or change it 2000ms

lib/Diagnostics/DiagnosticEngine.cpp
```diff
@@ -180,6 +182,14 @@ DiagnosticEngine::DiagIDType DiagnosticEngine::getBaseDiagID(DiagIDType Id) {
...
+  if (Lock.try_lock_for(WaitThreshold))
```
quic-seaswara, line 188: return Lock.try_lock_for(WaitThreshold) ?

PR #382, opened 2025-09-03
Fix tests for reverse iteration

quic-areg:
Adds CHECK-DAGs to iteration-sensitive tests. INIWriter now uses a std::map instead of a StringMap for consistent iteration.

### DIFF
--- a/include/eld/Support/INIWriter.h
+++ b/include/eld/Support/INIWriter.h
@@ -71,7 +71,7 @@ public:
   /// \returns layout of this ini file in the form of an std::ostream
   friend std::ostream &operator<<(std::ostream &os, INIWriter const &m) {
     for (auto &section : m.ini) {
-      os << "[" << section.first().str() << "]\n";
+      os << "[" << section.first << "]\n";
       os << *section.second;
       os << "\n";
     }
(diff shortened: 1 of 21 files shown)


include/eld/Support/INIWriter.h
```diff
@@ -105,7 +105,7 @@ class INIWriter {
   }
 
 private:
-  llvm::StringMap<INISection *> ini;
+  std::map<std::string, INISection *> ini;
```
quic-seaswara, line 108: This may cause a slowdown when using INI files ?

quic-seaswara:
> Adds CHECK-DAG's to iteration sensitive tests. INIWriter uses a std::map instead of a StringMap for consitent iteration.

spelling errors

include/eld/Support/INIWriter.h
quic-seaswara: Can we make the INI test unsupported with reverse_iterators ? instead of changing the data structure

quic-areg: > This may cause a slowdown when using INI files ?
Unless the config file has thousands of entries I would not expect a slowdown for config files. std::map is already used to represent the Sections in the ini file.

PR #390, opened 2025-09-09
Fix YamlLayoutPrinter crash when the link contains an empty archive

### LINKED ISSUE #387: YAMLLayoutPrinter crashes if the link contains an empty archive.
YAMLLayoutPrinter crashes if the link contains an empty archive.

parth-07:
This commit fixes YamlLayoutPrinter crash when the link contains an empty archive. The root cause of the issue was that 'ArchiveFile::getAllMembers' incorrectly assumed that this function will only be used for non-empty archives.

Resolves #387


lib/Input/ArchiveFile.cpp
```diff
@@ -95,7 +95,9 @@ bool ArchiveFile::isAlreadyReleased() const {
 
 // Returns all members of the archive file.
 const std::vector<Input *> &ArchiveFile::getAllMembers() const {
-  ASSERT(AFI, "AFI must not be null!");
+  static std::vector<Input *> EmptyInputVector;
```
quic-seaswara: Pls remove static

quic-seaswara (changes requested):

lib/Input/ArchiveFile.cpp
parth-07: We cannot remove static from here. The function returns a const-reference. If we create a local variable and then return it, or return `{}`, then we would be returning a dangling reference because the reference would be pointing to a local object on stack that is deleted when function returns.

quic-seaswara: Sorry I did not recognize the const reference return. I am fine with creating an empty archive file info instead.

parth-07: I have changed the solution to use empty archive file info instead.

PR #391, opened 2025-09-10
Fix YamlLayoutPrinter crash when the link contains an empty archive

### LINKED ISSUE #387: YAMLLayoutPrinter crashes if the link contains an empty archive.

parth-07:
This commit fixes YamlLayoutPrinter crash when the link contains an empty archive. The root cause of the issue was that 'ArchiveFile::getAllMembers' incorrectly assumed that this function will only be used for non-empty archives.

Resolves #387


lib/Readers/ArchiveParser.cpp
```diff
@@ -50,8 +50,11 @@ eld::Expected<uint32_t> ArchiveParser::parseFile(InputFile &inputFile) const {
   LLVMEXP_RETURN_DIAGENTRY_IF_ERROR(expArchiveReader);
   std::unique_ptr<llvm::object::Archive> archiveReader =
       std::move(*expArchiveReader);
-  if (archiveReader->isEmpty())
-    return {};
+  if (archiveReader->isEmpty()) {
+    if (!hasAFI)
+      archiveFile->initArchiveFileInfo();
```
quic-seaswara, line 58: Can we add a note that archive is empty ?

parth-07: I have added a note diagnostic for empty archives.

quic-seaswara (approved):

apazos:
We are validating the change, will report once testing is complete.

PR #394, opened 2025-09-15
Add support for displaying plugin stats in map file

Ris-Bali:
This patch adds support for displaying plugin stats in the linker map file


lib/Script/Plugin.cpp
```diff
@@ -112,6 +112,20 @@ bool Plugin::registerPlugin(void *Handle) {
...
+void Plugin::addLinkStat(std::string StatName, std::string Value) {
+  PluginStats.push_back({StatName, Value});
+}
+
+std::string Plugin::getLinkStats() {
+  std::ostringstream oss;
+  for (size_t i = 0; i < PluginStats.size(); ++i) {
+    oss << PluginStats[i].first << ": " << PluginStats[i].second;
+    if (i != PluginStats.size() - 1)
+      oss << ", ";
```
quic-seaswara: Why , let us use new lines

lib/Script/Plugin.cpp
```diff
@@ -112,6 +112,20 @@ bool Plugin::registerPlugin(void *Handle) {
   return true;
 }
 
+void Plugin::addLinkStat(std::string StatName, std::string Value) {
+  PluginStats.push_back({StatName, Value});
+}
+
+std::string Plugin::getLinkStats() {
```
quic-seaswara: const function, getPluginLinkStats()

include/eld/Script/Plugin.h
```diff
@@ -230,6 +234,7 @@ class Plugin {
   std::string PluginLibraryName;
   std::string PluginType;
   std::string PluginOptions;
+  std::vector<std::pair<std::string, std::string>> PluginStats;
```
quic-seaswara, line 238: What if the linkstats added need to be updated ?

include/eld/Core/LinkerScript.h
```diff
@@ -240,6 +240,9 @@ class LinkerScript {
 
   void updateRuleOp(plugin::LinkerWrapper *W, eld::Module *M, RuleContainer *R,
                     ELFSection *S, const std::string &Annotation = "");
+  void updateLinkStatsOp(plugin::LinkerWrapper *W, eld::Module *M,
+                         const std::string &StatName,
+                         const std::string &Annotation = "");
```
quic-seaswara: Do you save Annotation ?

quic-seaswara (changes requested):

lib/Script/Plugin.cpp
Ris-Bali: I think it would be better to separate the plugin stats for the same plugin by ", " I am using a new line to separate the stats of different plugins in the map file

include/eld/Core/LinkerScript.h
Ris-Bali: The annotation stores the value of the linkstat

include/eld/Script/Plugin.h
Ris-Bali: I have replaced the vector with an unordered map

include/eld/Core/LinkerScript.h
parth-07: What is the `Annotation` parameter here? Why does it have a default value? Additionally, what do you think about `StatName` and `StatValue` as the parameter names?

include/eld/LayoutMap/TextLayoutPrinter.h
```diff
@@ -105,6 +105,8 @@ class TextLayoutPrinter {
 
   bool printUpdateRulePluginOp(PluginOp *Pop) const;
 
+  bool printUpdateLinkStatsPluginOp(PluginOp *Pop) const;
```
parth-07: Please take the plugin as a reference. I believe we should try to minimize the pointer use whenever possible.

include/eld/Plugin/PluginOp.h
```diff
@@ -31,6 +31,7 @@ class PluginOp {
     ResetOffset,
     UpdateChunks,
     UpdateRule,
+    UpdateLinkStat,
```
parth-07, line 33: Can you please add it as per lexicographical order?

include/eld/Script/Plugin.h
```diff
@@ -172,6 +172,10 @@ class Plugin {
   /// Record fragment remove operation for fragment movement verification.
   void recordFragmentRemove(RuleContainer *R, Fragment *F);
 
+  void addLinkStat(std::string StatName, std::string Value);
```
parth-07:
```suggestion
  void addLinkStat(const std::string &StatName, const std::string &Value);
```

include/eld/Script/Plugin.h
```diff
@@ -230,6 +234,7 @@ class Plugin {
   std::string PluginLibraryName;
   std::string PluginType;
   std::string PluginOptions;
+  std::unordered_map<std::string, std::string> PluginStats;
```
parth-07: Please do not use unordered_map here. I would like the plugin stats to remain in the same order as the plugin inserts them.

lib/LayoutMap/TextLayoutPrinter.cpp
```diff
@@ -441,6 +444,13 @@ bool TextLayoutPrinter::printRemoveChunkPluginOp(PluginOp *Pop) const {
   return true;
 }
 
+bool TextLayoutPrinter::printUpdateLinkStatsPluginOp(PluginOp *Pop) const {
```
parth-07: Where is this function used?

lib/Script/Plugin.cpp
```diff
@@ -112,6 +112,19 @@ bool Plugin::registerPlugin(void *Handle) {
   return true;
 }
 
+void Plugin::addLinkStat(std::string StatName, std::string Value) {
+  PluginStats[StatName] = Value;
+}
+
+std::string Plugin::getPluginLinkStats() const {
```
parth-07: Plugin class should return the stats in a way such that the consumer can iterate over the stats. And the consumer (in this case, TextLayoutPrinter) should handle the formatting of the plugin stats. We should not mix the responsibilities.

test/Common/Plugin/AddPluginLinkStats/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+set(SOURCES AddPluginStats.cpp)
+
+if(NOT CYGWIN AND LLVM_ENABLE_PIC)
+  set(SHARED_LIB_SOURCES ${SOURCES})
+
+  set(bsl ${BUILD_SHARED_LIBS})
+
+  set(BUILD_SHARED_LIBS ON)
+
+  add_llvm_library(AddPluginStats ${SHARED_LIB_SOURCES} LINK_LIBS LW)
```
parth-07, line 10: Can we please house the test plugins artifacts in `lib/test` object directory instead?

test/Common/Plugin/AddPluginLinkStats/AddPluginStats.cpp
```diff
@@ -0,0 +1,23 @@
+#include "LinkerPlugin.h"
+#include "PluginVersion.h"
+#include <iostream>
+#include <string>
+
+class AddPluginStats : public eld::plugin::LinkerPlugin {
+public:
+  AddPluginStats() : eld::plugin::LinkerPlugin("AddPluginStats") {}
+  void Init(const std::string &Options) override;
```
parth-07: You do not need to add empty definitions in the plugins.

parth-07 (changes requested):

lib/Script/Plugin.cpp
parth-07: I agree with @quic-seaswara. Additionally, can we prefix the plugin stats with `[PluginName]` instead of suffixing the stat with the plugin name? For example, from the test:

```
SomeStat: 15 AddPluginStats
```

AddPluginStats at the end seems non-intuitive to me...

include/eld/Script/Plugin.h
quic-seaswara: I feel its ok to see a different order, is it critical to be deterministic here ?

parth-07: There are two major benefits of a consistent order of the plugin stats:

1) The stats in the right order can be more readable/easy-to-follow. For example, Input file count stat naturally precedes input sections count stat.
2) Less noise when comparing the map file diffs. If inconsistent order is used, then we may see false-positive diffs when diffing map-files.

include/eld/Core/LinkerScript.h
Ris-Bali: Annotation is being used to store the value of the stat. We can have an extra parameter for the value as well.

include/eld/Core/LinkerScript.h
```diff
@@ -240,6 +240,9 @@ class LinkerScript {
 
   void updateRuleOp(plugin::LinkerWrapper *W, eld::Module *M, RuleContainer *R,
                     ELFSection *S, const std::string &Annotation = "");
+  void updateLinkStatsOp(plugin::LinkerWrapper *W, eld::Module *M,
+                         const std::string &StatName,
+                         const std::string &StatValue = "");
```
parth-07: We do not need to give `StatValue` parameter a default value. It should never be empty, right?

lib/LayoutMap/TextLayoutPrinter.cpp
```diff
@@ -179,6 +188,10 @@ void TextLayoutPrinter::printStats(LayoutInfo::Stats &L,
               std::to_string(L.OutputFileSize.value()) + " bytes");
 
   printStat("LinkTime", L.LinkTime);
+  for (auto &W : ThisLayoutInfo->getPlugins()) {
+    Plugin *P = W->getPlugin();
+    printPluginStats(P->getPluginLinkStats(), P->getPluginName());
```
parth-07: Why are we passing plugin name instead of the plugin itself here?

lib/Script/Plugin.cpp
```diff
@@ -112,6 +112,22 @@ bool Plugin::registerPlugin(void *Handle) {
...
+  for (uint64_t i = 0; i < PluginStats.size(); ++i) {
+    if (PluginStats[i].first == StatName) {
+      PluginStats[i].second = Value;
+      return;
+    }
+  }
+  PluginStats.push_back({StatName, Value});
+}
+
+std::vector<std::pair<std::string, std::string>>
```
parth-07:
```suggestion
const std::vector<std::pair<std::string, std::string>> &
```

test/Common/Plugin/AddPluginLinkStats/AddPluginLinkStats.test
```diff
@@ -0,0 +1,12 @@
...
+#BEGIN_COMMENT
+# This test checks if plugin stats are added to the map file.
+#END_COMMENT
+#START_TEST
+RUN: %clang %clangopts -c %p/Inputs/1.c -o %t1.1.o -ffunction-sections
+RUN: %link %linkopts %t1.1.o -o %t2.out -L%libsdir/test --plugin-config %p/Inputs/plugin.config -Map %t1.1.map
+RUN: %filecheck %s < %t1.1.map
+
+#CHECK-DAG: SomeStat: 15 AddPluginStats
+#CHECK-DAG: OtherStat: 18 AddPluginStats
```
parth-07:
```suggestion
#CHECK: SomeStat: 15 AddPluginStats
#CHECK: OtherStat: 18 AddPluginStats
```

parth-07 (changes requested):

PR #397, opened 2025-09-16
Define __eh_frame_* symbols as standard symbols

quic-k:
Libunwind needs the following symbols for exception handling

__eh_frame_start __eh_frame_end __eh_frame_hdr_start __eh_frame_hdr_end

Instead of using a linker script just to define these symbols, let the linker define them as standard symbols.
Now we can create static executable with libunwind without a linker script.


lib/Target/GNULDBackend.cpp
```diff
@@ -321,6 +321,10 @@ bool GNULDBackend::initStandardSymbols() {
     }
   };
 
+  InitStandardSym("__eh_frame_start", "__eh_frame_end",
+                  m_Module.getSection(".eh_frame"));
```
parth-07, line 325: Are the eh_frame and eh_frame_hdr always present?

quic-k: i think .eh_frame is generated unless exception handling is explicitly disabled + some clang flags
.eh_frame_hdr is generated is --eh-frame-hdr flag is passed to linker

quic-k:
this is not related to Picolibc
for libunwind, I don't know
I think baremetal toolchains use linker script and non baremetal toolchains don't use these symbols(dynamic linking)


libunwind using __eh_frame_* for baremetal
[https://github.com/llvm/llvm-project/blob/main/libunwind/src/AddressSpace.hpp#L520-L526](https://github.com/llvm/llvm-project/blob/main/libunwind/src/AddressSpace.hpp#L520-L526)

quic-k:
added tests

test/Common/standalone/MagicSectionSymbols/MagicSectionSymbols.test
```diff
@@ -0,0 +1,19 @@
+#---MagicSectionSymbols.test--------------------------- Executable -----------------#
+#BEGIN_COMMENT
+# This test checks that start and end symbols for magic sections .eh_frame and .eh_frame_hdr are defined by the linker
+#END_COMMENT
+#START_TEST
+RUN: %clang %clangopts -o %t.o -c %p/Inputs/1.cpp %clangg0opts
```
parth-07: We should not use clangopts and clangg0opts together. clangg0opts is meant to be used when we want to compile and link with '-G0' option.

parth-07: Also, does the input have to be .cpp? If not, then can you please make it a .c file to be consistent with the rest of the test files?

quic-k: fixed

quic-seaswara (approved):

PR #403, opened 2025-09-17
Fix -1 e_flags with binary inputs

quic-areg:
Binary files have no architecture flags, so the
default value of -1 is used instead of 0.


quic-areg:
> ARM/AArch64 does not implement checkFlags

lld seems to not zero flags for binary inputs for Hexagon, Arm, AArch64 either so that should be ok

quic-seaswara (changes requested):
Please zero the flags for all architectures, it does not make sense to not have the flags 0'ed for specific architectures. 

$arm-none-eabi-ld.bfd -marmelf  --format=binary mydata.blob -o mydata.elf && llvm-readelf -h mydata.elf | grep Flags
  Flags:                             0x0

$ /local/mnt/workspace/shankare/tools/aarch64-gnu-tools/bin/aarch64-linux-gnu-ld.bfd -marmelf  --format=binary mydata.blob -o mydata.elf && llvm-readelf -h mydata.elf | grep Flags

$aarch64-linux-gnu-ld.bfd: warning: cannot find entry symbol _start; defaulting to 0000000000008000
  Flags:                             0x0

lib/Readers/BinaryFileParser.cpp
```diff
@@ -24,6 +24,10 @@ eld::Expected<void> BinaryFileParser::parseFile(InputFile &inputFile) {
   ELFSection *S = createDataSection(inputFile);
   ObjectFile *objFile = llvm::cast<ObjectFile>(&inputFile);
   objFile->addSection(S);
+  // Binary files carry no architecture flags. Explicitly set flags to 0 to
+  // avoid the sentinel value of -1.
+  if (m_Module.getLinker()->getBackend())
```
parth-07: @quic-seaswara @quic-areg Should we change this `if (backend)` to an assert? I feel that due to lack of invariants in the codebase, and the use of pointers, we tend to add more if-conditions than necessary and it makes the codebase more difficult to read.

test/RISCV/standalone/BinaryInputs/flags.test
```diff
@@ -0,0 +1,10 @@
+#---flags.test--------------------- Executable---------------------#
```
parth-07, line 1: We should probably add this test in Common/standalone.

test/RISCV/standalone/BinaryInputs/flags.test
```diff
@@ -0,0 +1,10 @@
...
+#BEGIN_COMMENT
+# Checks output flags are 0 for binary inputs
+#END_COMMENT
+#START_TEST
+RUN: %link %emulation %linkopts --format=binary %s -o %t.out
```
parth-07, line 6: Why do we have to explicitly use %emulation here?

quic-areg: no target can be inferred from a binary file alone:

`Error: target emulation unknown: -m or at least one .o file required`

lib/Readers/BinaryFileParser.cpp
quic-areg: There is already an assert in Module::getBackend() that would be hit here without the `if(backend)` in the case where there is no emulation specified and no .o files in the link

lib/Target/Hexagon/HexagonInfo.cpp
```diff
@@ -310,6 +311,8 @@ bool HexagonInfo::checkFlags(uint64_t pFlag,
 
 /// flags - the value of ElfXX_Ehdr::e_flags
 uint64_t HexagonInfo::flags() const {
+  if (m_OutputFlag == LINK_UNKNOWN)
+    return 0;
```
quic-seaswara, line 318: This will break the use case of 

```
rm -f empty.o && touch empty.o && ld.eld -marmelf empty.o && llvm-readelf -h a.out | grep Flags
rm -f empty.o && touch empty.o && ld.eld -mhexagonelf empty.o && llvm-readelf -h a.out | grep Flags
rm -f empty.o && touch empty.o && ld.eld -melf32lriscv empty.o && llvm-readelf -h a.out | grep Flags
```

The linux kernel builds infers the architecture using this method. I am not sure if we have a test will add it later.

lib/LinkerWrapper/GnuLdDriver.cpp
```diff
@@ -1300,6 +1300,8 @@ bool GnuLdDriver::createInputActions(llvm::opt::InputArgList &Args,
     case T::input_format: {
       actions.push_back(eld::make<eld::InputFormatAction>(arg->getValue(),
                                                           Config.getPrinter()));
+      if (llvm::StringRef(arg->getValue()) == "binary")
+        Config.options().setBinaryInput();
```
quic-seaswara: 1. setBinaryOutput
2. arg->getValue() is not StringRef() ?
3. case insenstive comparison ?

Why do we need this new entry in Config options() ?

lib/Target/ARM/ARMInfo.cpp
```diff
@@ -66,3 +66,14 @@ bool ARMInfo::InitializeDefaultMappings(Module &pModule) {
 }
 
 std::string ARMInfo::flagString(uint64_t flag) const { return "arm"; }
+
+uint64_t ARMInfo::flags() const {
+  if (m_Config.options().isBinaryInput() && !OutputFlags)
+    return 0;
+  return llvm::ELF::EF_ARM_EABI_VER5;
```
quic-seaswara: What about other files such as if the input just contains symdef files ?

lib/Target/ARM/ARMInfo.cpp
```diff
@@ -66,3 +66,14 @@ bool ARMInfo::InitializeDefaultMappings(Module &pModule) {
...
+}
+
+bool ARMInfo::checkFlags(uint64_t Flags, const InputFile *I) const {
+  OutputFlags = llvm::ELF::EF_ARM_EABI_VER5;
+  return true;
```
quic-seaswara, line 85: It looks like there is no check for flags, do you need this ? 

isn't flags() good enough ?

quic-seaswara (changes requested):
Instead of creating seperate commits this way, I would appreciate if you could seperate each target specific change to a seperate commit or a PR. 

Combining all of them together is making it hard to review.

lib/LinkerWrapper/GnuLdDriver.cpp
quic-areg: 1. the output is probably ELF? 
2. it's a const char *
3. sure 

I added this to check that we only zero flags in the case of binary inputs, and that we preserve eflags for empty .o files per your last comment.

It isn't a new entry in Config options; it's actually a really old one that was never used anywhere (it seems like we have lots of dead code like this)

lib/Target/ARM/ARMInfo.cpp
quic-areg: If the linker has a binary input and an ELF input then the result should have eflags.

lib/Target/Hexagon/HexagonInfo.cpp
quic-areg: I added a check that we only emit zero flags if there is a binary input

lib/Target/ARM/ARMInfo.cpp
quic-areg: the output will have arch flags

lib/Target/ARM/ARMInfo.cpp
quic-seaswara: You already account it in flags(), you dont need OutputFlags member variable for ARM I believe.

lib/Target/ARM/ARMInfo.cpp
quic-seaswara: Please add a test

lib/LinkerWrapper/GnuLdDriver.cpp
quic-seaswara: Please get rid of setBinaryInput()

--format=binary is a positional option, you can figure out if the input is binary or not by looking at the attribute (isBinary). See `include/eld/Input/InputTree.h`

quic-seaswara: Please create a ticket for dead code that you recognize when reading code, it is important that we start cleaning up

lib/Target/ARM/ARMInfo.cpp
quic-areg: I don't think so.

```cpp
uint64_t ARMInfo::flags() const {
  if (m_Config.options().isBinaryInput())
    return 0;
  return llvm::ELF::EF_ARM_EABI_VER5;
 }
 ```

if we have two inputs, a binary and an ELF, using only this code like you suggested and not what I have in `checkFlags`, then the result will have zero flags instead of `EABI_VER_5`:

```
$ git diff 
```
```diff
diff --git a/lib/Target/ARM/ARMInfo.cpp b/lib/Target/ARM/ARMInfo.cpp
index d89cc65..56e442f 100644
--- a/lib/Target/ARM/ARMInfo.cpp
+++ b/lib/Target/ARM/ARMInfo.cpp
@@ -68,12 +68,7 @@ bool ARMInfo::InitializeDefaultMappings(Module &pModule) {
 std::string ARMInfo::flagString(uint64_t flag) const { return "arm"; }
 
 uint64_t ARMInfo::flags() const {
-  if (m_Config.options().isBinaryInput() && !OutputFlags)
+  if (m_Config.options().isBinaryInput())
     return 0;
   return llvm::ELF::EF_ARM_EABI_VER5;
 }
-
-bool ARMInfo::checkFlags(uint64_t Flags, const InputFile *I) const {
-  OutputFlags = llvm::ELF::EF_ARM_EABI_VER5;
-  return true;
-}
\ No newline at end of file
diff --git a/lib/Target/ARM/ARMInfo.h b/lib/Target/ARM/ARMInfo.h
index 1c73622..ab3556c 100644
--- a/lib/Target/ARM/ARMInfo.h
+++ b/lib/Target/ARM/ARMInfo.h
@@ -81,11 +81,6 @@ public:
       m_Config.options().setWarnMismatch(false);
     }
   }
-
-  bool checkFlags(uint64_t Flag, const InputFile *I) const override;
-
-private:
-  mutable uint64_t OutputFlags = 0;
 };
 } // namespace eld
```
without `checkFlags`:
```
$ ld.eld -marmelf 1.o --format=binary data 
$ llvm-readelf -h a.out | grep Flags
  Flags:                             0x0
```
with `checkFlags`
```
$ ld.eld -marmelf 1.o --format=binary data 
$ llvm-readelf -h a.out | grep Flags
  Flags:                             0x5000000
```

test/Common/standalone/BinaryFormatInputFiles/flags.test
```diff
@@ -0,0 +1,10 @@
+#---flags.test--------------------- Executable---------------------#
+#BEGIN_COMMENT
+# Checks output flags are 0 for binary inputs
+#END_COMMENT
+#START_TEST
+RUN: %link %emulation %linkopts --format=binary %s -o %t.out
+RUN: %readelf -h %t.out | %filecheck %s
+#END_TEST
+
+CHECK: Flags: 0x0
```
quic-seaswara, line 10: Please improve the test to cover all the cases that you have modeled in the code.

lib/Target/ARM/ARMInfo.cpp
quic-seaswara: This is because of the check isBinaryInput(), we should not have that check.

--format=binary is a positional option.

ARMInfo::flags() { return m_OutputFlag; }

ARMInfo::checkFlags(int flag, InputFile *f) { 
  if (flag > m_OutputFlag) 
    m_OutputFlag = flag; 
}

lib/Target/ARM/ARMInfo.cpp
```diff
@@ -67,7 +67,12 @@ bool ARMInfo::InitializeDefaultMappings(Module &pModule) {
 
 std::string ARMInfo::flagString(uint64_t flag) const { return "arm"; }
 
+uint64_t ARMInfo::flags() const { return OutputFlags; }
+
 bool ARMInfo::checkFlags(uint64_t Flags, const InputFile *I, bool) const {
-  OutputFlags = llvm::ELF::EF_ARM_EABI_VER5;
+  if (!Flags && !OutputFlags && I->isBinaryFile())
+    OutputFlags = 0;
```
quic-seaswara: Why I->isBinaryFile() ?

lib/Target/ARM/ARMInfo.cpp
```diff
@@ -67,7 +67,12 @@ bool ARMInfo::InitializeDefaultMappings(Module &pModule) {
...
+  else
+    OutputFlags = llvm::ELF::EF_ARM_EABI_VER5;
```
quic-seaswara, line 84: Why OutputFlags is set to VER5 ? Can we just pick up whatever is recorded in the object file by comparing flags.

quic-areg: The existing behavior is to emit VER5 ignoring the object file. Picking up flags from the object file seems like a separate change that will require additional error checking/testing. Should I create an issue for this?

lib/Target/ARM/ARMInfo.cpp
quic-areg: The existing behavior is to ignore whatever is in the object file and use VER5. For this PR, we want to make an exception for binary files. Are there any other files we should make an exception for?

If we want to start reading flags from object files on ARM, I feel that should be a separate change.

test/Common/standalone/BinaryFormatInputFiles/flags.test
quic-areg: I added some tests in #409 I will add tests for binary inputs mixed with object inputs for this PR. Any other case I should test?

lib/LinkerWrapper/GnuLdDriver.cpp
quic-areg: #410

lib/Target/ARM/ARMInfo.cpp
quic-areg: #409

quic-seaswara (approved):

lib/Target/ARM/ARMInfo.cpp
quic-seaswara: Sure

lib/Target/ARM/ARMInfo.cpp
quic-seaswara: Sure

quic-seaswara:
Please document this behavior in linker documentation

lib/Readers/BinaryFileParser.cpp
```diff
@@ -25,6 +25,9 @@ eld::Expected<void> BinaryFileParser::parseFile(InputFile &inputFile) {
   ObjectFile *objFile = llvm::cast<ObjectFile>(&inputFile);
   objFile->addSection(S);
   addDescriptionSymbols(inputFile, S);
+  // Binary files carry no architecture flags. Explicitly set flags to 0.
+  if (m_Module.getLinker()->getBackend())
```
parth-07, line 29: We do not need this `if` condition as `getBackend()` must always be tree -- there is an assert for it to be true in `getBackend()` function as per the previous discussion.

lib/Target/ARM/ARMInfo.h
```diff
@@ -81,6 +81,11 @@ class ARMInfo : public TargetInfo {
       m_Config.options().setWarnMismatch(false);
     }
   }
+
+  bool checkFlags(uint64_t Flag, const InputFile *I, bool) const override;
+
+private:
+  mutable std::optional<uint64_t> OutputFlags;
```
parth-07, line 88: Why do we make it mutable instead of making the function non-const?

lib/Readers/BinaryFileParser.cpp
quic-areg: As per the previous discussion, the if condition is necessary for the case where there are no object files and no emulation provided to the linker, and therefore no backend is available:

without if:
```
ld.eld --format=binary data 
LLVM ERROR: Assertion `Linker->getBackend()` failed in ~/work/llvm-project/llvm/tools/eld/lib/Core/Module.cpp line 635: The value must be non-null.
Fatal: !!!UNEXPECTED LINKER BEHAVIOR!!!
Fatal: Please rerun link with /tmp/reproduce-d2a78b.sh and contact support
```

with if:
```
ld.eld --format=binary data 
Error: target emulation unknown: -m or at least one .o file required
Fatal: Linking had errors.
```

The linker should not crash.

lib/Target/ARM/ARMInfo.h
quic-areg: I agree; I will take a look at this in a follow-up

lib/Readers/BinaryFileParser.cpp
parth-07: Oh, thank you for the information.

parth-07 (approved):
Looks good!

PR #408, opened 2025-09-18
Improve compatible checks for RISC-V

### LINKED ISSUE #407: Ignore RISC-V elf compatibility checks if e_flags is zero and all sections in the input file are non-executable
**Description**
The psabi has a note under https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#elf-object-files that "The static linker may ignore the compatibility checks if all fields in the e_flags are zero and all sections in the input file are non-executable sections."

Currently, eld produces errors for the reproducers below which I think are reasonable enough patterns and don't have great workarounds short of manually editing the binary blob elfs. Combined with https://github.com/qualcomm/eld/issues/396, I think fixing this issue would mean that we can support those use cases in (IIUC) an ABI-conforming way.

**Reproducer**
Example 1:
```
cat > mydata.blob << '!'
asdf
!
cat > empty.c << '!'
int main(int argc, char *argv[]) { return 0; }
!

clang --target=riscv32-unknown-elf -mabi=ilp32f -c empty.c
llvm-objcopy --input-target=binary --output-target=elf32-littleriscv mydata.blob mydata.elf
# Error: Incompatible object files when reading empty.o, mixing RVC|FloatABISingle and
ld.eld -melf32lriscv mydata.elf empty.o
```

Example 2 (also requires https://github.com/qualcomm/eld/issues/396 to be resolved so `e_flags` is 0):
```
cat > mydata.blob << '!'
asdf
!
cat > empty.c << '!'
int main(int argc, char *argv[]) { return 0; }
!

clang --target=riscv32-unknown-elf -mabi=ilp32 -c empty.c
ld.eld -melf32lriscv -r --format=binary mydata.blob -o mydata.elf
# Error: Incompatible object files when reading empty.o, mixing RVC and RVC|FloatABISingle|FloatABIDouble|FloatABIQuad|RVE
ld.eld -melf32lriscv mydata.elf empty.o
```

quic-seaswara:
The psabi has a note under
https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc# elf-object-files that "The static linker may ignore the compatibility checks if all fields in the e_flags are zero and all sections in the input file are non-executable sections."

Currently, eld produces errors for the reproducers below which I think are reasonable enough patterns and don't have great workarounds short of manually editing the binary blob elfs.

Resolves #407

### DIFF
--- a/lib/Target/RISCV/RISCVInfo.cpp
+++ b/lib/Target/RISCV/RISCVInfo.cpp
@@ -86,7 +86,13 @@ bool RISCVInfo::isCompatible(uint64_t pFlag, const std::string &pFile) const {
   return true;
 }
 
-bool RISCVInfo::checkFlags(uint64_t flag, const InputFile *pInputFile) const {
+bool RISCVInfo::checkFlags(uint64_t flag, const InputFile *pInputFile,
+                           bool hasExecutableSections) const {
+  // If flag is empty and no executable sections found in the ELF file
+  // skip checking for compatibility
+  if (!flag && !hasExecutableSections)
+    return true;
+
   // Choose the default architecture from the input files, only if mcpu option
   // is not specified on the command line.
   if ((m_CmdLineFlag == UNKNOWN) && (m_OutputFlag == UNKNOWN))
--- a/test/RISCV/standalone/EFlags/EFlags.test
+++ b/test/RISCV/standalone/EFlags/EFlags.test
@@ -1,0 +1,17 @@
+UNSUPPORTED: riscv64
+#----------EFlags.test--------------------- Executable------------------#
+#BEGIN_COMMENT
+# This checks that the linker will error out appropriately when mixing
+# object files with incompatible flags
+#END_COMMENT
+#START_TEST
+RUN: %yaml2obj %p/Inputs/emptyelf.yaml -o %t.emptyelf.o
+RUN: %clang %clangopts -mabi=ilp32f -c %p/Inputs/1.c -o %t.1.o
+RUN: %eld %t.emptyelf.o %t.1.o -o %t1.out
+RUN: %readelf -h %t1.out | %filecheck %s -check-prefix=EFLAGS
+RUN: %yaml2obj %p/Inputs/flagszero_with_exec_sections.yaml -o %t.f.o
+RUN: %not %eld %t.f.o %t.1.o -o %t1.out 2>&1 | %filecheck %s
+#END_TEST
+
+EFLAGS: Flags: 0x3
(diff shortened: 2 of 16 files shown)


test/RISCV/standalone/EFlags/EFlags.test
```diff
@@ -0,0 +1,17 @@
...
+#START_TEST
+RUN: %yaml2obj %p/Inputs/emptyelf.yaml -o %t.emptyelf.o
+RUN: %clang %clangopts -mabi=ilp32f -c %p/Inputs/1.c -o %t.1.o
+RUN: %eld %t.emptyelf.o %t.1.o -o %t1.out
+RUN: %readelf -h %t1.out | %filecheck %s -check-prefix=EFLAGS
+RUN: %yaml2obj %p/Inputs/flagszero_with_exec_sections.yaml -o %t.f.o
+RUN: %not %eld %t.f.o %t.1.o -o %t1.out 2>&1 | %filecheck %s
+#END_TEST
+
+EFLAGS: 0x0
```
jonathonpenix: Is this check right? I think `e_flags` should be 0x3 here since t1.out is a mix of the emptyelf.o (which should have `e_flags` 0x0) and 1.o (which I think should have `e_flags` 0x3 since RVC + ilp32f)

Guessing we're matching some other field here in the readelf output ("Entry point address" maybe?)

quic-seaswara: thanks for noticing this stale test, I revised the test and failed to upload the new one

jonathonpenix (approved):
LGTM, thanks @quic-seaswara!

lib/Target/RISCV/RISCVInfo.cpp
```diff
@@ -86,7 +86,13 @@ bool RISCVInfo::isCompatible(uint64_t pFlag, const std::string &pFile) const {
...
 }
 
-bool RISCVInfo::checkFlags(uint64_t flag, const InputFile *pInputFile) const {
+bool RISCVInfo::checkFlags(uint64_t flag, const InputFile *pInputFile,
+                           bool hasExecutableSections) const {
+  // If flag is empty and no executable sections found in the ELF file
+  // skip checking for compatibility
+  if (!flag && !hasExecutableSections)
+    return true;
+
```
quic-akaryaki, line 95: I first thought InputFlags should needs to be set for the file, but it looks like it is never used, maybe a good time to clean it up.

quic-seaswara: we will add the plumbing to print the input file flags soon in the map file :)

PR #423, opened 2025-09-25
[docs] document e_flags behavior

quic-areg:


quic-areg:
image

quic-seaswara (approved):

quic-seaswara (approved):
This is a good start we need to explore each case of linking with binary file support and specify where in the image layout the binary output resides in a section.

Please expand this in a follow up.

PR #461, opened 2025-10-09
Make hexagon plugin test artifacts a dependency of check-eld target

### LINKED ISSUE #460: Hexagon plugin unit tests should be a dependency of `check-eld` target
Hexagon plugin unit tests are not dependency of `check-eld` target as of now and thus they need to be built separately. We should make it a dependency of `check-eld` so that they do not need to be built separately.

parth-07:
This commit makes the hexagon plugin test artifacts a depedency of check-eld targets.

Resolves #460

### DIFF
--- a/test/Hexagon/Plugin/CMakeLists.txt
+++ b/test/Hexagon/Plugin/CMakeLists.txt
@@ -1,10 +1,3 @@
-add_custom_target(PluginUnitTests)
-set_target_properties(PluginUnitTests PROPERTIES FOLDER "plugin unit tests")
-
-function(add_plugin pluginname)
-  add_dependencies(PluginUnitTests ${pluginname})
-endfunction()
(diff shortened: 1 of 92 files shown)


test/Hexagon/Plugin/CMakeLists.txt
```diff
@@ -1,8 +1,8 @@
-add_custom_target(PluginUnitTests)
-set_target_properties(PluginUnitTests PROPERTIES FOLDER "plugin unit tests")
+add_custom_target(HexagonPluginUnitTests)
```
quic-seaswara: what about target PluginUnitTests ?

parth-07: I have updated the code to use PluginUnitTests everywhere.

PR #463, opened 2025-10-09
[NFC] Remove `Assignment::level::OUTPUT_SECTION`

### LINKED ISSUE #462: `Assignment::level::OUTPUT_SECTION` is never used
`Assignment::level::OUTPUT_SECTION` and related code that use and store output section assignments are redundant and never actually used. These codes should be removed to make the codebase simpler.

parth-07:
This commit removes Assignment::level::OUTPUT_SECTION and the related codes that were storing and using assignments of type OUTPUT_SECTION. The Assignment::level::OUTPUT_SECTION is never being used. All assignments of type OUTPUT_SECTION were getting converted to level::SECTIONS_END later, and the assignments that are inside an output section are of type Assignment::level::INPUT_SECTION.

This commit also modifies to
`plugin::Script::Assignment::isOutsideOutputSection()` to unconditionally return false to maintain the existing behavior.

Resolves #462

### DIFF
--- a/include/eld/Script/Assignment.h
+++ b/include/eld/Script/Assignment.h
@@ -36,7 +36,6 @@ class Assignment : public ScriptCommand {
 public:
   enum Level {
     OUTSIDE_SECTIONS, // outside SECTIONS command
-    OUTPUT_SECTION,   // related to an output section
     INPUT_SECTION,    // related to an input section
     SECTIONS_END
   };
@@ -93,10 +92,6 @@ public:
   bool isOutsideSections() const {
     return AssignmentLevel == OUTSIDE_SECTIONS ||
            AssignmentLevel == SECTIONS_END;
-  }
-
-  bool isOutsideOutputSection() const {
-    return AssignmentLevel == OUTPUT_SECTION;
   }
 
   bool isInsideOutputSection() const {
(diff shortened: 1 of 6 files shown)


lib/Script/SectionsCmd.cpp
```diff
@@ -87,10 +87,8 @@ eld::Expected<void> SectionsCmd::activate(Module &CurModule) {
     case ScriptCommand::OUTPUT_SECT_DESC: {
       if (!Assignments.empty()) {
         iterator Assign, AssignEnd = Assignments.end();
-        for (Assign = Assignments.begin(); Assign != AssignEnd; ++Assign) {
-          llvm::cast<Assignment>(*Assign)->setLevel(Assignment::SECTIONS_END);
+        for (Assign = Assignments.begin(); Assign != AssignEnd; ++Assign)
```
quic-seaswara: we can convert this to 

for (auto &Assign : Assignments) 

and remove the unnecessary iterators.

include/eld/Script/Assignment.h
```diff
@@ -36,7 +36,6 @@ class Assignment : public ScriptCommand {
 public:
   enum Level {
     OUTSIDE_SECTIONS, // outside SECTIONS command
-    OUTPUT_SECTION,   // related to an output section
```
quic-seaswara, line 39: Wont we need something like before sections command and after sections command ?

parth-07: We will need this. This PR is a prerequisite for the PR that will add BEFORE_SECTIONS and AFTER_SECTIONS levels.

lib/Script/SectionsCmd.cpp
parth-07: I have made this change.

PR #470, opened 2025-10-13
Add initial support for creating shared library with symbol versioning

### LINKED ISSUE #646: Add support for building shared libraries with symbol versioning information
Currently, we do not support building shared libraries with symbol versioning information.

parth-07:
This commit adds initial support for building shared libraries with
symbol versioning information. Symbol versioning allows to add version
information to symbols. It is used to provide multiple definitions of
the same symbol for backward compatibility.

This commit adds the support such that the functionality can be enabled
conditionally at build time by specifying ELD_ENABLE_SYMBOL_VERSIONING
CMake option. This commit also updates the PR checkin builder to enable
symbol versioning functionality.

Symbol versioning is a target-independent feature but requires support
from compiler, linker and the loader. From the ELF perspective, symbol
versioning is implemented using 3 sections:
.gnu.version(SHT_GNU_versym), .gnu.version_d(SHT_GNU_verdef), and
.gnu.version_r(SHT_GNU_verneed). This commit only provides support for
.gnu.version and .gnu.version_d sections because only they are required
when building a shared library. .gnu.version_r is required when using a
shared library. Briefly, these sections can be descrbed as:

- .gnu.version: An array of version IDs. i'th entry of this array stores
                the version ID of the i'th symbol in the dynamic symbol
                table.
- .gnu.version_d: Stores the version definitions of the symbols that
                  are defined by this module.
- .gnu.version_r: Stores the version definitions of the undefined symbols that
                  are used by this module.

To add the support for creating shared libraries with symbol versioning
information, this commit:
- Adds support for creating symbol versioning sections .gnu.version and
  .gnu.version_d
- Improve version script support to properly handle versioned
  nodes.
- Add logic to properly assign output version IDs to symbols.
- Update dynamic symbol table and .hash section to properly emit
  versioned symbols.
- Emit symbol versioning sections and correctly fill symbol versioning
  information in the .dynamic section.

Additionally, this commit adds the trace flag --trace=symbol-versioning.

Resolves #646


include/eld/SymbolResolver/ResolveInfo.h
```diff
@@ -329,6 +351,7 @@ class ResolveInfo {
   llvm::StringRef SymbolName;
   ResolveInfo *SymbolAlias;
   InputFile *SymbolResolvedOrigin;
+  std::optional<uint16_t> SymbolVersionID;
```
quic-seaswara: This will increase memory utilization for regular builds where no symbol have version information.

can we sub class ResolveInfo for versioned symbols, the penalty for regular symbols that do not have any versioning information will make.

We can have a bit field for versioned symbols and use the ResolveInfo sub class for interpreting the information.

include/eld/Target/GNULDBackend.h
```diff
@@ -581,6 +583,13 @@ class GNULDBackend {
 
   void addSymbolScope(ResolveInfo *R, VersionSymbol *V) { SymbolScopes[R] = V; }
 
+  VersionSymbol *getSymbolScope(ResolveInfo *R) const {
```
quic-seaswara: const VersionSymbol * ?

parth-07: Done.

include/eld/SymbolResolver/ResolveInfo.h
partaror: I have added a map `GNULDBackend::OutputVersionIDs` to store symbol version IDs as discussed offline.

include/eld/Config/LinkerConfig.h
```diff
@@ -420,6 +420,12 @@ class LinkerConfig {
 
   std::string getSymDefString() const;
 
+  /// Returns true if the link should build a dynamic object file
+  /// or a dynamic executable file.
+#ifdef ELD_ENABLE_SYMBOL_VERSIONING
```
quic-akaryaki, line 425: You can reduce the number of ifdefs and leave only those that actually change behavior.

cmake/modules/LITModules.cmake
```diff
@@ -59,6 +59,10 @@ function(generate_ext_lit ARCH OPTIONNAME LIT_CFG_PATH)
     "${CMAKE_CURRENT_BINARY_DIR}/llvm-lit-${ARCH_LOWER}-${OPTIONNAME}")
 endfunction()
 
+if(ELD_ENABLE_SYMBOL_VERSIONING)
```
quic-akaryaki, line 62: I think you can just test ELD_ENABLE_SYMBOL_VERSIONING in lit.cfg.

include/eld/Driver/GnuLinkerOptions.td
```diff
@@ -688,7 +688,8 @@ defm trace
           "\t\t\t --trace=trampolines : trace trampolines\n"
           "\t\t\t --trace=wrap-symbols : trace symbol wrap options\n"
           "\t\t\t --trace=symdef : trace symbol resolution from symdef files\n"
-          "\t\t\t --trace=dynamic-linking : trace dynamic linking">,
+          "\t\t\t --trace=dynamic-linking : trace dynamic linking"
```
quic-akaryaki: \n

lib/Object/ObjectLinker.cpp
```diff
@@ -362,25 +382,58 @@ bool ObjectLinker::parseVersionScript() {
     }
   }
   auto &SymbolScopes = getTargetBackend().symbolScopes();
-  for (auto &G : ThisModule->getNamePool().getGlobals()) {
+  auto &NP = ThisModule->getNamePool();
```
quic-akaryaki, line 385: const

lib/Object/ObjectLinker.cpp
```diff
@@ -362,25 +382,58 @@ bool ObjectLinker::parseVersionScript() {
...
+  for (auto &G : NP.getGlobals()) {
     ResolveInfo *R = G.getValue();
     for (auto &VersionScriptNode : ThisModule->getVersionScriptNodes()) {
       if (VersionScriptNode->getGlobalBlock()) {
         for (auto *Sym : VersionScriptNode->getGlobalBlock()->getSymbols()) {
-          if (Sym->getSymbolPattern()->matched(*R)) {
+#ifdef ELD_ENABLE_SYMBOL_VERSIONING
+          bool isMatched = Sym->matched(*R, NP);
+#else
+          bool isMatched = Sym->getSymbolPattern()->matched(*R);
```
quic-akaryaki, line 394: I thin SYm->matched() will internally call getSymbolPattern()->matched(*R) if there is no versioning.

lib/Object/ObjectLinker.cpp
```diff
@@ -362,25 +382,58 @@ bool ObjectLinker::parseVersionScript() {
...
         continue;
+#endif
+      }
       if (VersionScriptNode->getLocalBlock()) {
         for (auto *Sym : VersionScriptNode->getLocalBlock()->getSymbols()) {
-          if (Sym->getSymbolPattern()->matched(*R)) {
+#ifdef ELD_ENABLE_SYMBOL_VERSIONING
+          bool isMatched = Sym->matched(*R, NP);
+#else
+          bool isMatched = Sym->getSymbolPattern()->matched(*R);
```
quic-akaryaki, line 419: Same as above.

lib/Object/ObjectLinker.cpp
```diff
@@ -362,25 +382,58 @@ bool ObjectLinker::parseVersionScript() {
...
+              ThisConfig.raise(Diag::trace_version_script_matched_scope)
+                  << "global" << R->name();
+            break;
+#endif
           } // end Symbol Match
         } // end Symbols
       } // end Global
-      if (SymbolScopes.find(R) != SymbolScopes.end())
+      if (SymbolScopes.find(R) != SymbolScopes.end()) {
+#ifdef ELD_ENABLE_SYMBOL_VERSIONING
```
quic-akaryaki, line 408: Why is there this difference?

lib/Script/ScriptSymbol.cpp
```diff
@@ -38,7 +38,11 @@ void ScriptSymbol::addResolveInfoToContainer(const ResolveInfo *Info) const {
 }
 
 bool ScriptSymbol::matched(const ResolveInfo &Sym) const {
+#ifdef ELD_ENABLE_SYMBOL_VERSIONING
+  llvm::StringRef Name = Sym.getNonVersionedName();
```
quic-akaryaki, line 42: getNonVersionedName() can be defined as name() when versioning is disabled.

lib/Target/GNULDBackend.cpp
```diff
@@ -929,7 +1001,8 @@ void GNULDBackend::sizeSymTab() {
         m_SymbolsToRemove.count(Sym)) {
       continue;
     }
-    strtab += (std::string(Sym->name()).size() + 1);
```
quic-akaryaki, line 932: `strtab += Sym->getName() + 1; ` will not work? There is even ResolveInfo::nameSize().

lib/Target/GNULDBackend.cpp
```diff
@@ -1016,11 +1093,14 @@ void GNULDBackend::emitSymbol64(llvm::ELF::Elf64_Sym &pSym, LDSymbol *pSymbol,
     pSym.st_name = 0;
   else {
     if (IsDynSymTab) {
-      auto optSymNameOffset =
-          getOutputFormat()->getOffsetInDynStrTab(std::string(pSymbol->name()));
-      ASSERT(optSymNameOffset.has_value(), "Symbol name (" +
-                                               std::string(pSymbol->name()) +
-                                               ") must be present in .dynstr!");
+#ifdef ELD_ENABLE_SYMBOL_VERSIONING
```
quic-akaryaki, line 1094: This is fifth or so fragment that can be simplified by supporting getNonVersionedName() when there is no versioning.

lib/Target/GNULDBackend.cpp
```diff
@@ -1114,14 +1196,19 @@ GNULDBackend::emitRegNamePools(llvm::FileOutputBuffer &pOutput) {
       continue;
     }
     if (config().targets().is32Bits())
-      emitSymbol32(symtab32[symIdx], S->outSymbol(), strtab, strtabsize,
```
quic-akaryaki, line 1117: Reformatting increases the number of unrelated changes.

lib/Target/GNULDBackend.cpp
```diff
@@ -5163,3 +5255,99 @@ bool GNULDBackend::setupTLS() {
...
+  const DiagnosticPrinter *DP = config().getPrinter();
+  if (DP->traceSymbolVersioning())
+    config().raise(Diag::trace_creating_symbol_versioning_section)
+        << ".gnu.version";
+  GNUVerSymSection = m_Module.createInternalSection(
+      Module::InternalInputType::SymbolVersioning,
+      LDFileFormat::Kind::SymbolVersion, ".gnu.version",
+      llvm::ELF::SHT_GNU_versym, llvm::ELF::SHF_ALLOC,
+      /*Align=*/sizeof(uint16_t),
+      /*EntrySize=*/2);
```
quic-akaryaki, line 5264: I wonder why Align is sizeof(u16) but EntrySize is just 2?

lib/Target/GNULDBackend.cpp
```diff
@@ -5163,3 +5255,99 @@ bool GNULDBackend::setupTLS() {
...
+      VernID = it->second;
+    } else {
+      VersionScriptNode *VSN = VernSymbol->getBlock()->getNode();
+      llvm::StringRef VerName = (VSN->isAnonymous() ? "" : VSN->getName());
+      auto it = VerNameToID.find(VerName.str());
+      ASSERT(it != VerNameToID.end(), "Map must contain the entry!");
+      VernID = it->second;
+    }
+
+    if (Pos == std::string::npos) {
```
quic-akaryaki: gratuitous braces.

quic-akaryaki: I may be missing something but Pos is already FullName.find('@'). The whole thing is just `isDefaultVersionSymbol = Pos != std::string::npos' ? You can define isDefaultVersionSymbol right here or remove it completely.

.github/workflows/ci.yml
```diff
@@ -89,6 +89,7 @@ jobs:
             -DLLVM_TARGETS_TO_BUILD="ARM;AArch64;RISCV;Hexagon;X86" \
             -DELD_TARGETS_TO_BUILD="ARM;AArch64;RISCV;Hexagon;x86_64" \
             -DCMAKE_BUILD_WITH_INSTALL_RPATH=ON \
+            -DELD_ENABLE_SYMBOL_VERSIONING=On \
```
quic-areg: ON

include/eld/Diagnostics/DiagnosticPrinter.h
```diff
@@ -47,6 +47,9 @@ class DiagnosticPrinter {
     TraceMergeStrings = 0x8000,
     TraceLinkerScript = 0x10000,
     TraceSymDef = 0x100000,
+#ifdef ELD_ENABLE_SYMBOL_VERSIONING
```
quic-areg, line 50: I think it is fine if there are no ifdefs here (and in other places)

include/eld/Diagnostics/DiagnosticPrinter.h
```diff
@@ -98,6 +101,13 @@ class DiagnosticPrinter {
 
   bool traceSymDef() { return (Trace & TraceSymDef); }
 
+  // Available when symbol versioning is enabled.
```
quic-areg: comment is not needed imo

lib/Target/GNULDBackend.cpp
```diff
@@ -743,6 +757,22 @@ bool GNULDBackend::SetSymbolsToBeExported() {
       continue;
     R->setExportToDyn();
   }
+
+#ifdef ELD_ENABLE_SYMBOL_VERSIONING
+  for (auto &R : m_Module.getSymbols()) {
+    auto found = SymbolScopes.find(R);
```
quic-areg: is this variable used?

include/eld/Diagnostics/DiagCommonKinds.inc
```diff
@@ -83,7 +83,8 @@ DIAG(mismatched_group, DiagnosticEngine::Fatal,
 DIAG(unsupported_version_node, DiagnosticEngine::Warning,
      "Only anonymous version nodes are supported. Ignoring version script `%0'")
 DIAG(unsupported_dependent_node, DiagnosticEngine::Warning,
-     "Dependent nodes are not supported. Ignoring version script `%0'")
+     "Dependent nodes are not suppored. Ignoring dependency for node '%0' "
```
quic-areg: supported

lib/Target/GNULDBackend.cpp
```diff
@@ -5163,3 +5255,99 @@ bool GNULDBackend::setupTLS() {
...
+      LDFileFormat::Kind::SymbolVersion, ".gnu.version",
+      llvm::ELF::SHT_GNU_versym, llvm::ELF::SHF_ALLOC,
+      /*Align=*/sizeof(uint16_t),
+      /*EntrySize=*/2);
+
+  // Add .gnu.version_d section creation
+  if (DP->traceSymbolVersioning())
+    config().raise(Diag::trace_creating_symbol_versioning_section)
+        << ".gnu.version_d";
+  GNUVerDefSection = m_Module.createInternalSection(
```
quic-areg, line 5270: shouldn't the entsize be nonzero?

lib/Target/GNULDBackend.cpp
```diff
@@ -5163,3 +5255,99 @@ bool GNULDBackend::setupTLS() {
...
+  for (const auto &VSNode : VSNodes) {
+    ASSERT(VSNode, "VSNode must not be null!");
+    if (VSNode->isAnonymous())
+      continue;
+    VerNameToID[VSNode->getName().str()] = NextVerID++;
+  }
+  uint16_t defaultVersionID = llvm::ELF::VER_NDX_GLOBAL;
+
+  // First, assign default version to any symbol without an explicit one.
+  for (std::size_t i = 0, e = DynamicSymbols.size(); i < e; ++i) {
```
quic-areg: index 0 should be VER_NDX_LOCAL:

eld
```
  000:   1 (*global*)      2 (V1)         
```

lld 
```
  000:   0 (*local*)       2 (V1)         
```

test/Common/standalone/VersionScript/VersionScript.test
```diff
@@ -166,12 +166,10 @@ MAPCHECK8: Local:
 MAPCHECK8: Pattern: *
 
 #Test that scripts with version nodes are warned as unsupported and ignored
```
quic-areg, line 168: update the comment?

lib/Fragment/GNUVerDefFragment.cpp
```diff
@@ -0,0 +1,116 @@
...
+  VerDefEntrySize = sizeof(typename ELFT::Verdef);
+  VerdAuxEntrySize = sizeof(typename ELFT::Verdaux);
+
+  const GeneralOptions &options = M.getConfig().options();
+  std::string baseVersion = "";
+  std::string SOName = options.soname();
+  if (!SOName.empty())
+    baseVersion = SOName;
+  else
+    baseVersion = options.outputFileName();
```
quic-areg: should this be the base name instead of potentially an absolute path?

include/eld/Config/LinkerConfig.h
parth-07: It was more troublesome to check which all ifdefs are essential and which are not. If extra ifdefs are harmless and ensure the previous behavior, then I would prefer to use ifdef at all places which are only required for symbol versioning. Hopefully, we will soon be able to remove all these ifdefs.

cmake/modules/LITModules.cmake
parth-07: The value of `ELD_ENABLE_SYMBOL_VERSIONING` depends upon the value passed by the user in the CMake configure command and thus it can be any of [`on`, `On`, `ON`, `True`, `true`, `1`, ...]. I have added an additional variable to make the check in lit python code simpler.

include/eld/Driver/GnuLinkerOptions.td
parth-07: Thank you for catching this. I have fixed it.

lib/Object/ObjectLinker.cpp
parth-07: Yes, you are right. However, I cannot directly use `Sym->matched(*R, NP)` when symbol versioning support is disabled because this function uses functions that are only available when symbol versioning support is enabled (for example, `ResolveInfo::getVersionName`)

lib/Object/ObjectLinker.cpp
parth-07: We are using `NP.getGlobals` at line 386, and `NamePool::getGlobals` is a non-const function.

lib/Script/ScriptSymbol.cpp
parth-07: I prefer to only have `ResolveInfo::getNonVersionedName` function accessible when the symbol versioning support is enabled. I feel that it can be error-prone to have and use symbol versioning-related functions when the support is explicitly disabled. Please let me know your thoughts on this.

lib/Object/ObjectLinker.cpp
parth-07: We need `break` here because we want the first version node that matches a symbol to be selected. If we `continue` here, then the last version node that matches a symbol is selected. It did not matter uptil now because when anonymous nodes are used, then we typically only have one version node.

lib/Object/ObjectLinker.cpp
parth-07: Same reasoning as above :).

Yes, you are right. However, I cannot directly use Sym->matched(*R, NP) when symbol versioning support is disabled because this function uses functions that are only available when symbol versioning support is enabled (for example, ResolveInfo::getVersionName)

lib/Target/GNULDBackend.cpp
parth-07: No, I have removed it now.

lib/Target/GNULDBackend.cpp
parth-07: `Sym->getName()` returns `const char*`, that's why `std::string` is used here.

lib/Target/GNULDBackend.cpp
parth-07: I expect these ifdefs to be temporary / short-lived. I do not want to optimize minimizing the number of ifdefs (and by mistake causing some unexpected behavior change) when I expect the ifdefs to be removed in the near future.

lib/Target/GNULDBackend.cpp
parth-07: `sizeof(uint16_t)` is also `2`. I agree that it would be simpler to just use 2 at both the places. I have made this change.

lib/Target/GNULDBackend.cpp
parth-07: Entry size of `.gnu.version_d` and `.gnu.version_r` sections is 0.

lib/Target/GNULDBackend.cpp
parth-07: It is a little more complicated. It would be:

```cpp
isDefaultVersionSymbol = (Pos == std::string::npos || (FullName.find("@@") != std::string::npos));
```

I have replaced the if-condition with this shorter variant.

lib/Target/GNULDBackend.cpp
parth-07: Thank you so much for catching this. I have fixed this and updated the test to check for this.

lib/Target/GNULDBackend.cpp
parth-07: I used the clang-format formatting. It changed more things than expected. I would be careful to use `git clang-format` the next time.

.github/workflows/ci.yml
parth-07: I have changed it to ON for consistency with the other options.

include/eld/Diagnostics/DiagnosticPrinter.h
parth-07: It was more troublesome to check where all ifdefs were truly necessary. If it is not a big issue, then I would like to keep ifdefs everywhere as of now. Hopefully, we will be able to remove all the ifdefs soon.

include/eld/Diagnostics/DiagnosticPrinter.h
parth-07: I have removed the comment.

include/eld/Diagnostics/DiagCommonKinds.inc
parth-07: Fixed.

include/eld/Diagnostics/DiagnosticPrinter.h
quic-areg: Why do we need them at all? If we need to gate this, can we gate this behind a hidden option with support off by default rather than macros?

lib/Fragment/GNUVerDefFragment.cpp
parth-07: Thank you for catching this. Yes, it should be base name instead of potentially an absolute path. I have fixed this and updated tests to check for this.

test/Common/standalone/VersionScript/VersionScript.test
parth-07: I have updated the test to its previous state and disabled the test for symbol-versioning lit feature.

include/eld/Diagnostics/DiagnosticPrinter.h
parth-07: Shankar decided to have a mechanism to enable / disable symbol versioning functionality at build time. This reduces the overhead of testing and accidentally changing a behavior which users may be relying on. We should be able to remove the ifdefs after adequate testing has been done.

lib/Object/ObjectLinker.cpp
quic-akaryaki: Wow

lib/Object/ObjectLinker.cpp
quic-akaryaki: Basically, you are ending up with a lot more code.

lib/Target/GNULDBackend.cpp
quic-akaryaki: So double work.

quic-seaswara (approved):

PR #476, opened 2025-10-14
Divide OUTSIDE_SECTIONS into BEFORE_SECTIONS and AFTER_SECTIONS

### LINKED ISSUE #473: `Assignment::Level::OUTSIDE_SECTIONS` must be divided into `BEFORE_SECTIONS` and `AFTER_SECTIONS`
`Assignment::Level::OUTSIDE_SECTIONS` must be divided into `Assignment::Level::BEFORE_SECTIONS` and `Assignment::Level::AFTER_SECTIONS`. This is because the order of assignment evaluation should be: Assignment before `SECTIONS` -> Assignment within `SECTIONS` -> Assignment after `SECTIONS`. Thus, we need to mark the assignment levels to correctly evaluate the assignments.

This issue only focuses on correctly marking the assignment level. The actual evaluation change should be handled separately.

parth-07:
This commit divides Assignment::Level::OUTSIDE_SECTIONS into BEFORE_SECTIONS and AFTER_SECTIONS assignment level types. It does not change the evaluation order of the assignment. The actual evaluation order will be fixed in a separate patch.

Resolves #473

### DIFF
--- a/include/eld/Script/Assignment.h
+++ b/include/eld/Script/Assignment.h
@@ -35,7 +35,8 @@ class ELFSection;
 class Assignment : public ScriptCommand {
 public:
   enum Level {
-    OUTSIDE_SECTIONS, // outside SECTIONS command
+    BEFORE_SECTIONS,  // Assignments before SECTIONS command
+    AFTER_SECTIONS,   // Assignments after SECTIONS command
     INPUT_SECTION,    // related to an input section
     SECTIONS_END
   };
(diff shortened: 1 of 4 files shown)


include/eld/Script/Assignment.h
```diff
@@ -35,7 +35,8 @@ class ELFSection;
 class Assignment : public ScriptCommand {
 public:
   enum Level {
-    OUTSIDE_SECTIONS, // outside SECTIONS command
+    BEFORE_SECTIONS,  // Assignments before SECTIONS command
+    AFTER_SECTIONS,   // Assignments after SECTIONS command
```
quic-seaswara, line 39: what category does defsym/symdef fall under ?

parth-07: It depends upon where the defsym is placed in the link command-line. If it is evaluated before the linker script with `SECTIONS` command, then it will fall under `BEFORE_SECTIONS` and if it is placed after, then it will be `AFTER_SECTIONS`. However, there is a catch, currently we evaluate linker scripts passed using `-T <script>` earlier than other input nodes such as `<script>` -- this is a bug which needs to be independently fixed.

PR #477, opened 2025-10-14
Re-evaluate OUTSIDE_SECTIONS assignments whenever layout resets

### LINKED ISSUE #468: Incorrect linker script symbol value due to layout reiteration
In some cases, layout reiteration that happens in eld causes incorrect linker script symbol value. This happens because layout reiteration causes reevaluation of some assignments with newer symbol values than what they should use.

Reproducer:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
int foo() { return 1; }

int val = 3;
EOF


cat > script.t << '!'
u = 0x100;
SECTIONS {
  v = u;
  .text : { *(.text*) }
  u = 0x300;
  .data : { *(.data*) }
}
!


LDs=(ld.eld)
SFs=(eld)

clang -o 1.o --target=riscv64-unknown-elf 1.c -c -ffunction-sections

for i in "${!SFs[@]}"; do
done

llvm-readelf -s 1.eld.out | grep 'ABS v'
EOF

**Output**
```
 13: 0000000000000300     0 NOTYPE  GLOBAL DEFAULT   ABS v
```

The expected value of `v` is 0x100.

parth-07:
This commit updates the layout step to re-evaluate OUTSIDE_SECTIONS assignments whenever the layout is reset during layout iteration. This is important to ensure correctness of re-evaluation of SECTIONS assignments.

Resolves #468

### DIFF
--- a/lib/Target/CreateProgramHeaders.hpp
+++ b/lib/Target/CreateProgramHeaders.hpp
@@ -168,6 +168,12 @@ bool GNULDBackend::createProgramHdrs() {
     noLoadSections.clear();
     resetNewSectionsAddedToLayout();
     enable_RELRO = true;
+
+    {
+      eld::RegisterTimer T("Evaluate Script Assignments", "Establish Layout",
+                           m_Module.getConfig().options().printTimingStats());
+      evaluateScriptAssignments(/*evaluateAsserts=*/false);
+    }
   };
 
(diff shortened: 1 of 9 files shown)


quic-seaswara (approved):
Please raise an issue for partial links and linker script assignments

PR #481, opened 2025-10-15
[RISCV][Xqci] Remove c.lui absolute relaxations

### LINKED ISSUE #455: Problem with Xqci Relaxations
Broadly, the following relaxations have not been implemented correctly, and are invalid:

- `qc.e.li` -> `c.lui`
- `qc.li` -> `c.lui`

Both for the same reason:

https://github.com/qualcomm/eld/blob/fde4266b25b7a08ff51c95352b9ee9547955be8a/lib/Target/RISCV/RISCVLDBackend.cpp#L600-L604

This check does not account for the future value of `S+A`, only its current value.

Further relaxations may affect the value of `S+A`, causing the low 12-bits to no longer be all zeros, even though this was the case when the relaxation was applied.

For the same reason, #416 (`qc.e.li` -> `lui`) is not correct.

There are two options going forwards:
1. Remove both relaxations
2. Implement a lot of checks to verify that required alignments mean the low 12-bits will not change after relaxation. 

For the 21.x branch (which has this bug), we need to do the first option.

For 22.x, we have more options, but it's not clear that these relaxation would be worth the maintenance burden without these checks.

The workaround before this fix lands is to use `--no-relax-c` with `--relax-xqci`.

lenary:
These are not valid, as we do not verify that the page-alignment that we checked for during relaxation will be preserved in the final object. Without these checks, we may end up creating a `c.lui` which cannot materialize the intended address, because the address is no longer aligned due to relaxation.

Fixes #455

(cherry picked from commit 75a9d8314467c841e74f28b5c2545f7aa41109cb)

### DIFF
--- a/lib/Target/RISCV/RISCVLDBackend.cpp
+++ b/lib/Target/RISCV/RISCVLDBackend.cpp
@@ -545,10 +545,8 @@ bool RISCVLDBackend::doRelaxationLui(Relocation *reloc, Relocator::DWord G) {
   return true;
 }
 
-bool RISCVLDBackend::doRelaxationQCLi(Relocation *reloc, Relocator::DWord G) {
+bool RISCVLDBackend::doRelaxationQCELi(Relocation *reloc, Relocator::DWord G) {
   /* Three similar relaxations can be applied here, in order of preference:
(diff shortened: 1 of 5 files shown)


lenary:
This is the 21.x version of #458

PR #501, opened 2025-10-23
[RISCV] Fix JAL overflow check

### LINKED ISSUE #492: Detect relocation overflow for JAL relocations
eld does not overflow on jal relocations.

```
cat > 1.s << \!
.extern bar
.section .text.foo,"ax",@progbits
foo:
jal bar
!

cat > 2.c << \!
__attribute((weak)) int bar() { return 0; }
!

cat > script.t << \!
SECTIONS {
  .foo (0) : { *(.text.foo) }
  .bar (0xF0000000) : { *(.text.bar) }
}
!

clang -c --target=riscv32 1.s 2.c -mno-relax -ffunction-sections -O2
ld.eld 1.o 2.o -T script.t
```

ld.lld: error: 1.o:(.text.foo+0x0): relocation R_RISCV_JAL out of range: 4026531840 is not in [-1048576, 1048575]; references 'bar'

quic-areg:
`R_RISCV_JAL` used to any 32-bit displacement:

We should check isInt<20>(Value >> 1) but we checked isInt<20>(Value >> 20).

Closes #492


quic-seaswara (changes requested):
Why do we need a shift by 1 ? 

https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc

<html>
<body>

17 | JAL | Static | J-Type | 20-bit PC-relative jump offset
-- | -- | -- | -- | --
S + A - P


</body>
</html>

lenary (approved):
LGTM

PR #504, opened 2025-10-29
Cherry-pick commits for eh_frame symbols

quic-k:


.github/workflows/ci.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       uses: actions/checkout@v4
       with:
         repository: llvm/llvm-project
-        ref: ${{ env.BASE_BRANCH_NAME }}
+        ref: main
```
parth-07: Why should we always fetch llvm from main branch? If the builder is building a patch on release/21.x branch, then shouldn't we use 21 branch of LLVM as well?

quic-k: hm right, but as there's no hexagon.picolibc branch in llvm, I will have to hardcode the base branch to release/21.x

PR #508, opened 2025-10-30
Trigger CI on push and manual dispatch

quic-k:
Enable CI on push to main and release branches
Add manual trigger support

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,11 @@
 name: CI
 
 on:
+  push: # trigger CI on push to main and release branches
+    branches:
+      - main
+      - release/21.x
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,6 +1,11 @@
 name: CI
 
 on:
+  push: # trigger CI on push to main and release branches
+    branches:
+      - main
+      - release/21.x
```
quic-seaswara, line 7: do you need to explictly list each branch ?

quic-k: yes, else it will trigger for all branches by default
we can probably add regex for release branches when we have multiple release branches

PR #510, opened 2025-10-30
Fix evaluation of non-dot assignments containing forward references

### LINKED ISSUE #505: Incorrect symbol value when using forward reference in linker script assignment.
We do not properly evaluate forward references in linker scripts. This can cause incorrect linker script symbol values.

Reproducer:

```bash
cat > 1.c << '!'
int foo() { return 1; }
!

cat > script.t << '!'
SECTIONS {
  u = v1;
  .text : { *(.text*) }
  v1 = v2;
  v2 = 0x100;
}
!


LDs=(ld.eld)
SFs=(eld)

clang -o 1.o --target=aarch64-unknown-elf 1.c -c -ffunction-sections

for i in "${!SFs[@]}"; do
done

llvm-readelf -s 1.eld.out | grep -i 'ABS u'
```

Output:
```
6: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT   ABS u
```

The value of `u` should be `0x100`.

parth-07:
This commit fixes evaluation of non-dot assignments containing forward references. At a high-level, eld evaluates assignments in sequence and consequently the expressions containing forward references needs to be reevaluated when the forward reference value is computed for the correct computation results. For example:

```
u = v1;    // Assignment 1
v1 = v2;   // Assignment 2
v2 = 0xa;  // Assignment 3
```

eld evaluates the assignments in order, that is, the evaluation happens as: [Assignment 1, Assignment 2, Assignment 3]. If we follow this evaluation order, the symbols `u` and `v1` will have incorrect values because the value of `v2` is unknown when the assignments of `u` and `v1` are evaluated.

This commit fixes evaluation of non-dot assignments containing forward references by making the below two changes:

1) Assignments which cannot be completely evaluated during the
   sequential evaluation of expressions are marked as pending
   assignments. After the layout is done, but before the relaxation,
   these pending assignments are recursively reevaluated until there
   is no improvement in pending assignments or a MaxIteration limit is
   reached.

2) During a layout iteration, assignments may get evaluated multiple times
   as layout needs to reset itself based on a few conditions
   (new segment needs to be created, and so on...). It is important to reset
   the symbol if a layout gets reset. Let's
   see why it is important with the help of an example:

```
   SECTIONS {
     FOO : {
       u = v; // Assignment 1
       v = 0xa; // Assignment 2
       *(.text.foo)
     }
     BAR : {
       v = 0xb; // Assignment 3
     }
     v = 0xc; // Assignment 4
  }
```

The sequential assignment evaluation order is: [A1, A2, A3, A4]. When A1 is evaluated, `v` is not defined, hence we mark A1 as pending assignment. However, if the layout gets reset after evaluating A2, then A1 will be evaluated again, but this time, `v` is defined (from assignment 2 evaluation) and thus A1 can be completely evaluated. This is wrong because A1 should get the value from the assignment 4 instead of assignment 2!

We fix this issue by resetting the symbol values
whenever a layout is reset.

The same issue happens when the layout needs to be recomputed after a relaxation pass. And the same solution of resetting the the symbol values works for this case as well.

This commit also adds a new trace category 'pending-assignments' for tracing pending assignment evaluation.

Resolves #505

### DIFF
--- a/include/eld/Diagnostics/DiagVerbose.inc
+++ b/include/eld/Diagnostics/DiagVerbose.inc
@@ -167,3 +167,7 @@ DIAG(verbose_loaded_plugin_config, DiagnosticEngine::Verbose,
      "Initialized plugin config for %0")
 DIAG(verbose_infer_target, DiagnosticEngine::Verbose,
      "Inferred target : %0")
+DIAG(verbose_performing_layout_iteration, DiagnosticEngine::Verbose,
+     "Performing layout iteration %0")
+DIAG(verbose_eval_pending_assignments, DiagnosticEngine::Verbose,
+     "Evaluating pending assignments")
--- a/include/eld/Target/GNULDBackend.h
+++ b/include/eld/Target/GNULDBackend.h
@@ -824,6 +824,28 @@ public:
 
   const ResolveInfo *findAbsolutePLT(ResolveInfo *I) const;
 
+  /// Returns true if the symbol is partially evaluated. That
+  /// is, the symbol assignment expression was partially evaluated.
+  bool isPartiallyEvaluated(const ResolveInfo *RI) const {
+    return PartiallyEvaluatedSymbols.count(RI);
+  }
+
+  void addPartiallyEvaluatedSymbol(const ResolveInfo *RI) {
+    PartiallyEvaluatedSymbols.insert(RI);
+  }
+
+  void removePartiallyEvaluatedSymbol(const ResolveInfo *RI) {
+    PartiallyEvaluatedSymbols.erase(RI);
+  }
+
+  void resetPartiallyEvalAssignmentsAndSymbols() {
+    PartiallyEvaluatedAssignments.clear();
+    PartiallyEvaluatedSymbols.clear();
+  }
+
+  /// Recursively evaluates the partially evaluated assignments.
+  void evaluatePendingAssignments();
+
 protected:
   virtual int numReservedSegments() const { return m_NumReservedSegments; }
 
@@ -1002,6 +1024,14 @@ private:
   // Setup TLS alignment and check for any layout issues
   bool setupTLS();
 
+  /// Evaluates the assignment and tracks the partially evaluated assignments.
+  /// Partially evaluated assignments can be requested to be (recursively)
+  /// reevaluated using evaluatePendingAssignments.
+  void evaluateAssignmentAndTrackPartiallyEvalAssignments(Assignment &A,
+                                                          const ELFSection *S);
+
+  void resetScriptSymbols();
+
 protected:
   Module &m_Module;
 
@@ -1132,6 +1162,12 @@ protected:
   bool m_NeedEhdr = false;
 
   bool m_NeedPhdr = false;
+
+  std::unordered_set<const ResolveInfo *> PartiallyEvaluatedSymbols;
+
+  /// Stores the assignments which needs to be reevaluated later.
+  std::vector<std::pair<Assignment *, const ELFSection *>>
+      PartiallyEvaluatedAssignments;
 };
 
 } // namespace eld
--- a/lib/SymbolResolver/IRBuilder.cpp
+++ b/lib/SymbolResolver/IRBuilder.cpp
@@ -547,6 +547,9 @@ LDSymbol *IRBuilder::addSymbol<IRBuilder::Force, IRBuilder::Unresolve>(
                                     Desc, /*isBitcode=*/false});
   }
 
+  if (Input && Input->isLinkerScript())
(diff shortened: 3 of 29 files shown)


include/eld/Diagnostics/DiagVerbose.inc
```diff
@@ -167,3 +167,7 @@ DIAG(verbose_loaded_plugin_config, DiagnosticEngine::Verbose,
      "Initialized plugin config for %0")
 DIAG(verbose_infer_target, DiagnosticEngine::Verbose,
      "Inferred target : %0")
+DIAG(verbose_performing_layout_iteration, DiagnosticEngine::Verbose,
+     "Performing layout iteration %0")
+DIAG(verbose_eval_pending_assignments, DiagnosticEngine::Verbose,
+     "Evaluating pending assignments")
```
quic-seaswara: add a %0 with the assignment expression

include/eld/Diagnostics/DiagVerbose.inc
```diff
@@ -167,3 +167,7 @@ DIAG(verbose_loaded_plugin_config, DiagnosticEngine::Verbose,
      "Initialized plugin config for %0")
 DIAG(verbose_infer_target, DiagnosticEngine::Verbose,
      "Inferred target : %0")
+DIAG(verbose_performing_layout_iteration, DiagnosticEngine::Verbose,
+     "Performing layout iteration %0")
```
quic-seaswara, line 171: Can we --trace=layout and add the layout diagnostics in tracing layout iterations ?

include/eld/Target/GNULDBackend.h
```diff
@@ -1132,6 +1162,12 @@ class GNULDBackend {
   bool m_NeedEhdr = false;
 
   bool m_NeedPhdr = false;
+
+  std::unordered_set<const ResolveInfo *> PartiallyEvaluatedSymbols;
```
quic-seaswara, line 1166: shouldn't this be PartiallyEvaluatedAssignments instead ? because everything is an assignment

lib/SymbolResolver/IRBuilder.cpp
```diff
@@ -627,6 +630,10 @@ LDSymbol *IRBuilder::addSymbol<IRBuilder::Force, IRBuilder::Resolve>(
     OutputSym->setValue(Value, false);
   }
 
+  // If symbol originates from a linker script, record it in NamePool.
+  if (Input && Input->isLinkerScript())
+    ThisModule.getNamePool().addScriptSymbol(Result.Info);
```
quic-seaswara, line 635: what about defsym ?

lib/SymbolResolver/IRBuilder.cpp
```diff
@@ -547,6 +547,9 @@ LDSymbol *IRBuilder::addSymbol<IRBuilder::Force, IRBuilder::Unresolve>(
                                     Desc, /*isBitcode=*/false});
   }
 
+  if (Input && Input->isLinkerScript())
+    ThisModule.getNamePool().addScriptSymbol(OutputSym->resolveInfo());
```
quic-seaswara, line 551: why in two places ?

include/eld/Target/GNULDBackend.h
parth-07: We need both `PartiallyEvaluatedAssignments` and `PartiallyEvaluatedSymbols`. With just storing assignments, we cannot determine if a symbol was partially evaluated or not, as we do not maintain a mapping of symbol nodes to assignments.

lib/SymbolResolver/IRBuilder.cpp
parth-07: One is used by non-PROVIDE linker script assignments, and the other is used by PROVIDE linker script assignments.

lib/SymbolResolver/IRBuilder.cpp
parth-07: defsym is handled in the exact same way as other linker script assignments.

include/eld/Diagnostics/DiagVerbose.inc
parth-07: We cannot add assignment expression here. This diagnostic marks the beginning of pending assignment evaluations. We do not report this diagnostic per assignment.

include/eld/Diagnostics/DiagVerbose.inc
parth-07: Yes, this sounds good to me. I think a lot of things can be covered under --trace=layout. Can we handle this as a separate task altogether that analysis what all diagnostics should fall under `--trace=layout`?

PR #515, opened 2025-11-03
x86_64: Implement GOTPLTN initialization

### LINKED ISSUE #568: x86_64: Add support for GOTPLTN entry initialization
When a PLT stub entry is created (referred as PLTN), a corresponding GOTPLT entry (GOTPLTN) needs to be created as well.
This GOTPLTN entry is required for the dynamic linking trampoline mechanism -

1. **Initial State**: GOTPLTN contains address of PLT entry + 6 (the `pushq` instruction)
2. **First Call**: Code jumps through GOTPLTN → lands at PLT entry + 6
3. **Resolution**: PLT entry pushes relocation index and jumps to PLT0
4. **Dynamic Linker**: Resolves symbol and updates GOTPLTN to point to actual function
5. **Subsequent Calls**: Jump through GOTPLTN directly to resolved function

bhoumik03:
Implement the GOTPLTN initialization required for the dynamic linking trampoline mechanism:

1. **Initial State**: GOTPLTN contains address of PLT entry + 6 (the `pushq` instruction)
2. **First Call**: Code jumps through GOTPLTN → lands at PLT entry + 6
3. **Resolution**: PLT entry pushes relocation index and jumps to PLT0
4. **Dynamic Linker**: Resolves symbol and updates GOTPLTN to point to actual function
5. **Subsequent Calls**: Jump through GOTPLTN directly to resolved function

Implementation Details:
- getContent() in x86_64GOTPLTN computes initial values so the GOTPLTN entries point to the pushq instruction of the corressponding PLTN entry
- setPLTEntry() in called in x86_64PLTN to link the GOTPLTN to the PLTN so it can compute the initial value

 Progress on #534 
 Fixes #568

### DIFF
--- a/lib/Target/x86_64/x86_64GOT.h
+++ b/lib/Target/x86_64/x86_64GOT.h
@@ -50,8 +50,7 @@ public:
       Value =
           symInfo()->outSymbol()->value() - GNULDBackend::getTLSTemplateSize();
 
-    return llvm::ArrayRef(reinterpret_cast<const uint8_t *>(&Value),
-                          sizeof(Value));
+    return llvm::ArrayRef(reinterpret_cast<uint8_t *>(&Value), sizeof(Value));
   }
 
   static x86_64GOT *Create(ELFSection *O, ResolveInfo *R) {
@@ -98,18 +97,43 @@ private:
   mutable uint8_t Value[24];
 };
 
+/** \class x86_64GOTPLTN
+ *  \brief GOT entry for PLT function.
+ *
+ *  Each function with a PLT entry has a corresponding GOTPLTN entry.
+ *  Initially points to PLT entry + 6 (the pushq instruction).
+ *  Updated by dynamic linker on first call to point to actual function for the
+ * case of lazy binding.
+ */
 class x86_64GOTPLTN : public x86_64GOT {
 public:
   x86_64GOTPLTN(ELFSection *O, ResolveInfo *R)
-      : x86_64GOT(GOT::GOTPLTN, O, R, /*Align=*/8, /*Size=*/8) {}
+      : x86_64GOT(GOT::GOTPLTN, O, R, /*Align=*/8, /*Size=*/8),
+        m_PLTEntry(nullptr) {}
 
   x86_64GOT *getFirst() override { return this; }
 
   x86_64GOT *getNext() override { return nullptr; }
 
-  static x86_64GOTPLTN *Create(ELFSection *O, ResolveInfo *R) {
-    return make<x86_64GOTPLTN>(O, R);
+  // Link to corresponding PLT entry for address calculation
+  void setPLTEntry(Fragment *plt) { m_PLTEntry = plt; }
+
+  virtual llvm::ArrayRef<uint8_t> getContent() const override {
+    Value = 0;
+    ELFSection *pltSection =
+        m_PLTEntry->getOwningSection()->getOutputELFSection();
+
+    // Point to PLT entry + 6 (pushq instruction)
+    Value = pltSection->addr() + m_PLTEntry->getOffset() + 6;
+
+    return llvm::ArrayRef(reinterpret_cast<uint8_t *>(&Value), sizeof(Value));
   }
+
+  static x86_64GOTPLTN *Create(ELFSection *O, ResolveInfo *R);
+
+private:
+  Fragment *m_PLTEntry;
+  mutable uint64_t Value;
 };
 
 class x86_64GDGOT : public x86_64GOT {
--- a/lib/Target/x86_64/x86_64LDBackend.cpp
+++ b/lib/Target/x86_64/x86_64LDBackend.cpp
@@ -217,9 +217,8 @@ x86_64PLT *x86_64LDBackend::createPLT(ELFObjectFile *Obj, ResolveInfo *R) {
 
   // The relocation index is set before getContent() is called during
   // layout, as it is embedded directly in the PLT entry's instruction bytes.
-  if (x86_64PLTN *pltn = llvm::dyn_cast<x86_64PLTN>(P)) {
-    pltn->setRelocIndex(m_RelaPLTIndex);
-  }
+  x86_64PLTN *pltn = llvm::cast<x86_64PLTN>(P);
+  pltn->setRelocIndex(m_RelaPLTIndex);
 
   Relocation &rela_entry = *Obj->getRelaPLT()->createOneReloc();
   rela_entry.setType(llvm::ELF::R_X86_64_JUMP_SLOT);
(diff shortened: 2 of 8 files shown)


lib/Target/x86_64/x86_64GOT.h
```diff
@@ -98,18 +98,47 @@ class x86_64GOTPLT0 : public x86_64GOT {
...
 
-  static x86_64GOTPLTN *Create(ELFSection *O, ResolveInfo *R) {
-    return make<x86_64GOTPLTN>(O, R);
+  // Link to corresponding PLT entry for address calculation
+  void setPLTEntry(Fragment *plt) { m_PLTEntry = plt; }
+
+  virtual llvm::ArrayRef<uint8_t> getContent() const override {
+    Value = 0;
+
+    if (m_PLTEntry) {
```
parth-07: Is there any valid case for which we would not have any m_PLTEntry here?

lib/Target/x86_64/x86_64LDBackend.cpp
```diff
@@ -203,21 +203,32 @@ x86_64PLT *x86_64LDBackend::createPLT(ELFObjectFile *Obj, ResolveInfo *R) {
...
-                                    createGOT(GOT::GOTPLTN, Obj, R),
-                                    Obj->getPLT(), R, hasNow);
-  // init the corresponding rel entry in .rela.plt
+  x86_64PLT *P =
+      x86_64PLTN::Create(*m_Module.getIRBuilder(),
+                         createGOT(GOT::GOTPLTN, Obj, R), getPLT(), R, hasNow);
+
+  // The relocation index is set before getContent() is called during
+  // layout, as it is embedded directly in the PLT entry's instruction bytes.
+  if (x86_64PLTN *pltn = llvm::dyn_cast<x86_64PLTN>(P)) {
```
parth-07: If cast should always be successful, then use `llvm::cast<..>` instead of `llvm::dyn_cast<...>`

lib/Target/x86_64/x86_64PLT.cpp
```diff
@@ -37,42 +38,35 @@ x86_64PLT0 *x86_64PLT0::Create(eld::IRBuilder &I, x86_64GOT *G, ELFSection *O,
...
-      true /* isPostLTOPhase */);
-  symbol->setShouldIgnore(false);
-  r1 = Relocation::Create(llvm::ELF::R_X86_64_JUMP_SLOT, 64,
-                          make<FragmentRef>(*P, 0), 0);
-  r1->setSymInfo(symbol->resolveInfo());
-  r2 = Relocation::Create(llvm::ELF::R_X86_64_JUMP_SLOT, 64,
-                          make<FragmentRef>(*P, 8), 8);
-  r2->setSymInfo(symbol->resolveInfo());
+  // Link GOTPLTN to this PLT entry so it can compute its initial value
+  if (x86_64GOTPLTN *gotplt = llvm::dyn_cast<x86_64GOTPLTN>(G)) {
```
parth-07: If cast should always be successful, then use `llvm::cast<..>` instead of `llvm::dyn_cast<...>`

lib/Target/x86_64/x86_64GOT.h
bhoumik03: No, we should always have it, removed the null check.

lib/Target/x86_64/x86_64GOT.h
```diff
@@ -98,18 +98,44 @@ class x86_64GOTPLT0 : public x86_64GOT {
...
+  virtual llvm::ArrayRef<uint8_t> getContent() const override {
+    Value = 0;
+    ELFSection *pltSection =
+        m_PLTEntry->getOwningSection()->getOutputELFSection();
+    if (pltSection) {
+      // Point to PLT entry + 6 (pushq instruction)
+      Value = pltSection->addr() + m_PLTEntry->getOffset() + 6;
+    }
+
+    return llvm::ArrayRef(reinterpret_cast<const uint8_t *>(&Value),
```
parth-07: Do you need to cast it to `const`? Ideally, `reinterpret_cast` should not be used for changing constness.

lib/Target/x86_64/x86_64LDBackend.cpp
```diff
@@ -217,7 +217,7 @@ x86_64PLT *x86_64LDBackend::createPLT(ELFObjectFile *Obj, ResolveInfo *R) {
 
   // The relocation index is set before getContent() is called during
   // layout, as it is embedded directly in the PLT entry's instruction bytes.
-  if (x86_64PLTN *pltn = llvm::dyn_cast<x86_64PLTN>(P)) {
+  if (x86_64PLTN *pltn = llvm::cast<x86_64PLTN>(P)) {
     pltn->setRelocIndex(m_RelaPLTIndex);
   }
```
parth-07:
```suggestion
  x86_64PLTN *pltn = llvm::cast<x86_64PLTN>(P))
  pltn->setRelocIndex(m_RelaPLTIndex);
```

lib/Target/x86_64/x86_64PLT.cpp
```diff
@@ -45,6 +45,11 @@ x86_64PLTN *x86_64PLTN::Create(eld::IRBuilder &I, x86_64GOT *G, ELFSection *O,
   x86_64PLTN *P = make<x86_64PLTN>(G, I, O, R, 16, 16);
   O->addFragmentAndUpdateSize(P);
 
+  // Link GOTPLTN to this PLT entry so it can compute its initial value
+  if (x86_64GOTPLTN *gotplt = llvm::cast<x86_64GOTPLTN>(G)) {
+    gotplt->setPLTEntry(P);
+  }
```
parth-07:
```suggestion
  x86_64GOTPLTN *gotplt = llvm::cast<x86_64GOTPLTN>(G))
  gotplt->setPLTEntry(P);
  
```

test/x86_64/linux/GOTPLTNStructure/GOTPLTNStructure.test
```diff
@@ -0,0 +1,22 @@
...
+#BEGIN_COMMENT
+# Verifies GOTPLTN entries point to PLT+6 (pushq instruction)
+#END_COMMENT
+#START_TEST
+RUN: %clang %clangopts -c %p/Inputs/1.c -o %t.1.o
+RUN: %clang %clangopts -c %p/Inputs/2.c -o %t.2.o
+RUN: %link %linkopts -shared -o %t.lib2.so %t.2.o
+#The script fixes the address of .plt section
+RUN: %link %linkopts -T %p/Inputs/script_fixPLTaddress.t -o %t.out %t.1.o %t.lib2.so
+RUN: (%readelf -x .got.plt %t.out && %readelf -S %t.out) | %filecheck %s
```
parth-07: Why do you need `%readelf -S` here?

lib/Target/x86_64/x86_64GOT.h
bhoumik03: You are correct, we don't need to cast it explicitly.

test/x86_64/linux/GOTPLTNStructure/GOTPLTNStructure.test
bhoumik03: It is not needed, removed it.

PR #516, opened 2025-11-03
x86_64: Add .rela.plt section creation for dynamic linking

### LINKED ISSUE #569: x86_64: Add .rela.plt section creation for dynamic linking
`rela.plt` contains ELF RELA relocation entries used by the dynamic linker to resolve symbols at runtime.

- **Entry size:** 24 bytes each  

## Entry layout

- **`r_offset`** — Virtual address where the relocation is applied  
- **`r_info`** — Encodes symbol table index and relocation type
- **`r_addend`** — Explicit signed addend used by RELA relocations

## Common relocation type

- **`R_X86_64_JUMP_SLOT`** — The most common type for `rela.plt`; used by the dynamic linker to write the resolved absolute address into the PLT/GOT slot so PLT calls go directly to the target function after resolution.

bhoumik03:
Implement .rela.plt section creation that provides the dynamic linker with information needed for runtime relocation resolution

- Add .rela.plt section creation in doPreLayout
- Fix getRelaEntrySize() from 12 to 24 bytes
- Implement getDynRelocType() to classify R_X86_64_JUMP_SLOT as JMP_SLOT type

Progress on #534 
Fixes #569


lib/Target/x86_64/x86_64GOT.h
```diff
@@ -98,18 +97,43 @@ class x86_64GOTPLT0 : public x86_64GOT {
   mutable uint8_t Value[24];
 };
 
+/** \class x86_64GOTPLTN
+ *  \brief GOT entry for PLT function (lazy binding).
```
parth-07: Do we not need .got.plt entry for non-lazy binding case?

lib/Target/x86_64/x86_64GOT.h
```diff
@@ -98,18 +97,43 @@ class x86_64GOTPLT0 : public x86_64GOT {
...
-  static x86_64GOTPLTN *Create(ELFSection *O, ResolveInfo *R) {
-    return make<x86_64GOTPLTN>(O, R);
+  // Link to corresponding PLT entry for address calculation
+  void setPLTEntry(Fragment *plt) { m_PLTEntry = plt; }
+
+  virtual llvm::ArrayRef<uint8_t> getContent() const override {
+    Value = 0;
+    ELFSection *pltSection =
+        m_PLTEntry->getOwningSection()->getOutputELFSection();
+    if (pltSection) {
```
parth-07: For which cases will the pltSection be nullptr here?

bhoumik03: It should not be nullptr. Removed the check.

lib/Target/x86_64/x86_64GOT.h
bhoumik03: Yes we do need it. Improved the comment.

lib/Target/x86_64/x86_64LDBackend.cpp
```diff
@@ -124,6 +124,12 @@ void x86_64LDBackend::doPreLayout() {
   // Create a dynamic section handler for non-static or forced-dynamic outputs.
   if (!config().isCodeStatic() || config().options().forceDynamic())
     m_pDynamic = make<x86_64ELFDynamic>(*this, config());
+
+  if (LinkerConfig::Object != config().codeGenType()) {
```
parth-07, line 128: Would this condition be true for static executables as well?

bhoumik03: Yes, .rela.plt section needs to be created in static executables for IRELATIVE relocations

parth-07: Who resolves IRELATIVE relocations in `.rela.plt` section for the case of non-pie executables?

bhoumik03: this is for PIE static executables

parth-07: Can we please modify the conditions such that we do not create/modify `.rela.plt` for non-pie executables? I think it will make the linker flow more clear. I know that .rela.plt will be discarded later for the non-pie executable case because of 0 size. Not creating .rela.plt for non-pie executables can be helpful for erroneous cases such as users using position-independent object files to create non-pie executables.

bhoumik03:
```
#include <stdio.h>

int main(){
    printf("Hello\n");
    return 0;
}
```

bfd places them in .rela.plt
```
clang -fuse-ld=bfd -static 1.c -o 1.bfd.out
```

lld places them in .rela.dyn

```
clang -fuse-ld=lld -static 1.c -o 1.bfd.out
```

bhoumik03: Yes, it is not needed for non-PIE executables, will modify the conditions for that case

parth-07: No, I think it is needed for non-pie static executables. It seems that glibc startup files contains code to resolve these relocations when a dynamic linker is not present.

PR #523, opened 2025-11-04
[NFC] Remove redundant error diagnostic error_no_backend_found

parth-07:
This commit removes redundant error diagnostic error_no_backend_found.
If backend initialization/creation has failed then initBackend function
already emits the error diagnostic for the same.


quic-seaswara:
It would make sense to remove the diag from init Backend and keep the linker call to initBackend as is.

parth-07:
> It would make sense to remove the diag from init Backend and keep the linker call to initBackend as is.

I have not modified the initBackend call. The removed code is a code that is currently unreachable right now.

PR #571, opened 2025-11-10
Fix FileCheck pattern for AbsoluteSymbolRelocation

quic-k:
Update abssymbolrelocation.test to avoid false negatives when matching readelf output. The previous CHECK-NOT pattern excluded "00000000" globally, which can incorrectly match the address column in Hexagon binaries. The new regex anchors to the data column, ensuring we only forbid zero-initialized relocation values while allowing zero addresses.

### DIFF
--- a/test/Hexagon/standalone/AbsoluteSymbolRelocation/abssymbolrelocation.test
+++ b/test/Hexagon/standalone/AbsoluteSymbolRelocation/abssymbolrelocation.test
@@ -2,4 +2,4 @@
 RUN: %clang %clangopts -c %p/Inputs/1.c -o %t1.o %clangg0opts
 RUN: %link %linkopts %t1.o -o %t2.out
 RUN: %readelf -x .data %t2.out | %filecheck %s
-#CHECK-NOT: 00000000
+#CHECK-NOT: {{^0x[0-9a-f]+}}{{[[:space:]]+}}00000000


test/Hexagon/standalone/AbsoluteSymbolRelocation/abssymbolrelocation.test
```diff
@@ -2,4 +2,4 @@
 RUN: %clang %clangopts -c %p/Inputs/1.c -o %t1.o %clangg0opts
 RUN: %link %linkopts %t1.o -o %t2.out
 RUN: %readelf -x .data %t2.out | %filecheck %s
-#CHECK-NOT: 00000000
+#CHECK-NOT: ^0x[0-9a-f]+\s*00000000
```
parth-07: This does not look right. It is making sure that the absolute string '^0x[0-9a-f]+\s*00000000' is not matched instead of this regex is not matched.

parth-07 (changes requested):

test/Hexagon/standalone/AbsoluteSymbolRelocation/abssymbolrelocation.test
quic-k: right, I have pushed a fix so Filecheck matches against regex

test/Hexagon/standalone/AbsoluteSymbolRelocation/abssymbolrelocation.test
```diff
@@ -2,4 +2,4 @@
 RUN: %clang %clangopts -c %p/Inputs/1.c -o %t1.o %clangg0opts
 RUN: %link %linkopts %t1.o -o %t2.out
 RUN: %readelf -x .data %t2.out | %filecheck %s
-#CHECK-NOT: 00000000
+#CHECK-NOT: {{^0x[0-9a-f]+}}{{[[:space:]].*}}00000000
```
parth-07: Wouldn't this also match non-zero values in .data. For example, `0x1000 200000`, where `0x1000` is the address and `200000` is the value. Would this be better?

```suggestion
#CHECK-NOT: {{^0x[0-9a-f]+}}{{[[:space:]]+}}00000000
```

quic-k: i think it still works as it still checks that data column is non zero, but this is better, thanks!

PR #574, opened 2025-11-11
[NFC] Add ELFSection::getLocation for diagnostics.

quic-areg:
Gives convenient location strings for diagnostics
in the form file:(section+offset) similar to lld.

### DIFF
--- a/lib/Readers/ELFSection.cpp
+++ b/lib/Readers/ELFSection.cpp
@@ -205,7 +205,7 @@ std::optional<std::string> ELFSection::getRMSectName() const {
 }
 
 std::string ELFSection::getLocation(uint64_t Offset,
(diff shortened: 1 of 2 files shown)


lib/Readers/ELFSection.cpp
```diff
@@ -203,3 +203,20 @@ std::optional<std::string> ELFSection::getRMSectName() const {
   }
   return RMSectName;
 }
+
+std::string ELFSection::getLocation(uint64_t Offset,
+                                    GeneralOptions &Options) const {
```
parth-07: Can we pass `Options` as `const`?

quic-areg: yes, thanks

PR #575, opened 2025-11-11
Update workflow to continue when no dashboard updates are available.

quic-shaksuma:
This commit fixes failed dashboard update workflow if no DB updates are available.

### DIFF
--- a/.github/workflows/update-build-dashboard-data.yml
+++ b/.github/workflows/update-build-dashboard-data.yml
@@ -38,3 +38,4 @@ jobs:
             git add .
             git commit -m "Update build status data for dashboard"
             git push -f origin HEAD:refs/heads/documentation
+        continue-on-error: true


.github/workflows/update-build-dashboard-data.yml
```diff
@@ -38,3 +38,4 @@ jobs:
             git add .
             git commit -m "Update build status data for dashboard"
             git push -f origin HEAD:refs/heads/documentation
+        continue-on-error: true
```
parth-07, line 41: This is not actually fixing the issue, right? Can we please update the commit and PR title?

quic-shaksuma: > This is not actually fixing the issue, right? Can we please update the commit and PR title?

The issue here is that the workflow is marked as failed when there is not an actual failure.
A no change commit unless accompanied with allow-empty will get rejected as expected, so, this patch is meant to fix the failed status of the workflow when the steps are behaving as expected and allow the workflow to complete.

I guess it depends on how we look at the issue here :)

parth-07: Thank you for the explanation.

PR #585, opened 2025-11-13
[NFC][ARM] remove signed-overflow helper

quic-areg:
We can use llvm::isInt<N> instead

### DIFF
--- a/lib/Target/ARM/ARMRelocator.cpp
+++ b/lib/Target/ARM/ARMRelocator.cpp
@@ -207,19 +207,6 @@ helper_thumb32_cond_branch_lower(Relocator::DWord pLower16,
   return (pLower16 & 0xd000U) | (j1 << 13) | (j2 << 11) | lo;
 }
 
-// Return true if overflow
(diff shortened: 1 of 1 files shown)


lib/Target/ARM/ARMRelocator.cpp
```diff
@@ -1082,7 +1069,7 @@ Relocator::Result thm_jump11(Relocation &pReloc, ARMRelocator &pParent) {
             ->getAddr(DiagEngine);
 
   Relocator::DWord X = S + A - P;
-  if (helper_check_signed_overflow(X, 12))
+  if (!llvm::isInt<12>(static_cast<int32_t>(X)))
```
parth-07, line 1072: Why are we casting it all to int32_t? Is there any issue if we keep them 64-bits?

quic-areg: The relocation arithmetic is done in 64 bits but the actual instruction encoding only cares about the lower bits. `arm-wraparound veneers` is an example of a test that would fail without a cast.

PR #588, opened 2025-11-13
Add plugin activity log file functionality

parth-07:
This commit adds primitive support for plugin activity log file functionality.
It adds the option '--plugin-activity-log <file>' to the linker. With this option,
the linker emits a JSON plugin activity log file that contains the
ordered records of relevent plugin actions for later analysis and debugging. Currently,
only the following operations are fully supported: RemoveSymbol, UpdateChunks,
and updateLinkStat. The support for remaining operations will be added
in follow-up patches.

One key difference between the plugin operation information in the map-file
and the plugin activity log file is that in the map-file, the plugin
operations are separated as per operation type and the plugin. On the
other hand, in the plugin activity log file, we have a single array of
the operations by all the plugins. This allows to determine the total
ordering of the operations that can be important for various analysis.


include/eld/Core/LinkState.h
```diff
@@ -0,0 +1,70 @@
+//===-
+// LinkState.h------------------------------------------------------------===//
```
quic-seaswara: nit : extra new line ?

include/eld/Core/LinkState.h
```diff
@@ -0,0 +1,70 @@
...
+// See https://github.com/qualcomm/eld/LICENSE.txt for license information.
+// SPDX-License-Identifier: BSD-3-Clause
+//===----------------------------------------------------------------------===//
+//
+//                     The MCLinker Project
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
```
quic-seaswara: Probably this file does not require MCLinker copyright

include/eld/Core/LinkState.h
```diff
@@ -0,0 +1,70 @@
...
+/// plugin is meaningful in the link state in which it is executed.
+/// Executing invalid actions for a link state can result in undefined
+/// behavior.
+enum LinkState : uint8_t {
+  Unknown,
+  Initializing,
+  BeforeLayout,
+  CreatingSections,
+  CreatingSegments,
+  AfterLayout,
```
quic-seaswara, line 36: add few comments along with the enum, it might help.

include/eld/Core/LinkState.h
```diff
@@ -0,0 +1,70 @@
...
+  switch (State) {
+    ADD_CASE(Unknown)
+    ADD_CASE(Initializing)
+    ADD_CASE(BeforeLayout)
+    ADD_CASE(CreatingSections)
+    ADD_CASE(CreatingSegments)
+    ADD_CASE(AfterLayout)
+  }
+#undef ADD_CASE
+  llvm_unreachable("Invalid LinkState");
```
quic-seaswara, line 52: This is correct, but it will make either the map file or reproduce tarball unavailable to debug. Can we figure out a better pattern ? 

Similar usages are found in the rest of the linker but we need to figure out if there is a way to make the map file and reproduce always available for development team to debug issues.

We can let this instance continue, but we need to figure out how to improve.

include/eld/Driver/GnuLinkerOptions.td
```diff
@@ -972,6 +972,12 @@ defm TrampolineMap
       MetaVarName<"<filename>">,
       Group<grp_mapopts>;
 
+defm PluginActivityFile
+    : smDashWithOpt<"plugin-activity-file", "PluginActivityFile",
```
quic-seaswara: plugin-activity-log ? can we create a seperate category for plugins ? This is housed under map file options or diagnostics, can it be moved ?

include/eld/Fragment/Fragment.h
```diff
@@ -123,6 +124,11 @@ class Fragment {
 
   bool originatesFromPlugin(const Module &Module) const;
 
+  /// Returns a string containing basic fragment information, such as, section
+  /// and input file name. It is intended to be used for printing to log files,
+  /// diagnostics, or for quick debugging.
+  std::string str(const GeneralOptions &Options) const;
```
quic-seaswara, line 135: toString ?

it might be easy to grep for the occurence and find out in the source code, str makes it very difficult.

lib/Core/LinkerScript.cpp
```diff
@@ -190,12 +190,16 @@ void LinkerScript::addSectionOverride(plugin::LinkerWrapper *W, eld::Module *M,
...
+  if (auto &PluginActLog = M->getPluginActivityLog())
+    PluginActLog->addPluginOperation(*Op);
 }
 
 void LinkerScript::removeSymbolOp(plugin::LinkerWrapper *W, eld::Module *M,
                                   const ResolveInfo *S) {
   RemoveSymbolPluginOp *Op = make<RemoveSymbolPluginOp>(W, "", S);
   LayoutInfo *layoutInfo = M->getLayoutInfo();
+  if (auto &PluginActLog = M->getPluginActivityLog())
+    PluginActLog->addPluginOperation(*Op);
```
quic-seaswara, line 231: We want to record what plugin's activity is this, so we need the linker wrapper(W)

lib/Plugin/PluginActivityLog.cpp
```diff
@@ -0,0 +1,150 @@
+//===-
+// PluginActivityLog.cpp------------------------------------------------------===//
```
quic-seaswara, line 1: nit : extra line

lib/Plugin/PluginActivityLog.cpp
```diff
@@ -0,0 +1,150 @@
...
+  auto JSONObj = getBaseActivityJSONObject(Op);
+  JSONObj["NewState"] = getLinkStateStrRef(Op.getNewState());
+  return JSONObj;
+}
+
+llvm::json::Object
+PluginActivityLog::getBaseActivityJSONObject(const PluginOp &Op) const {
+  return llvm::json::Object{{"Plugin", Op.getPluginName()},
+                            {"Type", Op.getPluginOpTypeStrRef()},
+                            {"Annotation", Op.getAnnotation()}};
```
quic-seaswara, line 148: dont emit if empty ?

lib/Plugin/PluginActivityLog.cpp
```diff
@@ -0,0 +1,150 @@
...
+llvm::json::Object
+PluginActivityLog::toJSON(const UpdateLinkStateOp &Op) const {
+  auto JSONObj = getBaseActivityJSONObject(Op);
+  JSONObj["NewState"] = getLinkStateStrRef(Op.getNewState());
+  return JSONObj;
+}
+
+llvm::json::Object
+PluginActivityLog::getBaseActivityJSONObject(const PluginOp &Op) const {
+  return llvm::json::Object{{"Plugin", Op.getPluginName()},
```
quic-seaswara, line 146: path of the plugin, and the command line options registered will be useful.

lib/Plugin/PluginActivityLog.cpp
quic-seaswara, line 148: can we indicate if the plugin was auto loaded or loaded seperately ?

parth-07:
> As usual, pls document :)

Can we please document this once the functionality is close to complete? This PR only adds primitive support of this feature.

include/eld/Core/LinkState.h
parth-07: clang-format changed this. I have now removed the extra new-line.

include/eld/Core/LinkState.h
parth-07: This PR is only moving this enum from `Module.h` to `LinkState.h`. Can we handle improving internal documentation in separate PRs?

include/eld/Core/LinkState.h
parth-07: > but it will make either the map file or reproduce tarball unavailable to debug.

It should not make the map-file or reproduce tarball unavailable to debug as this assert should never fail. Only linker can set the link states, hence, link state must always be valid. Can there be any case where linker may set incorrect link state? I believe that we should not avoid using assert / llvm_unreachable for the basic but important checks. As avoiding these checks can cause greater issues and make final debugging more complicated.

include/eld/Core/LinkState.h
parth-07: I have removed the license from this file. But is it okay to have a file without a license?

lib/Core/LinkerScript.cpp
parth-07: `PluginOp` object contains the linker wrapper.

lib/Plugin/PluginActivityLog.cpp
parth-07: I have removed the extra line.

lib/Plugin/PluginActivityLog.cpp
parth-07: This function avoids code-duplication for creating a JSON object for the fields that are common to each plugin operation. It does not make a decision regarding emitting  / don't-emiting the plugin activity.

lib/Plugin/PluginActivityLog.cpp
parth-07: Thank you for the suggestion. I like this idea. I am making the necessary changes.

lib/Plugin/PluginActivityLog.cpp
parth-07: Yes, it sounds good. I am making the necessary changes.

include/eld/Driver/GnuLinkerOptions.td
parth-07: Thank you for mentioning this. Yes, I will move it to a different category. I will need to see which category would fit best for it.

I previously used the name `--plugin-activity-log` but later changed it to `--plugin-activity-file` as the latter makes it clear that the input argument would be a filename. What do you think about it?

include/eld/Core/LinkState.h
parth-07: I have removed `MCLinker` from the file's license**

include/eld/Core/LinkState.h
parth-07: I will add the documentation for states in an another PR that adds the new link states (`ActBefore` states) to the `LinkState` enum.

lib/Plugin/PluginActivityLog.cpp
parth-07: I have added path of the plugin library and the registered command-line options.

lib/Plugin/PluginActivityLog.cpp
parth-07: Done. Now the plugin activity log file indicates if the plugin whether the plugin was autoloaded.

test/Common/Plugin/PluginActivityLogFile/PluginActivityLogFile.test
```diff
@@ -0,0 +1,153 @@
...
+PLUGIN_ACT-DAG:       "Fragment": "{{.*}}1.o(.text.foo1)",
+PLUGIN_ACT-DAG:       "Plugin": "OutSectIterPluginActLog",
+PLUGIN_ACT-DAG:       "Rule": "*(.text.foo*) #Rule 1, {{.*}}script.t",
+PLUGIN_ACT-DAG:       "Type": "RemoveChunk"
+PLUGIN_ACT:     },
+PLUGIN_ACT:     {
+PLUGIN_ACT-DAG:       "Annotation": "",
+PLUGIN_ACT-DAG:       "Fragment": "{{.*}}1.o(.text.foo2)",
+PLUGIN_ACT-DAG:       "Plugin": "OutSectIterPluginActLog",
+PLUGIN_ACT-DAG:       "Rule": "*(.text.foo*) #Rule 1, {{.*}}script.t",
```
quic-seaswara, line 51: Can we have the Rule that contains only rule strings ? 

Since it is a data structure now compared to text, probably adding 

RuleID : 
InputFile:

include/eld/Driver/GnuLinkerOptions.td
parth-07: I have moved this option to the `grp_pluginopts`.

test/Common/Plugin/PluginActivityLogFile/PluginActivityLogFile.test
parth-07: Can we make this more modular / structured when the need arises? I feel that more modular structure here can increase the work without necessarily giving the benefits.

I plan to add a `--filter` option to the tool that will read the plugin activity file. Users can use it as `ELDPluginRunAnalyzer plugin.activity.json --filter "/a/b/c/1.o"` to find all the plugin actions that contains the filename `/a/b/c/1.o`. 

The filter option will work by simply seeing if the JSON structure of the rule matches the filter-word. Thus, it can be used
to find relevant plugin operations without having individual fields for things such as input file.

quic-seaswara (approved):

PR #594, opened 2025-11-14
x86_64: RelaPltJumpSlot test fix

bhoumik03:
- Currently the RelaPltJumpSlot test fails. This is because of the non-deterministic nature of symbol indices that are part of the Info part of a .rela.plt entry. 
- This commit makes the test more robust my ignoring the symbol indices. 
- It still checks for the symbol names to ensure that the .rela.plt entries are for the correct symbols.

### DIFF
--- a/test/x86_64/linux/RelaPltJumpSlot/RelaPltJumpSlot.test
+++ b/test/x86_64/linux/RelaPltJumpSlot/RelaPltJumpSlot.test
@@ -17,5 +17,5 @@ CHECK: .rela.plt
 #The offset value for the JUMP_SLOT should be the address of corressponding GOTPLTN entry.
 #The first 0x18 bytes are part of GOTPLT0
 #Info - first 4 bytes contain the index of the symbol in .dynsym the next 4 bytes are for relocation type
-CHECK: [[#GOTPLT_ADDR+0x18]]{{.*}}0000000200000007{{.*}}R_X86_64_JUMP_SLOT{{.*}}foo + 0
-CHECK: [[#GOTPLT_ADDR+0x20]]{{.*}}0000000100000007{{.*}}R_X86_64_JUMP_SLOT {{.*}}foo2 + 0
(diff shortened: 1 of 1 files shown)


test/x86_64/linux/RelaPltJumpSlot/RelaPltJumpSlot.test
```diff
@@ -17,5 +17,5 @@ CHECK: .rela.plt
 #The offset value for the JUMP_SLOT should be the address of corressponding GOTPLTN entry.
 #The first 0x18 bytes are part of GOTPLT0
 #Info - first 4 bytes contain the index of the symbol in .dynsym the next 4 bytes are for relocation type
-CHECK: [[#GOTPLT_ADDR+0x18]]{{.*}}0000000200000007{{.*}}R_X86_64_JUMP_SLOT{{.*}}foo + 0
-CHECK: [[#GOTPLT_ADDR+0x20]]{{.*}}0000000100000007{{.*}}R_X86_64_JUMP_SLOT {{.*}}foo2 + 0
\ No newline at end of file
+CHECK-DAG: [[#GOTPLT_ADDR+0x18]]{{.*}}00000007{{.*}}R_X86_64_JUMP_SLOT {{.*}}foo + 0
+CHECK-DAG: [[#GOTPLT_ADDR+0x20]]{{.*}}00000007{{.*}}R_X86_64_JUMP_SLOT {{.*}}foo2 + 0
```
quic-seaswara, line 21: can we just have have 

CHECK-DAG:foo
CHECK-DAG:foo2

A test that can build and run will be useful to make sure what you are building works.

bhoumik03: We do need to check the offset value of the relocation entry. In case of JUMP_SLOT here, it should be the address of the corresponding GOTPLTN entry.

bhoumik03: The build as of the commit for .rela.plt would not have worked. I have included the run test in the PR #518

PR #598, opened 2025-11-14
[ARM] Report unencodable imm for R_ARM_ALU_PC_G0

quic-areg:
Report unencodable immediate errors instead of
overflows for R_ARM_ALU_PC_G0. The encoding space
is sparse rather than simply bounded, so this
diagnostic makes more sense than "overflow".
This matches the lld diagnostic.

### DIFF
--- a/lib/Readers/Relocation.cpp
+++ b/lib/Readers/Relocation.cpp
@@ -217,6 +217,19 @@ bool Relocation::issueOverflow(Relocator &pRelocator) const {
   return false;
 }
 
+bool Relocation::issueUnencodableImmediate(Relocator &R, int64_t Imm) const {
+  std::string Location = "";
+  if (FragmentRef *Ref = targetRef())
+    if (Fragment *F = Ref->frag())
+      if (ELFSection *S = F->getOwningSection())
+        Location = S->getLocation(Ref->offset(), R.config().options());
+  R.config().raise(Diag::error_unencodable_imm)
+      << Location << Imm << R.getName(type())
+      << getSymbolName(symInfo(), R.doDeMangle());
+  ASSERT(!Location.empty(), "expected a section location.");
+  return false;
+}
+
 bool Relocation::apply(Relocator &pRelocator) {
   // Issue undefined references for section magic symbols here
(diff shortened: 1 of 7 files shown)


lib/Readers/Relocation.cpp
```diff
@@ -217,6 +217,17 @@ bool Relocation::issueOverflow(Relocator &pRelocator) const {
   return false;
 }
 
+bool Relocation::issueUnencodableImmediate(Relocator &R, int64_t Imm) const {
+  std::string Location = "";
+  if (FragmentRef *Ref = targetRef())
+    if (Fragment *F = Ref->frag())
+      if (ELFSection *S = F->getOwningSection())
+        Location = S->getLocation(Ref->offset(), R.config().options());
+  R.config().raise(Diag::error_unencodable_imm)
```
quic-seaswara, line 226: we need what symbol is being referred to

quic-seaswara (changes requested):

lib/Readers/Relocation.cpp
quic-areg: follow-up?

quic-seaswara: isn't this just print symInfo() ?

lib/Readers/Relocation.cpp
```diff
@@ -217,6 +217,17 @@ bool Relocation::issueOverflow(Relocator &pRelocator) const {
   return false;
 }
 
+bool Relocation::issueUnencodableImmediate(Relocator &R, int64_t Imm) const {
+  std::string Location = "";
+  if (FragmentRef *Ref = targetRef())
+    if (Fragment *F = Ref->frag())
+      if (ELFSection *S = F->getOwningSection())
+        Location = S->getLocation(Ref->offset(), R.config().options());
```
quic-seaswara, line 225: Do we expect a Location at all times, if so should 226 be an assert ?

quic-areg: probably

lib/Readers/Relocation.cpp
quic-areg: ok

Issue #491, opened 2025-10-16
risc-v : check for overflows from relocations as applicable


quic-seaswara:
Improve relocation overflow checks in risc-v target

PR #599, opened 2025-11-14
[Relocator] Emit valid ranges in overflow diags

### LINKED ISSUE #493: Improve relocation overflows diagnostics to show range and the symbol value computed
It would be useful to know the symbol value and the range when reporting a relocation overflow.

Look at how lld does relocation overflows and improve eld framework to report similar diagnostics.

quic-areg:
Overflow errors now show the valid encoding range
as well as the symbol referred similar to lld.

Eg:

1.o:(.text+0x4): relocation 'R_RISCV_JAL' for 'foo' is out of range:
  1048576 is not in [-1048576, 1048575]

closes #493, #491

### DIFF
--- a/lib/Target/AArch64/AArch64Relocator.cpp
+++ b/lib/Target/AArch64/AArch64Relocator.cpp
@@ -31,7 +31,7 @@ static Relocator::Result emitSignedOrUnsignedRangeOverflow(Relocation &Rel,
                                                            unsigned Bits) {
   int64_t Min = llvm::minIntN(Bits);
   int64_t Max = llvm::maxUIntN(Bits);
-  Rel.issueOverflow(R, Value, Min, Max);
+  Rel.issueSignedOverflow(R, Value, Min, Max);
   return Relocator::Overflow;
 }
 
--- a/lib/Target/Relocator.cpp
+++ b/lib/Target/Relocator.cpp
@@ -321,7 +321,7 @@ Relocator::Result eld::checkSignedRange(Relocation &Rel, Relocator &R,
   if (Value == llvm::SignExtend64(Value, Bits))
     return Relocator::OK;
 
-  Rel.issueOverflow(R, Value, llvm::minIntN(Bits), llvm::maxIntN(Bits));
+  Rel.issueSignedOverflow(R, Value, llvm::minIntN(Bits), llvm::maxIntN(Bits));
   return Relocator::Overflow;
 }
 
@@ -331,6 +331,6 @@ Relocator::Result eld::checkUnsignedRange(Relocation &Rel, Relocator &R,
   if (Bits == 64 || Value >> Bits == 0)
     return Relocator::OK;
 
-  Rel.issueOverflow(R, Value, 0, llvm::maxUIntN(Bits));
+  Rel.issueUnsignedOverflow(R, Value, 0, llvm::maxUIntN(Bits));
   return Relocator::Overflow;
 }
--- a/include/eld/Readers/Relocation.h
+++ b/include/eld/Readers/Relocation.h
@@ -145,8 +145,11 @@ public:
 
   bool isMergeKind() const;
 
-  bool issueOverflow(Relocator &R, int64_t Value, int64_t Min,
-                     int64_t Max) const;
+  bool issueSignedOverflow(Relocator &R, int64_t Value, int64_t Min,
+                           int64_t Max) const;
+
+  bool issueUnsignedOverflow(Relocator &R, uint64_t Value, uint64_t Min,
+                             uint64_t Max) const;
 
   bool issueUnencodableImmediate(Relocator &R, int64_t Imm) const;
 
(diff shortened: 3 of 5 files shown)


test/RISCV/FromLLD/riscv-branch.s
```diff
@@ -13,8 +13,8 @@
 # LIMITS-32-NEXT: 7e001de3     bnez    zero, 0x1052
 
 # RUN: not %link %linkopts %t.rv32.o --defsym foo=_start+0x1000 --defsym bar=_start+4-0x1002 -o /dev/null 2>&1 | FileCheck --check-prefix=ERROR-RANGE %s
-# ERROR-RANGE: Relocation overflow when applying relocation `R_RISCV_BRANCH' for symbol `foo'
-# ERROR-RANGE: Relocation overflow when applying relocation `R_RISCV_BRANCH' for symbol `bar'
+# ERROR-RANGE: Error:     {{.*}}rv32.o:(.text): relocation 'R_RISCV_BRANCH' for 'foo' is out of range: 4096 is not in [-4096, 4095]
+# ERROR-RANGE: Error: {{.*}}rv32.o:(.text+0x4): relocation 'R_RISCV_BRANCH' for 'bar' is out of range: -4098 is not in [-4096, 4095]
```
lenary: Can we be careful about the order of information in this message?

ELD prints absolute paths, iirc, which can be quite long. So can C++ symbol names. Maybe these should go towards the end of the line like LLD does, so the fixed size stuff is not pushed too far right?

is

lib/Target/x86_64/x86_64Relocator.cpp
```diff
@@ -289,22 +289,43 @@ void x86_64Relocator::partialScanRelocation(Relocation &pReloc,
 
 uint32_t x86_64Relocator::getNumRelocs() const { return x86_64_MAXRELOCS; }
 
+static unsigned getEncodingBitWidth(EncodingType T) {
```
parth-07: Instead of adding this function, can we use `getNumberOfBits` function in the file `x86_64RelocationCompute`?

lib/Target/Relocator.cpp
```diff
@@ -314,3 +314,23 @@ std::optional<uint64_t> Relocator::getMaxTLSSegmentAlign() const {
...
+  return Relocator::Overflow;
+}
+
+Relocator::Result eld::checkUnsignedRange(Relocation &Rel, Relocator &R,
+                                          uint64_t Value, unsigned int Bits) {
+
+  if (Bits == 64 || Value >> Bits == 0)
+    return Relocator::OK;
+
+  Rel.issueOverflow(R, Value, 0, llvm::maxUIntN(Bits));
```
parth-07: `issueOverflow` function takes the max value as signed integer and print it as such. Wouldn't we end up printing incorrect max value for unsigned cases?

lib/Target/AArch64/AArch64Relocator.cpp
```diff
@@ -25,6 +25,16 @@ using namespace eld;
 
 namespace {
 
+static Relocator::Result emitSignedOrUnsignedRangeOverflow(Relocation &Rel,
+                                                           AArch64Relocator &R,
+                                                           int64_t Value,
+                                                           unsigned Bits) {
+  int64_t Min = llvm::minIntN(Bits);
+  int64_t Max = llvm::maxUIntN(Bits);
```
parth-07, line 33: Is it okay to always assume unsigned here?

If `N` is 32, then `llvm::maxUIntN(32)` will return the value `2^32 -1`. On the other hand, `llvm::maxIntN(32)` would be `2^31 -1`. In this case, isn't the `Max` value incorrect for signed relocations?

parth-07 (changes requested):

test/RISCV/FromLLD/riscv-branch.s
quic-areg: Thanks for the suggestion; moved symbol names to the right to match lld

lib/Target/x86_64/x86_64Relocator.cpp
quic-areg: Thanks; Done

lib/Target/Relocator.cpp
quic-areg: we now have `issueSignedOverflow` and `IssueUnsignedOverflow`

lib/Target/AArch64/AArch64Relocator.cpp
quic-areg: this function checks the range[-2^N, 2^N) for `R_AARCH64_ABS32` and `R_AARCH64_ABS16`. I believe the limits are correct.

see the test `aarch64-abs32.s` taken from lld.

quic-akaryaki (approved):

lib/Target/AArch64/AArch64Relocator.cpp
parth-07: Can we please improve the name of the function? `emitSignedOrUnsignedRangeOverflow` does not indicate that max value should be computed differently than the usual.

quic-areg: What name do you suggest? To me, `SignedOrUnisigned` seems appropriate for a bound that is part signed and part unsigned, and is distinct from the usual `Signed` and `Unsigned` bounds checks everywhere else.

PR #603, opened 2025-11-17
Fix invalid 'Referenced Chunk ... deleted' error when printing map-file

### LINKED ISSUE #587: Invalid / misleading error `Plugin Error : Referenced Chunk ... has been deleted` when the plugin errors out at `CreatingSections` link state
We emit invalid and very misleading error diagnostic `Plugin Error : Referenced Chunk ... has been deleted` while emitting text map-file when the plugin errors out at `CreatingSections` link state. This error happens even if the plugin has not modified anything in the link.

Reproducer:

```bash
#!/usr/bin/env bash

cat >ErrorOutAtCreatingSections.cpp <<\EOF
#include "Defines.h"
#include "PluginBase.h"
#include "LinkerWrapper.h"
#include "OutputSectionIteratorPlugin.h"
#include "PluginADT.h"
#include "PluginVersion.h"
#include <cassert>
#include <map>

class DLL_A_EXPORT ErrorOutAtCreatingSections
    : public eld::plugin::OutputSectionIteratorPlugin {
public:
  ErrorOutAtCreatingSections()
      : eld::plugin::OutputSectionIteratorPlugin("ErrorOutAtCreatingSections") {}

  void Init(std::string options) override {}

  void processOutputSection(eld::plugin::OutputSection O) override {}

  Status Run(bool trace) override {
    if (!getLinker()->isLinkStateCreatingSections())
      return Status::SUCCESS;
    return Status::ERROR;
  }

  std::string GetName() override { return "ErrorOutAtCreatingSections"; }

  std::string GetLastErrorAsString() override { return "Error!"; }

  void Destroy() override {}

  uint32_t GetLastError() override { return 0; }
};

ELD_REGISTER_PLUGIN(ErrorOutAtCreatingSections)
EOF

cat >1.c <<\EOF
int foo() { return 1; }

int bar() { return 5; }
EOF

cat >script.t <<\EOF
SECTIONS {
  FOO : {
    *(*foo*)
  }
}

PLUGIN_OUTPUT_SECTION_ITER("ErrorOutAtCreatingSections", "ErrorOutAtCreatingSections")
EOF

TOOLS_DIR=$(dirname $(which ld.eld))/..
clang++-18 -o libErrorOutAtCreatingSections.so ErrorOutAtCreatingSections.cpp -fPIC -shared -I${TOOLS_DIR}/include/ELD/PluginAPI -L ${TOOLS_DIR}/lib -lLW -stdlib=libc++
ld.eld -o 1.out 1.o -L . -T script.t -Map 1.map.txt
```

**Output**
```
Error: Plugin ErrorOutAtCreatingSections defined in library libErrorOutAtCreatingSections.so returned error Error! in state CreatingSections
Error: Plugin Error : Referenced Chunk .text.foo from 1.o, last seen from Rule *(*foo*) #Rule 1, script.t in Output Section FOO has been deleted!
Fatal: Linking had errors.
```

`Referenced Chunk ... has been deleted` error does not make sense and is confusing and wrongly indicates a plugin has caused this.

The root cause of this error is that when the plugin fails, we end the link before fragments are assigned any offsets. Later, when we print text map-file, the text map-file find fragments without any assigned offsets and reports this misleading error. This is because fragments must always have an offset as per the text map-file printer, unless something wrong has happened -- which it assumed only plugins can do.

parth-07:
This commit fixes the invalid and misleading error 'Referenced Chunk ... has been deleted' during printing of the text map-file when a plugin errors out at CreatingSections link state. This comit fixes this by adding checks in the TextLayoutPrinter to not compute fragment offset if the link state is before CreatingSegments. Additionally, now we print 'NA' for offsets in the text map-file when the map-file is being printed in CreatingSections link state.

Resolves #587


quic-seaswara:
If offsets are only later assigned, and no chunks have offsets, does it make sense to print "NA" or skip printing offsets all together. 

Just adding a note, that offsets are not assigned ?

PR #604, opened 2025-11-17
Update musl and picolibc workflows to record build status data.

quic-shaksuma:
Github Workflows updated to record build status data for picolibc
and musl builds using BuildStatusDataRecorder github composite action.
A file lock is also used to eliminate race condition in
updating the DB file.

### DIFF
--- a/.github/workflows/BuildStatusDataRecorder/action.yaml
+++ b/.github/workflows/BuildStatusDataRecorder/action.yaml
@@ -1,0 +1,105 @@
+name: Record Workflow
+description: Record pre/post build data for dashboard
+
+inputs:
+  enabled:
+    description: "Turn workflow recording On/Off"
+    default: true
+  project:
+    description: "Project name for build workflow recording"
+    required: true
+  stage:
+    description: "Pre/post build"
+    required: false
+  workflow-name:
+    description: "Build workflow name"
+    required: true
+  record-mode:
+    description: "Build data record or update mode"
+    required: true
+  status:
+    description: "Build workflow pass/fail status"
+    required: true
+  run-id:
+    description: "Build workflow run-id"
+    required: true
+  arch-name:
+    description: "Workflow matrix build arch name"
+    required: true
+
+runs:
+  using: "composite"
+  steps:
+    - name: Record ${{ inputs.stage }}build entry
+      if: ${{ inputs.enabled }} == "true"
+      shell: bash
+      run: |
+        cd ${{ inputs.project }}
+        DASH_DIR="$(mktemp -d)"
+        cp .github/workflows/scripts/record_builds.py ${DASH_DIR}/
+        git config --global user.name 'github-actions'
+        git config --global user.email 'github-actions@github.com'
+        # Switch to dashboard branch to access build-status.db
+        git fetch --all
+        git checkout dashboard
+        git reset --hard origin/dashboard
+        ls -l
+        : '
+        Multiple workflows and parallel matrix builds write to
+        build-status.db file. This causes race condition, stale DB or
+        git latest ref conflicts.
+        check_lock() is called throughout this workflow to make
+        sure build data is written to current DB file.
+        '
+        check_lock() {
+          FILE_PATH="data.lock"
+          # Checks for build data lock.
+          while [ -f "$FILE_PATH" ]; do
+            echo "Wait for data lock to be removed..."
+            sleep 10
+            git fetch --all
+            git reset --hard origin/dashboard
+            git pull
+            ls -l
+          done
+          echo "Data lock removed"
+        }
+        check_lock
+        # After data lock is released by other workflows, a lock is acquired for current workflow.
+        echo "Adding ${{ inputs.workflow-name }} ${{ inputs.arch-name }} data lock."
+        touch data.lock
+        git add .
+        git commit -m "Add data lock for ${{ inputs.workflow-name }} ${{ inputs.arch-name }} build status data to be written"
+        # Keep trying until a data lock is pushed successfully.
+        until git push -f origin HEAD:refs/heads/dashboard; do
+          sleep 5
+          check_lock
+          git fetch --all
+          git reset --hard origin/dashboard
+          touch data.lock
+          git add .
+          git commit -m "Add data lock for ${{ inputs.workflow-name }} ${{ inputs.arch-name }} build status data to be written"
+        done
+        # Record build status data for current workflow.
+        echo "Proceeding with data recording..."
+        cp ${DASH_DIR}/record_builds.py .
+        python record_builds.py --workflow ${{ inputs.workflow-name }} --${{ inputs.record-mode }} --${{ inputs.status }} --run-id ${{ inputs.run-id }} --arch ${{ inputs.arch-name }}
+        cp build-status.db ${DASH_DIR}/
+        rm -f record_builds.py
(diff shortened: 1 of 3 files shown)


.github/workflows/BuildStatusDataRecorder/action.yaml
```diff
@@ -0,0 +1,99 @@
...
+        DASH_DIR="$(mktemp -d)"
+        cp .github/workflows/scripts/record_builds.py ${DASH_DIR}/
+        git config --global user.name 'github-actions'
+        git config --global user.email 'github-actions@github.com'
+        # Switch to dashboard branch to access build-status.db
+        git fetch --all
+        git checkout dashboard
+        git reset --hard origin/dashboard
+        ls -l
+        : '
```
quic-seaswara, line 47: can we remove these ls -l and :'

.github/workflows/BuildStatusDataRecorder/action.yaml
```diff
@@ -0,0 +1,99 @@
...
+        git add .
+        git commit -m "Add data lock for ${{ inputs.workflow-name }} ${{ inputs.arch-name }} build status data to be written"
+        # Keep trying until a data lock is pushed successfully.
+        until git push -f origin HEAD:refs/heads/dashboard; do
+          sleep 5
+          check_lock
+          git fetch --all
+          git reset --hard origin/dashboard
+          touch data.lock
+          git add .
```
quic-seaswara, line 80: isnt this supposed to be outside the loop ?

.github/workflows/BuildStatusDataRecorder/action.yaml
```diff
@@ -0,0 +1,99 @@
...
+            sleep 10
+            git fetch --all
+            git reset --hard origin/dashboard
+            git pull
+            ls -l
+          done
+          echo "Data lock removed"
+        }
+        check_lock
+        # After data lock is released by other workflows, a lock is aquired for current workflow.
```
quic-seaswara: spell check workflow pls.

.github/workflows/BuildStatusDataRecorder/action.yaml
```diff
@@ -0,0 +1,99 @@
...
+        git add .
+        git commit -m "Add ${{ inputs.workflow-name }} ${{ inputs.arch-name }} build status data in DB"
+        # Keep trying until lock/conflicts are resolved and build data is pushed successfully.
+        until git push -f origin HEAD:refs/heads/dashboard; do
+          sleep 5
+          check_lock
+          git fetch --all
+          git reset --hard origin/dashboard
+          cp -f ${DASH_DIR}/build-status.db .
+          rm -f data.lock
```
quic-seaswara: git rm -f whereever applicable

.github/workflows/BuildStatusDataRecorder/action.yaml
```diff
@@ -0,0 +1,99 @@
+name: Record build date
+description: Record pre/post build data for dashboard
+
+inputs:
+  stage:
+    description: Pre/post build
+    required: true
+    default: '-'
```
quic-seaswara: required and default ?

.github/workflows/BuildStatusDataRecorder/action.yaml
```diff
@@ -0,0 +1,99 @@
+name: Record build date
```
quic-seaswara: record workflow ?

.github/workflows/BuildStatusDataRecorder/action.yaml
```diff
@@ -0,0 +1,99 @@
...
+    description: "Workflow matrix build arch name"
+    required: true
+
+runs:
+  using: "composite"
+  steps:
+    - name: Record ${{ inputs.stage }}build entry
+      shell: bash
+      run: |
+        cd llvm-project/llvm/tools/eld
```
quic-seaswara: probably project also can be a parameter ?

.github/workflows/BuildStatusDataRecorder/action.yaml
```diff
@@ -0,0 +1,99 @@
...
+        check_lock() {
+          FILE_PATH="data.lock"
+          # Checks for build data lock.
+          while [ -f "$FILE_PATH" ]; do
+            echo "Wait for data lock to be removed..."
+            sleep 10
+            git fetch --all
+            git reset --hard origin/dashboard
+            git pull
+            ls -l
```
quic-seaswara, line 63: Remove ls -l

.github/workflows/BuildStatusDataRecorder/action.yaml
```diff
@@ -0,0 +1,99 @@
...
+        git reset --hard origin/dashboard
+        ls -l
+        : '
+        Multiple workflows and parallel matrix builds write to
+        build-status.db file. This causes race condition, stale DB or
+        git latest ref conflicts.
+        check_lock() is called throughout this workflow to make
+        sure build data is written to current DB file.
+        '
+        check_lock() {
```
quic-seaswara, line 54: I would think this should do check_and_acquire_lock.

It will simplify the workflow code and also keep the logic in one place.

.github/workflows/BuildStatusDataRecorder/action.yaml
quic-shaksuma: > record workflow ?

Done

.github/workflows/BuildStatusDataRecorder/action.yaml
quic-shaksuma: > required and default ?

Removed

.github/workflows/BuildStatusDataRecorder/action.yaml
quic-shaksuma: > probably project also can be a parameter ?

Done

.github/workflows/BuildStatusDataRecorder/action.yaml
quic-shaksuma: > can we remove these ls -l and :'

ls is helpful as a debug info on repo tree state.
':' is needed for multi-line bash comment.

.github/workflows/BuildStatusDataRecorder/action.yaml
quic-shaksuma: > I would think this should do check_and_acquire_lock.
> 
> It will simplify the workflow code and also keep the logic in one place.

For current use-case, every time a "git push" is attempted, local repo tree should be in sync with remote and have same latest ref hash. If this does not happen, a git ref conflict happens and push is rejected.
The check for latest ref has to be done every time a push is to be done, e.g. for data.lock , for DB file update, etc.
A loop check this in case of conflict and resets the stale local tree/ref and proceeds with workflow steps until git push is successful.

.github/workflows/BuildStatusDataRecorder/action.yaml
quic-shaksuma: > Remove ls -l

Need for debug info

.github/workflows/BuildStatusDataRecorder/action.yaml
quic-shaksuma: > spell check workflow pls.

Done

.github/workflows/BuildStatusDataRecorder/action.yaml
quic-shaksuma: > git rm -f whereever applicable

Sure

PR #620, opened 2025-11-21
Emit memory reports to debug link time memory usage

### LINKED ISSUE #606: Emit Memory Utilization Reports for Linker Data Structures
Enhance `eld` to generate detailed memory utilization reports for linker data structures during the linking process. 

This will help developers understand memory allocation and detect inefficiencies or overflow conditions.

quic-seaswara:
This prints memory allocated by linker core data structures with a count of how many and total bytes used.

This will be followed up by improvements to report virtual memory / resident memory size.

Resolves #606

### DIFF
--- a/include/eld/BranchIsland/BranchIsland.h
+++ b/include/eld/BranchIsland/BranchIsland.h
@@ -34,6 +34,9 @@ class Relocation;
  */
 class BranchIsland {
 public:
+  static std::string getTypeName() { return "BranchIsland"; }
+
+public:
   typedef std::vector<Relocation *> RelocationListType;
   typedef RelocationListType::iterator reloc_iterator;
 
--- a/include/eld/Config/GeneralOptions.h
+++ b/include/eld/Config/GeneralOptions.h
@@ -44,6 +44,9 @@ class ZOption;
  */
 class GeneralOptions {
 public:
+  static std::string getTypeName() { return "GeneralOptions"; }
+
+public:
   typedef llvm::StringMap<std::string> SymbolRenameMap;
 
   typedef llvm::StringMap<uint64_t> AddressMapType;
@@ -834,6 +837,16 @@ public:
   void setTimingStatsFile(std::string StatsFile) {
     TimingStatsFile = StatsFile;
   }
+  //--------------------Memory Report--------------------------------
+  // --print-memory-report
+  bool showMemoryReport() const { return PrintMemoryReport; }
+
+  void setPrintMemoryReport() { PrintMemoryReport = true; }
+
+  // --emit-memory-report <file>
+  std::string getMemoryReportFile() const { return MemoryReportFile; }
+
+  void setMemoryReportFile(std::string F) { MemoryReportFile = F; }
   //--------------------Plugin Config--------------------------------
   void addPluginConfig(const std::string &Config) {
     PluginConfig.push_back(Config);
@@ -1233,6 +1246,7 @@ private:
   bool BExecuteOnly = false;              // --execute-only
   bool BPrintTimeStats = false;           // --print-stats
   bool BPrintAllUserPluginTimeStats = false;
+  bool PrintMemoryReport = false;         // --print-memory-report
   bool BDemangle = true;                  // --demangle-style
   bool ValidateArchOpts = false;          // check -mabi with backend
   bool DisableGuardForWeakUndefs = false; // hexagon specific option to
@@ -1265,7 +1279,8 @@ private:
   std::string Filter;
   std::string MapFile; // Mapfile
   std::string TarFile; // --reproduce output tarfile name
-  std::string TimingStatsFile;
+  std::string TimingStatsFile;           // --emit-timing-stats
+  std::string MemoryReportFile;          // --emit-memory-report
   std::string MappingFileName;           // --Mapping-file
   std::string MappingDumpFile;           // --dump-mapping-file
   std::string ResponseDumpFile;          // --dump-response-file
--- a/include/eld/Support/Memory.h
+++ b/include/eld/Support/Memory.h
@@ -30,35 +30,142 @@
 #include "llvm/Support/Allocator.h"
 #include "llvm/Support/PrettyStackTrace.h"
 #include "llvm/Support/StringSaver.h"
+#include "llvm/Support/raw_ostream.h"
+#include <atomic>
+#include <cstring> // For strstr, etc.
+#include <string>
+#include <string_view>
+#include <type_traits>
 #include <vector>
 
 namespace eld {
 
-// Use this arena if your object doesn't have a destructor.
 extern llvm::BumpPtrAllocator BAlloc;
 extern llvm::StringSaver Saver;
+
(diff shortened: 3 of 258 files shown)


include/eld/Config/GeneralOptions.h
```diff
@@ -43,6 +43,9 @@ class ZOption;
  *     - attribute of input files
  */
 class GeneralOptions {
+public:
+  static std::string getTypeName() { return "GeneralOptions"; }
```
partaror, line 47: Can we please return `std::string_view` / `llvm::StringRef` / `const char *` instead of `std::string` as these are more memory efficient?

include/eld/Config/GeneralOptions.h
```diff
@@ -834,6 +837,16 @@ class GeneralOptions {
...
+  //--------------------Memory Report--------------------------------
+  // --print-memory-report
+  bool showMemoryReport() const { return PrintMemoryReport; }
+
+  void setPrintMemoryReport() { PrintMemoryReport = true; }
+
+  // --emit-memory-report <file>
+  std::string getMemoryReportFile() const { return MemoryReportFile; }
+
+  void setMemoryReportFile(std::string F) { MemoryReportFile = F; }
```
partaror, line 849: I think we should try to use `llvm::StringRef` / `std::string_view` wherever we can as they avoid copying strings. What do you think about it?

```suggestion
  void setMemoryReportFile(llvm::StringRef F) { MemoryReportFile = F; }
```

include/eld/Config/GeneralOptions.h
```diff
@@ -834,6 +837,16 @@ class GeneralOptions {
...
     TimingStatsFile = StatsFile;
   }
+  //--------------------Memory Report--------------------------------
+  // --print-memory-report
+  bool showMemoryReport() const { return PrintMemoryReport; }
+
+  void setPrintMemoryReport() { PrintMemoryReport = true; }
+
+  // --emit-memory-report <file>
+  std::string getMemoryReportFile() const { return MemoryReportFile; }
```
partaror, line 847:
```suggestion
  llvm::StringRef getMemoryReportFile() const { return MemoryReportFile; }
```

include/eld/Config/GeneralOptions.h
```diff
@@ -1265,7 +1279,8 @@ class GeneralOptions {
   std::string Filter;
   std::string MapFile; // Mapfile
   std::string TarFile; // --reproduce output tarfile name
-  std::string TimingStatsFile;
+  std::string TimingStatsFile;           // --emit-timing-stats
+  std::string MemoryReportFile;          // --emit-memory-report
```
partaror, line 1283: Can we use one option `std::optional<MemoryReportFile>` instead of using two options `PrintMemoryReport` and `MemoryReportFile`?

include/eld/Support/Memory.h
```diff
@@ -30,36 +30,143 @@
...
+    auto &Alloc = getAlloc();
+    T *P = new (Alloc.Alloc.Allocate()) T(std::forward<U>(Args)...);
+    Alloc.onAllocate();
+    return P;
+  }
 };
 
-// Use this arena if your object has a destructor.
-// Your destructor will be invoked from freeArena().
 template <typename T, typename... U> T *make(U &&...Args) {
```
partaror, line 162: What do you think about adding the below macro for `make` (and slightly modifying `eld::make` definition) and directly using the type name supplied by the macro user to initialize the `TypeName` property of `SpecificAlloc`? I believe this will make the design significantly more maintainable and scalable.

```cpp
#define MAKE(Type, ...) \
  eld::make<Type>(#Type, __VA_ARGS__);

template<typename T, typename... U> T *make(llvm::StringRef TypeName, U &&... Args) {
  return ArenaForType<T>::create(TypeName, std::forward<T>(Args)...);
}
```

Now `ArenaForType` can directly use `TypeName` to initialize `SpecificAlloc`.

lib/LinkerWrapper/GnuLdDriver.cpp
```diff
@@ -149,6 +151,26 @@ bool GnuLdDriver::emitStats(eld::Module &M) const {
   return true;
 }
 
+bool GnuLdDriver::emitMemoryReport() const {
+  if (!Config.options().showMemoryReport())
+    return true;
+  std::string File = Config.options().getMemoryReportFile();
```
partaror, line 157:
```suggestion
  llvm::StringRef File = Config.options().getMemoryReportFile();
```

include/eld/BranchIsland/BranchIsland.h
```diff
@@ -33,6 +33,9 @@ class Relocation;
  *
  */
 class BranchIsland {
+public:
+  static std::string getTypeName() { return "BranchIsland"; }
```
quic-areg, line 37: Why can't we use macros to get these? Adding this to every single type feels needlessly complex and hard to scale.

include/eld/Support/Memory.h
```diff
@@ -30,36 +30,143 @@
...
+    return "parsing_failed";
+  return std::string(start, end - start);
+#elif defined(_MSC_VER)
+  const char *signature = __FUNCSIG__;
+  const char *start = strstr(signature, "parseTypeNameFromSignature<");
+  if (!start)
+    return "parsing_failed";
+  start += strlen("parseTypeNameFromSignature<");
+
+  // MSVC might add "class ", "struct ", etc.
```
quic-areg, line 92: Do we need this extra complexity for mscv?

include/eld/Config/GeneralOptions.h
```diff
@@ -834,6 +837,16 @@ class GeneralOptions {
   void setTimingStatsFile(std::string StatsFile) {
     TimingStatsFile = StatsFile;
   }
+  //--------------------Memory Report--------------------------------
```
quic-areg, line 840: is this feature useful to anyone other than an eld developer?

include/eld/Support/Memory.h
quic-seaswara: Yes.

include/eld/Config/GeneralOptions.h
quic-seaswara: Only eld developer.

quic-seaswara: I can make this option hidden if that is what you were hinting at.

include/eld/Support/Memory.h
quic-seaswara: i like this way better, though this will need every make<T>(X) to be changed to make(T, X)

include/eld/BranchIsland/BranchIsland.h
quic-seaswara: Sure. I think that is much better. Let me explore

include/eld/Support/Memory.h
partaror: Even currently, all make calls are of the form: `make<T>(X)`. We still have `T` in the make calls :).

PR #628, opened 2025-11-24
Support -plugin-opt=OX, -plugin-opt=-XXX, their aliases, and refactoring

quic-akaryaki:


quic-akaryaki:
Unfortunately, the cleanup of massive duplication in option processing could not get squeezed in this change. The problem is option tables are separate for each target, and each table includes general options. Therefore, general options cannot be parsed in a generic way.

quic-seaswara (approved):

PR #632, opened 2025-11-25
Mark MapStyleAll.test and SearchPathErrorMessage.test unsupported for windows

quic-garvgupt:
MapStyleAll.test is dependent on yaml module to be available on windows
builders otherwise it will fail. Therefore mark it unsupported for the
time being.
 
SearchPathErrorMessage.test checks for the non-existent library path
however, during path normalization in windowss`/a/b` path gets
normalized to `A:\b` which doesnt match the regex therefore mark it
unsupported as well for windows.

### DIFF
--- a/test/Common/standalone/Map/MapStyleAll/MapStyleAll.test
+++ b/test/Common/standalone/Map/MapStyleAll/MapStyleAll.test
@@ -2,6 +2,7 @@
 #BEGIN_COMMENT
 # Teach linker to emit multiple map files
 #END_COMMENT
+REQUIRES: yaml
 #START_TEST
 RUN: %clang %clangopts -c %p/Inputs/1.c -o %t1.1.o
 RUN: %link -MapStyle all %linkopts %t1.1.o -Map %t1.map -o %t2.out 2>&1
--- a/test/Common/standalone/SearchPathTest/SearchPathErrorMessage.test
+++ b/test/Common/standalone/SearchPathTest/SearchPathErrorMessage.test
@@ -2,6 +2,7 @@
 #BEGIN_COMMENT
 # This test check the error message when a search path can not be loaded.
 #END_COMMENT
+UNSUPPORTED: windows
 #START_TEST
 RUN: %clang %clangopts -c %p/Inputs/1.c -o %t.o
 RUN: %link %linkopts %t.o -L /a/b -o %t.test.out 2>&1 | %filecheck %s --check-prefix=MSG


test/Common/standalone/Map/MapStyleAll/MapStyleAll.test
```diff
@@ -2,6 +2,7 @@
 #BEGIN_COMMENT
 # Teach linker to emit multiple map files
 #END_COMMENT
+UNSUPPORTED: windows
```
quic-seaswara: You might want to say REQUIRES: yaml instead of unsupported

test/Common/standalone/SearchPathTest/SearchPathErrorMessage.test
```diff
@@ -2,6 +2,7 @@
 #BEGIN_COMMENT
 # This test check the error message when a search path can not be loaded.
 #END_COMMENT
+UNSUPPORTED: windows
```
quic-seaswara, line 5: can we fix the pattern ? 

we have tests that identify linux and windows paths. 

`{{[` -\(\)_A-Za-z0-9.\\/:]+}}`

quic-garvgupt: Pls correct me if I am wrong here but my understanding is that in this test, we are checking for paths ending with `/a/b`. If we remove it from regex altogether, wont it be too permissive?  I was thinking of adding regex something like this - {{([\(\)_A-Za-z0-9.\\/:]+a/b|A:/b)}}. This will either match the original pattern ending with `/a/b` (Linux case), or The exact Windows normalized path `A:/b`. Pls let me know your thoughts and apologies in advance if I am missing something trivial here

quic-seaswara: Ok I see, unsupported : windows is fine with me.

test/Common/standalone/Map/MapStyleAll/MapStyleAll.test
quic-garvgupt: Done

quic-garvgupt:
Can this PR be merged if everything looks good?

test/Common/standalone/SearchPathTest/SearchPathErrorMessage.test
quic-garvgupt: Done

PR #635, opened 2025-11-27
Move PAddr from ELFSection to OutputSectionEntry

quic-areg:
Save 8 bytes per section

### DIFF
--- a/lib/Readers/ELFSection.cpp
+++ b/lib/Readers/ELFSection.cpp
@@ -170,7 +170,7 @@ uint64_t ELFSection::pAddr() const {
 
 void ELFSection::setPaddr(size_t A) {
   OutputSectionEntry *OE = getOutputSection();
-  ASSERT(OE, "expected putput section");
+  ASSERT(OE, "expected output section");
   OE->setPaddr(A);
 }
 


lib/Readers/ELFSection.cpp
```diff
@@ -160,13 +161,25 @@ Fragment *ELFSection::getFirstFragmentInRule() const {
...
 }
 
+uint64_t ELFSection::pAddr() const {
+  OutputSectionEntry *OE = getOutputSection();
+  if (!OE)
+    return 0;
+  return OE->pAddr();
+}
+
+void ELFSection::setPaddr(size_t A) {
```
parth-07, line 171: It should be an assert to call `setPaddr` on input sections that do not have any corresponding `OutputSectionEntry`. Or perhaps, we should not expose this function through `ELFSection` and the users that want to set physical address should use:

```cpp
auto *OSE = S->getOutputSectionEntry();
ASSERT(OSE != nullptr, ...);
OSE->setPaddr(...);
```

parth-07 (changes requested):

lib/Readers/ELFSection.cpp
quic-areg: Added the assert

lib/Readers/ELFSection.cpp
```diff
@@ -160,13 +161,25 @@ Fragment *ELFSection::getFirstFragmentInRule() const {
...
+  OutputSectionEntry *OE = getOutputSection();
+  ASSERT(OE, "expected putput section");
```
quic-seaswara: nit : spell check

quic-areg:
```suggestion
  ASSERT(OE, "expected output section");
```

partaror:
```suggestion
  ASSERT(OE, "expected output section");
```

lib/Readers/ELFSection.cpp
```diff
@@ -160,13 +161,25 @@ Fragment *ELFSection::getFirstFragmentInRule() const {
   return nullptr;
 }
 
+uint64_t ELFSection::pAddr() const {
+  OutputSectionEntry *OE = getOutputSection();
+  if (!OE)
```
partaror, line 166: Can we please remove `setPaddr` and `pAddr` functions from `ELFSection` and enforce users to use these functions directly from `OutputSectionEntry`?

An output section can have multiple input sections within it. Input sections can have their own physical address as well, at least logically. Calling `setPaddr` on any of the input sections will seem like the call is changing the address of the particular `ELFSection` when instead it is changing the address of the output section of the `ELFSection`. I feel that `S->getOutputSectionEntry()->setPaddr(...)` will make things much more clear to the developer reading the code.

quic-areg: Sounds good. Can this be done in a follow-up?

partaror: Yes, follow-up sounds good!

PR #640, opened 2025-11-27
[x86_64] Add support for R_X86_64_TLSLD relocation

bhoumik03:
### Summary
Adds support for `R_X86_64_TLSLD` relocations in the x86_64 backend, enabling the Local Dynamic TLS access model for non-preemptible thread-local variables in shared objects.

The key benefit: when a function accesses multiple static TLS variables, they can share a single `__tls_get_addr` call by reusing the module base address.

### Changes
- Implement `scanLocalReloc()` handling for `R_X86_64_TLSLD`
- Add `getTLSModuleID()` to create and reuse single TLSLD GOT entry per module
- Emit one `R_X86_64_DTPMOD64` relocation per module (not per symbol)
- Override `sortRelocation()` to fix DTPMOD64 sorting among RELATIVE relocations
- Fix `hasSymInfo()` to handle relocations without symbol info


bhoumik03:
@parth-07 GitHub Actions workflows for this PR are awaiting approval.
Would appreciate your review.
Thanks!

partaror (approved):
LGTM.

Issue #534, opened 2025-11-06
Implement x86_64 target backend in ELD


quic-seaswara:
Summary
- Implement an x86_64 target backend for the ELD linker.
- Phase 1 targets static linking support (Linux + baremetal), with the goal of passing unit tests and producing working binaries.
- Future phases can extend to dynamic linking, symbol versioning, and broader GNU compatibility.

Goals
- Add a new x86_64 target backend that:
  - Builds in the ELD tree and integrates with ELD’s plugin/target architecture
  - Handles core ELF x86_64 relocations for static linking, including common PC-relative and absolute relocations
  - Supports basic TLS relocations needed for static/PIC cases
  - Produces detailed map files
  - Works with ELD’s linker script parser (GNU-compatible where applicable)
  - Normalizes target triples (e.g., x86_64-linux-gnu vs x86_64-unknown-linux-gnu)
  - Passes CI on Linux; optional Windows build check if feasible

Non-Goals (Phase 1)
- Full dynamic linking (GLIBC symbol versioning, PLT/GOT for shared libraries)
- Advanced GNU script edge cases beyond commonly used constructs
- Comprehensive Windows support (can be planned separately)

Scope & Deliverables
- New x86_64 target backend module(s) in ELD
- Core relocation handling (examples: 64-bit absolute, 32/32S, common PC-relative forms, minimal TLS)
- Section merging (.text, .data, .bss), alignment, start address handling aligned with triple/OS
- Linker script compatibility for common patterns; ensure parser handles “: ALIGN(…)” spacing expectations
- Detailed map-file output with target triple environment
- Reproduce tar functionality remains working when linking x86_64
- Docs: brief “Getting started” + “Supported features/limitations” + “Examples”
- Tests: LIT/unit tests for relocations, script handling, map output

User Scenarios
- Static link a simple C/C++ program targeting x86_64 (baremetal and Linux/musl)
- Link with a simple GNU-style linker script (SECTIONS, ALIGN, basic placement)
- Produce a readable map file confirming sections, symbols, and target triple environment

Work Breakdown (Checklist)
1) Environment & Planning
   - [x] Build ELD locally (Linux) and read target backend guidelines
   - [x] Confirm initial scope (static linking, minimal TLS, map files, scripts)
   - [ ] Pick initial triples to support (e.g., x86_64-unknown-linux-gnu, x86_64-none-elf)

2) Target Backend Skeleton
[...]

PR #641, opened 2025-11-27
[x86-64] Copy Relocations For Absolutely Referenced Data Symbols

### LINKED ISSUE #567: Support emission of copy relocations for x86-64 target
Currently, we do not support emitting copy relocations for x86-64 target. This is required when building a non-pie dynamic executable. 

Reproducer:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
#include <stdio.h>

extern int u;

int main() { 
  u = 11;
  printf("u: %d\n", u);
  return 0;
}
EOF

cat >u.c <<\EOF
int u;

int foo() { return u; }
EOF

musl-clang -o 1.o 1.c -c -ffunction-sections -fno-pic
musl-clang -o u.o u.c -c -ffunction-sections -fPIC
LD=eld musl-clang -o libu.eld.so u.o -shared

LD=eld musl-clang -o 1.eld.out 1.o libu.eld.so -no-pie
```

1.eld.out should have `R_X86_64_COPY` relocation for symbol `u`, but currently does not.

bhoumik03:
## Summary

This PR implements copy relocations (`R_X86_64_COPY`) for x86-64 in ELD, enabling proper linking of non-PIC executables that reference data symbols from shared libraries.
When a non-PIC executable references a data symbol from a shared library, the linker needs to resolve the reference without creating runtime relocations in read-only sections. The standard solution is to create a **copy relocation** that copies the symbol from the shared library into the executable's `.bss` section at load time.

## Changes
- Added logic in `scanGlobalReloc()` to detect when copy relocations are needed
- Calls `symbolNeedsCopyReloc()` to determine if a symbol requires copy relocation
- Handles both pointer-sized (`R_X86_64_64`) and truncated absolute relocations

Fixes #567 
Progress on #534


lib/Target/x86_64/x86_64Relocator.cpp
```diff
@@ -295,6 +312,39 @@ void x86_64Relocator::scanGlobalReloc(InputFile &pInputFile, Relocation &pReloc,
     }
     return;
   }
+
+  // Handle smaller absolute relocations similarly: if a dynamic relocation
+  // would be required, can use COPY reloc; otherwise need to error out as
+  // non-pointer-sized dynamic relocations should not be emitted
+  case llvm::ELF::R_X86_64_32:
```
partaror, line 319: Can you please add a test for smaller absolute relocations that uses copy relocation?

bhoumik03: The old version of the test created copy relocation against `R_X86_64_32S`.
I have extended the test for the case when a copy relocation needs to be created against `R_X86_64_64`.

bhoumik03: Copy relocations should be emitted for PC-relative imported-data relocations. See #669 for the reproducer.

PIE note: there’s a separate corner case in #669 where a PIE built from -fno-pic objects with PC-relative data refs may also require R_X86_64_COPY. @partaror — could you weigh in on whether eld should emit COPY for the PIE case or document/reject it?

parth-07: Hi @bhoumik03,

Thank you for creating the detailed issue. 

The corner case is interesting. We should definitely emit some warning if a user is trying to create a PIE from no-pic objects and we are able to properly detect this. This is erroneous / undefined behaviour. However, this should not be limited to this particular copy relocation case -- whenever we can detect such mismatch, we should report it.

PR #651, opened 2025-12-01
[ELFSection] Move DependentSections to side table

quic-areg:
DependentSections was a per-section vector only
used via the pluginAPI. Move it to a ObjectFile 
owned Map so sections with no dependencies pay nothing.

Saves 16 bytes per section.

### DIFF
--- a/include/eld/Input/ObjectFile.h
+++ b/include/eld/Input/ObjectFile.h
@@ -8,7 +8,9 @@
 #define ELD_INPUT_OBJECTFILE_H
 
 #include "eld/Input/InputFile.h"
+#include "llvm/ADT/DenseMap.h"
 #include "llvm/ADT/Hashing.h"
+#include "llvm/ADT/SmallVector.h"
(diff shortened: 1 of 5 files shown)


include/eld/Core/Module.h
```diff
@@ -739,6 +743,8 @@ class Module {
   llvm::StringSet<> OutputSectDescNameSet;
   llvm::SmallVector<const VersionScript *> VersionScripts;
   llvm::DenseMap<Fragment *, uint64_t> FragmentPaddingValues;
+  llvm::DenseMap<const ELFSection *, llvm::SmallVector<ELFSection *, 2>>
+      DependentSectionsMap;
```
quic-seaswara: Can we have these in the InputFile instead of Module ?

quic-areg: Moved it to ELFObjectFile

quic-seaswara (approved):

PR #653, opened 2025-12-02
[ArchiveFile] Store symbols by value and avoid storing strings.

quic-areg:
ArchiveFile::Symbol is now an aggregate held by
value in the symbol table and we record name offset/size rather than owning a std::string. This reduces the size from 40 bytes per symbol to 16 bytes per symbol.

Measured improvements linking eld (debug) with eld (release) with and without this patch:

- Wall clock time 5.88s to 4.13s  
- Max RSS from ~1.8 GB to ~1.5 GB

### DIFF
--- a/include/eld/Input/ArchiveFile.h
+++ b/include/eld/Input/ArchiveFile.h
@@ -25,26 +25,25 @@ struct ArchiveFileInfo;
 class DiagnosticPrinter;
 class ArchiveFile : public InputFile {
 public:
-  // -----------Declare Types ----------------
   struct Symbol {
-    enum SymbolStatus { Include, Exclude, Unknown };
+    enum SymbolStatus : uint8_t { Include, Exclude, Unknown };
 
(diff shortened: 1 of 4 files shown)


include/eld/Input/ArchiveFile.h
```diff
@@ -25,26 +25,25 @@ struct ArchiveFileInfo;
...
+
+    enum SymbolType : uint8_t {
+      NoType,
+      DefineData,
+      DefineFunction,
+      Common,
+      Weak
+    };
+
+    uint32_t NameOffset;
```
parth-07, line 39: What do you think about storing the name as `llvm::StringRef` instead of explicitly storing both the name offset and the name size?

quic-areg: That would double the size so I would prefer to not do it unless there is a real performance improvement, and I think there is not.

partaror: Oh, I see. Thank you for the explanation.

PR #656, opened 2025-12-05
Update CHECK_LINK_STATES to account for ActBefore<LinkState> states

### LINKED ISSUE #655: `AddBefore<LinkState>` states are missing in the `CHECK_LINK_STATES` and `Module::LinkState`
`AddBefore<LinkState>` states are missing in the `CHECK_LINK_STATES` checks and `Module:LinkState`. We should add them there for consistency.

parth-07:
This commit updates CHECK_LINK_STATE functionality in the plugin framework to account for the ActBefore<LinkState> states. Until now, these link states did not exist in the linker and thus the plugin framework could not have checks for these states.

Resolves #655

### DIFF
--- a/include/eld/Core/LinkState.h
+++ b/include/eld/Core/LinkState.h
@@ -30,10 +30,14 @@ namespace eld {
 enum LinkState : uint8_t {
   Unknown,
   Initializing,
+  ActBeforeRuleMatching,
   BeforeLayout,
+  ActBeforeSectionMerging,
   CreatingSections,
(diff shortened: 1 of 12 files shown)


parth-07:
> is this a change that affects backwards compatibility ?

No, this change does not affect backwards compatibility. If it does by any chance, then it is a bug in this change.

quic-seaswara:
Can  you pls verify this ? We dont want to break old plugins.

PR #670, opened 2025-12-21
Add initial support for creating executables that use symbol versioning

parth-07:
This commit adds primitive support for creating executables that use
shared libraries with symbol versioning information.

### DIFF
--- a/include/eld/Fragment/GNUVerNeedFragment.h
+++ b/include/eld/Fragment/GNUVerNeedFragment.h
@@ -1,0 +1,72 @@
+//===- GNUVerNeedFragment.h----------------------------------------------------===//
+// Part of the eld Project, under the BSD License
+// See https://github.com/qualcomm/eld/LICENSE.txt for license information.
+// SPDX-License-Identifier: BSD-3-Clause
+//===--------------------------------------------------------------------------===//
+//
+//                     The MCLinker Project
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===--------------------------------------------------------------------------===//
+
+#ifndef ELD_FRAGMENT_GNUVERNEEDFRAGMENT_H
+#define ELD_FRAGMENT_GNUVERNEEDFRAGMENT_H
+
+#include "eld/Fragment/Fragment.h"
+#include "llvm/ADT/StringRef.h"
+
+namespace eld {
+class DiagnosticEngine;
+class ELFSection;
+class InputFile;
+class ELFFileFormat;
+
+/** \class VerNeedFragment
+ *  \brief VerNeedFragment is a kind of Fragment containing input memory region
+ */
+// Region fragment expression
+class GNUVerNeedFragment : public Fragment {
+public:
+  GNUVerNeedFragment(ELFSection *S);
+
+  static bool classof(const Fragment *F) {
+    return F->getKind() == Fragment::Type::GNUVerNeed;
+  }
+
+  template <class ELFT>
+  eld::Expected<void>
+  computeVersionNeeds(const std::vector<InputFile *> &DynamicObjectFiles,
+                      ELFFileFormat *FileFormat, DiagnosticEngine &DE);
+
+  eld::Expected<void> emit(MemoryRegion &Mr, Module &M) override;
+
+  size_t size() const override;
+
+  size_t getNeedCount() { return VersionNeeds.size(); }
+
+private:
+  template <class ELFT> eld::Expected<void> emitImpl(uint8_t *Buf, Module &M);
+
+public:
+  struct VernAuxInfo {
+    uint32_t VersionNameOffset = 0;
+    uint32_t VersionID = 0;
+    uint32_t VersionNameHash = 0;
+  };
+
+  struct VerNeedInfo {
+    uint32_t SONameOffset = 0;
+    std::vector<VernAuxInfo> Vernauxs;
+  };
+
+protected:
+  std::vector<VerNeedInfo> VersionNeeds;
+  size_t VerNeedEntrySize = 0;
+  size_t VernAuxEntrySize = 0;
+};
+
+} // namespace eld
+
+#endif
--- a/include/eld/Input/ELFDynObjectFile.h
+++ b/include/eld/Input/ELFDynObjectFile.h
@@ -35,8 +35,130 @@ public:
 
   std::string getFallbackSOName() const;
 
+#ifdef ELD_ENABLE_SYMBOL_VERSIONING
+  void setVerDefSection(ELFSection *S) { VerDefSection = S; }
+
+  ELFSection *getVerDefSection() const { return VerDefSection; }
+
+  void setVerNeedSection(ELFSection *S) { VerNeedSection = S; }
+
+  ELFSection *getVerNeedSection() const { return VerNeedSection; }
+
+  void setVerSymSection(ELFSection *S) { VerSymSection = S; }
+
+  ELFSection *getVerSymSection() const { return VerSymSection; }
+
+  // Cache parsed Verdef entries indexed by vd_ndx.
+  void setVerDefs(std::vector<const void *> VDefs) { VerDefs = VDefs; }
+
+  // Verdef table indexed by version index (vd_ndx).
+  const std::vector<const void *> &getVerDefs() const { return VerDefs; }
+
+  // Cache parsed Vernaux name offsets for needed versions indexed by vna_other.
+  void setVerNeeds(std::vector<uint32_t> VNeeds) { VerNeeds = VNeeds; }
+
+  // Verneed name offsets indexed by version index (vna_other & VERSYM_VERSION).
+  const std::vector<uint32_t> &getVerNeeds() const { return VerNeeds; }
+
+  // Cache parsed versym entries (vs_index) from the input DSO.
+  void setVerSyms(std::vector<uint16_t> VSyms) { VerSyms = VSyms; }
+
+  // Raw versym entries (vs_index) indexed by dynsym index.
+  const std::vector<uint16_t> &getVerSyms() const { return VerSyms; }
+
+  // Record the input DSO's .dynstr section used by verdef/verneed names.
+  void setDynStrTabSection(ELFSection *S) { DynStrTabSection = S; }
+
+  // Returns the contents of the input DSO's .dynstr (empty if missing).
+  llvm::StringRef getDynStringTable() const;
+
+  template <class ELFT>
+  // Returns the version name associated with a dynsym entry.
+  // For VER_NDX_LOCAL/VER_NDX_GLOBAL this returns an empty StringRef.
+  llvm::StringRef getSymbolVersionName(uint32_t SymIdx,
+                                       ResolveInfo::Desc SymDesc) const {
+    uint16_t SymVerID = getSymbolVersionID(SymIdx);
+    llvm::StringRef VerName;
+    if (SymVerID == llvm::ELF::VER_NDX_LOCAL ||
+        SymVerID == llvm::ELF::VER_NDX_GLOBAL)
+      return VerName;
+    if (SymDesc == ResolveInfo::Desc::Undefined)
+      VerName = getDynStringTable().data() + VerNeeds[SymVerID];
+    else {
+      auto VerNameOffset =
+          reinterpret_cast<const typename ELFT::Verdef *>(VerDefs[SymVerID])
+              ->getAux()
+              ->vda_name;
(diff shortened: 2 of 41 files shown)


lib/Object/ObjectLinker.cpp
```diff
@@ -1574,6 +1574,10 @@ bool ObjectLinker::addSymbolToOutput(const ResolveInfo &PInfo) const {
       ThisModule->hasWrapReference(PInfo.name()) && ResolvedOrigin->isBitcode())
     return false;
 
+#ifdef ELD_ENABLE_SYMBOL_VERSIONING
+  if (PInfo.isDyn() && PInfo.outSymbol() && PInfo.outSymbol()->shouldIgnore())
+    return false;
```
quic-seaswara, line 1610: Why do we do this ?

include/eld/Input/ELFDynObjectFile.h
```diff
@@ -35,8 +35,110 @@ class ELFDynObjectFile : public ELFFileBase {
...
+
+  const std::vector<uint32_t> &getVerNeeds() const { return VerNeeds; }
+
+  void setVerSyms(std::vector<uint16_t> VSyms) { VerSyms = VSyms; }
+
+  const std::vector<uint16_t> &getVerSyms() const { return VerSyms; }
+
+  void setDynStrTabSection(ELFSection *S) { DynStrTabSection = S; }
+
+  llvm::StringRef getDynStringTable() const;
```
quic-seaswara, line 73: Move it out of symbol versioning for DynStrTabSection and getDynStringTable

include/eld/Input/ELFDynObjectFile.h
```diff
@@ -35,8 +35,110 @@ class ELFDynObjectFile : public ELFFileBase {
...
+
+  template <class ELFT>
+  llvm::StringRef getSymbolVersionName(uint32_t SymIdx,
+                                       ResolveInfo::Desc SymDesc) const {
```
quic-seaswara, line 79: can we pass in ResolveInfo here ?

include/eld/Input/ELFDynObjectFile.h
```diff
@@ -35,8 +35,110 @@ class ELFDynObjectFile : public ELFFileBase {
...
+    if (SymVerID == llvm::ELF::VER_NDX_LOCAL ||
+        SymVerID == llvm::ELF::VER_NDX_GLOBAL)
+      return VerName;
+    if (SymDesc == ResolveInfo::Desc::Undefined)
+      VerName = getDynStringTable().data() + VerNeeds[SymVerID];
+    else {
+      auto VerNameOffset =
+          reinterpret_cast<const typename ELFT::Verdef *>(VerDefs[SymVerID])
+              ->getAux()
+              ->vda_name;
```
quic-seaswara, line 91: Is there a function that we can use instead ?

include/eld/Fragment/GNUVerNeedFragment.h
```diff
@@ -0,0 +1,72 @@
...
+                      ELFFileFormat *FileFormat, DiagnosticEngine &DE);
+
+  eld::Expected<void> emit(MemoryRegion &Mr, Module &M) override;
+
+  size_t size() const override;
+
+  template <class ELFT>
+  eld::Expected<void> emitImpl(uint8_t *Buf, Module &M);
+
+  size_t needCount() { return VersionNeeds.size(); }
```
quic-seaswara: getNeedCount

include/eld/Fragment/GNUVerNeedFragment.h
```diff
@@ -0,0 +1,72 @@
...
+  eld::Expected<void>
+  computeVersionNeeds(const std::vector<InputFile *> &DynamicObjectFiles,
+                      ELFFileFormat *FileFormat, DiagnosticEngine &DE);
+
+  eld::Expected<void> emit(MemoryRegion &Mr, Module &M) override;
+
+  size_t size() const override;
+
+  template <class ELFT>
+  eld::Expected<void> emitImpl(uint8_t *Buf, Module &M);
```
quic-seaswara: This could be private ?

include/eld/Input/ELFDynObjectFile.h
```diff
@@ -35,8 +35,110 @@ class ELFDynObjectFile : public ELFFileBase {
...
+
+  ELFSection *getVerNeedSection() const { return VerNeedSection; }
+
+  void setVerSymSection(ELFSection *S) { VerSymSection = S; }
+
+  ELFSection *getVerSymSection() const { return VerSymSection; }
+
+  void setVerDefs(std::vector<const void *> VDefs) { VerDefs = VDefs; }
+
+  const std::vector<const void *> &getVerDefs() const { return VerDefs; }
```
quic-seaswara, line 55: do we want to have const std::vector, prefer std::vector

include/eld/Input/ELFDynObjectFile.h
```diff
@@ -35,8 +35,110 @@ class ELFDynObjectFile : public ELFFileBase {
...
+
+  void setDynStrTabSection(ELFSection *S) { DynStrTabSection = S; }
+
+  llvm::StringRef getDynStringTable() const;
+
+  template <class ELFT>
+  llvm::StringRef getSymbolVersionName(uint32_t SymIdx,
+                                       ResolveInfo::Desc SymDesc) const {
+    uint16_t SymVerID = getSymbolVersionID(SymIdx);
+    llvm::StringRef VerName;
```
quic-seaswara, line 81: uninitialized VerName

include/eld/Input/ELFDynObjectFile.h
```diff
@@ -35,8 +35,110 @@ class ELFDynObjectFile : public ELFFileBase {
...
+              ->getAux()
+              ->vda_name;
+      VerName = getDynStringTable().data() + VerNameOffset;
+    }
+    return VerName;
+  }
+
+  uint16_t getSymbolVersionID(uint32_t SymIdx) const {
+    assert(SymIdx < VerSyms.size() && "Invalid SymIdx");
+    return VerSyms[SymIdx] & ~llvm::ELF::VERSYM_HIDDEN;
```
quic-seaswara: Can we have a getRawSymbolVersionID function ? 

Should this be returning (VerSyms[SymIdx] & VERSYM_VERSION)

include/eld/Input/ELFDynObjectFile.h
```diff
@@ -35,8 +35,110 @@ class ELFDynObjectFile : public ELFFileBase {
...
+  }
+
+  bool isDefaultVersionedSymbol(uint32_t SymIdx) const {
```
quic-seaswara, line 104: is this output symbol index or input ?

include/eld/Input/ELFDynObjectFile.h
```diff
@@ -35,8 +35,110 @@ class ELFDynObjectFile : public ELFFileBase {
...
+    assert(SymIdx < VerSyms.size() && "Invalid SymIdx");
+    return VerSyms[SymIdx] == getSymbolVersionID(SymIdx);
+  }
+
+  uint16_t getOutputVernAuxID(uint16_t InputVerID) {
```
quic-seaswara: const function

include/eld/Input/ELFDynObjectFile.h
```diff
@@ -35,8 +35,110 @@ class ELFDynObjectFile : public ELFFileBase {
...
+    if (InputVerID >= OutputVernAuxIDMap.size())
+      OutputVernAuxIDMap.resize(InputVerID + 1, 0);
+    return OutputVernAuxIDMap[InputVerID];
```
quic-seaswara: Dont use [ as it will create an unnecessary bucket, find and return the value

include/eld/Input/ELFDynObjectFile.h
```diff
@@ -35,8 +35,110 @@ class ELFDynObjectFile : public ELFFileBase {
...
+  }
+
+  void setOutputVernAuxID(uint16_t InputVerID, uint16_t OutputVerID) {
```
quic-seaswara: Can we change Vern to Version everywhere ?

include/eld/Input/ELFDynObjectFile.h
```diff
@@ -35,8 +35,110 @@ class ELFDynObjectFile : public ELFFileBase {
...
+    if (InputVerID >= OutputVernAuxIDMap.size())
+      OutputVernAuxIDMap.resize(InputVerID + 1, 0);
+    OutputVernAuxIDMap[InputVerID] = OutputVerID;
+  }
+
+  const std::vector<uint16_t> &getOutputVernAuxIDMap() const {
```
quic-seaswara, line 125: Adding comments around functions would be helpful to see how it needs to be used

include/eld/Input/ELFDynObjectFile.h
```diff
@@ -35,8 +35,110 @@ class ELFDynObjectFile : public ELFFileBase {
...
+    return OutputVernAuxIDMap;
+  }
+
+  const void *getVerDef(uint16_t InputVerID) {
```
quic-seaswara: const function

include/eld/Input/ELFDynObjectFile.h
parth-07: Can we move it out of `ELD_ENABLE_SYMBOL_VERSIONING` block when it is required? Currently, we only need these functions for symbol versioning.

include/eld/Input/ELFDynObjectFile.h
parth-07: No, this function is called before the ResolveInfo for the symbol is properly created. And also, here we are referring to the input symbol. With ResolveInfo, we refer to the output symbol.

include/eld/Input/ELFDynObjectFile.h
parth-07: This functionality is directly from the LLVM. `vda_name` member is as specified in the official `Elf_Verdaux` data structure. Why would a function be better here?

include/eld/Input/ELFDynObjectFile.h
parth-07: It cannot be a `const` function, it may need to resize an internal member.

include/eld/Input/ELFDynObjectFile.h
parth-07: The line numbers 98-99 ensures that `[` will not create a new bucket.

lib/Object/ObjectLinker.cpp
parth-07: When for a default versioned symbol such as `foo@@V1`, both the canonical `foo@V1` and the non-canonical `foo` symbols are selected by the symbol resolution, then we need to discard the non-canonical symbol so that we do not have duplicates in the symbol table. The symbols that were resolving to the non-canonical `foo` are made to resolve to the canonical symbol `foo@V1`. All of this happens during the symbol normalization (`IRBuilder::normalizeSymbols`).

include/eld/Input/ELFDynObjectFile.h
parth-07: I have added the comments for some of the relevant functions.

include/eld/Fragment/GNUVerNeedFragment.h
parth-07: I have made it private.

include/eld/Input/ELFDynObjectFile.h
parth-07: Why should we prefer `std::vector` over `const std::vector`? `const` ensures that we do not accidentally modify values where the structure should never be modified.

include/eld/Input/ELFDynObjectFile.h
parth-07: `llvm::StringRef` is default-initialized to an empty StringRef object. This is intentional here because not all symbols have version names.

include/eld/Input/ELFDynObjectFile.h
parth-07: `ELFDynObjectFile` always handles input symbol indices, as it has no knowledge of output symbol indices.

include/eld/Input/ELFDynObjectFile.h
parth-07: > getRawSymbolVersionID

What would be the difference in raw symbol version ID and non-raw symbol version ID?

> Should this be returning (VerSyms[SymIdx] & VERSYM_VERSION)

Both this and the current `VerSyms[SymIdx] & ~llvm::ELF::VERSYM_HIDDEN` have the same exact behavior. I have changed to `VERSYM_VERSION` variant as it is more readable.

Issue #430, opened 2025-10-01
Assignments after `SECTIONS` command should be evaluated after evaluating assignments in `SECTIONS` command


parth-07:
We evaluate assignments after `SECTIONS` command before evaluating assignments in `SECTIONS` command. Assignments should be evaluated in the order in which they are visible to the linker.

Reproducer:

```bash
cat > 1.c << '!'
int foo()  { return 1; }

int bar() { return foo(); }
!

cat > script.t << '!'
v = 0xc;
SECTIONS {
  u = v; 
  FOO (0x1000): {
    start = .;
    *(.text.foo)
    end = .;
  }
  v = 0xa;
}
!


LDs=(ld.eld ld.lld riscv64-unknown-elf-ld.bfd)
SFs=(eld lld bfd)

clang -o 1.o --target=riscv64-unknown-elf 1.c -c -ffunction-sections

for i in "${!SFs[@]}"; do
  ${LDs[$i]} -o 1.${SFs[$i]}.out 1.o -T script.t -Map 1.${SFs[$i]}.map.txt
done
llvm-readelf -s 1.eld.out | grep -E 'ABS (u|v)'
```

**Output**
```
    10: 000000000000000a     0 NOTYPE  GLOBAL DEFAULT   ABS u
    11: 000000000000000c     0 NOTYPE  GLOBAL DEFAULT   ABS v
```

The output is incorrect, `u` should have value `0xc`.

parth-07:
Another more complex test-case for the same issue that must be resolved by the solution:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
int foo() { return 1; }
EOF

cat >script.t <<\EOF
SECTIONS {
  a = u;
}
u = 0x3; 
SECTIONS {
  b = u;
}
u = 0x7;
EOF

clang -o 1.o -target x86_64-linux-gnu 1.c -c
LDs=(ld.eld ld.lld ld.bfd)
SFs=(eld lld bfd)

for i in "${!SFs[@]}"; do
  ${LDs[$i]} -o 1.${SFs[$i]}.out 1.o -T script.t
done
```

Here we have a symbol assignment in-between `SECTIONS` command. The symbol assignment should be evaluated in-between `SECTIONS` command assignments as well.

PR #679, opened 2025-12-29
Add source assignment in symbol nodes

parth-07:
This commit changes how symbol expression nodes gets evaluated for
the absolute symbols. The key goal is to improve the layout convergence
when forward references is involved.
direction to improve the layout convergence.

When the linker script contains forward references, then we may need to
re-evaluate the linker script and the layout multiple times to reach
convergence. For example:

```
SECTIONS {
  u = v; // A1
  .foo (u) : { *(.text.foo) }
  v = 0x100; // A2
  .bar (v) : { *(.text.bar) }
  v = 0x300; // A3
}
```

The above linker script contains 3 assignments: A1, A2, and A3. The link
needs more than one pass of layout computation for reaching
convergence, because in the 1st pass, the assignment A1 is evaluted incorrectly
because `v` has not been assigned a value yet. To properly assign
layout here, the linker will recompute the layout iteratively. Let's
call this kind of layout pass as **FullLayoutPass**.

We also need to stop the layout pass in between and recompute the layout
from the beginning whenever some initial assumption changes such as a
new segment needs to be inserted. Let's call this kind of layout pass as
**InBetweenLayoutPass**.

There is an important difference between these two types of layout
recomputation. In the first type, that is, layout recomputation to
achieve layout convergence, we must use the symbol values from the
previous pass. However, in the second type, that is, layout recomputation
due to an initial assumption change such as a new segment needs to be inserted,
we must reset the symbol values before recomputing the layout, otherwise
we risk using stale/incorrect symbol values. Let's see why:

```
SECTIONS {
  u = v; // A1
  .foo : { *(.text.foo) }
  v = 0x100; // A2
  .data : { *(.text.data) }}
  v = 0x300; // A3
}
```

In the first evaluation of A1, u is assigned the value `0`.
On encountering '.data' output section, the layout is recomputed, and
this time when A1 is recomputed, u is assigned the value `0x100`
because A2 was evaluated in the last pass. The value `0x100` is
incorrect for `u` because the value of `v` that is to be used in A1 must
come from the last evaluation of `v`, that is, A3. Hence, reusing symbol values
from the last pass in this case leads to incorrect layout.

**We cannot add logic for when to reuse symbol values and when to
reset them because the two types of layout passes
(FullLayoutPass and InBetweenLayoutPass) have contradicting
requirements -- one requires preserving symbol values, and the other
requires resetting them, and InBetweenLayoutPass is invoked in-between
the FullLayoutPass.**

Adding a source assignment node with the symbol expression node makes
the recomputation simpler by obviating the need to reset the symbol
values in both the cases. **The key idea is that the value of a
symbol node of a script symbol is not the value of the corresponding
symbol, but instead is the result of the last assignment node for that
symbol.**

Adding the source assignment with a symbol node has additional benefits
as well:

1) It makes it easier to determine the culprit / closest assignment to
   use in diagnostics when a symbol value is not converging.

2) It makes it easier to add heuristics such as constant expression
   evaluation to speed up the layout convergence. For example:

```
u = v;
.foo (u) : { *(.text.foo) }
v = 0x2000;
```

  In this case, if the symbol v encodes the source assignment, then we
  can easily add heuristic to determine if `v` source assignment can be
  evaluated early.

3) Selectively recompute only those assignment nodes which needs to be recomputed.
   If we reset symbol values, then all the assignment nodes always needs
   to be recomputed in each layout pass.

Resolves #1037

### DIFF
--- a/include/eld/Target/GNULDBackend.h
+++ b/include/eld/Target/GNULDBackend.h
@@ -23,6 +23,7 @@
 #include "eld/Readers/ELFSection.h"
 #include "eld/Readers/SymDefReader.h"
 #include "eld/Script/Assignment.h"
+#include "eld/Script/Expression.h"
 #include "eld/Script/VersionScript.h"
 #include "eld/SymbolResolver/ResolveInfo.h"
 #include "eld/Target/ELFSegment.h"
@@ -902,6 +903,17 @@ public:
   }
 #endif
 
+  const Assignment *getLatestAssignment(llvm::StringRef SymName) {
+    auto it = SymbolNameToLatestAssignment.find(SymName);
+    if (it != SymbolNameToLatestAssignment.end())
+      return it->getValue();
+    return nullptr;
+  }
+
+  void updateLatestAssignment(llvm::StringRef SymName, const Assignment *A) {
+    SymbolNameToLatestAssignment[SymName] = A;
+  }
+
 protected:
   virtual int numReservedSegments() const { return m_NumReservedSegments; }
 
@@ -1230,6 +1242,8 @@ protected:
   GNUVerNeedFragment *GNUVerNeedFrag = nullptr;
   std::unordered_map<const ResolveInfo *, uint16_t> OutputVersionIDs;
 #endif
+
+  llvm::StringMap<const Assignment *> SymbolNameToLatestAssignment;
 };
 
 } // namespace eld
--- a/lib/Script/Assignment.cpp
+++ b/lib/Script/Assignment.cpp
@@ -224,6 +224,9 @@ bool Assignment::assign(Module &CurModule, const ELFSection *Section) {
     ThisSymbol->setScriptValueDefined();
   }
 
+  auto &Backend = CurModule.getBackend();
+  Backend.updateLatestAssignment(Name, this);
+
   if (CurModule.getPrinter()->traceAssignments())
     trace(llvm::outs());
   return true;
--- a/lib/Script/Expression.cpp
+++ b/lib/Script/Expression.cpp
@@ -9,6 +9,7 @@
 #include "eld/Core/LinkerScript.h"
 #include "eld/Core/Module.h"
 #include "eld/Readers/ELFSection.h"
+#include "eld/Script/Assignment.h"
 #include "eld/Script/ScriptFile.h"
 #include "eld/Support/MsgHandling.h"
 #include "eld/Support/Utils.h"
@@ -135,7 +136,6 @@ bool Symbol::hasDot() const {
 }
 
 eld::Expected<uint64_t> Symbol::evalImpl() {
-
   if (!ThisSymbol)
     ThisSymbol = ThisModule.getNamePool().findSymbol(Name);
 
@@ -144,6 +144,15 @@ eld::Expected<uint64_t> Symbol::evalImpl() {
     return std::make_unique<plugin::DiagnosticEntry>(
         Diag::undefined_symbol_in_linker_script,
         std::vector<std::string>{Name});
+
+  if (!SourceAssignment &&
+      ThisSymbol->resolveInfo()->outSymbol()->scriptDefined()) {
+    auto &Backend = ThisModule.getBackend();
+    const auto *A = Backend.getLatestAssignment(Name);
+    if (!A)
+      A = ThisModule.getAssignmentForSymbol(Name);
+    SourceAssignment = A;
+  }
 
   if (ThisSymbol->hasFragRef() && !ThisSymbol->shouldIgnore()) {
     FragmentRef *FragRef = ThisSymbol->fragRef();
@@ -154,6 +163,10 @@ eld::Expected<uint64_t> Symbol::evalImpl() {
            "using a symbol that points to a non allocatable section!");
     return Section->addr() + FragRef->getOutputOffset(ThisModule);
   }
+  if (hasDot())
+    return ThisSymbol->value();
+  if (SourceAssignment)
+    return SourceAssignment->value();
   return ThisSymbol->value();
 }
 
--- a/lib/Target/GNULDBackend.cpp
+++ b/lib/Target/GNULDBackend.cpp
@@ -3272,6 +3272,7 @@ bool GNULDBackend::layout() {
     return false;
   }
 
+  // FIXME: Adding more symbols this late can cause layout issues.
   {
     eld::RegisterTimer T("Define Magic Symbols", "Establish Layout",
                          m_Module.getConfig().options().printTimingStats());
--- a/include/eld/Script/Expression.h
+++ b/include/eld/Script/Expression.h
@@ -337,6 +337,7 @@ private:
   Expression *getRightExpression() const override { return nullptr; }
 
   mutable LDSymbol *ThisSymbol = nullptr;
+  const Assignment *SourceAssignment = nullptr;
 };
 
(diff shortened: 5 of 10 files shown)


lib/Script/Expression.cpp
```diff
@@ -133,10 +134,17 @@ bool Symbol::hasDot() const {
 }
 
 eld::Expected<uint64_t> Symbol::evalImpl() {
-
   if (!ThisSymbol)
     ThisSymbol = ThisModule.getNamePool().findSymbol(Name);
 
+  if (!SourceAssignment && ThisSymbol->resolveInfo()->isAbsolute()) {
```
quic-seaswara: should this be isScriptDefined() ?

lib/Target/GNULDBackend.cpp
```diff
@@ -3100,6 +3100,7 @@ bool GNULDBackend::layout() {
     return false;
   }
 
+  // FIXME: Adding more symbols this late can cause layout issues.
```
quic-seaswara, line 3275: magic symbols are defined before layout, why would this cause issue ?

lib/Script/Assignment.cpp
```diff
@@ -222,6 +222,9 @@ bool Assignment::assign(Module &CurModule, const ELFSection *Section) {
     ThisSymbol->setScriptValueDefined();
   }
 
+  auto &Backend = CurModule.getBackend();
+  Backend.updateLatestAssignment(Name, this);
```
quic-seaswara, line 228: wont this cause issues if the symbol assignment is PROVIDE ?

include/eld/Target/GNULDBackend.h
```diff
@@ -824,6 +825,17 @@ class GNULDBackend {
...
+  const Assignment *getLatestAssignment(llvm::StringRef SymName) {
+    auto it = SymbolNameToLatestAssignment.find(SymName);
+    if (it != SymbolNameToLatestAssignment.end())
+      return it->getValue();
+    return nullptr;
+  }
+
+  void updateLatestAssignment(llvm::StringRef SymName, const Assignment *A) {
+    SymbolNameToLatestAssignment[SymName] = A;
+  }
```
quic-seaswara, line 915: Why backend ? can we store this in LinkerScript ?

quic-seaswara (commented):
I could not follow which case does this patch handle ?

include/eld/Target/GNULDBackend.h
parth-07: They are used during the layout computation. Currently, the entire logic for layout computation is in GNULDBackend.

lib/Script/Assignment.cpp
parth-07: No, this does not change how PROVIDE assignments are evaluated. Why do you think it would cause issues with PROVIDE?

lib/Target/GNULDBackend.cpp
parth-07: We are defining these symbols after the layout is performed, right? The layout is performed above this, at line 3079, in the `relax` function call.

lib/Script/Expression.cpp
parth-07: I think `isScriptDefined` would work as well. Is there any case where an absolute symbol may not be script defined?

include/eld/Target/GNULDBackend.h
quic-seaswara: The latest assignment would then become PROVIDE which will not be provided because of the preceding assignment ?

parth-07: `Assignment::assign` for a PROVIDE assignment is only called if the assignment is indeed provided.

lib/Script/Expression.cpp
quic-seaswara: You can use assembly using .set

parth-07: Thank you for catching this. I have changed `isAbsolute()` to `isScriptDefined()` here.

parth-07:
@quic-areg @quic-seaswara Can you please review the patch?

parth-07:
clang-format check is incorrectly failing here. More information here: https://github.com/qualcomm/eld/pull/1077#issuecomment-4294965261

parth-07:
> I could not follow which case does this patch handle ?

I have now linked the issue in the PR description that describes the case that this PR handles: https://github.com/qualcomm/eld/issues/1037

quic-seaswara (approved):

quic-areg:
This will crash :

```
cat <<\! > 1.c
int foo() { return 1; }
int val = 3;
!

cat <<\! > ref.s
.global bar
.text
bar: ret
.data
.global ref_a
ref_a: .quad a
!

cat <<\! > 1.t
SECTIONS {
  .text : { *(.text*) }
  .data : { *(.data*) }
  .comment : { *(.comment) }
  .eh_frame : { *(.eh_frame) }
  .note.GNU-stack : { *(.note.GNU-stack) }
  PROVIDE(a = nonexistent_provide_sym);
}
!

clang -target aarch64-unknown-none-elf -c -o 1.o 1.c
clang -target aarch64-unknown-none-elf -c -o ref.o ref.s

ld.eld -m aarch64elf -mtriple aarch64-unknown-none-elf -o b.eld.out ref.o -T 1.t
ld.lld -m aarch64linux -o b.lld.out ref.o -T 1.t
aarch64-linux-gnu-ld -o b.gnu.out ref.o -T 1.t
```

parth-07:
> This will crash :

Thank you for finding this case. I was incorrectly assuming `ThisModule.getNamePool().findSymbol(Name)` to always return a valid pointer. I have fixed this case and improved testing.

quic-areg:
Not related to this PR, but we produce the wrong layout for this case:
eld: a=0x300,b=0x900
lld: a=0x900,b=0x900
```
b = a;
a = 0x300;
SECTIONS {
  .text : { *(.text*) }
  .data : { *(.data*) }
  .comment : { *(.comment) }
  .eh_frame : { *(.eh_frame) }
  .note.GNU-stack : { *(.note.GNU-stack) }
  a = 0x900;
}
```
Is this something that's already known?

quic-areg (approved):

parth-07:
> Not related to this PR, but we produce the wrong layout for this case: eld: a=0x300,b=0x900 lld: a=0x900,b=0x900
> 
> ```
> b = a;
> a = 0x300;
> SECTIONS {
>   .text : { *(.text*) }
>   .data : { *(.data*) }
>   .comment : { *(.comment) }
>   .eh_frame : { *(.eh_frame) }
>   .note.GNU-stack : { *(.note.GNU-stack) }
>   a = 0x900;
> }
> ```
> 
> Is this something that's already known?

Yes, it is a known issue. Currently, eld does not follow correct evaluation order for linker script assignments that are outside the SECTIONS command. Assignments after the SECTIONS command are evaluated before the assignments within the SECTIONS command and the outside assignments are evaluated both before and after evaluating the SECTIONS command. In your shared example, `b = a; a = 0x300` is evaluated twice, once before the SECTIONS command and once after the SECTIONS command, and due to this, the value of `a` is changed back to `0x300`. It was previously addressed in the PR that solved this issue: https://github.com/qualcomm/eld/issues/430. However, the PR had to be reverted because it caused a regression. Some of the layout convergence features are required to correctly re-land that PR.

quic-seaswara (approved):

PR #681, opened 2025-12-29
Fix parsing of `"archive:mem"` input section description pattern

### LINKED ISSUE #680: `"archive:mem"` input section description is not interpreted correctly when it is within quotes
`"archive:mem"` input section description file pattern is not interpreted correctly when it is within quotes.

Reproducer:

```
#!/usr/bin/env bash

cat >1.c <<\EOF
int foo() { return 1; }
EOF

cat >main.c <<\EOF
int foo();
int main() { return foo(); }
EOF

cat >script.t <<\EOF
SECTIONS {
  .foo : { "lib1.a:1.o"(.text*) }
  .text : { *(.text*) }
}
EOF


clang -target x86_64-linux-pc-gnu -o 1.o 1.c -c -ffunction-sections
clang -target x86_64-linux-pc-gnu -o main.o main.c -c -ffunction-sections
llvm-ar cr lib1.a 1.o

LDs=(ld.eld)
SFs=(eld)

for i in "${!SFs[@]}"; do
  ${LDs[$i]} -o main.${SFs[$i]}.out main.o lib1.a -T script.t -Map main.${SFs[$i]}.map.txt
done
```

The generated output file does not contain `.foo` output section, and the map-file shows that the `"lib1.a:1.o"` pattern is not interpreted correctly.

parth-07:
This commit fixes the parsing / interpretation of `"archive:mem"` file pattern. Previously, it was getting incorrectly stored because we were not properly handling quotes when a single quoted token in the linker script (`"archive:member"`) is stored as two separate patterns in the linker implementation.

Resolves #680

### DIFF
--- a/include/eld/Script/StrToken.h
+++ b/include/eld/Script/StrToken.h
@@ -44,10 +44,18 @@ public:
 
   std::string getDecoratedName() const;
 
+  void setLeftQuoted() { LeftQuoted = true; }
+  bool isLeftQuoted() const { return LeftQuoted; }
+
+  void setRightQuoted() { RightQuoted = true; }
+  bool isRightQuoted() const { return RightQuoted; }
+
 protected:
   std::string Name;
   bool Quoted = false;
   Kind ScriptFileKind;
+  bool LeftQuoted = false;
+  bool RightQuoted = false;
 };
 
 } // namespace eld
--- a/lib/Script/ScriptFile.cpp
+++ b/lib/Script/ScriptFile.cpp
@@ -537,14 +537,23 @@ StrToken *ScriptFile::createParserStr(const char *PText, size_t PLength) {
 }
 
 StrToken *ScriptFile::createParserStr(llvm::StringRef S) {
-  bool IsQuoted = false;
+  bool isLeftQuoted = false;
+  bool isRightQuoted = false;
   if (S.starts_with("\"")) {
-    S = S.substr(1, S.size() - 2);
-    IsQuoted = true;
+    S = S.substr(1);
+    isLeftQuoted = true;
+  }
+  if (S.ends_with("\"")) {
+    S = S.drop_back(1);
+    isRightQuoted = true;
   }
   StrToken *Tok = make<eld::StrToken>(S.str());
-  if (IsQuoted)
+  if (isLeftQuoted && isRightQuoted)
     Tok->setQuoted();
+  if (isLeftQuoted)
+    Tok->setLeftQuoted();
+  if (isRightQuoted)
+    Tok->setRightQuoted();
   return Tok;
 }
 
--- a/lib/Script/StrToken.cpp
+++ b/lib/Script/StrToken.cpp
@@ -22,10 +22,10 @@ StrToken::StrToken(const std::string &PString, StrToken::Kind K)
 
 std::string StrToken::getDecoratedName() const {
   std::string R = "";
-  if (isQuoted())
+  if (isQuoted() || LeftQuoted)
     R = "\"";
   R += Name;
-  if (isQuoted())
+  if (isQuoted() || RightQuoted)
     R += "\"";
   return R;
 }
--- a/lib/Script/WildcardPattern.cpp
+++ b/lib/Script/WildcardPattern.cpp
@@ -35,6 +35,10 @@ WildcardPattern::WildcardPattern(StrToken *PPattern, SortPolicy PPolicy,
       MBPatternIsPrefix(false), MBPatternIsSuffix(false), CurID(0) {
   if (PPattern->isQuoted())
     setQuoted();
+  if (PPattern->isLeftQuoted())
+    setLeftQuoted();
+  if (PPattern->isRightQuoted())
+    setRightQuoted();
   createGlobPattern(llvm::StringRef(PPattern->name()));
 }
 
--- a/test/Common/standalone/linkerscript/QuotedArchiveMemberPattern/Inputs/1.c
+++ b/test/Common/standalone/linkerscript/QuotedArchiveMemberPattern/Inputs/1.c
@@ -1,0 +1,2 @@
+int foo() { return 1; }
+
--- a/test/Common/standalone/linkerscript/QuotedArchiveMemberPattern/Inputs/main.c
+++ b/test/Common/standalone/linkerscript/QuotedArchiveMemberPattern/Inputs/main.c
@@ -1,0 +1,3 @@
+int foo();
+int main() { return foo(); }
+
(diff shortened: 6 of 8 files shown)


lib/ScriptParser/ScriptParser.cpp
```diff
@@ -661,6 +661,10 @@ InputSectDesc::Spec ScriptParser::readInputSectionDescSpec(StringRef Tok) {
   if (!Tok.contains(':'))
     FilePat = createAndRegisterWildcardPattern(Tok);
   else {
+    // We cannot store the original quote information because we represent
+    // the single token "archive:mem" in the linker script as two separate
+    // patterns in the linker codebase.
+    Tok = unquote(Tok);
```
quic-seaswara: Should we use createParserStr to preserve the property that the input token had a quote. 

The map file might not show information about quotes from the linker script or the linker script generator too.

partaror: `createParserStr` alone would not be helpful here. We do use `createParserStr` within `createAndRegisterWildcardPattern`. The issue is that a single quoted token (`"archive:mem"`) is stored as two separate tokens -- archive file pattern and member pattern -- in the linker codebase. It would be incorrect to store both archive file pattern and the member pattern with quotes if the original combined pattern was with quotes because `"archive":"mem"` would be interpreted differently from `"archive:mem"`.

parth-07: `createParserStr` alone would not be helpful here. We do use `createParserStr` within `createAndRegisterWildcardPattern`. The issue is that a single quoted token (`"archive:mem"`) is stored as two separate tokens -- archive file pattern and member pattern -- in the linker codebase. It would be incorrect to store both archive file pattern and the member pattern with quotes if the original combined pattern was with quotes because `"archive":"mem"` would be interpreted differently from `"archive:mem"`.

quic-seaswara: Yes, we have to seperate the file pattern from the member pattern, but I could not follow why "archive":"mem" needs to be interpreted differently from "archive:mem".

The quotes need to be still printed in the map file to preserve how we read the linker script right ?

quic-seaswara: The problem in the current function createParserStr is that it assumes that when a string starts with quote, it ends with quote as well. So the pattern split appears to have an issue.

parth-07: > Yes, we have to seperate the file pattern from the member pattern, but I could not follow why "archive":"mem" needs to be interpreted differently from "archive:mem".
>
> The quotes need to be still printed in the map file to preserve how we read the linker script right ?

If the linker script contains `"libfoo.a:1.o"` and we keep the quotes on both the archive pattern and the member pattern and consequently store them as `"libfoo.a"` and `"1.o"`, then when we print the rule in the map-file as `"libfoo.a":"1.o"` because we added the quotes in both the archive pattern and the member pattern. `"libfoo.a":"1.o"` will be interpreted differently from `"libfoo.a:1.o"`. Thus, we will need extra conditions to properly print the rule if we preserve the quotes in both the archive pattern and the member pattern.

Should I make the change to add quotes on both the archive and the member pattern and add the logic to properly print the rule description when both the archive pattern and the member pattern contain quotes?

quic-seaswara: The current code assumes that if the string started with a quote, it will also end up with a quote. 

So this pattern "libfoo.a:1.o" will need to be treated as 

* "libfoo.a
* 1.o"

StrToken will need to know of if the state of the string has a single quote/double quote, just for printing purposes only.

The representation for linker script matching will still be libfoo.a and 1.o.

quic-seaswara: ??

quic-seaswara (changes requested):

lib/ScriptParser/ScriptParser.cpp
parth-07: I have updated `StrToken` such that it can store the single quote information.

quic-seaswara (commented):
```
diff --git a/lib/Script/ScriptFile.cpp b/lib/Script/ScriptFile.cpp
index 8c048bb9..00feb8f9 100644
--- a/lib/Script/ScriptFile.cpp
+++ b/lib/Script/ScriptFile.cpp
@@ -538,13 +538,28 @@ StrToken *ScriptFile::createParserStr(const char *PText, size_t PLength) {

 StrToken *ScriptFile::createParserStr(llvm::StringRef S) {
   bool IsQuoted = false;
+  bool IsLeftQuoted = false;
+  bool IsRightQuoted = false;
   if (S.starts_with("\"")) {
-    S = S.substr(1, S.size() - 2);
-    IsQuoted = true;
+    if (S.ends_with("\""))  {
+      S = S.substr(1, S.size() - 2);
+      IsQuoted = true;
+    }
+    else {
+      S = S.substr(1, S.size() - 1);
+      IsLeftQuoted = true;
+    }
+  } else if (S.ends_with("\"")) {
+    S = S.substr(0, S.size() - 1);
+    IsRightQuoted = true;
   }
   StrToken *Tok = make<eld::StrToken>(S.str());
   if (IsQuoted)
     Tok->setQuoted();
+  if (IsLeftQuoted)
+    Tok->setLeftQuote();
+  if (IsRightQuoted)
+    Tok->setRightQuote();
   return Tok;
 }

diff --git a/lib/ScriptParser/ScriptParser.cpp b/lib/ScriptParser/ScriptParser.cpp
index 4e0b2047..126a8405 100644
--- a/lib/ScriptParser/ScriptParser.cpp
+++ b/lib/ScriptParser/ScriptParser.cpp
@@ -661,18 +661,11 @@ InputSectDesc::Spec ScriptParser::readInputSectionDescSpec(StringRef Tok) {
   if (!Tok.contains(':'))
     FilePat = createAndRegisterWildcardPattern(Tok);
   else {
-    bool isQuoted = Tok.starts_with("\"");
-    Tok = unquote(Tok);
     std::pair<llvm::StringRef, llvm::StringRef> Split = Tok.split(':');
     FilePat = createAndRegisterWildcardPattern(Split.first);
     if (!Split.second.empty()) {
       ArchiveMem = createAndRegisterWildcardPattern(Split.second);
     }
-    if (isQuoted) {
-      FilePat->setLeftQuote();
-      if (ArchiveMem)
-        ArchiveMem->setRightQuote();
-    }
     llvm::StringRef peekTok = peek();
     if (!atEOF() && peekTok != "(" &&
         computeLineNumber(peekTok) == getCurrentLineNumber()) {
```

I was approaching the problem in the way above. 

The core issue is in createParserStr. 

Can you please look into it ?

quic-seaswara (approved):

PR #682, opened 2025-12-30
Fix Hexagon linux tests marked unsupported on Windows locally

Ris-Bali:
This patch changes `lit.local.cfg` to check for `hexagon-unknown-linux` instead of `hexagon-unknown-elf` in the target triple condition. This fixes tests being incorrectly marked as unsupported on Windows local builds while maintaining compatibility with Linux and CI builders.

### DIFF
--- a/test/Hexagon/linux/lit.local.cfg
+++ b/test/Hexagon/linux/lit.local.cfg
@@ -1,4 +1,4 @@
 config.suffixes = ['.test']
 
-if 'hexagon-unknown-elf' in config.target_triple:
+if 'hexagon-unknown-linux' in config.target_triple:
   config.unsupported = True


test/Hexagon/linux/lit.local.cfg
```diff
@@ -1,4 +1,4 @@
 config.suffixes = ['.test']
 
-if 'hexagon-unknown-elf' in config.target_triple:
+if 'hexagon-unknown-linux' in config.target_triple:
```
quic-seaswara, line 3: This will make all tests under linux as unsupported.

Ris-Bali:
This seems to be a local configuration issue closing it.

PR #683, opened 2026-01-02
Fix incorrect NEEDED entry due to unneeded shared lib reference

### LINKED ISSUE #668: Shared library is incorrectly marked as needed when it is referenced by an another (un)needed shared library
We incorrectly mark a shared library as needed when it is referenced by an another (un)needed shared library.

Reproducible example:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
int foo() { return 1; }
EOF

cat >2.c <<\EOF
int foo();
int bar() { return 3; }

int baz() { return foo(); }
EOF

cat >main.c <<\EOF
int main() { return 0; }
EOF

clang-18 -o lib1.so 1.c -fPIC -shared
clang-18 -o lib2.so 2.c -fPIC -shared
clang-18 -o main.o main.c -c -ffunction-sections
LDs=(ld.eld)
SFs=(eld)

for i in "${!SFs[@]}"; do
  ${LDs[$i]} -o main.${SFs[$i]}.out main.o --as-needed lib2.so lib1.so 
done
```

```bash
$ llvm-readelf --dynamic main.eld.out
# Dynamic section at offset 0x1000 contains 9 entries:
#  Tag                Type        Name/Value
#  0x0000000000000001 (NEEDED)    Shared library: [lib1.so]
#  0x0000000000000004 (HASH)      0x4001b0
#  0x0000000000000006 (SYMTAB)    0x40017
```

This issue causes a simple hello-world x86-64 program to unnecessarily have  `libgcc_s.so.1` and `ld-linux-x86-64.so.2` as `DT_NEEDED` libraries. Neither GNU nor lld have these `DT_NEEDED` dependency for the same hello-world program.

parth-07:
This commit fixes the incorrect NEEDED entry when --as-needed is used and a shared library symbol is referenced by an another unneeded shared library.

The root cause of the issue was the logic that was used to determine whether a shared library symbol ResolveInfo should have an outSymbol or not.

Resolves #668

### DIFF
--- a/lib/SymbolResolver/IRBuilder.cpp
+++ b/lib/SymbolResolver/IRBuilder.cpp
@@ -20,6 +20,7 @@
 #include "eld/Fragment/RegionFragmentEx.h"
 #include "eld/Input/ArchiveMemberInput.h"
 #include "eld/Input/ELFDynObjectFile.h"
+#include "eld/Input/ELFFileBase.h"
 #include "eld/Input/ELFObjectFile.h"
 #include "eld/Object/ObjectBuilder.h"
 #include "eld/Support/Memory.h"
@@ -33,6 +34,7 @@
 #include "eld/Target/Relocator.h"
 #include "llvm/ADT/StringRef.h"
 #include "llvm/BinaryFormat/ELF.h"
+#include "llvm/Support/Casting.h"
 #include <unordered_map>
 
 using namespace eld;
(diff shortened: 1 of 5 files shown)


lib/SymbolResolver/IRBuilder.cpp
quic-seaswara, line 371: Should the condition be changed here ?

parth-07: No, this does not need to be changed.

parth-07:
Hi @quic-seaswara,

```
$LD 1.o  lib2.so --as-needed lib3.so --no-as-needed -o two
llvm-readelf -d two | grep NEEDED
```

Here lib2.so should always be marked as NEEDED because it is not enclosed within `--as-needed`. 

In issue #668, eld marks as NEEDED an used shared library **that is enclosed within `--as-needed`**. lld and bfd do not have the bug described in #668.

PR #686, opened 2026-01-05
Add UpdateSymbol plugin as a dependency to check-eld

### LINKED ISSUE #685: Test `UpdateSymbol` target is not added as a dependency to `check-eld` target
Test target `UpdateSymbol` should be added as a dependency to `check-eld` target. Otherwise, we have to explicitly build it before running the `check-eld` target.

parth-07:
Resolves #685


quic-seaswara (commented):
has changes more than required ?

parth-07:
> has changes more than required ?

Thank you for catching this. I accidentally added changes of the other patch here. I have fixed it now.

PR #688, opened 2026-01-05
Reiterate layout step until section addresses converge

### LINKED ISSUE #687: Incorrect layout when there is a forward reference in a dot assignment
We have incorrect layout when there is a forward reference in a dot assignment.

Reproducer:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
int foo() { return 1; }

int val = 3;
EOF


cat > script.t << '!'
SECTIONS {
  . = u;
  .text : { *(.text*) }
  .data : { *(.data*) }
  .comment : { *(.comment*) }
  u = 0x3000;
}
!


LDs=(ld.eld ld.lld aarch64-none-elf-ld.bfd)
SFs=(eld lld bfd)

clang -o 1.o --target=aarch64-unknown-elf 1.c -c -ffunction-sections

for i in "${!SFs[@]}"; do
done
```

Running `llvm-readelf -S 1.eld.out` outputs:
```
There are 7 section headers, starting at offset 0x1280:

Section Headers:
  [Nr] Name              Type            Address          Off    Size   ES Flg Lk Inf Al
  [ 0]                   NULL            0000000000000000 000000 000000 00      0   0  0
  [ 1] .text             PROGBITS        0000000000000000 001000 000008 00  AX  0   0  4
  [ 2] .data             PROGBITS        0000000000000008 001008 000004 00  WA  0   0  4
  [ 3] .comment          PROGBITS        0000000000000000 00100c 0000ce 01  MS  0   0  1
  [ 4] .shstrtab         STRTAB          0000000000000000 0010da 000030 00      0   0  1
  [ 5] .symtab           SYMTAB          0000000000000000 001110 000108 18      6   8  8
  [ 6] .strtab           STRTAB          0000000000000000 001218 00002d 00      0   0  1
```

The address of `.text` section should be `0x3000` instead of `0x1000`.

parth-07:
This commit modifies the layout step to reiterate until the section addresses converge. The reiteration has an upper limit of 4. On reaching this limit, a warning is emitted and the layout reiteration is stopped.

Please note that this reiteration limit is for layout iterations before relaxations take place. After a layout relaxation pass, the reiteration count is reset to 0.

Resolves #687

### DIFF
--- a/include/eld/Diagnostics/DiagLayouts.inc
+++ b/include/eld/Diagnostics/DiagLayouts.inc
@@ -61,5 +61,6 @@ DIAG(warning_zero_sized_fragment_for_non_zero_symbol, DiagnosticEngine::Warning,
      "Zero sized fragment %0 for non zero sized symbol %1 from input file %2")
 DIAG(error_offset_not_assigned_for_output_section, DiagnosticEngine::Error,
      "Requested offset for output section %0 has not yet been assigned")
-DIAG(warn_empty_segment, DiagnosticEngine::Warning,
-     "Empty segment: '%0'")
+DIAG(warn_empty_segment, DiagnosticEngine::Warning, "Empty segment: '%0'")
+DIAG(note_section_address_not_converging, DiagnosticEngine::Note,
+     "Section '%0' address did not converge after %1 layout passes")
--- a/include/eld/Target/GNULDBackend.h
+++ b/include/eld/Target/GNULDBackend.h
@@ -72,6 +72,7 @@ class ScriptMemoryRegion;
 class StubFactory;
 class SymDefReader;
 class TimingFragment;
+class OutputSectionEntry;
 
 /** \class GNULDBackend
  *  \brief GNULDBackend provides a common interface for all GNU Unix-OS
@@ -762,6 +763,21 @@ public:
 
   // -------------------- Finalize Layout callback ----------------
   virtual bool finalizeLayout() { return true; }
+
+  struct SectionAddrs {
+    uint64_t vma = 0;
+    uint64_t lma = 0;
+  };
+
+  struct LayoutSnapshot {
+    llvm::DenseMap<const OutputSectionEntry *, SectionAddrs> outSections;
+  };
+
+  // Capture current layout snapshot keyed by OutputSectionEntry*.
+  LayoutSnapshot captureLayoutSnapshot() const;
+
+  const OutputSectionEntry *findDivergence(const LayoutSnapshot &Prev,
+                                           const LayoutSnapshot &cur) const;
 
   // --- Exclude symbol support
   void markSymbolForRemoval(const ResolveInfo *S);
--- a/lib/Target/GNULDBackend.cpp
+++ b/lib/Target/GNULDBackend.cpp
@@ -3280,6 +3280,39 @@ bool GNULDBackend::layout() {
   return config().getDiagEngine()->diagnose();
 }
 
+GNULDBackend::LayoutSnapshot GNULDBackend::captureLayoutSnapshot() const {
+  LayoutSnapshot S;
+  const SectionMap &SM = m_Module.getScript().sectionMap();
+  for (auto It = SM.begin(), End = SM.end(); It != End; ++It) {
+    const OutputSectionEntry *O = *It;
+    const ELFSection *Sec = O->getSection();
+    ASSERT(Sec, "Must not be null!");
+    SectionAddrs A;
+    A.vma = Sec->addr();
+    A.lma = Sec->pAddr();
+    S.outSections.insert({O, A});
+  }
+  return S;
+}
+
+const OutputSectionEntry *
+GNULDBackend::findDivergence(const LayoutSnapshot &Prev,
+                             const LayoutSnapshot &cur) const {
+  const SectionMap &SM = m_Module.getScript().sectionMap();
+  for (auto It = SM.begin(), End = SM.end(); It != End; ++It) {
+    const OutputSectionEntry *O = *It;
+    auto PrevIt = Prev.outSections.find(O);
+    auto CurIt = cur.outSections.find(O);
+    if (PrevIt == Prev.outSections.end() || CurIt == cur.outSections.end())
+      return O;
+    const SectionAddrs &A = PrevIt->second;
+    const SectionAddrs &B = CurIt->second;
+    if (A.vma != B.vma || A.lma != B.lma)
+      return O;
+  }
+  return nullptr;
+}
+
 // Create or return an already created relocation output section for partial
 // linking.
 ELFSection *GNULDBackend::getOutputRelocationSection(ELFSection *S,
@@ -4074,15 +4107,33 @@ bool GNULDBackend::relax() {
   while (!finished) {
     auto start = std::chrono::steady_clock::now();
     {
-      eld::RegisterTimer T("Assign Address", "Establish Layout",
-                           m_Module.getConfig().options().printTimingStats());
-      bool hasError = createProgramHdrs();
-      if (hasError)
-        m_Module.setFailure(true);
-      if (updateTargetSections()) {
+      // Bounded convergence loop over address assignment.
+      LayoutSnapshot prevSnap, curSnap;
+      prevSnap = captureLayoutSnapshot();
+      constexpr int maxIterations = 4;
+      const OutputSectionEntry *changed = nullptr;
+      for (int i = 0; i < maxIterations; ++i) {
+        eld::RegisterTimer T("Assign Address", "Establish Layout",
+                             m_Module.getConfig().options().printTimingStats());
         bool hasError = createProgramHdrs();
         if (hasError)
           m_Module.setFailure(true);
+        if (updateTargetSections()) {
+          bool hasError = createProgramHdrs();
+          if (hasError)
+            m_Module.setFailure(true);
+        }
+        curSnap = captureLayoutSnapshot();
+        changed = findDivergence(prevSnap, curSnap);
+        if (!changed)
+          break;
+        prevSnap = std::move(curSnap);
+      }
+      if (changed) {
+        // Emit a note for the first observed changed section.
+        if (const ELFSection *S = changed->getSection())
+          config().raise(Diag::note_section_address_not_converging)
+              << S->name() << maxIterations;
       }
       if (LinkerConfig::Object != config().codeGenType()) {
         if (!setupProgramHdrs()) {
--- a/test/Common/standalone/DotForwReference/DotForwReference.test
+++ b/test/Common/standalone/DotForwReference/DotForwReference.test
@@ -1,0 +1,35 @@
+#---DotForwReference.test--------------------- Executable,LS ------------------#
+#BEGIN_COMMENT
+# Validates that a forward reference to a symbol in dot assignments within
+# SECTIONS is handled correctly.
+#END_COMMENT
+#START_TEST
+RUN: %clang %clangopts -o %t1.1.o %p/Inputs/1.c -c -ffunction-sections
+RUN: %link %linkopts -o %t1.1.out %t1.1.o -T %p/Inputs/script.1.t
+RUN: %readelf -S %t1.1.out | %filecheck %s
+RUN: %link %linkopts -o %t1.1.phdr.out %t1.1.o -T %p/Inputs/script.1.phdr.t
+RUN: %readelf -S %t1.1.phdr.out | %filecheck %s
+RUN: %link %linkopts -o %t1.2.out %t1.1.o -T %p/Inputs/script.2.t 2>&1 \
+RUN:   --trace=assignments | %filecheck %s --allow-empty --check-prefix TRACE
+RUN: %readelf -S %t1.2.out | %filecheck %s --check-prefix CHECK2
+RUN: %link %linkopts -o %t1.2.phdr.out %t1.1.o -T %p/Inputs/script.2.phdr.t 2>&1 \
+RUN:   --trace=assignments | %filecheck %s --allow-empty --check-prefix TRACE
+RUN: %readelf -S %t1.2.phdr.out | %filecheck %s --check-prefix CHECK2
+RUN: %link %linkopts -o %t1.1.3.out %t1.1.o -T %p/Inputs/script.3.t 2>&1 \
+RUN:   | %filecheck %s --check-prefix=NON_CONVERGENCE_NOTE
+RUN: %link %linkopts -o %t1.1.3.out %t1.1.o -T %p/Inputs/script.3.phdr.t 2>&1 \
+RUN:   | %filecheck %s --check-prefix=NON_CONVERGENCE_NOTE
+#END_TEST
+
+CHECK: .text PROGBITS {{0+}}3000
+
+TRACE: Trace: OUTPUT_SECTION(EVALUATE) >> .text VMA : 0x0 LMA : 0x0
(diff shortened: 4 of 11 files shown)


lib/Target/CreateScriptProgramHeaders.hpp
```diff
@@ -87,319 +87,333 @@ bool GNULDBackend::createScriptProgramHdrs() {
...
-  reset_state();
+    changedOutputSection = nullptr;
+    if (linkerScriptHasMemoryCommand)
+      clearMemoryRegions();
+  };
 
-  while (out != outEnd) {
-    bool useSetLMA = false;
+  int preRelaxPassIter = 0;
+  do {
```
quic-areg: This is a massive loop that requires indenting everything and makes the diff really large. I wonder if this can be avoided somehow? Can we isolate the single-pass into a new function and then have a small loop in relax() or somewhere else?

lib/Target/GNULDBackend.cpp
```diff
@@ -4017,6 +4017,8 @@ bool GNULDBackend::maySkipRelocProcessing(Relocation *pReloc) const {
 bool GNULDBackend::relax() {
   bool finished = false;
   int iteration = 0;
+  // Limit the number of layout passes to avoid non-converging layouts.
+  const int maxPreRelaxationPasses = 4;
```
quic-areg: Can we define this in one place?

parth-07:
> Why are you exposing maxPreRelaxationPasses to the function ?

We need to use `maxPreRelaxationPasses` in createProgramHdrs` and `createScriptProgramHdrs` function to determine
the number of times the layout needs to be computed when it is not getting converged.

> I feel createProgramHeaders is becoming overly iterative.
> The logic can be contained in GNULDBackend.

To push the logic in GNULDBackend, we will need to refactor `createProgramHdrs` and `createScriptProgramHdrs` so that
we can run the inner loop of these functions directly from GNULDBackend. However, in our offline discussions, you told not
to refactor those functions as of now as that will increase the testing requirement.

include/eld/Target/GNULDBackend.h
```diff
@@ -1184,6 +1189,8 @@ class GNULDBackend {
   GNUVerDefFragment *GNUVerDefFrag = nullptr;
   std::unordered_map<const ResolveInfo *, uint16_t> OutputVersionIDs;
 #endif
+  OutputSectionEntry * changedOutputSection = nullptr;
```
quic-areg: Layout iterations only continue when section addresses change, but what about symbol-only forward references?

```ld
cat > 1.c << \!
int foo() {return 0;}
!

cat > cycle.t << \!
SECTIONS {
  . = 0x1000;
  .text : {*(.text*)}
  a = b + 4;
  b = a + 4;
}
!

cat > sym.t << \!
PHDRS {TEXT PT_LOAD;}
SECTIONS {
  . = 0x1000;
  .text : {*(.text*)} : TEXT
  a = b + 4;
  b = c + 4;
  c = 0x2000;
}
!

clang -c 1.c

ld.eld 1.o -T sym.t # a,b,c = 0x8,    0x2004, 0x2000
ld.lld 1.o -T sym.t # a,b,c = 0x2008, 0x2004, 0x2000

ld.lld 1.o -T cycle.t # error: assignment to symbol a does not converge
ld.eld 1.o -T cycle.t # a = 0x14; b=0x18
```

lib/Target/CreateScriptProgramHeaders.hpp
parth-07: >  Can we isolate the single-pass into a new function and then have a small loop in relax() or somewhere else?

That will also result in a huge diff, same as it is right now.

> I wonder if this can be avoided somehow?

We can keep the indentation of the original loop as before, but still enclose it within do .. while(), however, then the indentation would be incorrect and reading the code in future might be misleading in the future due to incorrect indentation.

include/eld/Target/GNULDBackend.h
parth-07: > Layout iterations only continue when section addresses change, but what about symbol-only forward references?

I plan to handle symbol address convergence in the next patch. The subsequent patches will add support for symbol address convergence, improved diagnostics related to layout iterations and convergence, and some heuristics to reduce the number of iterations wherever possible.

lib/Target/CreateScriptProgramHeaders.hpp
```diff
@@ -58,7 +58,7 @@ bool GNULDBackend::createScriptProgramHdrs() {
       _segmentsForSection[m_phdr->getOutputSection()].push_back(
           _segments[EhdrInLoadSegment->getSpec()->name()]);
   };
-
+  bool linkerScriptHasMemoryCommand = m_Module.getScript().hasMemoryCommand();
   auto reset_state = [&](OutputSectionEntry *OS = nullptr) {
```
quic-areg, line 62: we reuse the segment table across passes; segments from a transient layout can stick around:

In the first pass we create a new `PT_LOAD` for `.bar` due to the forward reference that remains even after `v` is fixed:

```bash
cat > 1.c << \!
[[gnu::section(".foo")]] int foo = 1;
[[gnu::section(".bar")]] int bar = 2;
!

cat > fwd.t << \!
SECTIONS {
  . = u;
  .foo : {*(.foo)}
  . = v;
  .bar : {*(.bar)}
  u = 0x1000;
  v = u + 0x10;
}
!

cat > nofwd.t << \!
SECTIONS {
  . = 0x1000;
  .foo : {*(.foo)}
  . = 0x1010;
  .bar : {*(.bar)}
}
!

clang -c 1.c
ld.eld 1.o -T fwd.t -o fwd     # two PT_LOADs with .foo and .bar separate.
ld.eld 1.o -T nofwd.t -o nofwd # single PT_LOAD with .foo and .bar together.

ld.lld 1.o -T fwd.t -o fwd
ld.lld 1.o -T fwd.t -o nofwd

llvm-readelf -l fwd nofwd
```

lib/Target/GNULDBackend.cpp
parth-07: Thank you for catching this. This variable is not used at all. I added a member variable of the same name and that is being used everywhere. I have removed it now.

include/eld/Target/GNULDBackend.h
```diff
@@ -1184,6 +1189,8 @@ class GNULDBackend {
...
+  const int maxPreRelaxationPasses = 4;
 };
```
quic-areg, line 1211: It looks like lld's max passes is 5. Is there any reason we picked a different number?

include/eld/Diagnostics/DiagLayouts.inc
```diff
@@ -62,4 +62,6 @@ DIAG(warning_zero_sized_fragment_for_non_zero_symbol, DiagnosticEngine::Warning,
 DIAG(error_offset_not_assigned_for_output_section, DiagnosticEngine::Error,
      "Requested offset for output section %0 has not yet been assigned")
 DIAG(warn_empty_segment, DiagnosticEngine::Warning,
-     "Empty segment: '%0'")
\ No newline at end of file
+     "Empty segment: '%0'")
+DIAG(note_section_address_not_converging, DiagnosticEngine::Note,
```
quic-areg, line 65: should this be a warning instead?

lib/Target/CreateScriptProgramHeaders.hpp
```diff
@@ -87,319 +87,333 @@ bool GNULDBackend::createScriptProgramHdrs() {
...
+        config().raise(Diag::cannot_set_at_address) << cur->name();
+        hasError = true;
+      }
 
-    if (cur->isFixedAddr() && (vma != cur->addr())) {
-      config().raise(Diag::cannot_set_at_address) << cur->name();
-      hasError = true;
-    }
+      setVMA(*cur, vma, /*ignoreChangedSection=*/false);
+      cur->setAddr(vma);
```
quic-areg: do we need both `setVMA` and `cur->setAddr()`?

lib/Target/CreateScriptProgramHeaders.hpp
quic-areg: I was thinking to first refactor the single-pass in a separate patch, and then this PR just becomes adding a loop

If refactoring is not an option, can the loop be a `while()` loop instead of a `do...while`? Since the function is massive it's not obvious what condition we are looping over until we find the end of the loop many lines down.

Regardless of whatever decision is taken here I think the two `.hpp` files could benefit from refactoring somewhere down the line.

lib/Target/CreateScriptProgramHeaders.hpp
parth-07: > we reuse the segment table across passes; segments from a transient layout can stick around:

Thank you for mentioning this. This can indeed create some issues.

> ld.eld 1.o -T fwd.t -o fwd     # two PT_LOADs with .foo and .bar separate.
> ld.eld 1.o -T nofwd.t -o nofwd # single PT_LOAD with .foo and .bar together.

The behavior here is the same as before this patch. 

I will change the patch functionality so that the segments from a transient layout do not cause issues.

include/eld/Diagnostics/DiagLayouts.inc
parth-07: This is a note because with warning the link will fail if `--fatal-warning` is used.

lib/Target/CreateScriptProgramHeaders.hpp
parth-07: No, thank you for catching this. I am fixing this.

include/eld/Target/GNULDBackend.h
parth-07: LLD's has two different kind of max passes, one is `30` and the other is `5`. LLD uses the combination of these two during the layout. eld layout mechanism is different from LLD. `4` is not fixed here, and can be changed if a greater / smaller value seems better in the future. I believe that `4` is appropriate here as a good upper-bound, especially after adding some of the heuristics to converge faster that are planned as a follow-up after this patch. And if the layout converges in the 2nd iteration, then the remaining iterations will be skipped.

lib/Target/CreateScriptProgramHeaders.hpp
parth-07: >  ld.eld 1.o -T fwd.t -o fwd # two PT_LOADs with .foo and .bar separate.
>  ld.eld 1.o -T nofwd.t -o nofwd # single PT_LOAD with .foo and .bar together.

With the latest patchset, we have single PT_LOAD segment in both the cases. Thank you for pointing this out.

lib/Target/CreateScriptProgramHeaders.hpp
parth-07: I have changed the design so that now we don't have large number of changes in this PR.

> Regardless of whatever decision is taken here I think the two .hpp files could benefit from refactoring somewhere down the line.

Totally agree on this.

include/eld/Target/GNULDBackend.h
```diff
@@ -757,6 +758,24 @@ class GNULDBackend {
   // -------------------- Finalize Layout callback ----------------
   virtual bool finalizeLayout() { return true; }
 
+  // -------------------- Layout convergence snapshot -------------------
```
quic-areg: As a tangent, I am generally not a fan of this kind of comment.

`  // -------------------- Layout convergence snapshot -------------------`
`  // -------------------- Finalize Layout callback ----------------`

The length of the `----` portions are always inconsistent. The comment would look a lot better as 

`// Layout convergence`

Or in some cases, omitted entirely.

include/eld/Target/GNULDBackend.h
```diff
@@ -757,6 +758,24 @@ class GNULDBackend {
...
+
+  // Capture current layout snapshot keyed by OutputSectionEntry*.
+  LayoutSnapshot captureLayoutSnapshot() const;
+
+  // Compare two snapshots for exact address convergence.
+  // Returns nullptr if converged; otherwise returns a changed OutputSectionEntry.
+  const OutputSectionEntry *isConverged(const LayoutSnapshot &Prev,
+                                        const LayoutSnapshot &cur) const;
+
   // --- Exclude symbol support
```
quic-areg, line 782: Tangent: This is an example where the comment is better omitted. It just repeats the function name, which is already self-describing.

include/eld/Target/GNULDBackend.h
```diff
@@ -757,6 +758,24 @@ class GNULDBackend {
   // -------------------- Finalize Layout callback ----------------
   virtual bool finalizeLayout() { return true; }
 
+  // -------------------- Layout convergence snapshot -------------------
+  struct SectionAddrs {
+    uint64_t vma = 0;
+    uint64_t lma = 0;
+  };
+
+  struct LayoutSnapshot {
```
quic-areg, line 772: can this be 
```cpp
using LayoutSnapShot = llvm::DenseMap...
```
?

include/eld/Diagnostics/DiagLayouts.inc
quic-areg: non-convergence implies potentially incorrect output, and that should be a warning imo. A note may be be ignored allowing potentially bad code to ship.

include/eld/Target/GNULDBackend.h
```diff
@@ -757,6 +758,24 @@ class GNULDBackend {
...
+    llvm::DenseMap<const OutputSectionEntry *, SectionAddrs> outSections;
```
quic-seaswara, line 773: can we make this a vector of all the output sections ? 

Since we know the order in this case, it  is much easier to compare and return ? 

```
struct { 
   OutputSectionEntry *OS;
   uint64_t vma;
   uint64_t lma;
   operator ==(OSE &right) {
   ... 
   }
} OSE;
std::vector<OSE> OutputSections;
```

quic-seaswara (commented):
I generally like the direction we are going with this.

include/eld/Target/GNULDBackend.h
parth-07: I agree with your suggestion. I have omitted the comment here entirely.

include/eld/Target/GNULDBackend.h
parth-07: Agreed! I have removed the redundant function comment.

include/eld/Target/GNULDBackend.h
parth-07: This is a `struct` intentionally. The next patch will add another member to the struct for tracking symbol values.

include/eld/Target/GNULDBackend.h
parth-07: I tried doing this. It seems that the order of output sections can change in the `createProgramHdrs` call due to this code:

```cpp
// If there is an AT Table. Lets try to check the dot value with the
// section specified in the AT Table.
if (atTable.size() && m_AtTableIndex < atTable.size()) {
  ELFSection *atSection = atTable[m_AtTableIndex];
  if (curIsDebugSection ||
       ((atSection->addr() < dotSymbol->value()) ||
       (scriptvma && (*scriptvma > atSection->addr())))) {
    out = script.sectionMap().insert(out, atSection);
```

The `out = script.sectionMap().insert(out, atSection);` call can change the output section order.

quic-seaswara: ah! at sections are very rare and the feature is going away FYI. I would like to see that we have the order of output sections as a vector.

partaror: > the feature is going away 

But the feature is not away yet. Can we remove the at feature currently? I do not want the layout convergence functionality to break at some cases as it can get difficult to track.

Issue #693, opened 2026-01-06
Add compatibility with `LLVM_DISTRIBUTION_COMPONENTS`


jonathonpenix:
**Background**
`LLVM_DISTRIBUTION_COMPONENTS` is the documented/recommended way of building an LLVM distribution: https://github.com/llvm/llvm-project/blob/main/llvm/docs/BuildingADistribution.rst

Right now, I think eld is incompatible with this (adding eld into `LLVM_DISTRIBUTION_COMPONENTS` results in cmake failures) as it doesn't export the relevant install targets.

**Reproducer**
It's a bit of a silly example, but:

```
cmake -G Ninja \
   -DCMAKE_BUILD_TYPE=Release \
   -DCMAKE_INSTALL_PREFIX=$PWD/install \
   -DLLVM_ENABLE_ASSERTIONS:BOOL=ON \
   -DLLVM_ENABLE_PROJECTS='llvm;clang' \
   -DLLVM_EXTERNAL_PROJECTS=eld \
   -DLLVM_EXTERNAL_ELD_SOURCE_DIR=${PWD}/llvm-project/eld \
   -DLLVM_TARGETS_TO_BUILD='ARM;AArch64;RISCV' \
   -DELD_TARGETS_TO_BUILD='ARM;AArch64;RISCV' \
   -DCMAKE_CXX_FLAGS='-stdlib=libc++' \
   -DLLVM_DISTRIBUTION_COMPONENTS='clang;dsymutil;LTO;ld.eld;LW' \
   -B $PWD/build \
   -S ./llvm-project/llvm
```

produces the following errors:

```
CMake Error at cmake/modules/LLVMDistributionSupport.cmake:286 (message):
  Specified distribution component 'ld.eld' doesn't have an install target
Call Stack (most recent call first):
  CMakeLists.txt:1502 (llvm_distribution_add_targets)


CMake Error at cmake/modules/LLVMDistributionSupport.cmake:294 (message):
  Specified distribution component 'ld.eld' doesn't have an install-stripped
  target.  Its installation target creation should be changed to use
  add_llvm_install_targets, or you should manually create the
  'install-ld.eld-stripped' target.
Call Stack (most recent call first):
  CMakeLists.txt:1502 (llvm_distribution_add_targets)
```

PR #694, opened 2026-01-07
Add basic LLVM_DISTRIBUTION_COMPONENTS compatibility

jonathonpenix:
LLVM_DISTRIBUTION_COMPONENTS seems to require that each component have
a) install-* and install-*-stripped targets and b) a component associated with
the things to install.

So, there's a few separate changes made to add this support into eld:
- For each of the components we want to support, `add_llvm_install_targets`
  is used to create the install* and install-*-stripped targets. I think this
  is the easiest (and a recommended) way to do this.
- I think eld has five components we need to handle: ld.eld, LW,
  PluginAPIHeaders, linker-script (templates), and YAMLMapParser. Each
  of these had the appropriate install targets and components added, as
  needed. Hopefully I'm not missing any additional components we need to
  consider for distributions.

  It might make sense to more aggressively couple some of these, but I
  think the granularity here matches pretty well with available CMake
  options (to ex: disable the YAMLMapParser) and current toolchains with
  eld included (where the PluginAPIHeaders, templates, etc. may be stripped
  out). So, I left this as-is.
- A dependency was added between the ld.eld and LW install targets--without
  these ld.eld would always fail at runtime about the missing LW library, if
  only ex: `ninja install-ld.eld` was run. Which, doesn't seem that helpful.
  So, make ld.eld's install depend on LW's.

Expected usage looks something like:

`-DLLVM_DISTRIBUTION_COMPONENTS='...;ld.eld;LW;PluginAPIHeaders;linker-script;YAMLMapParser'`

and these items can generally be added/removed based on what people want in
the distribution (modulo ld.eld's dependency on LW).

And, as per the install-* target requirement mentioned above, each of
these components can now be individually installed with ex: install-ld.eld.

Fixes https://github.com/qualcomm/eld/issues/693

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -212,6 +212,13 @@ install(
   PATTERN "*.template"
   PATTERN ".git" EXCLUDE
   PATTERN ".svn" EXCLUDE)
+if (NOT LLVM_ENABLE_IDE)
+  # LLVM_DISTRIBUTION_COMPONENTS requires that each component have both a
+  # component and an install-component target, so add appropriate dummy targets.
+  add_custom_target(linker-script)
+  add_llvm_install_targets(install-linker-script
+                           COMPONENT linker-script)
+endif()
 
 set(ELD_INSTALL_YAML_MAP_PARSER
     ON
@@ -224,5 +231,13 @@ if(ELD_INSTALL_YAML_MAP_PARSER)
     set(ELD_YAML_MAP_PARSER_DEST_DIR "bin")
   endif()
   install(PROGRAMS utils/YAMLMapParser/YAMLMapParser.py
-          DESTINATION ${ELD_YAML_MAP_PARSER_DEST_DIR})
+          DESTINATION ${ELD_YAML_MAP_PARSER_DEST_DIR}
+          COMPONENT YAMLMapParser)
+  if (NOT LLVM_ENABLE_IDE)
+    # LLVM_DISTRIBUTION_COMPONENTS requires that each component have both a
+    # component and an install-component target, so add appropriate dummy targets.
+    add_custom_target(YAMLMapParser)
+    add_llvm_install_targets(install-YAMLMapParser
+                             COMPONENT YAMLMapParser)
+  endif()
 endif()
--- a/include/eld/CMakeLists.txt
+++ b/include/eld/CMakeLists.txt
@@ -21,4 +21,12 @@ if(ENABLE_ELD_PLUGIN_SUPPORT)
       DESTINATION include/ELD/PluginAPI/
       COMPONENT PluginAPIHeaders)
   endif()
+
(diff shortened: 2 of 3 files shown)


quic-k:
will this patch allow me to do `ninja install-ld.eld` same as `ninja install-clang`?

jonathonpenix:
Yep!

Playing with this a bit though now that you mention it, it looks like doing just a `ninja install-ld.eld` doesn't also install `LW` which kind of breaks things (errors when running eld unless `ninja install-LW` is also run). Will look at that and see if that can be changed.

quic-k:
> Yep!
> 

great, this is useful
> Playing with this a bit though now that you mention it, it looks like doing just a `ninja install-ld.eld` doesn't also install `LW` which kind of breaks things (errors when running eld unless `ninja install-LW` is also run). Will look at that and see if that can be changed.

probably need to add LW as ld.eld's dependency

jonathonpenix:
> probably need to add LW as ld.eld's dependency

Yep! I think that works

jonathonpenix:
> I just don't know how this is supposed to work so take my comment with a grain of salt. Is it true that whoever creates "distributions" needs to know the full list of components to get the "default" i.e. full distribution? So if we ever add a new component, those people will need be alerted, like "we added a new component in eld, please consider if you want it your distribution". ... I'm just thinking if someone would build and install elf just for YAMLMapParser...

Yep, that is correct. To some degree I think that is just a part of the `LLVM_DISTRIBUTION_COMPONENTS` model--the person specifying it is expected to understand which components are available and which they want in their final distribution. And, there's not really any requirement that it be a "reasonable" distribution besides that dependencies for the selected components are satisfied.

> Maybe it would be easier create one component (ELD) that includes everything?

Overall, I'm a bit split on this--I do think as-is things are a bit verbose and needing to ex: explicitly specify both ld.eld and LW seems a bit unnecessary given that I don't think LW is that helpful outside of ld.eld, and ld.eld requires LW, but:

- This level of granularity mirrors how some past toolchains have interacted with eld (ex: plugin headers included in the normal build/install, but stripped from the final distribution). It's unclear to me if that was an _intentional_ decision, but that was the behavior IIUC--not sure how important this is to preserve.
- There are some components that I think ex: CPULLVM _wouldn't_ want. The linker-script templates for example are not helpful for us--as-is we can drop this from the distribution. As mentioned above, past toolchains haven't actually distributed the plugin headers (but eld `FORCE`s these on)--unclear to me what behavior we want going forward. But, I think ex: making these regular options that could be excluded from both the normal `install` and `install-distribution` targets would be ok too (and having the singular eld install-distribution target)

@quic-seaswara and others, do you have thoughts here?

jonathonpenix:
Closing this in favor of https://github.com/qualcomm/eld/pull/740

PR #707, opened 2026-01-09
Support z separate/no-separate code

quic-areg:
`-z separate-code` puts the text segment in its
own pages disjoint from any other segments.


include/eld/Input/ZOption.h
```diff
@@ -41,6 +41,8 @@ class ZOption {
     NoDelete,
     NoExecStack,
     NoGnuStack,
+    separateCode,
```
quic-seaswara: SeperateCode

lib/Target/CreateProgramHeaders.hpp
```diff
@@ -178,6 +178,25 @@ bool GNULDBackend::createProgramHdrs() {
...
+      return false;
+    if (!prevOut)
+      return false;
+    auto PrevSegIt = _segmentsForSection.find(prevOut);
+    if (PrevSegIt == _segmentsForSection.end() || PrevSegIt->second.empty())
+      return false;
+    bool PrevExec = (PrevSegIt->second.front()->flag() & llvm::ELF::PF_X);
+    return PrevExec != IsExec;
+  };
+
```
quic-seaswara, line 199: Can we rename this to something better since the lambda does nore than aligning segments to a page ? 

Why does aligning segments to a page not work with -z seperate-code ?

quic-areg: > Can we rename this to something better since the lambda does nore than aligning segments to a page ?

If it does anything more than determine whether or not segments should be page aligned then I did something wrong. Taking another look, I am not seeing that it does anything other than that. What am I missing? What name do you suggest?

> Why does aligning segments to a page not work with -z seperate-code ?

I am not sure if I understand your question. It does "work". Starting code segments at page aligned offsets ensures that they are not co-resident on the same page as other segments. That is the point of this lambda function: when transitioning to/from a code segment to another kind of segment, with `-z separate-code` the next segment should be page aligned. That is my understanding anyways

include/eld/Input/ZOption.h
```diff
@@ -41,6 +41,8 @@ class ZOption {
     NoDelete,
     NoExecStack,
     NoGnuStack,
+    SeparateCode,
+    NoseparateCode,
```
partaror: Would this be better?

```suggestion
    NoSeparateCode,
```

lib/Target/CreateProgramHeaders.hpp
```diff
@@ -178,6 +178,25 @@ bool GNULDBackend::createProgramHdrs() {
 
   reset_state();
 
+  auto SeparateKind = config().options().getSeparateSegmentKind();
+  bool ShouldSeparate =
+      !linkerScriptHasSectionsCommand &&
```
partaror, line 183: It seems that the `-z separate-code` option only takes affect when the linker script does not have `SECTIONS` command. Should we report a warning if this option is used along with the `SECTIONS` command?

partaror (commented):
This option is affecting all the targets, however, the test is added just for x86. Can we add a test in the common/Standalone subdirectory as well?

quic-areg:
>  it looks like there is some behavioral difference with these switches with linker script, omagic support, and shared library support

It is expected that `readelf -l` would differ with the absence/presence of these options.

mixing `-z separate-code` with partial linking, omagic, or linker scripts does not make sense. We could warn for these combinations though neither lld nor bfd currently do.

lib/Target/CreateProgramHeaders.hpp
quic-seaswara: The lambda checks for the variety of things 

- align segments to a page 
- checks for the last section added to the segment and sees if it has execute permissions (SeperateSegmentKind for code) 

Should this be called DoPageAlignment or NeedSeperateSegmentOrPageAlign ?

quic-areg: > checks for the last section added to the segment and sees if it has execute permissions (SeperateSegmentKind for code)

Because it is relevant to determine if we `ShouldPageAlignSegments`? 

It checks if we should page align segments based on:
1. `--align-segments/no-align-segments`
2. if we have a linker script
3. `-z separate-code`. This requires looking at transitions to/from executable segments

> Should this be called DoPageAlignment or NeedSeperateSegmentOrPageAlign

I don't see how these are better names. They sound like (worse) synonyms to me.

include/eld/Input/ZOption.h
quic-areg: Done

lib/Target/CreateProgramHeaders.hpp
quic-areg: I will take this up in a follow-up

PR #715, opened 2026-01-15
Enable target-independent tests that use LLVM assembly

quic-akaryaki:
Some lld tests are written in LLVM assembly (`.ll`) and need target triple and data layout. These tests are usually compiled with `llvm-as` and run only on x86 because they have the x86 triple and data layout hard coded inside. `llvm-as` can override data layout but unfortunately not the target triple.

`.ll` tests can be made target independent if `opt` is used instead of `llvm-as` as `opt` can override both target triple and data layout, which we need to provide in the command line. Note that `opt` does code optimization in addition to just parsing so this may not works for some tests or need additional adjustments.

With this changes we remove `REQUIRES: x86` from the two `.ll` tests sourced from lld.

Another options is to write tests in C, which is often easier but the may take longer to run.

### DIFF
--- a/test/lit.cfg
+++ b/test/lit.cfg
@@ -194,6 +194,7 @@ cp = 'cp'
 tail = 'tail'
 xlen = 4
 dirname = 'dirname'
+datalayout = ''
 
 # Control RISC-V Compression extension option, on by default
 clangnorvcopts = ''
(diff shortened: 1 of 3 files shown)


test/lld/ELF/opt-level.ll
```diff
@@ -1,5 +1,5 @@
-; REQUIRES: x86
-; RUN: %llvm-as -o %t.o %s
+; RUN: opt --mtriple=%triple --data-layout=%datalayout -o %t.o %s
```
quic-seaswara: %opt

quic-seaswara: Can you add a dependency on opt for ninja check-eld ?

test/lit.cfg
```diff
@@ -252,10 +253,13 @@ config.test_target = config.eld_targets_to_build.split('_', 1)[0]
 
 if config.test_target == 'Hexagon':
     clang = 'clang -target hexagon'
+    if not config.target_triple:
+        config.target_triple = "hexagon-unknown-none-elf"
```
quic-seaswara, line 257: Do we need this ?

test/lld/ELF/opt-level.ll
quic-akaryaki: > %opt

Done

quic-akaryaki: I don't think we need to add dependencies between projects. It may be far fetched but we should allow people to build eld outside of LLVM cmake tree.

test/lit.cfg
quic-akaryaki: This should set a "default" triple in case a particular test is not tightly tied to a specific triple. In case of multi-target builds, the default triple should be set from the config files. But if it's still not set, we need some default value.

quic-seaswara (approved):

PR #717, opened 2026-01-16
Fix ELDExpected test windows failure

Ris-Bali:
This test fixes the eldexpected test windows failure occuring due to presence of the binary in the config subdirectory in case of multi-config setups.


quic-seaswara (approved):
Please take an action item on identifying why we need do things differently only for ELDExpected. 

I will approve this to be merged for now, but the root cause needs to be identified.

PR #729, opened 2026-01-23
Fix windows nightly failures

Ris-Bali:
This patch fixes windows nightly failures related to `/dev/null` path issue and multiconfig build related path issue.

### DIFF
--- a/test/Hexagon/Plugin/ConfigFile/DefaultINIConfigFilePathTest/DefaultINIConfigFilePathTest.test
+++ b/test/Hexagon/Plugin/ConfigFile/DefaultINIConfigFilePathTest/DefaultINIConfigFilePathTest.test
@@ -1,3 +1,4 @@
+#UNSUPPORTED: windows
 #---DefaultINIConfigFilePathTest.test----------------------- Executable,LS --------------------#
 #BEGIN_COMMENT
 # Tests that plugins are able to load .config from search dirs and .ini files from default path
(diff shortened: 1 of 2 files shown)


test/Hexagon/Plugin/ConfigFile/DefaultINIConfigFilePathTest/DefaultINIConfigFilePathTest.test
```diff
@@ -1,3 +1,4 @@
+#UNSUPPORTED: windows
```
quic-akaryaki, line 1: Why is it not working on windows? Is there a plan to fix the root cause and reenable it?

test/lld/ELF/new-pass-manager.ll
```diff
@@ -1,8 +1,8 @@
 ; RUN: %opt --mtriple=%triple --data-layout=%datalayout %s -o %t.o
 
 ; Test debug-pass-manager option
-; RUN: %link %linkopts --plugin-opt=debug-pass-manager -o /dev/null %t.o 2>&1 | FileCheck %s
-; RUN: %link %linkopts --lto-debug-pass-manager -o /dev/null %t.o 2>&1 | FileCheck %s
+; RUN: %link %linkopts --plugin-opt=debug-pass-manager -o %t.temp %t.o 2>&1 | FileCheck %s
+; RUN: %link %linkopts --lto-debug-pass-manager -o %t.temp %t.o 2>&1 | FileCheck %s
```
quic-akaryaki, line 5: Do we have other tests that use /dev/null as output path?

PR #736, opened 2026-01-23
Use llvm-project@main instead of BASE_BRANCH_NAME

navaneethshan:
Stops failures caused by BASE_BRANCH_NAME pointing to ELD-only branches during manual workflow_dispatch runs. ELD continues to use BASE_BRANCH_NAME normally.

### DIFF
--- a/.github/workflows/ci-win.yml
+++ b/.github/workflows/ci-win.yml
@@ -26,7 +26,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: llvm/llvm-project
-          ref: ${{ env.BASE_BRANCH_NAME }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-win.yml
```diff
@@ -26,7 +26,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: llvm/llvm-project
-          ref: ${{ env.BASE_BRANCH_NAME }}
+          ref: main # Update on release branches
```
partaror, line 29: @quic-seaswara What do you think about adding a weekly windows builder for the release branches?

Issue #737, opened 2026-01-25
eld fails to build/install with CPack


jonathonpenix:
**Description**
[CPack](https://cmake.org/cmake/help/book/mastering-cmake/chapter/Packaging%20With%20CPack.html) is a packaging tool commonly used with cmake (used by ATfE, Zephyr's LLVM SDK, soon-to-be-used by cpullvm). Right now, trying to use CPack to package something including eld results in failures like below:

```
CPack: - Install project: example []
CMake Error at /local/mnt/workspace/jpenix/toolchains/eld/llvm-project/eld/tools/eld/ld_eld_symlink.cmake:5 (include):
  include could not find requested file:

    /tools/eld/eld_install.cmake
```
The issue here is that ld_eld_symlink.cmake uses `${CMAKE_BINARY_DIR}`, but that isn't defined by default when CPack is running, so we end up with a broken path. It seems this is (sort of, at least) expected behavior: https://discourse.cmake.org/t/cpack-script-not-recognizing-cmake-binary-dir/2191

I think there's a few other variables that might have the same issue there.

**Reproducer**
The below should fail with the error shown above. Removing the eld bits from the top-level CMakeLists.txt will let the package step succeed.
```
cat > CMakeLists.txt << '!'
cmake_minimum_required(VERSION 3.16)

project(example)

set(LLVM_ENABLE_PROJECTS clang;lld CACHE STRING "")
set(LLVM_TARGETS_TO_BUILD AArch64;ARM;RISCV CACHE STRING "")
set(LLVM_ENABLE_LIBCXX ON CACHE BOOL "")

set(llvmproject_src_dir ${CMAKE_CURRENT_SOURCE_DIR}/llvm-project)
set(LLVM_EXTERNAL_PROJECTS eld)
set(LLVM_EXTERNAL_ELD_SOURCE_DIR ${llvmproject_src_dir}/eld)

add_subdirectory(
    ${llvmproject_src_dir}/llvm llvm
)

include(CPack)
!

cmake -G Ninja \
      -DCMAKE_BUILD_TYPE=Release \
      -DCMAKE_C_COMPILER_LAUNCHER=ccache \
      -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
      -DCMAKE_INSTALL_PREFIX=$PWD/install \
      -B $PWD/build \
      -S .

cmake --build build
# (or, directly invoke `cpack -G...`, etc.)
ninja -C build/ package
```

**Thoughts**
[...]

Issue #710, opened 2026-01-12
ELDExpectedUsage test binary, target-specific wrappers end up in the wrong build location when LLVM is invoked as a subproject


jonathonpenix:
**Description**
When llvm is invoked as a subproject, the target-specific wrappers (riscv-link, etc.) and `tests/ELDExpectedUsage` end up in unexpected places for the build (the final install looks ok I think). See the example below (the linker wrappers end up in `./build/bin` rather than `./build/llvm/bin`)

This isn't the end of the world (I don't think the target-specific wrappers being in a different spot causes any direct issues besides being a bit confusing), but does cause test failures for ELDExpectedUsage (see Comments section below).

Invoking LLVM as a subproject can be helpful when assembling a full toolchain--this is how ex: [Arm's toolchain](https://github.com/arm/arm-toolchain/blob/6066f7868a93f108b9f0d7f9e355623a4a75724f/arm-software/embedded/CMakeLists.txt#L368-L370) and the [Zephyr SDK](https://github.com/zephyrproject-rtos/sdk-ng/blob/f3febf2d6888bdb603b8d6f12be49f5d9ec6e9b8/scripts/llvm/CMakeLists.txt#L320-L322) (which is based off an older version of the Arm toolchain) work. We've also considered similar things for cpullvm. So, I think it is might be worth trying to fix this to support this sort of use case.

**Reproducer**
```
cat > CMakeLists.txt << '!'
cmake_minimum_required(VERSION 3.16)

project(example)

set(LLVM_ENABLE_PROJECTS clang;lld CACHE STRING "")
set(LLVM_TARGETS_TO_BUILD AArch64;ARM;RISCV CACHE STRING "")
set(LLVM_ENABLE_LIBCXX ON CACHE BOOL "")

set(llvmproject_src_dir ${CMAKE_CURRENT_SOURCE_DIR}/llvm-project)
set(LLVM_EXTERNAL_PROJECTS eld)
set(LLVM_EXTERNAL_ELD_SOURCE_DIR ${llvmproject_src_dir}/eld)

add_subdirectory(
    ${llvmproject_src_dir}/llvm llvm
)
!

cmake -G Ninja \
     -DCMAKE_BUILD_TYPE=Release \
     -DCMAKE_C_COMPILER_LAUNCHER=ccache \
     -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
     -DCMAKE_INSTALL_PREFIX=$PWD/install \
     -B $PWD/build \
[...]

PR #738, opened 2026-01-25
Create symlinks via `llvm_add_tool_symlink`/`llvm_install_symlink`

jonathonpenix:
eld's symlink creation currently has a few issues/quirks in how it interacts with CPack and where symlinks are placed in builds [1][2].

I think the easiest way to fix these issues is just to leverage LLVM's cmake utilities `llvm_add_tool_symlink` and `llvm_install_symlink`. Also, I think there's a few other advantages:
- The end result (created symlinks or copied/renamed binaries) should be the same as before while allowing us to simplify eld's cmake. We are already tightly coupled to LLVM/LLVM's cmake anyway
- eld should behave a bit more consistently to other LLVM tools now--for example, if someone does want to force symlinks on Windows in LLVM, eld will do the same

The one (maybe negative, intended) functional change here is that the symlinks now won't have their own targets. But:
- It isn't clear to me how helpful that it is to have per-symlink targets (any/all symlinks will be built/installed with ld.eld)
- This should be similar to how ex: lld's symlinks work
- Worst case, I think we can add them back without reverting any of this if needed So, I omitted them for now.

[1] https://github.com/qualcomm/eld/issues/737
[2] https://github.com/qualcomm/eld/issues/710


jonathonpenix:
FWIW, we have been testing cpullvm with this (and the other open cmake-y patches) for a little bit now. If you want, you can check out a builds with these applied here: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/21575041496

quic-seaswara (approved):

PR #740, opened 2026-01-26
Add basic LLVM_DISTRIBUTION_COMPONENTS compatibility with a unified component

jonathonpenix:
This is an alternate implementation of https://github.com/qualcomm/eld/pull/694 with everything unified under one component.

Note that there are additional commits doing the cleanup mentioned in https://github.com/qualcomm/eld/pull/694 (making it possible to not install the 'templates/' dir, always enabling the PluginHeaders). These are included here as I think it is worth deciding on what can/can't be omitted from the install/distribution (whether that be via individual components, or CMake options). But, they can be separate PRs--this is just for convenience.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -205,7 +205,7 @@ if(ELD_INSTALL_LINKER_SCRIPT_TEMPLATES)
   install(
     DIRECTORY templates/
     DESTINATION templates
-    COMPONENT linker-script
+    COMPONENT ld.eld
     FILES_MATCHING
     PATTERN "*.template"
     PATTERN ".git" EXCLUDE
@@ -223,5 +223,6 @@ if(ELD_INSTALL_YAML_MAP_PARSER)
     set(ELD_YAML_MAP_PARSER_DEST_DIR "bin")
   endif()
   install(PROGRAMS utils/YAMLMapParser/YAMLMapParser.py
-          DESTINATION ${ELD_YAML_MAP_PARSER_DEST_DIR})
+          DESTINATION ${ELD_YAML_MAP_PARSER_DEST_DIR}
+          COMPONENT ld.eld)
 endif()
--- a/include/eld/CMakeLists.txt
+++ b/include/eld/CMakeLists.txt
@@ -7,7 +7,7 @@ configure_file(${CMAKE_CURRENT_SOURCE_DIR}/PluginAPI/PluginBase.h.inc
 install(
   DIRECTORY PluginAPI/
   DESTINATION include/ELD/PluginAPI/
-  COMPONENT PluginAPIHeaders
+  COMPONENT ld.eld
   FILES_MATCHING
   PATTERN "*.h"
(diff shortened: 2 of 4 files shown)


CMakeLists.txt
```diff
@@ -200,14 +200,17 @@ add_subdirectory(docs)
...
 
-install(
-  DIRECTORY templates/
-  DESTINATION templates
-  COMPONENT linker-script
-  FILES_MATCHING
-  PATTERN "*.template"
-  PATTERN ".git" EXCLUDE
-  PATTERN ".svn" EXCLUDE)
+option(ELD_INSTALL_LINKER_SCRIPT_TEMPLATES "Include linker script templates in the install" ON)
```
quic-akaryaki, line 203: Are you going to override this option in non-Hexagon builds? Why wouldn't just conditionally enable templates installation based on whether Hexagon is in the target list?

jonathonpenix: > Are you going to override this option in non-Hexagon builds?

For cpullvm at least, that is my current plan if we go this route.

> Why wouldn't just conditionally enable templates installation based on whether Hexagon is in the target list?

I don't have a particularly strong reason. I guess my thought is that today, everyone gets them regardless of target and I don't know who does/doesn't find them helpful. Historically, our Arm/RISC-V toolchains have stripped them out from our releases (so, I'm continuing down that route), but for a broader group maybe it is helpful for someone even if they aren't using Hexagon (just to have a reference linker script of any kind), maybe someone with Hexagon enabled wouldn't otherwise want these (I haven't looked at any hexagon toolchains to see if these are included or are similarly stripped out). So it seemed safer to just let it be on-by-default and people can toggle it off as desired.

quic-akaryaki: What if there were ARM/RISC-V linker scripts? I guess there are two things: 1) someone is not interested in in eld linkers scripts 2) someone is distributing a linker that does not target hexagon. I think your current option works for case (1) but I think your reasoning is more about option (2). Fine with me both ways.

jonathonpenix: > What if there were ARM/RISC-V linker scripts?

I'm not sure--I guess it depends on what the linker scripts were. I think it can be helpful to have something people can reference (even if it isn't directly for the arch they're targeting!), but I think I'd be more interested in something like what arm does where it provides [full samples that are expected to work](https://github.com/arm/arm-toolchain/tree/arm-software/arm-software/embedded/samples). But at that point, the reference linker scripts that'd we'd include are probably coming from picolibc itself (or derivatives for other libcs), not eld.

>I guess there are two things: 1) someone is not interested in in eld linkers scripts 2) someone is distributing a linker that does not target hexagon. I think your current option works for case (1) but I think your reasoning is more about option (2)

That's fair, I think for me it is essentially "Because I'm in case (2), then case (1)". Plus, some historical baggage I'm following (I don't know why we originally started removing the templates in our releases in the first place)

Issue #540, opened 2025-11-06
Make builds recorder and dashboard JS generic


quic-shaksuma:
Make builds recorder and dashboard JS generic

PR #743, opened 2026-01-26
Add filter selector check-box for build charts.

### LINKED ISSUE #541: Add individual filter selector in dashboard
Add individual filter selector for each workflow.
This will allow date filter for example to be applied only on selected workflow's pie chart.

quic-shaksuma:
This commit adds individual filter application check-box for each build status chart.

Closes #541

### DIFF
--- a/dash/dash.html
+++ b/dash/dash.html
@@ -29,12 +29,12 @@
   <div class="all-builds flex-parent">
     <div class="Build1 flex-child" style="max-width:400px;border:2px solid skyblue;border-style:inset hidden hidden inset;border-radius: 18px;">
       <canvas id="picoPie"></canvas><br>
-      &nbsp;&nbsp;&nbsp;&nbsp;Picolibc Status <sub><span id="pico_green" class="status status-pass"></span><span id="pico_red" class="status status-fail"></span></sub>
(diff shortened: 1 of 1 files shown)


dash/dash.html
```diff
@@ -164,15 +164,19 @@ <h1 id='page-title'>Build Status</h1>
 
     function filterPies(start_date, end_date) {
       // Picolibc
-      let newPicoData = getUpdatedData(picolibc_build_states, start_date, end_date)
-      console.log("Updating Picolibc Pie Chart for Selected Date Range.");
-      pico_pie.data.datasets[0].data = [newPicoData[0], newPicoData[1]];
-      pico_pie.update();
+      if ($('#select_picolibc').prop('checked')) {
```
partaror, line 167: Would we need to add an entry here for each workflow type such as picolibc, musl, ...? If so, wouldn't this lead to too much code duplication?

quic-shaksuma: Yes, this needs cleanup, there is another issue for cleaning the code [#540](https://github.com/qualcomm/eld/issues/540) and that will be in another PR.
Although, the code duplication can only be avoided to a limit as the dashboard does not use a server to generate dynamic renderings.

This commit just adds the filter selection to existing layout.

partaror: Sounds good!

partaror (approved):

PR #752, opened 2026-01-28
Fix SUBALIGN linker script feature

### LINKED ISSUE #343: linkerscript SUBALIGN applies to each input section
SUBALIGN may be used to adjust the alignment of each input section

```
cat > script.t << '!'
SECTIONS {
  text (0x14) : SUBALIGN(64) {
    *(.text*)
  }
}
!

cat > test.c << '!'
int foo() { return 1; }
int bar() { return 0; }
!

clang -falign-functions=4 --target=riscv32-unknown-elf -c test.c -ffunction-sections
ld.eld test.o -T script.t -Map map.txt -o eld.out
```

**Bug** 

```
$ llvm-readelf -s eld.out | grep -E 'foo|bar'
     9: 00000014    18 FUNC    GLOBAL DEFAULT     1 foo
    10: 00000028    18 FUNC    GLOBAL DEFAULT     1 bar
```

**Correct Behavior**

```
$ llvm-readelf -s bfd.out | grep -E 'foo|bar'
     9: 00000040    18 FUNC    GLOBAL DEFAULT     1 foo
    10: 00000080    18 FUNC    GLOBAL DEFAULT     1 bar
```

parth-07:
This commit fixes the SUBALIGN linker script feature. SUBALIGN feature is used to modify the alignment of all input sections present in an output section. Until now, we were only changing the alignment of the first fragment of each rule of the output sections with SUBALIGN specified. This commit fixes the SUBALIGN behavior.

SUBALIGN affects the alignment of the first fragment as is seen by the linker when traversing the output sections for the input sections that may contain more than 1 fragment such as .eh_frame, PLT / GOT sections, and .comment section.

Resolves #343

### DIFF
--- a/include/eld/Diagnostics/DiagLDScript.inc
+++ b/include/eld/Diagnostics/DiagLDScript.inc
@@ -124,6 +124,8 @@ DIAG(error_insert_output_section, DiagnosticEngine::Error,
      "Cannot INSERT output section %0 %1 %2 in script %3")
 DIAG(warn_non_power_of_2_value_to_align_builtin, DiagnosticEngine::Warning,
      "%0: non-power-of-2 value 0x%1 passed to ALIGN builtin function")
+DIAG(warn_subalign_less_than_section_alignment, DiagnosticEngine::Warning,
+     "SUBALIGN(0x%0) is less than the section alignment (0x%1) for section '%2'")
 DIAG(error_non_power_of_2_value_to_align_output_section, DiagnosticEngine::Error,
      "%0: non-power-of-2 value 0x%1 passed to ALIGN in '%2' output section description, value must "
      "be 0 or a power of 2")
--- a/lib/Object/ObjectLinker.cpp
+++ b/lib/Object/ObjectLinker.cpp
@@ -64,6 +64,7 @@
 #include "eld/SymbolResolver/ResolveInfo.h"
 #include "eld/Target/ELFFileFormat.h"
 #include "eld/Target/GNULDBackend.h"
+#include "eld/Target/LDFileFormat.h"
 #include "eld/Target/Relocator.h"
 #include "llvm/ADT/ScopeExit.h"
 #include "llvm/ADT/StringSwitch.h"
@@ -84,6 +85,7 @@
 #include <chrono>
 #include <mutex>
 #include <sstream>
+#include <unordered_set>
 
 using namespace llvm;
 using namespace eld;
@@ -1002,7 +1004,7 @@ bool ObjectLinker::sortSections(RuleContainer *I, bool SortRule) {
 bool ObjectLinker::createOutputSection(ObjectBuilder &Builder,
                                        OutputSectionEntry *Output,
                                        bool PostLayout) {
-  uint64_t OutAlign = 0x0, InAlign = 0x0;
+  uint64_t OutAlign = 0x0;
   bool IsPartialLink = (LinkerConfig::Object == ThisConfig.codeGenType());
 
   ELFSection *OutSect = Output->getSection();
@@ -1012,15 +1014,6 @@ bool ObjectLinker::createOutputSection(ObjectBuilder &Builder,
   OutputSectionEntry::iterator In, InBegin, InEnd;
   InBegin = Output->begin();
   InEnd = Output->end();
-  bool HasSubAlign = false;
-
-  // force input alignment from ldscript if any
-  if (Output->prolog().hasSubAlign()) {
-    Output->prolog().subAlign().eval();
-    Output->prolog().subAlign().commit();
-    InAlign = Output->prolog().subAlign().result();
-    HasSubAlign = true;
-  }
 
   // force output alignment from ldscript if any
   if (Output->prolog().hasAlign()) {
@@ -1061,12 +1054,6 @@ bool ObjectLinker::createOutputSection(ObjectBuilder &Builder,
         Builder.updateSectionFlags(InSect, F->getOwningSection());
       }
       InSect->setAddrAlign(Alignment);
-    }
-    if (HasSubAlign && (InSect->getAddrAlign() < InAlign)) {
-      if (InSect->getFragmentList().size()) {
-        InSect->getFragmentList().front()->setAlignment(InAlign);
-        InSect->setAddrAlign(InAlign);
-      }
     }
     if (InSect->getFragmentList().size() && !FirstNonEmptyRule)
       FirstNonEmptyRule = *In;
@@ -1323,6 +1310,9 @@ bool ObjectLinker::mergeSections() {
   {
     eld::RegisterTimer T("Create Output Section", "Merge Sections",
                          ThisConfig.options().printTimingStats());
+    // Prepass: apply SUBALIGN to first fragments per input section per output
+    // section serially to avoid races in parallel output creation.
+    applySubAlign();
     std::vector<OutputSectionEntry *> OutSections;
     for (Out = OutBegin; Out != OutEnd; ++Out) {
       OutSections.push_back(*Out);
@@ -1364,6 +1354,41 @@ bool ObjectLinker::initializeOutputSectionsAndRunPlugin() {
       sortSections(In, false);
   }
   return runOutputSectionIteratorPlugin();
+}
+
+void ObjectLinker::applySubAlign() {
+  std::unordered_set<ELFSection *> seen;
+
+  for (auto *O : ThisModule->getScript().sectionMap()) {
+    auto &prolog = O->prolog();
+    if (!prolog.hasSubAlign())
+      continue;
+    uint64_t subAlign;
+    prolog.subAlign().eval();
+    prolog.subAlign().commit();
+    subAlign = prolog.subAlign().result();
+
+    for (RuleContainer *R : *O) {
+      ELFSection *inSect = R->getSection();
+      for (Fragment *F : inSect->getFragmentList()) {
+        ELFSection *owningSect = F->getOwningSection();
+        if (owningSect->getKind() == LDFileFormat::Kind::OutputSectData) {
+          continue;
+        }
+        if (owningSect && seen.insert(owningSect).second) {
+          // Warn if SUBALIGN is reducing the section alignment
+          if (ThisConfig.showLinkerScriptWarnings() && F->alignment() > subAlign) {
+            ThisConfig.raise(Diag::warn_subalign_less_than_section_alignment)
+                << utility::toHex(subAlign) << utility::toHex(F->alignment())
+                << owningSect->getLocation(0, ThisConfig.options());
+          }
+          F->setAlignment(subAlign);
+          inSect->setAddrAlign(std::max(
+              static_cast<uint64_t>(inSect->getAddrAlign()), subAlign));
+        }
+      }
+    }
+  }
 }
 
 void ObjectLinker::assignOffset(OutputSectionEntry *Out) {
--- a/include/eld/Object/ObjectLinker.h
+++ b/include/eld/Object/ObjectLinker.h
@@ -323,6 +323,8 @@ public:
 
   bool initializeOutputSectionsAndRunPlugin();
 
+  void applySubAlign();
+
   // Get Plugin list for Relocation registration callback.
   LinkerScript::PluginVectorT getLinkerPluginWithLinkerConfigs();
 
--- a/lib/Fragment/StringFragment.cpp
+++ b/lib/Fragment/StringFragment.cpp
@@ -35,7 +35,10 @@ eld::Expected<void> StringFragment::emit(MemoryRegion &Mr, Module &M) {
     return {};
   Out = Mr.begin() + getOffset(M.getConfig().getDiagEngine()) -
         this->paddingSize();
-  uint64_t PaddingValue = *M.getFragmentPaddingValue(this);
+  std::optional<uint64_t> optPaddingValue = M.getFragmentPaddingValue(this);
+  if (!optPaddingValue)
+    return {};
+  uint64_t PaddingValue = *optPaddingValue;
   uint32_t ValueSize = Fragment::getPaddingValueSize(PaddingValue);
   if (PaddingValue == 0)
     return {};
--- a/lib/LayoutMap/TextLayoutPrinter.cpp
+++ b/lib/LayoutMap/TextLayoutPrinter.cpp
@@ -109,6 +109,10 @@ void TextLayoutPrinter::printOnlyLayoutSection(GNULDBackend const &Backend,
   outputStream().write_hex(Section->size());
   outputStream() << "\t# Alignment: 0x";
   outputStream().write_hex(Section->getAddrAlign());
+  if (OS->prolog().hasSubAlign()) {
+    outputStream() << ", SubAlign: 0x";
+    outputStream().write_hex(OS->prolog().subAlign().resultOrZero());
+  }
   if (Section->isAlloc())
     printSegments(Backend, OS);
   outputStream() << "\n";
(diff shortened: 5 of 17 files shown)


lib/Object/ObjectLinker.cpp
```diff
@@ -1291,6 +1295,36 @@ bool ObjectLinker::initializeOutputSectionsAndRunPlugin() {
...
+    }
+
+    // Track input sections seen for this output.
+    static std::unordered_set<ELFSection *> seen;
+
+    for (RuleContainer *R : *O) {
+      ELFSection *InSect = R->getSection();
+      for (Fragment *F : InSect->getFragmentList()) {
+        ELFSection *owningSect = F->getOwningSection();
+        if (owningSect && seen.insert(owningSect).second) {
```
quic-areg, line 1378: early return if either of these are false?

parth-07: There is no statement after this if-command, and thus, I am not sure if early return would be beneficial here.

lib/Object/ObjectLinker.cpp
```diff
@@ -1291,6 +1295,33 @@ bool ObjectLinker::initializeOutputSectionsAndRunPlugin() {
...
+  for (auto *O : ThisModule->getScript().sectionMap()) {
+    auto &Prolog = O->prolog();
+    std::optional<uint64_t> optSubAlign;
+    if (Prolog.hasSubAlign()) {
+      Prolog.subAlign().eval();
+      Prolog.subAlign().commit();
+      optSubAlign = Prolog.subAlign().result();
+    }
+
+    for (RuleContainer *R : *O) {
```
quic-seaswara, line 1371: if (!Prolog.hasSubAlign())
   continue;

lib/Object/ObjectLinker.cpp
```diff
@@ -1291,6 +1295,33 @@ bool ObjectLinker::initializeOutputSectionsAndRunPlugin() {
...
+      ELFSection *InSect = R->getSection();
+      for (Fragment *F : InSect->getFragmentList()) {
+        ELFSection *owningSect = F->getOwningSection();
+        if (owningSect && seen.insert(owningSect).second && optSubAlign) {
+          F->setAlignment(optSubAlign.value());
+          InSect->setAddrAlign(
+              std::max(static_cast<uint64_t>(InSect->getAddrAlign()),
+                       optSubAlign.value()));
+        }
```
quic-seaswara, line 1388: Can we do this at the time of calling mergeSections.

I would think we would implement by first 

- inspecting if the output section has subalign 
- update rule container with the maximum alignment of the section

lib/Object/ObjectLinker.cpp
```diff
@@ -1291,6 +1295,33 @@ bool ObjectLinker::initializeOutputSectionsAndRunPlugin() {
...
+      optSubAlign = Prolog.subAlign().result();
+    }
+
+    for (RuleContainer *R : *O) {
+      ELFSection *InSect = R->getSection();
+      for (Fragment *F : InSect->getFragmentList()) {
+        ELFSection *owningSect = F->getOwningSection();
+        if (owningSect && seen.insert(owningSect).second && optSubAlign) {
+          F->setAlignment(optSubAlign.value());
+          InSect->setAddrAlign(
```
quic-seaswara: Should we emit a warning for sections that bumping subalign can cause issues 

Examples of them may be dynamic sections or eh frame sections.

lib/Object/ObjectLinker.cpp
quic-seaswara: You might want also want to make sure the map file is annotated with this information for debugging reasons.

lib/Object/ObjectLinker.cpp
parth-07: We cannot skip output sections that do not have `SUBALIGN`. The way we define `SUBALIGN` functionality in eld is: subalign changes the alignment of the fragments in the output sections that are the first fragments **seen from their corresponding input sections.** in the output section table.

So, if a plugin moves fragments of a section into two output sections, we need to determine which of them appears first in the output section table, and thus we need to go through all the output sections and fragments.

lib/Object/ObjectLinker.cpp
parth-07: > Can we do this at the time of calling mergeSections.

merging input sections handles entirely different responsibility than handling subAlign. Why should we do this together with merging input sections? I would like to follow the unix philosophy, "do one thing well", wherever possible.

> You might want also want to make sure the map file is annotated with this information for debugging reasons.

I will add map-file improvements as a follow-up patch.

lib/Object/ObjectLinker.cpp
parth-07: I am not aware of cases where making alignment larger can cause issues. Why would increasing alignment of .dynamic / .eh_frame sections cause issues?

I think we should emit warning if subalign is used to decrease the alignment of an input section.

lib/Object/ObjectLinker.cpp
quic-seaswara: Merging input sections already takes care of choosing the maximum alignment, and SUBALIGN is just one additional kind for choosing maximum alignment.

I am very concerned about link time. 

* The usage of subalign is very rare and we are iterating through all the output sections for rare use-cases. 
* We can mention as undefined behavior for now if users use SUBALIGN with linker plugin chunk movements. 

Independent of the current approach, this patch traverses through every output section (allocatable, non allocatable sections), which can be improved too. 

Overall comment : If you are covering some use-case, please add tests for all use cases. 

For example 

* DISCARD sections
* Emit relocs 
* Empty sections with SUBALIGN, Sections with explicit data - BYTE/QUAD/... 
* Partial linking
* Shared libraries 
* Non allocatable sections (probably we should warn here and not do anything)

quic-seaswara: I feel like map file improvements need to go in the same patch just because it would be a way to check of what the linker did explicitly.

parth-07: > Merging input sections already takes care of choosing the maximum alignment, and SUBALIGN is just one additional kind for choosing maximum alignment.

Merging input sections only sets the alignment of the rule-container internal input section, it is not responsible of reading linker script directives to change the alignments and it does not change the alignment of fragments / owning input sections.

> I am very concerned about link time.

The link time would be the same whether we apply subalign alignments as part of the merging input sections, or
as a separate step after it. This is because the number of operations remains the same.

> The usage of subalign is very rare and we are iterating through all the output sections for rare use-cases.

It is one pass over the fragments. We have many such passes over the fragments. I do not expect to see any noticeable
time difference. Do you think it would be better to go over the output sections once and check whether there is any output section with subalign, and only traverse fragments if there is indeed a use of subalign in the link?

We can change the code to **only traverse fragments of output sections with SUBALIGN** but that will slightly change the meaning of SUBALIGN for us. Please let me know if that would be fine. 

> We can mention as undefined behavior for now if users use SUBALIGN with linker plugin chunk movements.

Disabling plugin chunk movements depending upon whether SUBALIGN is used or not seems to be a very big restriction to me, and I believe that we should avoid it if possible. Additionally, even without the SUBALIGN, if plugins move the fragments incorrectly, then the image will be invalid and we do not report any warning / error for it.

> Independent of the current approach, this patch traverses through every output section (allocatable, non allocatable sections), which can be improved too.
> Overall comment : If you are covering some use-case, please add tests for all use cases.

I will add more tests, but I think as a functionality-wise, `SUBALIGN` should work for all the output sections. Other similar linker script functionality such as output section alignment can be specified for all the outputs sections as well.

quic-seaswara: yes, that is what I am trying to say do this only for output sections with SUBALIGN. 

Restrict plugins to not move chunks between output sections with SUBALIGN. This is totally ok for me. We can relax this case if we have a genuine requirement. 

It does not make sense to use SUBALIGN on sections such as debug. We should warn on those usecases.

parth-07: > yes, that is what I am trying to say do this only for output sections with SUBALIGN.

I have changed the code to iterate input sections of only output sections that have SUBALIGN.

> Restrict plugins to not move chunks between output sections with SUBALIGN. This is totally ok for me. We can relax this case if we have a genuine requirement.

The issue of incorrectly moving chunks of the same input section is not limited to SUBALIGN usage. If a plugin moves the chunk of the same input section to different output sections, then the image would be wrong even with no SUBALIGN usage.

> It does not make sense to use SUBALIGN on sections such as debug. We should warn on those usecases.

Can we add handle such diagnostic improvements separately? Also, I believe that if there is no harm on using SUBALIGN for non-alloc sections, then we do not need to report a warning for it. We do not report any warning for using ALIGN on non-alloc sections as well.

parth-07: > For example
>DISCARD sections
> Emit relocs
> Empty sections with SUBALIGN, Sections with explicit data - BYTE/QUAD/...
> Partial linking
> Shared libraries
> Non allocatable sections

I have added tests for all these cases. I am not currently warning for non-allocatable sections because we also do not warn if `ALIGN` linker script directive or the output section address linker script directive is used on non-allocatable sections.

I have added a test and a diagnostic for when SUBALIGN reduces an input section alignment.

include/eld/Diagnostics/DiagLDScript.inc
```diff
@@ -122,6 +122,8 @@ DIAG(error_cannot_specify_lma_and_memory_region, DiagnosticEngine::Error,
      "cannot specify AT and LMA memory region for output section : %0 defined in script %1")
 DIAG(warn_non_power_of_2_value_to_align_builtin, DiagnosticEngine::Warning,
      "%0: non-power-of-2 value 0x%1 passed to ALIGN builtin function")
+DIAG(warn_subalign_less_than_section_alignment, DiagnosticEngine::Warning,
+     "SUBALIGN(0x%0) is less than section alignment (0x%1) for section '%2'")
```
quic-seaswara: Add the input file as well.

lib/Object/ObjectLinker.cpp
```diff
@@ -1329,6 +1319,41 @@ bool ObjectLinker::initializeOutputSectionsAndRunPlugin() {
   return runOutputSectionIteratorPlugin();
 }
 
+void ObjectLinker::applySubAlign() {
+  std::unordered_set<ELFSection *> seen;
```
quic-seaswara, line 1360: This variable can moved to line 1329 ?

lib/Object/ObjectLinker.cpp
```diff
@@ -1329,6 +1319,41 @@ bool ObjectLinker::initializeOutputSectionsAndRunPlugin() {
...
+      ELFSection *inSect = R->getSection();
+      for (Fragment *F : inSect->getFragmentList()) {
+        ELFSection *owningSect = F->getOwningSection();
+        if (owningSect->getKind() == LDFileFormat::Kind::OutputSectData) {
+          continue;
+        }
+        if (owningSect && seen.insert(owningSect).second) {
+          // Warn if SUBALIGN is reducing the section alignment
+          if (ThisConfig.showLinkerScriptWarnings() && F->alignment() > subAlign) {
+            ThisConfig.raise(Diag::warn_subalign_less_than_section_alignment)
```
quic-seaswara, line 1381: Should this be an error ?

lib/Object/ObjectLinker.cpp
```diff
@@ -1329,6 +1319,41 @@ bool ObjectLinker::initializeOutputSectionsAndRunPlugin() {
   return runOutputSectionIteratorPlugin();
 }
 
+void ObjectLinker::applySubAlign() {
+  std::unordered_set<ELFSection *> seen;
+
+  for (auto *O : ThisModule->getScript().sectionMap()) {
```
quic-seaswara, line 1362: follow up : multithread by output section

lib/Object/ObjectLinker.cpp
quic-seaswara: I agree we dont do this, but its an area that we can make the user understand what he might be doing or improve his build.

We should try and improve, but I agree this can be a follow up too.

lib/Object/ObjectLinker.cpp
```diff
@@ -1329,6 +1319,41 @@ bool ObjectLinker::initializeOutputSectionsAndRunPlugin() {
...
+    prolog.subAlign().eval();
+    prolog.subAlign().commit();
+    subAlign = prolog.subAlign().result();
+
+    for (RuleContainer *R : *O) {
+      ELFSection *inSect = R->getSection();
+      for (Fragment *F : inSect->getFragmentList()) {
+        ELFSection *owningSect = F->getOwningSection();
+        if (owningSect->getKind() == LDFileFormat::Kind::OutputSectData) {
+          continue;
```
quic-seaswara, line 1376: awesome!

Should we do this for internal sections ?

lib/Object/ObjectLinker.cpp
parth-07: We cannot do this computation multi-threaded, otherwise the subalign behavior will become non-deterministic for the input sections which consist of multiple fragments.

lib/Object/ObjectLinker.cpp
parth-07: No, it has to be before the `for (auto : sectionMap())` loop

lib/Object/ObjectLinker.cpp
parth-07: No, as per GNU ld docs (https://sourceware.org/binutils/docs/ld/Forced-Input-Alignment.html), SUBALIGN is allowed to reduce the input section alignment.

lib/Object/ObjectLinker.cpp
parth-07: We may need to take this as case-by-case basis. Most internal sections do indeed represent a section and thus `SUBALIGN` should affect them. `OutputSectData` is an exception because it is not a section at all as per GNU ld docs.

include/eld/Diagnostics/DiagLDScript.inc
parth-07: Thank you for the suggestion. I have improved this diagnostic.

include/eld/Diagnostics/DiagLDScript.inc
```diff
@@ -122,6 +122,8 @@ DIAG(error_cannot_specify_lma_and_memory_region, DiagnosticEngine::Error,
      "cannot specify AT and LMA memory region for output section : %0 defined in script %1")
 DIAG(warn_non_power_of_2_value_to_align_builtin, DiagnosticEngine::Warning,
      "%0: non-power-of-2 value 0x%1 passed to ALIGN builtin function")
+DIAG(warn_subalign_less_than_section_alignment, DiagnosticEngine::Warning,
+     "SUBALIGN(0x%0) is less than the section alignment (0x%1) for section '%2(%3)'")
```
quic-areg: can we use `ELFSection::getLocation` for diagnostics for sections? it will conveniently print the input file too, and section offset in hex if any

parth-07: Thank you for the suggestion. I have changed the code to use `ELFSection::getLocation`

lib/Object/ObjectLinker.cpp
```diff
@@ -1356,6 +1346,41 @@ bool ObjectLinker::initializeOutputSectionsAndRunPlugin() {
...
+        }
+        if (owningSect && seen.insert(owningSect).second) {
+          // Warn if SUBALIGN is reducing the section alignment
+          if (ThisConfig.showLinkerScriptWarnings() && F->alignment() > subAlign) {
+            ThisConfig.raise(Diag::warn_subalign_less_than_section_alignment)
+                << utility::toHex(subAlign) << utility::toHex(F->alignment())
+                << owningSect->getLocation(0, ThisConfig.options());
+          }
+          F->setAlignment(subAlign);
```
quic-seaswara, line 1385: We should do 

F->setOwningSection->setAlignment(subAlign)  right ?

partaror: No. `Fragment::paddingSize()` is the key function that determines a fragment's padding requirements and it uses Fragment's alignment instead of the owning input section alignment. In the current design, it seems that we prefer fragment's alignment over the input section's alignment in the layout phase.

PR #756, opened 2026-01-28
[NFC] Store MergeableStrings by value

quic-areg:
Avoids making N heap allocations for N strings and provides better cache locality using values instead of pointers.

`make<T>` is also not thread safe, so eliminating those calls will allow mergeable string sections to be read and split in parallel.


quic-seaswara:
What is the advantage of storing by value ? Can you add some information in the commit message description or the PR on what we are trying to achieve ?

quic-areg:
> What is the advantage of storing by value ? Can you add some information in the commit message description or the PR on what we are trying to achieve ?

updated the description.

PR #763, opened 2026-01-29
[NFC] Fix release build warnings from compiled-out ASSERTs

quic-areg:
`ASSERT`s are now no-ops in release builds, leading to new warnings such as unused variables for assert-only locals, or missing return paths from unreachable sections previously denoted by `ASSERT(0)`.

### DIFF
--- a/lib/LinkerWrapper/DWARF.cpp
+++ b/lib/LinkerWrapper/DWARF.cpp
@@ -270,9 +270,13 @@ bool getSize(llvm::DWARFDie &Die, uint64_t &Size, uint32_t PointerSize) {
     Size = ArrayCount * BaseTypeSize;
     return true;
   }
-  default:
-    ASSERT(false, "found unknown type " + TagString(Die.getTag()).str() +
-                      " @Offset 0x" + llvm::utohexstr(Die.getOffset(), true));
+  default: {
+    std::string Msg =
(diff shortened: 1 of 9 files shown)


quic-seaswara:
Can we do #ifndef NDEBUG instead ?

lib/LinkerWrapper/DWARF.cpp
```diff
@@ -270,9 +270,13 @@ bool getSize(llvm::DWARFDie &Die, uint64_t &Size, uint32_t PointerSize) {
...
-  default:
-    ASSERT(false, "found unknown type " + TagString(Die.getTag()).str() +
-                      " @Offset 0x" + llvm::utohexstr(Die.getOffset(), true));
+  default: {
+    std::string Msg =
+        (llvm::Twine("found unknown type ") + TagString(Die.getTag()) +
+         " @Offset 0x" + llvm::utohexstr(Die.getOffset(), true))
+            .str();
+    llvm_unreachable(Msg.c_str());
+  }
```
quic-seaswara, line 279: pls clang-format

quic-areg: it's already clang-formatted

quic-seaswara: oh

PR #765, opened 2026-01-29
Revert "Eval AFTER_SECTIONS assignments after assignments within SECTIONS cmd"

parth-07:
This reverts commit 2a8d83ddc1b2c4556bd61168eb416395d0561872.

This commit is getting reverted because it caused the regression #769


quic-seaswara (commented):
can you test and make sure that there are no other patches that need to be reverted inorder to merge this.

Please open a issue on why you reverting and tie this revert to the issue.

PR #772, opened 2026-01-30
Fix crash when AArch64 objects contain unsupported relocations.

nezetic:
Due to a missing assert, eld crashes when trying to link AArch64 objects including unsupported relocations.

This commit add the missing assert, as well as two additional checks on the size (that triggered when the size was pointing to invalid data in memory).

A unit test is making sure eld is asserting on invalid/unsupported relocations.

### DIFF
--- a/lib/Readers/Relocation.cpp
+++ b/lib/Readers/Relocation.cpp
@@ -39,8 +39,10 @@ Relocation *Relocation::Create() { return make<Relocation>(0, nullptr, 0, 0); }
 Relocation *Relocation::Create(Type pType, Size pSize, FragmentRef *pFragRef,
                                Address pAddend) {
   DWord targetData = 0;
+  pSize /= 8;
+  ASSERT(pSize <= sizeof(targetData), "unexpected relocation size.");
   if (pSize != 0)
-    pFragRef->memcpy(&targetData, pSize / 8);
+    pFragRef->memcpy(&targetData, pSize);
   return make<Relocation>(pType, pFragRef, pAddend, targetData);
 }
 
@@ -64,7 +66,9 @@ Relocation::Relocation(const Relocator *pRelocator, Relocation::Type pType,
                        FragmentRef *pTargetRef, Relocation::Address pAddend)
     : m_pSymInfo(nullptr), m_TargetAddress(pTargetRef), m_Addend(pAddend),
       m_TargetData(0), m_Type(pType) {
-  Size relocSize = pRelocator->getSize(pType);
+  // Size is returned in bits
+  Size relocSize = pRelocator->getSize(pType) / 8;
+  ASSERT(relocSize <= sizeof(m_TargetData), "unexpected relocation size.");
   if (relocSize != 0)
     pTargetRef->memcpy(&m_TargetData, relocSize);
 }
--- a/lib/Target/AArch64/AArch64Relocator.cpp
+++ b/lib/Target/AArch64/AArch64Relocator.cpp
@@ -134,6 +134,17 @@ AArch64Relocator::AArch64Relocator(AArch64GNUInfoLDBackend &pParent,
 
 AArch64Relocator::~AArch64Relocator() {}
 
+bool AArch64Relocator::isRelocSupported(Relocation &pReloc) const {
+  Relocation::Type type = pReloc.type();
+  // valid types are 0x0, 0x100-0x239
+  if ((type < 0x100 || type > 0x239) && (type != 0x0) &&
+      (type != R_AARCH64_COPY_INSN))
+    return false;
+
+  assert(ApplyFunctions.find(type) != ApplyFunctions.end());
+  return true;
+}
+
 bool AArch64Relocator::isInvalidReloc(Relocation &pReloc) const {
   if (!config().isCodeIndep())
     return false;
@@ -151,14 +162,8 @@ bool AArch64Relocator::isInvalidReloc(Relocation &pReloc) const {
 }
 
 Relocator::Result AArch64Relocator::applyRelocation(Relocation &pRelocation) {
-  Relocation::Type type = pRelocation.type();
-
-  // valid types are 0x0, 0x100-0x239
-  if ((type < 0x100 || type > 0x239) && (type != 0x0) &&
-      (type != R_AARCH64_COPY_INSN))
+  if (!isRelocSupported(pRelocation))
     return Relocator::Unknown;
-
-  assert(ApplyFunctions.find(type) != ApplyFunctions.end());
 
   ResolveInfo *symInfo = pRelocation.symInfo();
 
@@ -175,6 +180,7 @@ Relocator::Result AArch64Relocator::applyRelocation(Relocation &pRelocation) {
     }
   }
 
+  Relocation::Type type = pRelocation.type();
   return ApplyFunctions[type].func(pRelocation, *this);
 }
 
@@ -186,6 +192,8 @@ const char *AArch64Relocator::getName(Relocator::Type pType) const {
 uint32_t AArch64Relocator::getNumRelocs() const { return AARCH64_MAXRELOCS; }
 
 Relocator::Size AArch64Relocator::getSize(Relocation::Type pType) const {
+  if (ApplyFunctions.find(pType) == ApplyFunctions.end())
+    return 0;
   return ApplyFunctions[pType].size;
 }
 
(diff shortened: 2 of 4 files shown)


nezetic:
I looked at other targets (Hexagon / RISCV) and implemented the _isRelocSupported_.

Now _getSize_ return 0 (like the RISCV target) if the relocation is unknown, but the linker raise a fatal error earlier anyway (during _scanRelocation_).

I also updated the unit test accordingly.

```bash
$ ../../obj/bin/ld.eld  -march=aarch64 -static unsupported.o -o unsupported.out
Fatal: Found unsupported reloc type 65535 in section .rela.text in input file unsupported.o
Fatal: Linking had errors.
```

quic-seaswara (approved):

test/AArch64/standalone/Relocations/unsupported.test
```diff
@@ -0,0 +1,37 @@
+#---unsupported.test--------------------- SharedLibrary------------------#
```
quic-seaswara: Can you change the name of the test to something like UnsupportedReloc, and change the comment from SharedLibrary to Executable. 

That said, it might be useful to run this test for 

- Partial Linking
- Shared library
- Static executable

and make sure that all of them report the same error.

test/AArch64/standalone/Relocations/unsupported.test
```diff
@@ -0,0 +1,37 @@
...
+#BEGIN_COMMENT
+# Test that linker rejects unsupported relocation types.
+#END_COMMENT
+#START_TEST
+# RUN: yaml2obj %s -o %t.o
+# RUN: not %link %linkopts -march=aarch64 -static %t.o -o %t.out >%t.log 2>&1 || true
```
quic-seaswara: %not

test/AArch64/standalone/Relocations/unsupported.test
```diff
@@ -0,0 +1,37 @@
+#---unsupported.test--------------------- SharedLibrary------------------#
+#BEGIN_COMMENT
+# Test that linker rejects unsupported relocation types.
+#END_COMMENT
+#START_TEST
+# RUN: yaml2obj %s -o %t.o
```
quic-seaswara: %yaml2obj

test/AArch64/standalone/Relocations/unsupported.test
quic-seaswara: Do you need || true ?

quic-seaswara (changes requested):

test/AArch64/standalone/Relocations/unsupported.test
quic-seaswara: Also enhance this test to add a linker script where this section gets discarded. It might be helpful to see if the linker reports a invalid reloc from a discarded section.

lib/Readers/Relocation.cpp
```diff
@@ -39,8 +39,9 @@ Relocation *Relocation::Create() { return make<Relocation>(0, nullptr, 0, 0); }
 Relocation *Relocation::Create(Type pType, Size pSize, FragmentRef *pFragRef,
                                Address pAddend) {
   DWord targetData = 0;
-  if (pSize != 0)
-    pFragRef->memcpy(&targetData, pSize / 8);
+  pSize /= 8;
+  if (pSize != 0 && pSize <= sizeof(targetData))
```
quic-areg: would this extra condition be better as an `assert`?

lib/Target/AArch64/AArch64Relocator.cpp
```diff
@@ -134,6 +134,20 @@ AArch64Relocator::AArch64Relocator(AArch64GNUInfoLDBackend &pParent,
...
+  // valid types are 0x0, 0x100-0x239
+  if ((pType < 0x100 || pType > 0x239) && (pType != 0x0) &&
+      (pType != R_AARCH64_COPY_INSN))
+    return false;
+
+  assert(ApplyFunctions.find(pType) != ApplyFunctions.end());
+  return true;
+}
+
+bool AArch64Relocator::isRelocSupported(Relocation &pReloc) const {
```
quic-areg: is this overload useful?

lib/Target/AArch64/AArch64Relocator.cpp
```diff
@@ -153,12 +167,9 @@ bool AArch64Relocator::isInvalidReloc(Relocation &pReloc) const {
...
   Relocation::Type type = pRelocation.type();
 
-  // valid types are 0x0, 0x100-0x239
-  if ((type < 0x100 || type > 0x239) && (type != 0x0) &&
-      (type != R_AARCH64_COPY_INSN))
+  if (!isRelocSupported(type)) {
     return Relocator::Unknown;
-
-  assert(ApplyFunctions.find(type) != ApplyFunctions.end());
+  }
```
quic-areg: braces can be omitted

lib/Target/AArch64/AArch64Relocator.cpp
```diff
@@ -186,6 +197,8 @@ const char *AArch64Relocator::getName(Relocator::Type pType) const {
 uint32_t AArch64Relocator::getNumRelocs() const { return AARCH64_MAXRELOCS; }
 
 Relocator::Size AArch64Relocator::getSize(Relocation::Type pType) const {
+  if (!isRelocSupported(pType))
```
quic-areg: in the case where there is an unsupported relocation, shouldn't we have errored out at this point? Is this code reachable?

test/AArch64/standalone/Relocations/unsupported.test
quic-areg:
```suggestion
# RUN: %not %link %linkopts %t.o -o %t.out 2>&1 | %filecheck %s
```
`-march` can be inferred, redirecting to a log is not needed, `|| true` overrides the linker's exit code (we expect it to be non-zero)

nezetic: Good catch, modified the tests

lib/Target/AArch64/AArch64Relocator.cpp
nezetic: _getSize_ is called during IR building, before _scanRelocation_ (so before the relocation is rejected). I implemented the same logic here as in the RISCV backend.

```C++
Relocation::Size RISCVRelocator::getSize(Relocation::Type pType) const {
  if (RelocDescs.count(pType) == 0)
    return 0;
  return getRISCVReloc(pType).Size;
}```

File: lib/Target/AArch64/AArch64Relocator.cpp (hunk 6 above)
Comment (nezetic) [reply lib/Target/AArch64/AArch64Relocator.cpp]: Indeed, fixed

File: lib/Target/AArch64/AArch64Relocator.cpp (hunk 5 above)
Comment (nezetic) [reply lib/Target/AArch64/AArch64Relocator.cpp]: I removed the overload, as it was only needed in _getSize_, which is now implemented as in the RISCV target.

I wanted to keep _isRelocSupported_ as in the other targets (so with a Relocation as argument).

File: lib/Readers/Relocation.cpp (hunk 4 above)
Comment (nezetic) [reply lib/Readers/Relocation.cpp]: Indeed, I replaced them with asserts, and found another OOB issue (size in bits being handled as bytes).

File: test/AArch64/standalone/Relocations/unsupported.test (hunk 2 above)
Comment (nezetic) [reply test/AArch64/standalone/Relocations/unsupported.test]: Fixed

File: test/AArch64/standalone/Relocations/unsupported.test (hunk 3 above)
Comment (nezetic) [reply test/AArch64/standalone/Relocations/unsupported.test]: Fixed

File: test/AArch64/standalone/Relocations/unsupported.test (hunk 2 above)
Comment (nezetic) [reply test/AArch64/standalone/Relocations/unsupported.test]: Good catch, this was a leftover from the previous tests catching the assert. I improved the tests based on feedback.

File: test/AArch64/standalone/Relocations/unsupported.test (hunk 1 above)
Comment (nezetic) [reply test/AArch64/standalone/Relocations/unsupported.test]: I improved the tests, handling both dynamic / static executable, as well as shared libraries.
I also added a test making sure invalid relocations are ignored in discarded sections (using _--gc-sections_).

Regarding partial linking, the "unknown" relocations are just copied through, so no error is raised.
It seems this is how other targets are handling them too.

File: lib/Readers/Relocation.cpp (hunk 4 above)
Comment (nezetic) [reply lib/Readers/Relocation.cpp]: Indeed, I replaced them with asserts, and found another OOB issue (size in bits being handled as bytes).

**Review by quic-areg — APPROVED** [review]
LGTM

File: lib/Target/AArch64/AArch64Relocator.cpp [hunk 8]
```diff
@@ -134,6 +134,17 @@ AArch64Relocator::AArch64Relocator(AArch64GNUInfoLDBackend &pParent,
...
 AArch64Relocator::~AArch64Relocator() {}
 
+bool AArch64Relocator::isRelocSupported(Relocation &pReloc) const {
+  Relocation::Type type = pReloc.type();
+  // valid types are 0x0, 0x100-0x239
+  if ((type < 0x100 || type > 0x239) && (type != 0x0) &&
+      (type != R_AARCH64_COPY_INSN))
+    return false;
+
+  assert(ApplyFunctions.find(type) != ApplyFunctions.end());
```
partaror, line 144: What do you think about returning false if `ApplyFunctions` does not contain this specific relocation?

nezetic: There are more entries in _ApplyFunctions_ that what the original code consider as _supported_ relocations.
I wanted to fix the OOB issues without changing the logic.

partaror: Okay, I see. Thank you for the explanation.

PR #777, opened 2026-02-02
Revert "[ELFSection] Remove unnecessary Annotations member."

parth-07:
This reverts commit 8f235b406e3019928d9e926da001c15d063de177. This commit is getting reverted because it caused a performance regression while computing section annotations when there are large number of plugin operations.

### DIFF
--- a/include/eld/Readers/ELFSection.h
+++ b/include/eld/Readers/ELFSection.h
@@ -147,6 +147,12 @@ public:
 
   static std::string getELFPermissionsStr(uint32_t Permissions);
 
+  std::string getSectionAnnotations() const;
+
+  bool hasAnnotations() const;
+
+  void addSectionAnnotation(const std::string &Annotation);
+
   bool hasOffset() const;
 
   /// FIXME: We change the offset for input sections so this will not return the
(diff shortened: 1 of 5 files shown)


include/eld/Readers/ELFSection.h
```diff
@@ -318,17 +324,15 @@ class ELFSection : public ELFSectionBase {
...
   llvm::SmallVector<Fragment *, 0> Fragments;
   llvm::SmallVector<Relocation *, 0> Relocations;
 
   /// FIXME: These vectors can be moved out of this class?
   llvm::SmallVector<const ELFSection *, 0> GroupSections;
 };
 
-#ifndef _WIN32
-static_assert(sizeof(ELFSection) <= 224, "ELFSection grew too large!");
-#endif
```
quic-areg, line 330: I think this static_assert is still useful, and that it does not make sense to store this annotations member in ELFSection. Can we move it to a side table somewhere else?

quic-areg (commented):
Thanks for catching this

include/eld/Readers/ELFSection.h
quic-seaswara: @quic-areg Yes, can you please work on that ?

quic-seaswara (approved):

PR #779, opened 2026-02-02
Add support for print cmd

quic-seaswara:
The PRINT command lets a linker script emit formatted messages on the linker's standard output stream while the script is being parsed and evaluated. This is useful for debugging script expressions, inspecting symbol values, or annotating map files during development.

The syntax is very similar to using 'C' printf.

### DIFF
--- a/docs/userguide/documentation/linker_script.rst
+++ b/docs/userguide/documentation/linker_script.rst
@@ -66,159 +66,220 @@ This is the standard method used by most linkers like GNU ld.
 Basic script syntax
 --------------------
 
- - Symbols
+Symbols
+^^^^^^^
 
-   Symbol names must begin with a letter, underscore, or period. They can include letters, numbers, underscores, hyphens, or periods.
+Symbol names must begin with a letter, underscore, or period. They can
+include letters, numbers, underscores, hyphens, or periods.
 
- - Comments
+Comments
+^^^^^^^^
 
-   Comments can appear in linker scripts
+Comments can appear in linker scripts.
 
- - Strings
+Strings
+^^^^^^^
 
-   Character strings can be specified as parameters with or without delimiter characters.
+Character strings can be specified as parameters with or without
+delimiter characters.
 
- - Expressions
+Expression basics
+^^^^^^^^^^^^^^^^^
 
-   Expressions are similar to C, and support all C arithmetic operators. They are evaluated as type long or unsigned long
+Expressions are similar to C, and support all C arithmetic operators.
+They are evaluated as type ``long`` or ``unsigned long``.
 
- - Location Counter
+Location counter basics
+^^^^^^^^^^^^^^^^^^^^^^^
 
-   | A period is used as a symbol to indicate the current location counter. It is used in the SECTIONS command only, where it designates locations in the output section:
-   | . = ALIGN(0x1000); . = . + 0x1000;
-   | Assigning a value to the location counter symbol changes the location counter to the specified value.
-   | The location counter can be moved forward by arbitrary amounts to create gaps in an output section.
-   | It cannot, however, be moved backwards.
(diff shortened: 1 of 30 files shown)


test/Hexagon/standalone/linkerscript/printcmd/Inputs/print_s_success.t
```diff
@@ -0,0 +1,7 @@
+/* foo is defined in 1.c */
+PRINT("S: %s\n", foo);
```
parth-07, line 2: Why does not `%s` mean string here if we are following printf style? How will I choose the decimal/octal/hex representation for a symbol value?

docs/userguide/documentation/linker_script.rst
```diff
@@ -743,3 +805,205 @@ the parser can determine whether the command is an assignment command or not.
...
+Field width and precision have the usual ``printf`` syntax::
+
+  %[flags][width][.precision][length]conversion
+
+with the following constraints:
+
+- ``width`` and ``precision`` must be decimal integer constants.
+- ``*`` is **not** supported for either width or precision; using it
+  produces a diagnostics error.
+- Supported length modifiers (``hh``, ``h``, ``l``, ``ll``, ``j``, ``z``,
```
parth-07, line 885: Why don't we error on length modifiers if they are not supported?

lib/Script/Expression.cpp
```diff
@@ -1232,6 +1234,240 @@ void AssertCmd::getSymbolNames(std::unordered_set<std::string> &SymbolTokens) {
...
+      // Unknown escape sequence: keep as-is.
+      Out.push_back(C);
+      continue;
+    }
+  }
+  return Out;
+}
+
+eld::Expected<std::string>
+PrintCmd::formatString(llvm::StringRef Fmt,
```
parth-07, line 1280: Would it be better to do this parsing in ScriptParser so that we can properly emit the errors with the line / column number information?

test/Hexagon/standalone/linkerscript/printcmd/Inputs/print_s_success.t
quic-seaswara: It is close to printf style but not. %s is a shortcut to print a symbol, which is the name of the symbol. You dont want to print that as hexadecimal or octal imo.

partaror: Wouldn't it better to use `"foo"` (within quotes) when printing symbol name, and `foo` (without quotes) to mean the symbol value?

docs/userguide/documentation/linker_script.rst
quic-seaswara: We ignore it, no need to error out. Since we treat everything as 64bit.

lib/Script/Expression.cpp
quic-seaswara: I see there are two ways of doing this, parse in ScriptParser and Expression, and I started doing this, but the code became very convoluted, I then thought it is  much easier to parse the semantics of print in Expression since we need to print the string as it was passed. 

We can keep it like this IMO for now.

test/Hexagon/standalone/linkerscript/printcmd/Inputs/print_s_success.t
quic-seaswara: That is the advantage of printing the symbol name for example %s = 0x%x would be a pattern I expect you to use to print symbol names with their values. 

```
SECTIONS {
  .text : { *(.text*) }
  q = 0xF0000;
  PRINT("PRINTCMD %s = 0x%x\n", q, q);
}

$ ld.eld 1.o -T p.t --defsym bar=0x100
PRINTCMD q = 0xf0000
```

PR #781, opened 2026-02-03
Add primitive support for link-time memory reports

parth-07:
This commit adds primitive support for link-time memory reports. The memory report contains memory usage information for each (most!) timer that we have in the codebase (eld::RegisterTimer).

The main motivation for link-time memory reports is to help find out which linker areas to focus on for reducing the link memory footprint.

The memory usage information contains the current resident set size, the resident set size change in this timer, and the peak resident set size seen so far. All these information are computed by parsing the virtual file '/proc/self/status'. As expected, this solution would not work for windows and thus this feature is only available for eld-on-linux. The virtual file '/proc/self/status' may be represented slightly different across different linux distributions so we might see some issues in different linux distributions. Thus, this feature should be considered experimental for now.

Memory usage information is not recorded for timers that are created a large number of times, for example, VisitSymbol and VisitSections. This is because each read to the virtual file '/proc/self/status' is a system call and making large number of this system call can take considerable time. This is fine because we can always improve/rearrange timers such that we get the memory information that we need.

### DIFF
--- a/include/eld/Config/GeneralOptions.h
+++ b/include/eld/Config/GeneralOptions.h
@@ -18,6 +18,7 @@
 #include "eld/Support/FileSystem.h"
 #include "eld/Support/Memory.h"
 #include "eld/Support/MsgHandling.h"
+#include "eld/Support/RegisterTimer.h"
 #include "eld/SymbolResolver/ResolveInfo.h"
 #include "llvm/ADT/ArrayRef.h"
 #include "llvm/ADT/DenseMap.h"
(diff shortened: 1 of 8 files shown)


quic-seaswara:
> > A monitor thread allows you to continously measure memory growth through the time of link.
> 
> But this patch aims to measure memory consumption in the key link phases that we can control through the hooks. Why do we want to continuously measure memory growth? Also, by continuously measuring memory growth, do you mean tracking memory consumption at random intervals?

Yes, it can be extended to measure at random intervals.

PR #790, opened 2026-02-05
[NFC] Compute rule matching information once

quic-seaswara:
I thought this would improve performance a bit, but it does not change anything. I think this is still a good cleanup.

### DIFF
--- a/lib/Object/ObjectBuilder.cpp
+++ b/lib/Object/ObjectBuilder.cpp
@@ -271,14 +271,93 @@ void ObjectBuilder::assignInputFromOutput(eld::InputFile *Obj) {
   SectionMap &SectionMap = ThisModule.getScript().sectionMap();
   ObjectFile *ObjFile = llvm::dyn_cast<ObjectFile>(Obj);
   std::vector<Section *> Sections = getInputSectionsForRuleMatching(ObjFile);
+
+  struct SectionMatchData {
+    Section *Section = nullptr;
+    ELFSection *ELFSect = nullptr;
+    InputFile *Input = nullptr;
+    bool IsCommonSection = false;
+    bool IsArchive = false;
+    bool IsMergeStr = false;
+    bool IsWritable = false;
+    std::string SectName;
+    uint64_t InputSectionHash = 0;
+    uint64_t InputFileHash = 0;
+    uint64_t NameHash = 0;
+    std::string PInputFile;
+    std::string Name;
+  };
+
+  std::vector<SectionMatchData> MatchData;
+  MatchData.reserve(Sections.size());
+
+  auto ensurePatternMapInitialized = [&](InputFile *Input) {
+    auto *InputImpl = Input->getInput();
+    if (InputImpl->isPatternMapInitialized())
+      return;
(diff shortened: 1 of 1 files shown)


lib/Object/ObjectBuilder.cpp
```diff
@@ -271,14 +271,93 @@ void ObjectBuilder::assignInputFromOutput(eld::InputFile *Obj) {
   SectionMap &SectionMap = ThisModule.getScript().sectionMap();
   ObjectFile *ObjFile = llvm::dyn_cast<ObjectFile>(Obj);
   std::vector<Section *> Sections = getInputSectionsForRuleMatching(ObjFile);
+
+  struct SectionMatchData {
```
parth-07, line 275: What is the benefit of computing and storing all of these information in advance instead of directly using them in the loop below that is traversing the sections? We are paying extra memory overhead, which we can avoid, for storing all these information.

lib/Object/ObjectBuilder.cpp
```diff
@@ -271,14 +271,93 @@ void ObjectBuilder::assignInputFromOutput(eld::InputFile *Obj) {
...
+  };
+
+  std::vector<SectionMatchData> MatchData;
+  MatchData.reserve(Sections.size());
+
+  auto ensurePatternMapInitialized = [&](InputFile *Input) {
+    auto *InputImpl = Input->getInput();
+    if (InputImpl->isPatternMapInitialized())
+      return;
+    std::lock_guard<std::mutex> Guard(Mutex);
```
parth-07, line 298: `assignInputFromOutput` function runs per InputFile. Why do we need a mutex here for resizing the input?

lib/Object/ObjectBuilder.cpp
```diff
@@ -271,14 +271,93 @@ void ObjectBuilder::assignInputFromOutput(eld::InputFile *Obj) {
...
+    bool IsCommonSection = false;
+    if (CommonELFSection *CommonSection =
+            llvm::dyn_cast<CommonELFSection>(Section)) {
+      Input = CommonSection->getOrigin();
+      IsCommonSection = true;
+    }
+    if (Section->getOldInputFile())
+      Input = Section->getOldInputFile();
+
+    ensurePatternMapInitialized(Input);
```
parth-07, line 322: Do we need to call this function per section when all the sections have the same input file?

lib/Object/ObjectBuilder.cpp
```diff
@@ -435,27 +486,25 @@ void ObjectBuilder::storePatternsForInputFile(InputFile *Obj,
          "Common internal input file contains common symbols from various "
          "different input files. StorePatternsForInputFile should be "
          "called on the actual input files.");
+  std::string const &InputName = Obj->getInput()->getName();
+  uint64_t NameHash = Obj->getInput()->getArchiveMemberNameHash();
+  std::string const &InputFile = Obj->getInput()->getResolvedPath().native();
```
parth-07, line 491: Can we please use more descriptive names here? Here we have both `InputName` and `InputFile` variables that both mean input file names...

PR #791, opened 2026-02-06
Fix resolving script inputs path when sysroot is specified

### LINKED ISSUE #786: eld looks for version scripts with an absolute path inside of the --sysroot dir
**Description**
When eld is given `--sysroot=/some/dir` and `--version-script=/some/absolute/path` it looks for the file at `/some/dir/some/absolute/path`. In my particular case (cross compiling hwasan with eld, see below), this gives errors. From what I can see, lld and bfd both look into '/some/absolute/path' rather than prepending the sysroot like eld.

**Reproducer**
It's a bit of a silly/incomplete example, but below should work to reproduce the issue.

A real example in the wild can be seen when trying to build hwasan using eld (https://github.com/qualcomm/cpullvm-toolchain/pull/134, logs are [here](https://productionresultssa14.blob.core.windows.net/actions-results/82aa121c-080c-4840-a861-e96cb99a8c80/workflow-job-run-c61aff6f-f8f4-5291-b017-f30f0283516a/logs/job/job-logs.txt?rsct=text%2Fplain&se=2026-02-05T02%3A45%3A20Z&sig=DKKFqQ%2B2mrJ%2F1i9Uu0knLErk8MGOYKkr1Oi49QLQZGE%3D&ske=2026-02-05T05%3A07%3A47Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2026-02-05T01%3A07%3A47Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2026-02-05T02%3A35%3A15Z&sv=2025-11-05))
```
cat > empty.c << '!'
int main() {}
!
touch myvers.vers

# eld errors with the below when given an absolute path:
# Fatal: cannot read file /mysysroot/path/to/working/directory/myvers.vers, because of error No such file or directory
clang --target=riscv32-linux-gnu --sysroot=/mysysroot -shared -fuse-ld=eld -nostartfiles -nostdlib empty.c -Wl,--version-script=$PWD/myvers.vers

# lld and bfd both look in the absolute path, not considering the sysroot.
# For this example, lld produces this error (which, make sense--it found the right file, which is empty):
# ld.lld: error: /path/to/working/directory/myvers.vers:1: unexpected EOF
clang --target=riscv32-linux-gnu --sysroot=/mysysroot -shared -fuse-ld=eld -nostartfiles -nostdlib empty.c -Wl,--version-script=$PWD/myvers.vers

# eld looks at the right file when given a relative path (though interestingly it doesn't error on the empty file):
clang --target=riscv32-linux-gnu --sysroot=/mysysroot -shared -fuse-ld=eld -nostartfiles -nostdlib empty.c -Wl,--version-script=myvers.vers
```

parth-07:
This commit fix the behavior of searching script inputs such that
if the script input is found in the specified absolute / relative path,
then the sysroot is not used to find the script. This change is done to
make to eld behavior consistent with the other major linkers (such as bfd).

Resolves #786

### DIFF
--- a/lib/Input/Input.cpp
+++ b/lib/Input/Input.cpp
@@ -102,11 +102,6 @@ bool Input::resolvePath(const LinkerConfig &PConfig) {
   }
 
   if (Type == Input::Script) {
-    if (PSearchDirs.hasSysRoot() &&
-        (FileName.size() > 0 && FileName[0] == '/')) {
-      ResolvedPath = PSearchDirs.sysroot();
-      ResolvedPath->append(FileName);
-    }
     if (!llvm::sys::fs::exists(ResolvedPath->native())) {
       const sys::fs::Path *P = PSearchDirs.find(FileName, Input::Script);
       if (P != nullptr)
--- a/test/Common/standalone/sysRoot/VersionAndDynamicScripts/Inputs/1.c
+++ b/test/Common/standalone/sysRoot/VersionAndDynamicScripts/Inputs/1.c
@@ -1,0 +1,4 @@
+// Simple library with exported functions
+int exported_func_v1() { return 1; }
+int exported_func_v2() { return 2; }
+int hidden_func() { return 3; }
--- a/test/Common/standalone/sysRoot/VersionAndDynamicScripts/Inputs/dynamic.list
+++ b/test/Common/standalone/sysRoot/VersionAndDynamicScripts/Inputs/dynamic.list
@@ -1,0 +1,4 @@
+{
+  exported_func_v1;
+  exported_func_v2;
+};
--- a/test/Common/standalone/sysRoot/VersionAndDynamicScripts/Inputs/version.script
+++ b/test/Common/standalone/sysRoot/VersionAndDynamicScripts/Inputs/version.script
@@ -1,0 +1,6 @@
+{
+  global:
+    exported_func_v1;
+  local:
+    *;
+};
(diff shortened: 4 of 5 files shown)


lib/Input/Input.cpp
```diff
@@ -102,11 +102,6 @@ bool Input::resolvePath(const LinkerConfig &PConfig) {
   }
 
   if (Type == Input::Script) {
-    if (PSearchDirs.hasSysRoot() &&
-        (FileName.size() > 0 && FileName[0] == '/')) {
```
quic-seaswara, line 106: FileName[0] != '/'

Test :

```
$ ls /tmp/myvers.vers 
/tmp/myvers.vers
$echo $PWD
/usr2/xxx/
$ld.eld --sysroot=/tmp -L=/ --version-script myvers.vers /tmp/1.o
```

parth-07: The changes in this patch correctly handle this case. If we make the `FileName[0] != '/'` change then we will incorrectly search in the sysroot location even when the script file is present in the specified relative path. We should only search in the `-L` locations if the input is not found in the specified absolute / relative paths.

I have added test for this case.

jonathonpenix:
Thanks for fixing this @parth-07!

Just a heads up, I _think_ we're seeing Windows failures with this over in cpullvm--it's a WIP branch, but see ex: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/21848478483/job/63049668083?pr=150

Looks like the eld Windows nightly is in progress, so I'll keep an eye on that too and see if the failure shows up there.

PR #793, opened 2026-02-06
UndefinedBehaviorSanitizer: signed integer overflow issue

sam-svutkoor:
eld/lib/Target/RISCV/RISCVHelper.h:243:18: runtime error: signed integer overflow: 2147483628 + 2048 cannot be represented in type 'int' 
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior RISCVHelper.h:243:18 
this commit fixes the build failure for picolibc riscv64-bm sanity
Removes signed overflow UBSan complaining

### DIFF
--- a/lib/Target/RISCV/RISCVHelper.h
+++ b/lib/Target/RISCV/RISCVHelper.h
@@ -240,7 +240,8 @@ template <typename T> T encodeUJ(T Result) {
 }
 
 inline int32_t encodeU_HI20(int32_t Result) {
-  return (Result + 0x800) & 0xFFFFF000;
+  uint32_t U = static_cast<uint32_t>(Result);
+  return static_cast<int32_t>((U + 0x800u) & 0xFFFFF000u);
 }
 
 template <typename T> T encodeU_ABS20(T Result) {


lib/Target/RISCV/RISCVHelper.h
```diff
@@ -240,7 +240,9 @@ template <typename T> T encodeUJ(T Result) {
 }
 
 inline int32_t encodeU_HI20(int32_t Result) {
-  return (Result + 0x800) & 0xFFFFF000;
+  uint32_t U = static_cast<uint32_t>(Result);
+  return static_cast<int32_t>((U + 0x800u) & 0xFFFFF000u);
+  //return (Result + 0x800) & 0xFFFFF000;
```
quic-seaswara: Remove comment

lib/Target/RISCV/RISCVHelper.h
```diff
@@ -240,7 +240,9 @@ template <typename T> T encodeUJ(T Result) {
 }
 
 inline int32_t encodeU_HI20(int32_t Result) {
-  return (Result + 0x800) & 0xFFFFF000;
+  uint32_t U = static_cast<uint32_t>(Result);
+  return static_cast<int32_t>((U + 0x800u) & 0xFFFFF000u);
```
quic-seaswara, line 244: Is this just silencing the error, will this cause any change in functionality ?

sam-svutkoor: Because encodeU_HI20 is logically a bit‑encoding operation, not a signed arithmetic one, moving the add into unsigned space does not affect correctness. 
It simply guarantees defined wraparound behavior and removes the signed‑overflow UB reported by UBSan. Thanks,

lenary (approved):
LGTM

I think this is right, but it's not very neat. I think anything that would make this a neater change would be quite involved, though.

Issue #697, opened 2026-01-07
support z notext


quic-areg:
-z notext permits relocations in RO segments

quic-seaswara:
-z notext and -z text are related. I am fine if you want to close this and open a new one.

PR #801, opened 2026-02-10
[AArch64] Add support for pointer authentication relocations

nezetic:
Implements basic PAuth support for AArch64 following the PAuth ABI Extension specification (2025Q1):
https://github.com/ARM-software/abi-aa/blob/main/pauthabielf64/pauthabielf64.rst

**Adds:**
- R_AARCH64_AUTH_ABS64 (0x244): authenticated absolute relocations
- R_AARCH64_AUTH_RELATIVE (0x411): authenticated relative relocations

**Testing:**
Covers PIE/non-PIE/static executables, as well as various corner cases from the spec like undefined weak references, COPY relocations and non-allocating sections.

**Not included:**
GOT/PLT signing, additional AUTH relocation types (future work).

**Notes:**
I changed the way *GNULDBackend::recordRelativeReloc* store relative relocations. The original map is only iterated once (in a linear way), but reversing the key / values allows to implement an efficient *GNULDBackend::findRelativeReloc*.

No changes to existing non-AUTH relocation behavior.

### DIFF
--- a/include/eld/Target/GNULDBackend.h
+++ b/include/eld/Target/GNULDBackend.h
@@ -661,7 +661,7 @@ public:
   virtual LDSymbol *getGOTSymbol() const { return m_pGOTSymbol; }
 
   void recordRelativeReloc(Relocation *R, const Relocation *N) {
-    m_RelativeRelocMap[R] = N;
+    m_RelativeRelocMap[N] = R;
   }
 
   // Patching sections.
@@ -1171,7 +1171,7 @@ protected:
   // Dynamic linking
   ELFObjectFile *m_DynamicSectionHeadersInputFile = nullptr;
   LDSymbol *m_pGOTSymbol = nullptr;
-  llvm::DenseMap<Relocation *, const Relocation *> m_RelativeRelocMap;
+  llvm::DenseMap<const Relocation *, Relocation *> m_RelativeRelocMap;
 
   // Patching.
   llvm::DenseMap<ResolveInfo *, const ResolveInfo *> m_AbsolutePLTMap;
--- a/lib/Target/AArch64/AArch64ELFDynamic.cpp
+++ b/lib/Target/AArch64/AArch64ELFDynamic.cpp
@@ -31,7 +31,8 @@ void AArch64ELFDynamic::reserveTargetEntries() {
 void AArch64ELFDynamic::applyTargetEntries() {
   uint32_t relaCount = 0;
   for (auto &R : m_Backend.getRelaDyn()->getRelocations()) {
-    if (R->type() == llvm::ELF::R_AARCH64_RELATIVE)
+    if (R->type() == llvm::ELF::R_AARCH64_RELATIVE ||
+        R->type() == llvm::ELF::R_AARCH64_AUTH_RELATIVE)
       relaCount++;
   }
   applyOne(llvm::ELF::DT_RELACOUNT, relaCount);
--- a/lib/Target/AArch64/AArch64LDBackend.cpp
+++ b/lib/Target/AArch64/AArch64LDBackend.cpp
@@ -773,8 +773,12 @@ void AArch64LDBackend::recordPLT(ResolveInfo *I, AArch64PLT *P) {
   m_PLTMap[I] = P;
 }
 
+Relocation *AArch64LDBackend::findRelativeReloc(const Relocation *pReloc) const {
+  return m_RelativeRelocMap.lookup(pReloc);
+}
+
 Stub *AArch64LDBackend::getBranchIslandStub(Relocation *pReloc,
-                                                   int64_t targetValue) const {
+                                            int64_t targetValue) const {
   (void)pReloc;
   (void)targetValue;
   return *getStubFactory()->getAllStubs().cbegin();
--- a/lib/Target/AArch64/AArch64RelocationFunctions.h
+++ b/lib/Target/AArch64/AArch64RelocationFunctions.h
@@ -128,7 +128,7 @@
                 MappedType(&ld64_got_lo12, "R_AARCH64_LD64_GOT_LO12_NC", 32)), \
       ValueType(0x139,                                                         \
                 MappedType(&ld64_gotpage_lo15,                                 \
-			   "R_AARCH64_LD64_GOTPAGE_LO15", 32)),                \
+                           "R_AARCH64_LD64_GOTPAGE_LO15", 32)),                \
       ValueType(0x13a,                                                         \
                 MappedType(&unsupport, "R_AARCH64_PLT32", 32)),                \
       ValueType(0x13b,                                                         \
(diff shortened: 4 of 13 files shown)


include/eld/Target/GNULDBackend.h
```diff
@@ -1159,7 +1159,7 @@ class GNULDBackend {
   // Dynamic linking
   ELFObjectFile *m_DynamicSectionHeadersInputFile = nullptr;
   LDSymbol *m_pGOTSymbol = nullptr;
-  llvm::DenseMap<Relocation *, const Relocation *> m_RelativeRelocMap;
+  llvm::DenseMap<const Relocation *, Relocation *> m_RelativeRelocMap;
```
quic-seaswara, line 1174: can we simply make the key and the value Relocation * ?

lib/Target/AArch64/AArch64ELFDynamic.cpp
```diff
@@ -31,7 +31,8 @@ void AArch64ELFDynamic::reserveTargetEntries() {
 void AArch64ELFDynamic::applyTargetEntries() {
   uint32_t relaCount = 0;
   for (auto &R : m_Backend.getRelaDyn()->getRelocations()) {
-    if (R->type() == llvm::ELF::R_AARCH64_RELATIVE)
+    if (R->type() == llvm::ELF::R_AARCH64_RELATIVE ||
```
quic-areg, line 34: I think this may lead to a potentially wrong `DT_RELACOUNT` value with eld (lld does not count `AARCH64_AUTH_RELATIVE` here).

nezetic: My understanding if that `DT_RELACOUNT` represents `the number of relative relocations in .rela.dyn`. `AARCH64_AUTH_RELATIVE ` relocations will end up in this section, so they must be counted.

> Relocation section '.rela.dyn' at offset 0x1e8 contains 11 entries:
0000000000001130  0000000000000411 R_AARCH64_AUTH_RELATIVE           341
0000000000001138  0000000000000411 R_AARCH64_AUTH_RELATIVE           1132

It's important to note LLD implement a very tedious (and optional) _Relocation Compression_ (section 11) , that is not in the scope of this PR.

I did not perform a 1:1 comparaison with LLD, but made sure ELD is passing similar tests.

quic-seaswara: @nezetic is this what you are mentioning for PAUTH relocations 

Arm’s PAuth ABI spec documents “AUTH variant dynamic relocations,” additional dynamic section tags, and a separate compressed relocation section (SHT_AARCH64_AUTH_RELR), reinforcing that authenticated relatives are accounted for outside the legacy DT_RELACOUNT mechanism.

I am not sure about this, I feel this behavior in the linker is outside the spec. Are we doin gthis until we get RELR support ? 

How does GNU linker handle this ?

quic-seaswara:
@efriedma-quic  can you please review this change ?

lib/Target/AArch64/AArch64LDBackend.cpp
```diff
@@ -808,6 +808,13 @@ void AArch64GNUInfoLDBackend::recordPLT(ResolveInfo *I, AArch64PLT *P) {
   m_PLTMap[I] = P;
 }
 
+Relocation *AArch64GNUInfoLDBackend::findRelativeReloc(const Relocation *pReloc) const {
+  auto Entry = m_RelativeRelocMap.find(pReloc);
+  if (Entry == m_RelativeRelocMap.end())
+    return nullptr;
+  return Entry->second;
```
efriedma-quic:
```suggestion
  return m_RelativeRelocMap.lookup(pReloc);
```

lib/Target/AArch64/AArch64RelocationFunctions.h
```diff
@@ -263,6 +263,7 @@
       ValueType(0x23d,                                                         \
                 MappedType(&unsupport,                                         \
                            "R_AARCH64_TLSLE_LDST128_DTPREL_LO12_NC", 32)),     \
+      ValueType(0x244, MappedType(&abs, "R_AARCH64_AUTH_ABS64", 64)),          \
```
efriedma-quic: Can we arrange this header in the same order as llvm/BinaryFormat/ELFRelocs/AArch64.def, to make future updates easier?

lib/Target/AArch64/AArch64Relocator.cpp
```diff
@@ -203,18 +238,23 @@ void AArch64Relocator::scanLocalReloc(InputFile &pInput, Relocation &pReloc,
...
   switch (pReloc.type()) {
+  case llvm::ELF::R_AARCH64_AUTH_ABS64:
   case llvm::ELF::R_AARCH64_ABS64:
-    // If building PIC object (shared library or PIC executable),
+    // If building PIC object (shared library or PIC executable)
+    // or dealing with a signed relocation,
     // a dynamic relocations with RELATIVE type to this location is needed.
     // Reserve an entry in .rel.dyn
-    if (config().isCodeIndep()) {
+    if (config().isCodeIndep() || isAuthReloc(pReloc)) {
```
efriedma-quic: Using isAuthReloc() instead of just `pReloc.type() == llvm::ELF::R_AARCH64_AUTH_ABS64` feels a little strange.

test/AArch64/standalone/Relocations/PAuthReloc.s
```diff
@@ -0,0 +1,159 @@
...
+# STATIC-NEXT:       {{0x[0-9A-F]+}} R_AARCH64_AUTH_RELATIVE - 0x34567{{[0-9A-F]+}}
+# STATIC-NEXT:       {{0x[0-9A-F]+}} R_AARCH64_AUTH_RELATIVE - 0xFFFFFFFFCBA98{{[0-9A-F]+}}
+# STATIC-NEXT:       {{0x[0-9A-F]+}} R_AARCH64_AUTH_RELATIVE - {{0x[0-9A-F]+}}
+# STATIC-NEXT:       {{0x[0-9A-F]+}} R_AARCH64_AUTH_RELATIVE - {{0x[0-9A-F]+}}
+# STATIC-NEXT:       {{0x[0-9A-F]+}} R_AARCH64_AUTH_RELATIVE - {{0x[0-9A-F]+}}
+# STATIC-NEXT:       {{0x[0-9A-F]+}} R_AARCH64_AUTH_RELATIVE - {{0x[0-9A-F]+}}
+# STATIC-NEXT:       {{0x[0-9A-F]+}} R_AARCH64_AUTH_RELATIVE - {{0x[0-9A-F]+}}
+# STATIC-NEXT: }
+
+# Test PAuth relocations with static executable (without PIE)
```
efriedma-quic, line 136: In a static executable, what code actually processes the rela.dyn section?

test/AArch64/standalone/Relocations/PAuthReloc.s
```diff
@@ -0,0 +1,159 @@
...
+#--- a.s
+.global bar
+.type bar, @function
+bar:
+
+.global foo
+foo:
+
+#--- main.s
+.section .test, "aw"
```
efriedma-quic, line 168: Maybe worth adding a test to show what happens if you try to use R_AARCH64_AUTH_ABS64 in a readonly section?

lib/Target/AArch64/AArch64LDBackend.cpp
nezetic: Updated the code 👍

lib/Target/AArch64/AArch64RelocationFunctions.h
nezetic: I updated the header based on AArch64.def, adding all AUTH relocations in the same order

lib/Target/AArch64/AArch64Relocator.cpp
nezetic: Indeed, I simplified the code a bit, testing only with R_AARCH64_AUTH_ABS64 where I'm only matching this one (_abs_ function and relocation scan switch cases).

test/AArch64/standalone/Relocations/PAuthReloc.s
nezetic: In the case of a baremetal binary (a firmware), a custom dynamic linker (as it's already the case for things like TZ).

test/AArch64/standalone/Relocations/PAuthReloc.s
nezetic: I added a few relocations in a read-only section in the tests.

As far I understand the specifications, there are no restrictions regarding read-only sections.

efriedma-quic: The spec might not say anything, but eld defaults to `-z text`, which should forbid such relocations?

quic-seaswara: I agree with @efriedma-quic  this is a bug in eld.  Please open a bug when you get a chance. Probably this test can be modified to add -z notext ?

quic-seaswara: Reopened : https://github.com/qualcomm/eld/issues/697

nezetic: I'm not sure to understand the question.

I'm not changing the inner logic of how ABS64 relocations are handled by ELD.

If I add non-auth relocations in a read-only section in my tests, ELD will link without any error (creating R_AARCH64_RELATIVE instead of R_AARCH64_AUTH_RELATIVE).

```asm
.section .rodata, "a"
  .p2align 3
  .quad (__etext + 10)
  .quad (.rodata + 8)
  .quad (bar + 100)
```

```
Relocation section '.rela.dyn' at offset 0x1e8 contains 14 entries:
0000000000000380  0000000000000403 R_AARCH64_RELATIVE                3a2
0000000000000388  0000000000000403 R_AARCH64_RELATIVE                388
0000000000001150  0000000000000411 R_AARCH64_AUTH_RELATIVE           399
0000000000001158  0000000000000411 R_AARCH64_AUTH_RELATIVE           1152
0000000000001160  0000000000000411 R_AARCH64_AUTH_RELATIVE           ffffffffffffffb0
0000000000001168  0000000000000411 R_AARCH64_AUTH_RELATIVE           12345a10
0000000000001170  0000000000000411 R_AARCH64_AUTH_RELATIVE           34567c32
0000000000001178  0000000000000411 R_AARCH64_AUTH_RELATIVE           ffffffffcba98afe
0000000000001180  0000000000000411 R_AARCH64_AUTH_RELATIVE           ffffffff8000114f
0000000000001199  0000000000000411 R_AARCH64_AUTH_RELATIVE           39c
00000000000011a2  0000000000000411 R_AARCH64_AUTH_RELATIVE           1155
0000000000000390  0000000100000101 R_AARCH64_ABS64        0000000000000000 bar + 64
0000000000001190  0000000100000244 R_AARCH64_AUTH_ABS64   0000000000000000 bar + 0
0000000000001188  0000000200000244 R_AARCH64_AUTH_ABS64   0000000000000000 foo + 1111
```

efriedma-quic: Maybe eld just doesn't diagnose that at the moment, in general?  I guess if that's the case, it's fine.

quic-seaswara (approved):

lib/Target/AArch64/AArch64Relocator.cpp
```diff
@@ -618,19 +682,50 @@ Relocator::Result unsupport(Relocation &pReloc, AArch64Relocator &pParent) {
 // R_AARCH64_ABS64: S + A
 // R_AARCH64_ABS32: S + A
 // R_AARCH64_ABS16: S + A
+// R_AARCH64_AUTH_ABS64: PAUTH(S + A)
 Relocator::Result abs(Relocation &pReloc, AArch64Relocator &pParent) {
   ResolveInfo *rsym = pReloc.symInfo();
   Relocator::DWord A = pReloc.addend();
   Relocator::DWord S = pParent.getSymValue(&pReloc);
+  Relocator::Address targetVal = S + A;
```
efriedma-quic, line 690: It looks like this changes the behavior for non-pauth code: the value of "S" is modified by the code below (`if (rsym && rsym->reserved() & Relocator::ReservePLT)` etc.).  I assume that's not intentional?

nezetic: No, this was never intended :(

I submitted a PR to fix this regression (#1334).

PR #809, opened 2026-02-11
Correct sysroot-prepend conditions for script inputs

### LINKED ISSUE #786: eld looks for version scripts with an absolute path inside of the --sysroot dir

### LINKED ISSUE #808: Sysroot is incorrectly prepended to inputs specified in INPUT / GROUP command
Currently, we incorrectly prepend sysroot to inputs specified in INPUT / GROUP linker script commands. 

Reproducible example:

```bash
#!/usr/bin/env bash

mkdir -p A/lib64
mkdir -p A/script

cat >1.c <<\EOF
int foo() { return 1; }
EOF

cat >script.t <<\EOF
GROUP(/lib64/lib1.so)
EOF

cat >main.c <<\EOF
int main() { return 0; }
EOF

mv lib1.so A/lib64/lib1.so

ld.eld --sysroot=A -T script.t main.o 
ld.lld --sysroot=A -T script.t main.o
ld.bfd --sysroot=A -T script.t main.o 
```

lld and bfd errors out whereas eld incorrectly finds the lib1.so file.

Another example:

```bash
#!/usr/bin/env bash

mkdir -p A/lib64
mkdir -p A/script

cat >1.c <<\EOF
int foo() { return 1; }
EOF

cat >script.t <<EOF
GROUP($(pwd)/A/lib64/lib1.so)
EOF

cat >main.c <<\EOF
int main() { return 0; }
EOF

mv lib1.so A/lib64/lib1.so

ld.eld --sysroot=A -T script.t main.o 
ld.lld --sysroot=A -T script.t main.o
ld.bfd --sysroot=A -T script.t main.o 
```

lld and bfd are correctly able to find lib1.so but eld canont find the lib1.so file.

parth-07:
This commit fixes the logic that determines when the script input path should be prepended with sysroot. The summary of the new behavior:

- sysroot is only prepended to the script inputs when all the below constraints are true:
  - the input is specified in INPUT/GROUP linker script command.
  - The input path begins with '/'.
  - The linker script that contains this INPUT/GROUP command is within the sysroot directory.
- In all other cases, sysroot is not prepended to the script inputs.

Resolves #786
Resolves #808

### DIFF
--- a/lib/Input/Input.cpp
+++ b/lib/Input/Input.cpp
@@ -102,8 +102,7 @@ bool Input::resolvePath(const LinkerConfig &PConfig) {
   }
 
   if (Type == Input::Script) {
-    if (PSearchDirs.hasSysRoot() &&
-        (FileName.size() > 0 && FileName[0] == '/')) {
+    if (shouldPrependSysrootToScriptInput(PConfig)) {
       ResolvedPath = PSearchDirs.sysroot();
       ResolvedPath->append(FileName);
     }
@@ -149,6 +148,30 @@ bool Input::resolvePath(const LinkerConfig &PConfig) {
   // driver.
   Name = FileName;
   return true;
+}
+
+bool Input::shouldPrependSysrootToScriptInput(
+    const LinkerConfig &Config) const {
+  auto &searchDirs = Config.directories();
+  if (!searchDirs.hasSysRoot())
+    return false;
+  if (Type != Input::Script)
+    return false;
+  if (FileName.empty() || FileName[0] != '/')
+    return false;
+
+  // Only apply sysroot for INPUT/GROUP entries when we know which script they
+  // came from and that script is inside sysroot.
+  if (!ParentScriptFile)
+    return false;
+
+  Input *scriptInput = ParentScriptFile->getInput();
+
+  std::string scriptPath = scriptInput->getResolvedPath().getFullPath();
+  std::string sysrootPath = searchDirs.sysroot().getFullPath();
+
+  return scriptPath.size() >= sysrootPath.size() &&
+         scriptPath.compare(0, sysrootPath.size(), sysrootPath) == 0;
 }
 
 void Input::setInputFile(InputFile *Inp) { IF = Inp; }
--- a/lib/Script/InputCmd.cpp
+++ b/lib/Script/InputCmd.cpp
@@ -56,6 +56,9 @@ void InputCmd::dump(llvm::raw_ostream &Outs) const {
 }
 
 eld::Expected<void> InputCmd::activate(Module &CurModule) {
+  // Prefer ScriptCommand context (handles included scripts).
+  InputFile *parentScriptFile = &ThisScriptFile.getLinkerScriptFile();
+
   for (auto &S : ThisStringList) {
     InputToken *Token = llvm::cast<InputToken>(S);
     Token = ThisScriptFile.findResolvedFilename(Token);
@@ -64,9 +67,11 @@ eld::Expected<void> InputCmd::activate(Module &CurModule) {
     else
       ThisBuilder.getAttributes().unsetAsNeeded();
     switch (Token->type()) {
-    case InputToken::File:
-      ThisBuilder.createDeferredInput(Token->name(), Input::Script);
+    case InputToken::File: {
+      Input *I = ThisBuilder.createDeferredInput(Token->name(), Input::Script);
+      I->setParentScriptFile(parentScriptFile);
       break;
+    }
     case InputToken::NameSpec: {
       ThisBuilder.createDeferredInput(Token->name(), Input::Namespec);
       break;
--- a/include/eld/Driver/GnuLinkerOptions.td
+++ b/include/eld/Driver/GnuLinkerOptions.td
@@ -136,7 +136,7 @@ defm output_file : smDashTwoWithOpt<"o", "output", "output_file",
                    MetaVarName<"<outputfile>">,
                    Group<grp_general>;
 defm sysroot : smDash<"sysroot", "sysroot", "Set the system root">,
-              MetaVarName<"<sysroot-path>">,
+               MetaVarName<"<sysroot-path>">,
                Group<grp_general>;
 def version : Flag<["--"], "version">,
               HelpText<"Print the Linker version">,
--- a/include/eld/Input/Input.h
+++ b/include/eld/Input/Input.h
@@ -76,6 +76,12 @@ public:
 
   void setMemberNameHash(uint64_t Hash) { MemberNameHash = Hash; }
 
+  /// Set the parent script file for inputs from INPUT/GROUP commands.
+  void setParentScriptFile(InputFile *File) { ParentScriptFile = File; }
+
+  /// Get the parent script file.
+  InputFile *getParentScriptFile() const { return ParentScriptFile; }
+
   uint32_t getInputOrdinal() { return InputOrdinal; }
 
   Attribute &getAttribute() { return Attr; }
@@ -163,6 +169,12 @@ private:
   // Check if a path is valid and emit any errors
   bool isPathValid(const std::string &Path) const;
 
+  // Return true if sysroot should be prepended for a Script-typed input whose
+  // filename starts with '/'. This is used for INPUT/GROUP entries coming from
+  // a linker script: sysroot is applied only when the parent script is within
+  // the sysroot directory.
+  bool shouldPrependSysrootToScriptInput(const LinkerConfig &Config) const;
+
 protected:
   InputFile *IF = nullptr;
   MemoryArea *MemArea = nullptr;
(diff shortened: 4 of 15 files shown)


lib/Input/Input.cpp
```diff
@@ -151,6 +150,29 @@ bool Input::resolvePath(const LinkerConfig &PConfig) {
...
+
+  // Only apply sysroot for INPUT/GROUP entries when we know which script they
+  // came from and that script is inside sysroot.
+  if (!ParentScriptFile)
+    return false;
+
+  Input *scriptInput = ParentScriptFile->getInput();
+
+  std::string scriptPath = scriptInput->getResolvedPath().getFullPath();
+  std::string sysrootPath = searchDirs.sysroot().getFullPath();
```
quic-seaswara, line 171: Will this work with reproduce option ?

lib/Input/Input.cpp
```diff
@@ -151,6 +150,29 @@ bool Input::resolvePath(const LinkerConfig &PConfig) {
   return true;
 }
 
+bool Input::shouldPrependSysrootToInput(const LinkerConfig &Config) {
```
quic-seaswara: const function

Please change the name of the function to say its for script inputs.

lib/Input/Input.cpp
```diff
@@ -151,6 +150,29 @@ bool Input::resolvePath(const LinkerConfig &PConfig) {
...
+  auto &searchDirs = Config.directories();
+  if (!searchDirs.hasSysRoot())
+    return false;
+  if (Type != Input::Script)
+    return false;
+  if (FileName.empty() || FileName[0] != '/')
```
quic-seaswara, line 160: I am not sure how this will be effective on windows. Needs to be tested on windows too.

lib/Script/InputCmd.cpp
```diff
@@ -64,9 +67,11 @@ eld::Expected<void> InputCmd::activate(Module &CurModule) {
...
     switch (Token->type()) {
-    case InputToken::File:
-      ThisBuilder.createDeferredInput(Token->name(), Input::Script);
+    case InputToken::File: {
+      Input *I = ThisBuilder.createDeferredInput(Token->name(), Input::Script);
+      I->setParentScriptFile(parentScriptFile);
       break;
+    }
     case InputToken::NameSpec: {
       ThisBuilder.createDeferredInput(Token->name(), Input::Namespec);
```
quic-seaswara, line 76: Can we add the parent script file in all cases, and check if the command is INPUT or GROUP ?

lib/Input/Input.cpp
parth-07: Yes, I think so. We already had this code. Previously, we used to unconditionally prepend the sysroot for script inputs that began with `/`. Now, we only  prepend sysroot for script inputs that begins with `/` if the parent script is within the sysroot directory.

lib/Input/Input.cpp
parth-07: Done.

lib/Script/InputCmd.cpp
parth-07: Can you please explain this in more detail? Currently, we only add parent script for INPUT and GROUP inputs. If I understand correctly, then for the other cases, we do not have any associated parent script.

lib/Input/Input.cpp
parth-07: Okay!

lib/Input/Input.cpp
quic-seaswara: Can you please check ?

lib/Script/InputCmd.cpp
quic-seaswara: I see. Do we need to add this for InputToken::NameSpec too ?

parth-07: No, we do not need this for NameSpec. But we do need to fix our behavior for NameSpec with sysroot. GNU LD finds the NameSpec library in all the standard paths inside sysroot. (Paths such as /usr/lib, /lib, /lib64, and so on).

quic-seaswara: Please open a follow up ticket on this, We do have this code in `eld::ELDEmulateELF` but that code is not getting triggered.

lib/Input/Input.cpp
parth-07: I just checked. The linker crashes :'). However, this is not a regression with this patch. We crash without this patch as well.

Reproducer (nightly linker crashes with this):

```bash
#!/usr/bin/env bash

mkdir -p A/lib64
mkdir -p A/script

cat >1.c <<\EOF
int foo() { return 1; }
EOF

cat >script.t <<\EOF
GROUP(/lib64/lib1.so)
EOF

cat >main.c <<\EOF
int main() { return 0; }
EOF

mv lib1.so A/lib64/lib1.so
mv script.t A/script.t

ld.eld --sysroot=A -T A/script.t main.o --reproduce a.tar
rm -rf unpacked && mkdir unpacked
tar -xvf a.tar -C unpacked --strip-components=1
cd unpacked
bash -x response.txt
```

I believe this should be treated as a separate issue.

lib/Input/Input.cpp
```diff
@@ -151,6 +150,30 @@ bool Input::resolvePath(const LinkerConfig &PConfig) {
...
+  // came from and that script is inside sysroot.
+  if (!ParentScriptFile)
+    return false;
+
+  Input *scriptInput = ParentScriptFile->getInput();
+
+  std::string scriptPath = scriptInput->getResolvedPath().getFullPath();
+  std::string sysrootPath = searchDirs.sysroot().getFullPath();
+
+  return scriptPath.size() >= sysrootPath.size() &&
```
quic-areg, line 173: what about a hypothetical `/sdk/sysroot` and a sibling path `/sdk/sysroot_extra`? we should probably use `llvm::sys::fs::equivalent` or similar instead of a raw prefix check

parth-07: Here we just need to determine whether the script is within or outside the sysroot. If a script is contained within /sdk/sysroot_extra and the sysroot is /sdk/sysroot, then as per the linker the script is correctly outside the sysroot. The behavior seems correct to me. Can you please let me know if I have missed something?

lib/Input/Input.cpp
parth-07: I have created an issue to track this #818

lib/Script/InputCmd.cpp
parth-07: I have created an issue to track this: #819

lib/Input/Input.cpp
quic-seaswara: Do we need to follow up with windows testing ?

PR #811, opened 2026-02-11
Respect --thread-count by setting parallel strategy

quic-areg:
ELD's --thread-count option wasn't affecting `llvm::parallelFor()` because `llvm::parallel::strategy` was never set, so LLVM always defaulted to hardware concurrency. Set the strategy from --thread-count so `parallelFor()` uses the requested thread limit.

### DIFF
--- a/lib/Core/Module.cpp
+++ b/lib/Core/Module.cpp
@@ -35,6 +35,7 @@
 #include "llvm/Support/Allocator.h"
 #include "llvm/Support/ErrorOr.h"
 #include "llvm/Support/MemoryBuffer.h"
+#include "llvm/Support/Parallel.h"
 #include "llvm/Support/StringSaver.h"
 #include "llvm/Support/ThreadPool.h"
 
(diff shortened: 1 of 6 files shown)


lib/Core/Module.cpp
```diff
@@ -844,11 +847,17 @@ void Module::addReferencedSymbol(Section &RefencingSection,
...
     return LinkerThreadPool;
-  LinkerThreadPool = eld::make<llvm::StdThreadPool>(
-      llvm::hardware_concurrency(ThisConfig.options().numThreads()));
+  LinkerThreadPool = eld::make<llvm::StdThreadPool>(ThreadingStrategy);
   return LinkerThreadPool;
 }
 
+void Module::initThreading() {
+  unsigned NumThreads =
+      ThisConfig.useThreads() ? ThisConfig.options().numThreads() : 1;
```
parth-07: Can we change it to this:


```suggestion
  unsigned NumThreads = ThisConfig.options().numThreads();
```

PR #813, opened 2026-02-11
Improve memory usage

quic-seaswara:
This PR reduces the memory usage from 

without patch (VSS/RSS) : 15.2g  13.1g 
with patch (VSS/RSS) : 14.4g  12.4g

4% improvement in memory usage. 

This was a constructed linker benchmark that links 200 translation units with 20000 functions with a linker script.

### DIFF
--- a/include/eld/Input/ELFObjectFile.h
+++ b/include/eld/Input/ELFObjectFile.h
@@ -16,6 +16,7 @@
 namespace eld {
 
 class ELFSection;
+class LDSymbol;
 class TimingSection;
 class RelocMap;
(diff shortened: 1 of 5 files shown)


quic-seaswara:
> Can you please add in the commit message regarding what was used as a test program for linking?

Added to PR description. We can probably house the script in a benchmark branch.

quic-areg:
How does this affect `ELFSection`s not owned by `ELFObjectFile`s such as internal sections or plugin created sections?

Ideally this PR should be split up.

PR #814, opened 2026-02-11
Read merge string sections in parallel

quic-areg:
Read merge string sections after section discovery and use
llvm::parallelFor to parse the strings concurrently when threads
are enabled. This reduces time spent in merge string parsing on
inputs with many SHF_MERGE|SHF_STRINGS sections.

### DIFF
--- a/lib/Fragment/MergeStringFragment.cpp
+++ b/lib/Fragment/MergeStringFragment.cpp
@@ -10,6 +10,8 @@
 #include "eld/Core/Module.h"
 #include "eld/LayoutMap/LayoutInfo.h"
 #include "eld/Readers/ELFSection.h"
+#include "llvm/ADT/Sequence.h"
+#include "llvm/Support/Parallel.h"
 
 using namespace eld;
 
@@ -119,4 +121,49 @@ void MergeStringFragment::assignOutputOffsets() {
 
 bool MergeableString::isAlloc() const {
   return Fragment->getOwningSection()->isAlloc();
+}
+
+void eld::addMergeStringSection(
+    ELFSection *sec, llvm::SmallVectorImpl<MergeStringFragment *> &fragments) {
+  if (sec->getContents().empty())
+    return;
+  fragments.push_back(make<MergeStringFragment>(sec));
+}
+
+Expected<void> eld::readMergeStringSections(
+    LinkerConfig &cfg, Module &m, InputFile &file,
+    llvm::SmallVectorImpl<MergeStringFragment *> &fragments) {
+  if (fragments.empty())
+    return {};
+
+  llvm::SmallVector<uint8_t, 0> readStatus(fragments.size(), 1);
+  auto readOne = [&](size_t i) {
+    readStatus[i] = fragments[i]->readStrings(cfg);
+  };
+
+  const bool useThreads = cfg.useThreads();
+  if (useThreads && fragments.size() > 1) {
+    llvm::parallelFor(0, fragments.size(), readOne);
+  } else {
+    for (auto i : llvm::seq<size_t>(0, fragments.size()))
+      readOne(i);
+  }
+
+  for (auto [i, frag] : llvm::enumerate(fragments)) {
+    if (readStatus[i])
+      continue;
+    auto *s = frag->getOwningSection();
+    return std::make_unique<plugin::DiagnosticEntry>(plugin::DiagnosticEntry(
+        Diag::err_cannot_read_section, {s->name().str()}));
+  }
+
+  auto *layoutInfo = m.getLayoutInfo();
+  for (auto *frag : fragments) {
+    auto *s = frag->getOwningSection();
+    s->addFragment(frag);
+    if (layoutInfo)
+      layoutInfo->recordFragment(&file, s, frag);
+  }
+
+  return {};
 }
--- a/lib/Readers/ELFExecObjParser.cpp
+++ b/lib/Readers/ELFExecObjParser.cpp
@@ -89,6 +89,7 @@ eld::Expected<void> ELFExecObjParser::readSections(ELFReaderBase &ELFReader) {
     config.raise(Diag::trace_file) << inputFile->getInput()->decoratedPath();
 
   ObjectFile *ObjFile = llvm::dyn_cast<ObjectFile>(inputFile);
+  llvm::SmallVector<MergeStringFragment *, 0> mergeStrFragments;
   // handle sections
   for (Section *sect : ObjFile->getSections()) {
     if (!sect)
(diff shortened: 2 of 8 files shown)


lib/Target/LDFileFormat.cpp
```diff
@@ -44,16 +50,17 @@ LDFileFormat::Kind LDFileFormat::getELFSectionKind(
       return LDFileFormat::MergeStr;
   }
 
+  if (Name.starts_with(".debug") || Name.starts_with(".zdebug") ||
+      Name.starts_with(".line") || Name.starts_with(".stab"))
+    return LDFileFormat::Debug;
```
quic-seaswara, line 55: we can move this up and 

do if (config.options().stripDebug()) return LDFileFormat::Ignore; else LDFileFormat::Debug

lib/Readers/ELFExecObjParser.cpp
```diff
@@ -145,25 +146,7 @@ eld::Expected<void> ELFExecObjParser::readSections(ELFReaderBase &ELFReader) {
...
     case LDFileFormat::MergeStr: {
-      if (S->isCompressed()) {
-        eld::Expected<bool> expReadCompressedSection =
-            ELFReader.readCompressedSection(S);
-        ELDEXP_RETURN_DIAGENTRY_IF_ERROR(expReadCompressedSection);
-        if (!expReadCompressedSection.value())
-          return std::make_unique<plugin::DiagnosticEntry>(
-              plugin::DiagnosticEntry(Diag::err_cannot_read_section,
-                                      {S->name().str()}));
-      }
```
quic-seaswara, line 157: compressed section no longer supported ?

lib/Fragment/MergeStringFragment.cpp
```diff
@@ -120,3 +122,49 @@ void MergeStringFragment::assignOutputOffsets() {
...
+
+  auto *layoutInfo = m.getLayoutInfo();
+  for (auto *frag : fragments) {
+    auto *s = frag->getOwningSection();
+    s->addFragment(frag);
+    if (layoutInfo)
+      layoutInfo->recordFragment(file, s, frag);
+  }
+
+  return {};
```
quic-seaswara, line 168: clear mergeStrFragments after they are read ?

lib/Fragment/MergeStringFragment.cpp
```diff
@@ -120,3 +122,49 @@ void MergeStringFragment::assignOutputOffsets() {
...
+    for (auto i : llvm::seq<size_t>(0, fragments.size()))
+      readOne(i);
+  }
+
+  for (auto [i, frag] : llvm::enumerate(fragments)) {
+    if (readStatus[i])
+      continue;
+    auto *s = frag->getOwningSection();
+    return std::make_unique<plugin::DiagnosticEntry>(plugin::DiagnosticEntry(
+        Diag::err_cannot_read_section, {s->name().str()}));
```
quic-seaswara, line 157: need the input file along with cannot read section

lib/Fragment/MergeStringFragment.cpp
```diff
@@ -120,3 +122,49 @@ void MergeStringFragment::assignOutputOffsets() {
...
+
+void eld::addMergeStringSection(
+    ELFSection *sec, llvm::SmallVectorImpl<MergeStringFragment *> &fragments) {
+  if (sec->getContents().empty())
+    return;
+  fragments.push_back(make<MergeStringFragment>(sec));
+}
+
+Expected<void> eld::readMergeStringSections(
+    LinkerConfig &cfg, Module &m, InputFile *file,
```
quic-seaswara: curious to know why InputFile passed by pointer, const InputFile ?

lib/Fragment/MergeStringFragment.cpp
```diff
@@ -120,3 +122,49 @@ void MergeStringFragment::assignOutputOffsets() {
...
+
+  llvm::SmallVector<uint8_t, 0> readStatus(fragments.size(), 1);
+  auto readOne = [&](size_t i) {
+    readStatus[i] = fragments[i]->readStrings(cfg);
+  };
+
+  const bool useThreads =
+      cfg.options().threadsEnabled() && cfg.options().numThreads() > 1;
+  if (useThreads && fragments.size() > 1) {
+    llvm::parallelFor(0, fragments.size(), readOne);
```
quic-seaswara, line 146: probably the fragments can be sorted to reduce skew ? 

follow up could be delay reading the string fragments until after garbage collection is complete (discarded sections need not be processed), might save time and memory too!

lib/Fragment/MergeStringFragment.cpp
```diff
@@ -120,3 +122,49 @@ void MergeStringFragment::assignOutputOffsets() {
...
+  if (fragments.empty())
+    return {};
+
+  llvm::SmallVector<uint8_t, 0> readStatus(fragments.size(), 1);
+  auto readOne = [&](size_t i) {
+    readStatus[i] = fragments[i]->readStrings(cfg);
+  };
+
+  const bool useThreads =
+      cfg.options().threadsEnabled() && cfg.options().numThreads() > 1;
```
quic-seaswara: cfg.useThreads()

lib/Fragment/MergeStringFragment.cpp
quic-areg: > probably the fragments can be sorted to reduce skew ?

I will test this out and push a follow-up if it is effective, but I don't expect much improvement.

lib/Fragment/MergeStringFragment.cpp
quic-areg: the vector should be cleaned up automatically as soon as the calling function returns, so I don't see a good reason to do it manually

lib/Target/LDFileFormat.cpp
quic-areg: That would make debug_str sections debug instead of mergestr, and we want them to be mergestr

lib/Fragment/MergeStringFragment.cpp
quic-areg: This is the existing diagnostic. I want to keep the scope of this change as tight as possible.

lib/Fragment/MergeStringFragment.cpp
quic-areg: Sure, it can be passed as a reference instead. We can't make it `const` because `recordFragment` does not take a `const InputFile`.

lib/Fragment/MergeStringFragment.cpp
quic-areg: Thanks, updated

lib/Readers/ELFExecObjParser.cpp
quic-areg: Thank you for catching this! I am surprised how test/Common/standalone/CompressedSections/ReadCompressedSections.test was able to pass without this though

lib/Fragment/MergeStringFragment.cpp
quic-seaswara: sure

PR #816, opened 2026-02-11
[NFC] Remove pattern map that stores match results

quic-seaswara:
This map can grow and is per input file and when there are large amount of input files, and when the linker runs with a large linker script memory consumption goes up significantly.

This should not have effect on performance as the code is already multithreaded.

### DIFF
--- a/include/eld/Input/Input.h
+++ b/include/eld/Input/Input.h
@@ -18,9 +18,7 @@
 #include "eld/Script/WildcardPattern.h"
 #include "eld/Support/MemoryArea.h"
 #include "eld/Support/Path.h"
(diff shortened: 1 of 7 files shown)


quic-seaswara:
> > This map can grow and is per input file and when there are large amount of input files, and when the linker runs with a large linker script memory consumption goes up significantly.
> > This should not have effect on performance as the code is already multithreaded.
> 
> The commit message can be written better.
> 
> > This should not have effect on performance as the code is already multithreaded.
> 
> This should probably be measured. Can we generate a wildcard heavy test?

Agree with all.

PR #820, opened 2026-02-12
Optimize rule-matching by reordering the OutSection-Rule-Sections loop

parth-07:
This commit improves the performance of linker script rule-matching by reordering the nested OutputSection-LinkerScriptRule-InputSections loop.

Currently, the nested loop is as follows:

```
for (auto O : OutSections) {
  for (auto R : O.rules()) {
    for (auto S : InputSections) {
      // ...
    }
  }
}
```

This patch changes the nested loop to:

```
for (auto S : InputSections) {
  for (auto O : OutSections) {
    for (auto R : O.rules()) {
      // ...
    }
  }
}
```

The new loop structure has the following benefits:

- Once an input section is matched to a rule, we waste no iterations over it. On the contrary, in the old design, all the input sections are always iterated over for EACH rule.
- We do a lot of computation for a section. In the old design, all the computations are repeated for EACH rule, whereas in the new design all the computations happen once-per-section in the outermost loop.

On a contrived test case that make good use of linker script rule-matching, this patch results in ~15% performance improvement for the rule-matching phase.

### DIFF
--- a/lib/Object/ObjectBuilder.cpp
+++ b/lib/Object/ObjectBuilder.cpp
@@ -263,6 +263,114 @@ void ObjectBuilder::doPluginIterateSections(eld::InputFile *Obj,
 }
 
 void ObjectBuilder::assignInputFromOutput(eld::InputFile *Obj) {
+  if (ThisConfig.options().useOldRuleMatching()) {
+    assignInputFromOutputLegacy(Obj);
+    return;
+  }
+  bool IsPartialLink = (ThisConfig.codeGenType() == LinkerConfig::Object);
+  bool IsGnuCompatible =
+      (ThisConfig.options().getScriptOption() == GeneralOptions::MatchGNU);
+  bool LinkerScriptHasSectionsCommand =
+      ThisModule.getScript().linkerScriptHasSectionsCommand();
+  SectionMap &SectionMap = ThisModule.getScript().sectionMap();
+  ObjectFile *ObjFile = llvm::dyn_cast<ObjectFile>(Obj);
+  std::vector<Section *> Sections = getInputSectionsForRuleMatching(ObjFile);
+  // For each input section.
+  for (Section *Sect : Sections) {
+    ELFSection *ELFSect = llvm::dyn_cast<eld::ELFSection>(Sect);
+
+    bool IsRetrySect = false;
+
+    // Skip sections with merge strings and if there is no linker scripts
+    // provided.
+    if (IsPartialLink && (ELFSect && ELFSect->isMergeStr()) &&
+        !LinkerScriptHasSectionsCommand)
+      continue;
+    InputFile *Input = Obj;
+    bool IsCommonSection = false;
+    if (CommonELFSection *CommonSection =
+            llvm::dyn_cast<CommonELFSection>(Sect)) {
+      Input = CommonSection->getOrigin();
+      IsCommonSection = true;
+    }
+    if (Sect->getOldInputFile())
+      Input = Sect->getOldInputFile();
+    std::string const &PInputFile =
+        Input->getInput()->getResolvedPath().native();
+    std::string const &Name = Input->getInput()->getName();
+    bool IsArchive = Input->isArchive() ||
+                     llvm::dyn_cast<eld::ArchiveMemberInput>(Input->getInput());
+    // Hash of all the required things for Match.
+    uint64_t InputFileHash = Input->getInput()->getResolvedPathHash();
+    uint64_t NameHash = Input->getInput()->getArchiveMemberNameHash();
+    std::string SectName = Sect->name().str();
+    uint64_t InputSectionHash = Sect->sectionNameHash();
+    if (ELFSection *ELFSect = llvm::dyn_cast<ELFSection>(Sect)) {
+      if (auto OptRThisSectionName =
+              ObjFile->getRuleMatchingSectName(ELFSect->getIndex())) {
+        SectName = OptRThisSectionName.value();
+        InputSectionHash = llvm::hash_combine(SectName);
+      }
+    }
+    // For all output sections.
+    for (auto *Out : SectionMap) {
+      if (ELFSect) {
+        // If the rule needs to match on permissions, skip if the rule doesnot
+        // satisfy.
+        switch (Out->prolog().constraint()) {
+        case OutputSectDesc::NO_CONSTRAINT:
+          break;
+        case OutputSectDesc::ONLY_IF_RO:
+          if (ELFSect->isWritable())
+            continue;
+          break;
+        case OutputSectDesc::ONLY_IF_RW:
+          if (!ELFSect->isWritable())
+            continue;
+          break;
+        }
+      }
+      bool shouldSkipMatch = false;
+      // For each input rule.
+      for (auto *In : *Out) {
+        // If the section has an already assigned output section, skip.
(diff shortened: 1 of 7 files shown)


lib/Object/ObjectBuilder.cpp
```diff
@@ -271,81 +271,77 @@ void ObjectBuilder::assignInputFromOutput(eld::InputFile *Obj) {
...
-  // For all output sections.
-  for (auto *Out : SectionMap) {
-    // For each input rule.
-    for (auto *In : *Out) {
-      auto Start = std::chrono::system_clock::now();
-      // For each input section.
-      for (Section *Section : Sections) {
-        ELFSection *ELFSect = llvm::dyn_cast<eld::ELFSection>(Section);
+  // For each input section.
+  for (Section *Section : Sections) {
```
quic-seaswara: Change the variable Section to Sec or something like that.

lib/Object/ObjectBuilder.cpp
```diff
@@ -361,16 +357,22 @@ void ObjectBuilder::assignInputFromOutput(eld::InputFile *Obj) {
...
-          if (IsRetrySect && !In->isSpecial())
+          if (IsRetrySect && !In->isSpecial()) {
+            IsRetrySect = false;
             RetrySections.erase(Section);
+          }
           // Retry the match until the closest match is found.
-          if (In->isSpecial())
+          if (In->isSpecial()) {
+            IsRetrySect = true;
             RetrySections.insert(std::make_pair(Section, true));
```
quic-seaswara: we dont need this map.

lib/Object/ObjectBuilder.cpp
```diff
@@ -271,81 +271,77 @@ void ObjectBuilder::assignInputFromOutput(eld::InputFile *Obj) {
...
-        std::string SectName = Section->name().str();
-        uint64_t InputSectionHash = Section->sectionNameHash();
-        if (ELFSection *ELFSect = llvm::dyn_cast<ELFSection>(Section)) {
-          if (auto OptRThisSectionName =
-                  ObjFile->getRuleMatchingSectName(ELFSect->getIndex())) {
-            SectName = OptRThisSectionName.value();
-            InputSectionHash = llvm::hash_combine(SectName);
-          }
+        if (Section->getOutputSection() && !IsRetrySect) {
+          shouldBreak = false;
```
quic-seaswara: shouldSkipMatch instead of shouldBreak ?

lib/Object/ObjectBuilder.cpp
```diff
@@ -271,81 +271,77 @@ void ObjectBuilder::assignInputFromOutput(eld::InputFile *Obj) {
...
+          if (ELFSect->isWritable())
+            continue;
+          break;
+        case OutputSectDesc::ONLY_IF_RW:
+          if (!ELFSect->isWritable())
+            continue;
+          break;
+        }
+      }
+      bool shouldBreak = false;
```
quic-areg:
```suggestion
```
This shouldBreak is currently a no-op

lib/Object/ObjectBuilder.cpp
```diff
@@ -271,81 +271,77 @@ void ObjectBuilder::assignInputFromOutput(eld::InputFile *Obj) {
...
+    if (ELFSection *ELFSect = llvm::dyn_cast<ELFSection>(Section)) {
+      if (auto OptRThisSectionName =
+              ObjFile->getRuleMatchingSectName(ELFSect->getIndex())) {
+        SectName = OptRThisSectionName.value();
+        InputSectionHash = llvm::hash_combine(SectName);
+      }
+    }
+    // For all output sections.
+    for (auto *Out : SectionMap) {
+      if (ELFSect) {
```
quic-areg, line 317:
```suggestion
      if (Section->getOutputSection() && !IsRetrySect)
        break;
      if (ELFSect) {
```

lib/Object/ObjectBuilder.cpp
```diff
@@ -361,16 +357,22 @@ void ObjectBuilder::assignInputFromOutput(eld::InputFile *Obj) {
...
+          }
         } // end match
-      } // end each input section
-      In->addMatchTime(std::chrono::system_clock::now() - Start);
-    } // end each rule
-  } // end each output section
+        // RuleMatchTimes[In] += std::chrono::system_clock::now() - Start;
+      } // end each rule
+      if (shouldBreak)
+        break;
```
quic-areg, line 368:
```suggestion
```

lib/Object/ObjectBuilder.cpp
```diff
@@ -361,16 +357,22 @@ void ObjectBuilder::assignInputFromOutput(eld::InputFile *Obj) {
...
             RetrySections.insert(std::make_pair(Section, true));
+          }
         } // end match
-      } // end each input section
-      In->addMatchTime(std::chrono::system_clock::now() - Start);
-    } // end each rule
-  } // end each output section
+        // RuleMatchTimes[In] += std::chrono::system_clock::now() - Start;
+      } // end each rule
+      if (shouldBreak)
```
quic-areg:
```suggestion
```

lib/Object/ObjectBuilder.cpp
quic-areg: Is this still used anymore? We write to it but do we read it anywhere?

lib/Object/ObjectBuilder.cpp
parth-07: Why do you think it is a no-op? It's value is used at the end of the for (auto *Out : SectionMap) loop body to break out of this loop.

lib/Object/ObjectBuilder.cpp
parth-07: `if (Section->getOutputSection() && !IsRetrySect)` check needs to stay within the input-section-description loop (`for (auto *In : *Out)`). Otherwise, we will keep traversing input section descriptions even though the section has already matched an input section description.

lib/Object/ObjectBuilder.cpp
parth-07: We need `shouldBreak` here to signal that more output sections do not need to be traversed for this input section.

lib/Object/ObjectBuilder.cpp
parth-07: We need shouldBreak here to signal that more output sections do not need to be traversed for this input section.

lib/Object/ObjectBuilder.cpp
parth-07: I have removed this map.

lib/Object/ObjectBuilder.cpp
parth-07: I have removed this map.

lib/Object/ObjectBuilder.cpp
parth-07: Done.

lib/Object/ObjectBuilder.cpp
parth-07: Done.

lib/Object/ObjectBuilder.cpp
parth-07: Previously the inner-loop was not correctly setting it to true. It really was a no-op before. I have fixed the inner-loop to correctly set it true when we should break the outer loop as well.

quic-seaswara (approved):

Steven6798 (approved):

PR #822, opened 2026-02-13
Add debian packaging for eld

androm3da:


androm3da:
FIXME/TODO:

* maintainer email address
* can we build against `llvm-19-dev`?  Prob some compat issues here: llvm 19 is older, the `-dev` package is not like the source tree so unlikely to work as-is.

PR #828, opened 2026-02-17
Fix address of `_GLOBAL_OFFSET_TABLE_`

efriedma-quic:
On AArch64, it's supposed to be the address of the beginning of the GOT. The address itself doesn't actually matter that much; there isn't anything in particular located at the beginning of the GOT. But to make R_AARCH64_LD64_GOTPAGE_LO15 work correctly, it needs to be consistent with the address used by R_AARCH64_LD64_GOTPAGE_LO15, and it needs to be less than or equal to every GOT entry.

### DIFF
--- a/lib/Target/AArch64/AArch64LDBackend.cpp
+++ b/lib/Target/AArch64/AArch64LDBackend.cpp
@@ -202,44 +202,6 @@ Relocation::Type AArch64GNUInfoLDBackend::getCopyRelType() const {
   return llvm::ELF::R_AARCH64_COPY;
 }
 
-void AArch64GNUInfoLDBackend::defineGOTSymbol(Fragment &pFrag) {
-  // define symbol _GLOBAL_OFFSET_TABLE_
-  auto SymbolName = "_GLOBAL_OFFSET_TABLE_";
-  if (m_pGOTSymbol != nullptr) {
-    m_pGOTSymbol =
-        m_Module.getIRBuilder()
-            ->addSymbol<IRBuilder::Force, IRBuilder::Unresolve>(
-                pFrag.getOwningSection()->getInputFile(), SymbolName,
-                ResolveInfo::Object, ResolveInfo::Define, ResolveInfo::Local,
-                0x0, // size
-                0x0, // value
-                make<FragmentRef>(pFrag, 0x0), ResolveInfo::Hidden);
-  } else {
-    m_pGOTSymbol =
-        m_Module.getIRBuilder()
-            ->addSymbol<IRBuilder::Force, IRBuilder::Resolve>(
-                m_Module.getInternalInput(Module::Script), SymbolName,
-                ResolveInfo::Object, ResolveInfo::Define, ResolveInfo::Local,
-                0x0, // size
-                0x0, // value
-                make<FragmentRef>(pFrag, 0x0), ResolveInfo::Hidden);
-  }
-  if (m_Module.getConfig().options().isSymbolTracingRequested() &&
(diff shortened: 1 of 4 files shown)


lib/Target/AArch64/AArch64LDBackend.cpp
```diff
@@ -231,12 +231,18 @@ void AArch64GNUInfoLDBackend::defineGOTSymbol(Fragment &pFrag) {
...
-      frag = *GOTPLT->getFragmentList().begin();
-  if (frag)
+  if (auto *GOT = getGOT()) {
+    Fragment *frag = nullptr;
+    if (getGOT()->hasSectionData()) {
+      frag = *getGOT()->getFragmentList().begin();
+    } else {
+      // At this point, there may not be any fragments; the GOT doesn't have
+      // any entries in its header. Make an empty fragment as a placeholder.
+      frag = make<FillFragment>(m_Module, 0, 0, GOT, 1);
```
quic-seaswara: What if we have only .got.plt and no .got ? 

Why do you need an empty fragment, we should not be creating additional fragments.

efriedma-quic: I'm creating an empty fragment because I don't know how else to define m_pGOTSymbol correctly: at this point in the compilation, we haven't laid out the GOT, so I don't know how to figure out which GOT entry is "first".  So, I'm creating a fragment representing the point before the first GOT entry.

I think I can rearrange this code so the fragment is only created if we have GOT entries.

efriedma-quic: Pushed update that only creates the fragment when we have GOT relocations.

I still don't have any ideas for how to define m_pGOTSymbol without the fragment.  (Note that `Obj->getGOT()` and `getGOT()` refer to different objects: the second expands out to `m_DynamicSectionHeadersInputFile->getGOT()`.)

quic-seaswara: Thank you, I see the issue. m_DynamicSectionHeadersInputFile points to the header and each object has a GOT section that gets merged into the output section to support multithreading along with deterministic linking.

We can move setting up the value of GLOBAL_OFFSET_TABLE to the fragment that is pointing to the GOT in function finalizeTargetSymbols. 

You can fetch the GOT section by using m_Module.getSection(".got"), and point GLOBAL_OFFSET_TABLE to the first fragment in the section.

efriedma-quic: I couldn't figure out the whole fragment thing... but I managed to get a variation working without examining fragments directly.  Let me know if this is okay.

lib/Target/AArch64/AArch64LDBackend.cpp
```diff
@@ -606,6 +568,9 @@ bool AArch64GNUInfoLDBackend::finalizeTargetSymbols() {
         getRelaPLT()->getOutputSection()->getSection()->addr() +
         getRelaPLT()->getOutputSection()->getSection()->size());
   }
+  if (m_pGOTSymbol) {
+    m_pGOTSymbol->setValue(getGOT()->pAddr());
+  }
```
quic-seaswara, line 573: pAddr -> points to the physical address, you might want to say addr() here for virtual address.

quic-seaswara (changes requested):

lib/Target/AArch64/AArch64LDBackend.cpp
efriedma-quic: Done.

PR #839, opened 2026-02-21
Remove the redundant return values

### LINKED ISSUE #611: `Relocation::issueOverflow` return value is redundant
The return value of `Relocation::issueOverflow` is redundant and should perhaps be removed.

def3r:
Change the return type of `Relocation::issueSignedOverflow` and `Relocation::issueUnsignedOverflow` to void

It looks like `Relocation::issueUnencodableImmediate` also falls under the same category. Should I include it too?

closes #611

### DIFF
--- a/include/eld/Readers/Relocation.h
+++ b/include/eld/Readers/Relocation.h
@@ -145,10 +145,10 @@ public:
 
   bool isMergeKind() const;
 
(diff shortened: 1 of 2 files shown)


quic-seaswara (approved):
Thanks for the patch!

PR #872, opened 2026-02-26
Make tests independent of the default triple

ShikharJ-Corp:
This commit improves a few test so that they pass independent of the
default triple that was used to configure/build linker.

### DIFF
--- a/test/ARM/standalone/GCexternList/GCexternList.test
+++ b/test/ARM/standalone/GCexternList/GCexternList.test
@@ -6,5 +6,5 @@ RUN: %link %linkopts -march arm %t1.o %tlib.a --entry main -o %t.out --extern-li
 RUN: %nm -n %t.out  | %filecheck %s
 
 #CHECK:         U externSym
-#CHECK:00000014 T foo
+#CHECK: T foo
 
--- a/test/ARM/standalone/Relocs/R_ARM_THM_JUMP24/R_ARM_THM_JUMP24.test
+++ b/test/ARM/standalone/Relocs/R_ARM_THM_JUMP24/R_ARM_THM_JUMP24.test
@@ -5,7 +5,7 @@
 #END_COMMENT
 #START_TEST
 RUN: %clang -c %p/Inputs/x.s -o %t1.1.o
-RUN: %link %linkopts -march arm %t1.1.o --defsym arm_callee1=1 -o %t2.out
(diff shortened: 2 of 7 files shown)


test/ARM/standalone/GCexternList/GCexternList.test
```diff
@@ -6,5 +6,5 @@ RUN: %link %linkopts -march arm %t1.o %tlib.a --entry main -o %t.out --extern-li
 RUN: %nm -n %t.out  | %filecheck %s
 
 #CHECK:         U externSym
-#CHECK:00000014 T foo
+#CHECK:{{[0-9a-f]+}} T foo
```
parth-07:
```suggestion
#CHECK: T foo
```

If the symbol value is not important here, then we should remove it altogether.

test/ARM/standalone/SymDefAddend/SymDefAddend.test
```diff
@@ -8,6 +8,6 @@ RUN: %link -march arm %t1.1.o --defsym arm_val=1 -o %t2.out
 RUN: %objdump -s %t2.out | %filecheck %s
 
 #CHECK: Contents of section .data:
-#CHECK-NEXT: 0000 02000000
+#CHECK-NEXT: {{[0-9a-fA-F]+}} 02000000
```
parth-07, line 11: Do you know why this changes from 0 to a different number for windows?

test/ARM/standalone/Relocs/R_ARM_THM_JUMP24/R_ARM_THM_JUMP24.test
```diff
@@ -8,5 +8,5 @@ RUN: %clang -c %p/Inputs/x.s -o %t1.1.o
 RUN: %link %linkopts -march arm %t1.1.o --defsym arm_callee1=1 -o %t2.out
 RUN: %objdump -d %t2.out | %filecheck %s
 
-#CHECK:  0: f7ff bffe
+#CHECK: f000 b800{{.*}}b.w{{.*}}__arm_callee1_T2T_veneer@island-1
```
parth-07: This seems to be failing for me locally -- the test is not running on the builder at all because of an incorrect non-doc changes step that is skipping building and running of tests. In particular, I am getting this output instead:

```
00000000 <thumb_caller>:
       0: f7ff bffe    	b.w	0x0 <thumb_caller>      @ imm = #-0x4
```

This needs a closer look because of the difference in behavior.

PR #877, opened 2026-02-26
Fix pauth relocations tests

nezetic:
Nightly tests failed due to the order of relocations being different than expected.

Use -DAG instead of -NEXT to make tests more robust.

### DIFF
--- a/test/AArch64/standalone/Relocations/PAuthReloc.s
+++ b/test/AArch64/standalone/Relocations/PAuthReloc.s
@@ -10,25 +10,24 @@
 # RUN: %llvm-mc -filetype=obj -triple=aarch64 main.s -o main.o
 
 # Test PAuth relocations with PIE
-# RUN: %link %linkopts main.o -pie a.so -o main
+# RUN: %link %linkopts -z notext main.o -pie a.so -o main
 # RUN: %readelf --elf-output-style LLVM -r main 2>&1 | %filecheck %s --check-prefix=UNPACKED
 
 # UNPACKED:          Section ({{.+}}) .rela.dyn {
-# UNPACKED-NEXT:       {{0x[0-9A-F]+}} R_AARCH64_AUTH_RELATIVE - {{0x[0-9A-F]+}}
-# UNPACKED-NEXT:       {{0x[0-9A-F]+}} R_AARCH64_AUTH_RELATIVE - {{0x[0-9A-F]+}}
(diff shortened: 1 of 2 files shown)


test/AArch64/standalone/Relocations/PAuthReloc.s
```diff
@@ -92,67 +92,67 @@
...
+# STATIC_NOPIE-DAG:       {{0x[0-9A-F]+}} R_AARCH64_AUTH_RELATIVE - {{0x[0-9A-F]+}}
+# STATIC_NOPIE-DAG:       {{0x[0-9A-F]+}} R_AARCH64_AUTH_RELATIVE - {{0x[0-9A-F]+}}
+# STATIC_NOPIE-DAG:       {{0x[0-9A-F]+}} R_AARCH64_AUTH_RELATIVE - {{0x[0-9A-F]+}}
+# STATIC_NOPIE-DAG:       {{0x[0-9A-F]+}} R_AARCH64_AUTH_RELATIVE - {{0x[0-9A-F]+}}
+# STATIC_NOPIE-DAG:       {{0x[0-9A-F]+}} R_AARCH64_AUTH_RELATIVE - {{0x[0-9A-F]+}}
+# STATIC_NOPIE-DAG:       {{0x[0-9A-F]+}} R_AARCH64_AUTH_RELATIVE - {{0x[0-9A-F]+}}
+# STATIC_NOPIE-DAG:       {{0x[0-9A-F]+}} R_AARCH64_AUTH_RELATIVE - {{0x[0-9A-F]+}}
+# STATIC_NOPIE-DAG:       {{0x[0-9A-F]+}} R_AARCH64_AUTH_RELATIVE - {{0x[0-9A-F]+}}
+# STATIC_NOPIE-DAG:       {{0x[0-9A-F]+}} R_AARCH64_AUTH_RELATIVE - {{0x[0-9A-F]+}}
+# STATIC_NOPIE-DAG: }
```
quic-seaswara: remove the last line that has a DAG. 

May be do you want to use CHECK-COUNT instead ? It might be more robust.

nezetic: I removed the lines with  "-DAG: }" and replaced the STATIC-NOPIE-NEXT with a -COUNT. I kept the -DAG elsewhere as -COUNT will be matching the relocations where values are explicitly checked.

PR #878, opened 2026-02-26
x86 is a 64 bit target

quic-areg:
x86 needs to be counted as a 64-bit target for
tests that are intended only to run on 64/32 bit
targets to work properly.

### DIFF
--- a/test/lit.cfg
+++ b/test/lit.cfg
@@ -123,7 +123,7 @@ if getattr(config, 'has_symbol_versioning', 'false') == 'true':
     config.available_features.add('symbol_versioning')
 
 # `config.eld_targets_to_build` is actually the name of the test config
-# for a single target, eg. 'hexgaon_default'.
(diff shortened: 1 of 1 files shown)


test/lit.cfg
```diff
@@ -122,8 +122,11 @@ except ImportError:
 if getattr(config, 'has_symbol_versioning', 'false') == 'true':
     config.available_features.add('symbol_versioning')
 
-if ('RISCV64' in config.eld_targets_to_build or
-     'AArch64' in config.eld_targets_to_build):
+# `config.eld_targets_to_build` is actually the name of the test config
+# for a single target, eg. 'hexgaon_default'.
```
quic-seaswara: hexagon

quic-areg:
```suggestion
# for a single target, eg. 'hexagon_default'.
```

PR #879, opened 2026-02-26
Add REQUIRES: shell to LTOTempObjNotWritable test

quic-garvgupt:
The LTOTempObjNotWritable test fails on Windows because cmd.exe does not support the shell‑specific syntax used in the test. In particular, cmd.exe interprets the parentheses around %not as separate tokens, leading to incorrect command parsing.

This change adds a REQUIRES: shell directive to the test, marking it as unsupported in environments that lack a Bash‑like shell.


quic-seaswara:
This requires a testing link with this patch that the build is successful with running tests.

@apazos can you please let me know if this can be merged ?

apazos:
Not yet, it needs the test PR (draft) pushed to CPULLVM and clean precheck-ins.

PR #881, opened 2026-02-26
[aarch64] Fix TLS IE/GD support

### LINKED ISSUE #868: Incorrect offset for TLS accesses on AArch64 (initial-exec tls model)
Having an issue with picolibc's TLS tests once again, when compiled with `-ftls-model=initial-exec`.

This is basically a replica of #302 , same test and same issue, but this time since we're using initial-exec, the issue is with the `R_AARCH64_TLSIE_ADR_GOTTPREL_PAGE21` relocation. 

Recap:
> picolibc's TLS tests (https://github.com/picolibc/picolibc/blob/main/test/tls.c) are currently failing when built with eld. I think this is due to incorrect offsets being used when accessing TLS variables where other TLS variables with sufficiently large alignments are present.

> It looks like eld is always using a 16-byte offset (for the TCB I guess) rather than also accounting for possible alignment padding which is what lld/bfd/picolibc all seem to be doing or expecting. I'm not sure what the official source is, but a few possibly-relevant snippets from codebases about this:

Copying over some of Jonathon's test code and modifying it a bit:

**Reproducer**
```
cat > test.c << '!'
__thread int a = 12;
__attribute__((aligned(128)))
__thread int b = 13;

int foo(void) { return a; }
!

clang --target=aarch64-none-elf -c test.c -ftls-model=initial-exec -fPIC
ld.eld -maarch64linux test.o -o eld.a.out
ld.lld test.o -o lld.a.out --no-relax
ld.bfd test.o -o bfd.a.out
```
 (note: lld decides to optimize out the GOT relocation (presumably a relaxation to local exec? I can't seem to turn it off with --no-relax))

**Output**

(to reproduce the output: `llvm-objdump -dj .text -dj .got <linker>.a.out`

eld:
```
Disassembly of section .text:
0000000000400120 <foo>:
  400120: b0000009      adrp    x9, 0x401000 <b+0x400f80>
  400124: f9404529      ldr     x9, [x9, #0x88]
  400128: d53bd048      mrs     x8, TPIDR_EL0
  40012c: b8696900      ldr     w0, [x8, x9]
  400130: d65f03c0      ret

Disassembly of section .got:

0000000000401088 <.got>:
  401088: 00000010      udf     #0x10     <-----
  40108c: 00000000      udf     #0x0
```
ld.bfd:
```
Disassembly of section .text:
0000000000400158 <foo>:
  400158: f00000e9      adrp    x9, 0x41f000 <foo+0x1eea8>
  40015c: f947c929      ldr     x9, [x9, #0xf90]
  400160: d53bd048      mrs     x8, TPIDR_EL0
  400164: b8696900      ldr     w0, [x8, x9]
  400168: d65f03c0      ret

Disassembly of section .got:

000000000041ff88 <_GLOBAL_OFFSET_TABLE_>:
                ...
  41ff90: 00000080      udf     #0x80    <-----
  41ff94: 00000000      udf     #0x0
```
ld.lld:
```
Disassembly of section .text:
00000000002101c8 <foo>:
  2101c8: d2a00009      movz    x9, #0x0, lsl #16
  2101cc: f2801009      movk    x9, #0x80          <-----
  2101d0: d53bd048      mrs     x8, TPIDR_EL0
  2101d4: b8696900      ldr     w0, [x8, x9]
  2101d8: d65f03c0      ret
```
The important thing to note is that in ld's GOT and in lld's optimized code, the offset is 0x80 while the offset in the GOT for eld is 0x10. Similar difference in #302 .

quic-seaswara:
This fixes an issue when object files built with TLS models with large alignments would not produce the right values when used with *static* links

- initial-exec
- global-dynamic

Fix description :
------------------

- AArch64IEGOT was incorrectly created with GOT::TLS_LE
- compute TLS offsets as done in static links to properly set the GOT slots reserved for IE/GD models
- updateTLSIEGOTOffsets() applies the correct static TLS block offset

Resolves #868

### DIFF
--- a/lib/Target/AArch64/AArch64Relocator.cpp
+++ b/lib/Target/AArch64/AArch64Relocator.cpp
@@ -1109,9 +1109,15 @@ Relocator::Result tls_gottprel_page(Relocation &pReloc,
                                     AArch64Relocator &pParent) {
   DiagnosticEngine *DiagEngine = pParent.config().getDiagEngine();
   Relocator::DWord A = pReloc.addend();
-  Relocator::DWord X = pParent.getSymValue(&pReloc) + 0x10;
 
   if (!(pReloc.symInfo()->reserved() & Relocator::ReserveGOT)) {
+    const auto &OptTLSBlockVarOffset = pParent.getStaticTLSBlockVarOffset();
+    if (!OptTLSBlockVarOffset.has_value()) {
+      pParent.config().raise(Diag::no_pt_tls_segment);
+      return Relocator::Result::BadReloc;
+    }
+    Relocator::DWord X =
+        pParent.getSymValue(&pReloc) + OptTLSBlockVarOffset.value();
     // Convert to movz
     uint32_t movz = 0xD2A00000 | (pReloc.target() & 0x0000001F);
     pReloc.target() = helper_reencode_movzk_imm(movz, X >> 16);
@@ -1133,9 +1139,15 @@ Relocator::Result tls_gottprel_page(Relocation &pReloc,
 Relocator::Result tls_gottprel_lo(Relocation &pReloc,
                                   AArch64Relocator &pParent) {
   Relocator::DWord A = pReloc.addend();
-  Relocator::DWord X = pParent.getSymValue(&pReloc) + 0x10;
 
   if (!(pReloc.symInfo()->reserved() & Relocator::ReserveGOT)) {
+    const auto &OptTLSBlockVarOffset = pParent.getStaticTLSBlockVarOffset();
+    if (!OptTLSBlockVarOffset.has_value()) {
+      pParent.config().raise(Diag::no_pt_tls_segment);
+      return Relocator::Result::BadReloc;
+    }
+    Relocator::DWord X =
+        pParent.getSymValue(&pReloc) + OptTLSBlockVarOffset.value();
     // Convert to movk
     uint32_t movk = 0xF2800000 | (pReloc.target() & 0x0000001F);
     pReloc.target() = helper_reencode_movzk_imm(movk, X);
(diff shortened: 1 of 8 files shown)


lib/Target/AArch64/AArch64Relocator.cpp
```diff
@@ -1206,9 +1224,15 @@ Relocator::Result tls_tlsdesc_page(Relocation &pReloc,
...
 
   if (!(pReloc.symInfo()->reserved() & Relocator::ReserveGOT)) {
+    const auto &OptTLSBlockVarOffset = pParent.getStaticTLSBlockVarOffset();
+    if (!OptTLSBlockVarOffset.has_value()) {
+      pParent.config().raise(Diag::no_pt_tls_segment);
+      return Relocator::Result::BadReloc;
+    }
+    Relocator::DWord X =
+        pParent.getSymValue(&pReloc) + OptTLSBlockVarOffset.value();
     // Convert to movk, save to x0
```
efriedma-quic, line 1236: I guess this isn't directly related to this patch, but why are we hardcoding x0 here?

(It would be nice to have some additional test coverage for the relocations.)

quic-seaswara: It appears that its a incorrect comment. I dont see a difference. Will delete it.

Yes you are right, the relocations need additonal test coverage, which I will work on a follow up by comparing with lld.

If you have pointers of additional testing that will be useful, would appreciate it.

efriedma-quic: I think the comment is correctly describing the code: 0xF2800000 is a movk to x0.

Compare to tls_gottprel_lo, which uses `uint32_t movk = 0xF2800000 | (pReloc.target() & 0x0000001F);`.

quic-seaswara:
Looks like the bot ran the tests successfully.

Will go ahead and merge.

PR #883, opened 2026-02-26
Create nightly Zephyr workflow for riscv32

quic-areg:
Add a nightly GitHub Actions workflow that builds Zephyr and runs tests using the ELD nightly toolchain.

Use picolibc and libgcc from the Zephyr SDK because the nightly toolchain does not ship compiler-rt or C libraries.

Follow-ups:
- Filter known-failing tests so the workflow can pass.
- Support remaining targets.
- Use docker to simplify installing dependencies.

### DIFF
--- a/.github/workflows/zephyr-nightly.yml
+++ b/.github/workflows/zephyr-nightly.yml
@@ -1,0 +1,201 @@
+name: Zephyr Nightly with ELD
+
+on:
+  # pull_request: {} # Uncomment only to test this WF file update.
+  schedule:
+    - cron: '0 9 * * *'
+  workflow_dispatch:
+
+jobs:
+  zephyr-build-and-test:
+    if: github.repository == 'qualcomm/eld' || github.event_name == 'workflow_dispatch'
+    runs-on: ubuntu-latest
+    env:
+      ZEPHYR_WORKSPACE: ${{ github.workspace }}/zephyrproject
+
+    steps:
+      - name: Checkout eld
+        uses: actions/checkout@v4
+        with:
+          path: llvm-project/llvm/tools/eld
+
+      - name: Record pre-build entry
+        if: github.event_name == 'schedule'
+        uses: ./llvm-project/llvm/tools/eld/.github/workflows/BuildStatusDataRecorder
+        with:
+          enabled: true
+          project: "llvm-project/llvm/tools/eld"
+          stage: "pre-"
+          workflow-name: "zephyr"
+          record-mode: "record"
+          status: "fail"
+          run-id: ${{github.run_id}}
+          arch-name: "riscv32"
+          branch-name: "main"
+
+      - name: Install dependencies
+        run: |
+          export DEBIAN_FRONTEND=noninteractive
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends \
+            libc++-dev libc++abi-dev libclang-rt-20-dev \
+            git cmake ninja-build gperf dfu-util device-tree-compiler wget \
+            python3-dev python3-pip python3-setuptools python3-tk python3-venv python3-wheel xz-utils \
+            file make gcc gcc-multilib g++-multilib libsdl2-dev libmagic1 qemu-system-riscv32
+
+          python3 -m venv ${{ env.ZEPHYR_WORKSPACE }}/.venv
+          . ${{ env.ZEPHYR_WORKSPACE }}/.venv/bin/activate
+          python -m pip install -U pip
+          python -m pip install -U west semver patool
+
+          cd ${{ github.workspace }}
+          west init -m https://github.com/zephyrproject-rtos/zephyr --mr main ${{ env.ZEPHYR_WORKSPACE }}
+          cd ${{ env.ZEPHYR_WORKSPACE }}
+          west update
+          west zephyr-export
+          west packages pip --install
+
+      - name: Fetch nightly toolchain
+        uses: ./llvm-project/llvm/tools/eld/.github/workflows/FetchNightlyToolset
+
+      - name: Set up Zephyr environment
+        run: |
+          echo "ZEPHYR_VENV=${{ env.ZEPHYR_WORKSPACE }}/.venv" >> $GITHUB_ENV
+
+      - name: Install Zephyr SDK
+        run: |
+          cd ${{ env.ZEPHYR_WORKSPACE }}
+          . ${{ env.ZEPHYR_WORKSPACE }}/.venv/bin/activate
+          west sdk install -t riscv64-zephyr-elf
+
+      - name: Install RISC-V toolchain payload
+        run: |
+          SDK_DIR=$(ls -d "$HOME"/zephyr-sdk-* 2>/dev/null | head -n 1)
+          if [ -z "$SDK_DIR" ]; then
+            echo "Failed to locate Zephyr SDK under $HOME"
+            ls -la "$HOME" || true
+            exit 1
+          fi
+          cd "$SDK_DIR"
+          ./setup.sh -t riscv64-zephyr-elf
+
+          PICOLIBC_SYSROOT=${SDK_DIR}/riscv64-zephyr-elf/picolibc
+          PICOLIBC_LIB_DIR=${PICOLIBC_SYSROOT}/riscv64-zephyr-elf/lib/rv32imac_zicsr_zifencei/ilp32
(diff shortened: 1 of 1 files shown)


.github/workflows/zephyr-nightly.yml
```diff
@@ -0,0 +1,173 @@
...
+    env:
+      ZEPHYR_WORKSPACE: ${{ github.workspace }}/zephyrproject
+
+    steps:
+      - name: Checkout eld
+        uses: actions/checkout@v4
+        with:
+          path: llvm-project/llvm/tools/eld
+
+      - name: Install dependencies
```
jonathonpenix, line 36: FWIW, Zephyr maintains/publishes docker images that might be useful here/save you from having to maintain some of this. See https://github.com/zephyrproject-rtos/docker-image (then I'm not sure, but something like [this](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idcontainer) would work maybe?)

Will leave it up to you if you want to try this (or, maybe you have already? I'd be interested in hearing how it went if you did)

.github/workflows/zephyr-nightly.yml
```diff
@@ -0,0 +1,173 @@
...
+            -x LLVM_TOOLCHAIN_PATH=${{ github.workspace }}/inst \
+            -x TOOLCHAIN_HAS_PICOLIBC=ON \
+            -x SYSROOT_DIR=${{ env.ELD_PICOLIBC_SYSROOT }} \
+            -x TOOLCHAIN_C_FLAGS=--gcc-install-dir=${{ env.ELD_GCC_INSTALL_DIR }} \
+            -x CONFIG_LLVM_USE_ELD=y \
+            -x CONFIG_PICOLIBC_USE_MODULE=n \
+            -x CONFIG_PICOLIBC_USE_TOOLCHAIN=y \
+            -x CONFIG_COMPILER_RT_RTLIB=n \
+            -x CONFIG_LIBGCC_RTLIB=y \
+            -x CMAKE_EXE_LINKER_FLAGS="-L${{ env.ELD_PICOLIBC_LIB_DIR }}"
```
jonathonpenix, line 169: Been a while since I've run this so I'm not sure if they're still relevant but, a few options that might be helpful:
- `--retry-failed 3`: We had some flakiness with some qemu tests in the past and this would help sift through some of them. Not sure if still needed.
- `-x=TOOLCHAIN_HAS_NEWLIB=n`: Even when we'd force picolibc as module  or minimal libc, etc. there would still be some failures related to newlib so we set this. Not sure if you'll still see these?

.github/workflows/zephyr-nightly.yml
```diff
@@ -0,0 +1,173 @@
...
+            -j 1 \
+            -v \
+            -x ZEPHYR_SDK_INSTALL_DIR=${{ env.ELD_SDK_DIR }} \
+            -x ZEPHYR_TOOLCHAIN_VARIANT=llvm \
+            -x LLVM_TOOLCHAIN_PATH=${{ github.workspace }}/inst \
+            -x TOOLCHAIN_HAS_PICOLIBC=ON \
+            -x SYSROOT_DIR=${{ env.ELD_PICOLIBC_SYSROOT }} \
+            -x TOOLCHAIN_C_FLAGS=--gcc-install-dir=${{ env.ELD_GCC_INSTALL_DIR }} \
+            -x CONFIG_LLVM_USE_ELD=y \
+            -x CONFIG_PICOLIBC_USE_MODULE=n \
```
jonathonpenix: Do you need to force this off? IIRC there's some tests that try to run under this (and, there's some users that use it today), and having the extra coverage in theory can't hurt (if it isn't break anything by not disabling this).

.github/workflows/zephyr-nightly.yml
```diff
@@ -0,0 +1,173 @@
...
+            ${{ env.ZEPHYR_WORKSPACE }}/zephyr/twister-out/twister.log
+            ${{ env.ZEPHYR_WORKSPACE }}/zephyr/twister-out/twister.json
+
+      - name: Upload build artifacts
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: zephyr-build-artifacts
+          path: |
+            ${{ env.ZEPHYR_WORKSPACE }}/zephyr/build/zephyr/zephyr.elf
```
jonathonpenix, line 186: Is this just the .elf/map from the hello world?

Just curious if this is going to be used for something or just for debugging, etc.

.github/workflows/zephyr-nightly.yml
```diff
@@ -0,0 +1,173 @@
...
+          else
+            echo "✗ hello_world did not produce expected output"
+            exit 1
+          fi
+
+      - name: Run twister tests for qemu_riscv32 with ELD
+        run: |
+          cd ${{ env.ZEPHYR_WORKSPACE }}/zephyr
+          . ${{ env.ZEPHYR_WORKSPACE }}/.venv/bin/activate
+          ./scripts/twister -W -p qemu_riscv32 \
```
quic-shaksuma, line 153: Add a timeout here as well?

.github/workflows/zephyr-nightly.yml
```diff
@@ -0,0 +1,173 @@
...
+    runs-on: ubuntu-latest
+    env:
+      ZEPHYR_WORKSPACE: ${{ github.workspace }}/zephyrproject
+
+    steps:
+      - name: Checkout eld
+        uses: actions/checkout@v4
+        with:
+          path: llvm-project/llvm/tools/eld
+
```
quic-shaksuma, line 21: It would be nice to start recording build dashboard data using the BuildStatusDataRecorder action. Refer to other workflows e.g. picolibc

.github/workflows/zephyr-nightly.yml
quic-areg: Yes; I did not have any specific use case in mind beyond debugging

.github/workflows/zephyr-nightly.yml
quic-areg: So far I am seeing a consistent number of failures: `- 4670 of 5707 executed test cases passed (81.83%)`, so I think we should exclude retries for now. The workflow takes about 5.5 hours to complete currently.

I see newlib tests getting filtered, but I will try running it again with this flag to see if anything changes.

jonathonpenix: Sure, sounds good re: about retries.

> I see newlib tests getting filtered, but I will try running it again with this flag to see if anything changes.

On this, you might just be able to check for failing tests with 'newlib' in the name--if there aren't any, it should be fine. Not sure if that has been cleaned up upstream or something.

.github/workflows/zephyr-nightly.yml
```diff
@@ -0,0 +1,173 @@
+name: Zephyr Nightly with ELD
+
+on:
+  # pull_request: {} # Uncomment only to test this WF file update.
+  schedule:
+    - cron: '0 9 * * *'
+  workflow_dispatch:
+
+jobs:
+  zephyr-build-and-test:
```
jonathonpenix, line 10: Does this need to be gated on the repo too similar to what was done in https://github.com/qualcomm/eld/pull/902?

.github/workflows/zephyr-nightly.yml
quic-areg: Thanks for the suggestion; I will include this in a follow-up.

.github/workflows/zephyr-nightly.yml
quic-areg: There is no difference with or without this option, but it is probably better to be explicit so I added it.

.github/workflows/zephyr-nightly.yml
quic-areg: Removed `CONFIG_PICOLIBC_USE_MODULE`

.github/workflows/zephyr-nightly.yml
quic-areg: Done

quic-areg:
> How much time does it take to run all the tests, should we run this on self-hosted ?

My last run took 3h40m

.github/workflows/zephyr-nightly.yml
quic-areg: Done

quic-seaswara:
@quic-areg  please investigate the failures here 

https://github.com/qualcomm/eld/actions/runs/22692839505/job/65792368843?pr=883

I am not sure if there is a non determinism here.

PR #884, opened 2026-02-26
Add support for parsing SORT(CONSTRUCTORS)

quic-seaswara:
As per LLD comments, this is a NFC for ELF targets

### DIFF
--- a/lib/ScriptParser/ScriptParser.cpp
+++ b/lib/ScriptParser/ScriptParser.cpp
@@ -647,6 +647,7 @@ void ScriptParser::readOverlayMemberOutputSectionDescription(
     } else if (readInclude(Tok)) {
     } else if (readOutputSectionData(Tok)) {
     } else if (readAssignment(Tok)) {
+    } else if (readSortConstructors(Tok)) {
     } else {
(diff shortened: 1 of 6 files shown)


lib/ScriptParser/ScriptParser.cpp
parth-07, line 773: We need `readSortConstructors` here as well.

lib/ScriptParser/ScriptParser.cpp
```diff
@@ -647,6 +647,7 @@ void ScriptParser::readOverlayMemberOutputSectionDescription(
     } else if (readInclude(Tok)) {
     } else if (readOutputSectionData(Tok)) {
     } else if (readAssignment(Tok)) {
+    } else if (readSortConstructors(Tok)) {
     } else {
       readInputSectionDescription(Tok);
```
parth-07, line 652: I believe that because `SORT(CONSTRUCTORS)` is an input section description, it should be parsed as part of `readInputSectionDescription` function.

quic-seaswara: Thanks for catching that.

PR #890, opened 2026-02-27
Fix __eh_fram_* symbols

quic-k:


parth-07 (commented):
What do you think about adding tests in eld to make sure that nobody accidentally cause a regression in the template linker scripts?

PR #902, opened 2026-03-03
[CI] Don't run Nightly Workflows on Forks

lenary:
These are a bunch of ELD workflows that relate to ELD testing, and that downstream forks should not be running.

When `qualcomm/eld` is forked, then the fork gets a `main` branch that matches the `qualcomm/eld` `main` branch at the time it was forked, and because this is the fork's default branch, it cannot be deleted. This prevents spamming people's notifications with failing nightly build notifications which aren't relevant.

### DIFF
--- a/.github/workflows/ci-win.yml
+++ b/.github/workflows/ci-win.yml
@@ -9,6 +9,7 @@ on:
 
 jobs:
   build-windows:
+    if: github.repository == 'qualcomm/eld'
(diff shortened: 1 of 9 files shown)


lenary:
I don't know if I got all the right ones, I was looking for things that were looking for `main`, or things that had cron schedules in them. I might have missed some or included too many.

PR #905, opened 2026-03-03
Don't run release-artifact-builder.yaml in forks

jonathonpenix:
This is an extension of what was done in https://github.com/qualcomm/eld/pull/902-- this runs when updating one's fork and (for me at least) always fails and doesn't seem like something we actually want running in people's forks.

Not sure if this was omitted accidentally or if I just have something goofy in my environment that it wasn't running for others.

### DIFF
--- a/.github/workflows/release-artifact-builder.yaml
+++ b/.github/workflows/release-artifact-builder.yaml
@@ -12,6 +12,7 @@ on:
 
 jobs:
   release-toolchain:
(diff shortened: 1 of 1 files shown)


jonathonpenix:
Sorry, one more ping on this, this builder doesn't actually even work in eld itself right now as far as I can tell: https://github.com/qualcomm/eld/actions/workflows/release-artifact-builder.yaml

PR #907, opened 2026-03-03
[AArch64] Improve IFunc support for AArch64 static executables

### LINKED ISSUE #903: eld does not handle `IFUNC` symbols properly for static non-pie executables
`IFUNC` symbol is a GNU extension. It allows a function definition to be resolved at runtime. More information about `IFUNC` symbols can be found here: https://gcc.gnu.org/onlinedocs/gcc-5.3.0/gcc/Function-Attributes.html#index-g_t_0040code_007bifunc_007d-function-attribute-3095

We do not handle `IFUNC` symbols properly for static non-pie executables. In particular, we do not create the required `IRELATIVE` relocation, and PLT and .got.plt entry for the ifunc symbol when it is stored in a function pointer instead of being directly called. 

There are also some other minor issues/improvements which we may want to fix in the future:

1) `__rela_iplt_[start|end]` symbols created by eld have attributes `OBJECT`, `GLOBAL`, `HIDDEN`, `ABS`. It should instead have attributes `NOTYPE`, `LOCAL`, `HIDDEN`, <section-relative>.
2) `__rela_iplt_end` symbol unnecessarily have an associated size.
3) LLD stores PLT entries for ifunc symbols in `.iplt` section instead of `.plt`. This change is not required for correctness, and seems to be for better organization / easier code maintenance.

Reproducible example:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
#include <stdio.h>

char *foo1() {
  return "foo";
}

char *foo2() {
  return "bar";
}

static char *(*resolve_foo (void)) (void) {
  return foo1;
}

char *foo();
EOF

cat >main.c <<\EOF
#include <stdio.h>
char *foo();

int main() {
  char * (*foop)(void) = foo;
  printf("Hello %s\n", (foop()));
//  printf("Hello %s\n", foo());
  return 0;
}
EOF
rm ./ld
cat >ld <<\EOF
#!/usr/bin/env bash

ld.eld "$@"
EOF

chmod +x ./ld
aarch64-none-linux-gnu-gcc -o 1.eld.out 1.c main.c -O0 -static -B .
aarch64-none-linux-gnu-gcc -o 1.lld.out 1.c main.c -O0 -static -fuse-ld=lld
aarch64-none-linux-gnu-gcc -o 1.bfd.out 1.c main.c -O0 -static 

qemu-aarch64 ./1.eld.out
qemu-aarch64 ./1.lld.out
qemu-aarch64 ./1.bfd.out
```

Output:

```
+ qemu-aarch64 ./1.eld.out
Hello �
+ qemu-aarch64 ./1.lld.out
Hello foo
+ qemu-aarch64 ./1.bfd.out
Hello foo
```

eld does not correctly call `foo`. This is because eld does not create a `IRELATIVE` relocation, and PLT and .got.plt entries for the symbol `foo`
when it is not called directly as in the case above.

androm3da:
This commit improves GNU IFunc support for static executables.
GNU IFunc allows a function definition to be resolved at runtime.
More information about IFUNC symbols can be found here:
https://gcc.gnu.org/onlinedocs/gcc-5.3.0/gcc/Function-Attributes.html#index-g_t_0040code_007bifunc_007d-function-attribute-3095

For each non-preemptible ifunc symbol that is referred in the codebase,
we create a .plt entry and a corresponding .got.plt entry. In
particular, we create .plt/.got.plt entry for an ifunc symbol when it is
referred by either of these relocation types:
- R_AARCH64_ABS*
- R_AARCH64_(CONDBR19|JUMP26|CALL26)
- R_AARCH_ADR_PREL_PG_*
- R_AARCH64_ADR_GOT_*

For ABS, CALL*, PREL_PG_* relocations, the value of ifunc symbol is
taken as the address of its .plt entry.

For ADR_GOT_* relocations, the value of the ifunc symbol is taken as the
address of its .got.plt entry.

Resolves #903

### DIFF
--- a/lib/Target/AArch64/AArch64Relocator.cpp
+++ b/lib/Target/AArch64/AArch64Relocator.cpp
@@ -7,6 +7,7 @@
 
 #include "AArch64Relocator.h"
 #include "AArch64InsnHelpers.h"
+#include "AArch64PLT.h"
 #include "AArch64RelocationFunctions.h"
 #include "AArch64RelocationHelpers.h"
 #include "eld/Config/LinkerConfig.h"
@@ -344,13 +345,14 @@ void AArch64Relocator::scanGlobalReloc(InputFile &pInput, Relocation &pReloc,
   ELFObjectFile *Obj = llvm::dyn_cast<ELFObjectFile>(&pInput);
   // rsym - The relocation target symbol
   ResolveInfo *rsym = pReloc.symInfo();
-
   switch (pReloc.type()) {
   case llvm::ELF::R_AARCH64_AUTH_ABS64:
   case llvm::ELF::R_AARCH64_ABS16:
   case llvm::ELF::R_AARCH64_ABS32:
   case llvm::ELF::R_AARCH64_ABS64: {
     std::lock_guard<std::mutex> relocGuard(m_RelocMutex);
+    if (handleScanForNonPreemptibleIFunc(rsym, Obj))
+      return;
     bool isAuthAbs = pReloc.type() == llvm::ELF::R_AARCH64_AUTH_ABS64;
     // Absolute relocation type, symbol may needs PLT entry or
     // dynamic relocation entry
@@ -447,18 +449,12 @@ void AArch64Relocator::scanGlobalReloc(InputFile &pInput, Relocation &pReloc,
   case llvm::ELF::R_AARCH64_JUMP26:
   case llvm::ELF::R_AARCH64_CALL26: {
     std::lock_guard<std::mutex> relocGuard(m_RelocMutex);
+    if (handleScanForNonPreemptibleIFunc(rsym, Obj))
+      return;
     // return if we already create plt for this symbol
     if (rsym->reserved() & ReservePLT)
       return;
 
-    // create IRELATIVE for IFUNC symbol
-    if (rsym->type() == ResolveInfo::IndirectFunc && config().isCodeStatic()) {
-      m_Target.createPLT(Obj, rsym, true);
-      rsym->setReserved(rsym->reserved() | ReservePLT);
-      AArch64LDBackend &backend = getTarget();
-      backend.defineIRelativeRange(*rsym);
-      return;
-    }
     // if symbol is defined in the output file and it's not
     // preemptible, no need plt
     if (!getTarget().isSymbolPreemptible(*rsym)) {
@@ -477,6 +473,8 @@ void AArch64Relocator::scanGlobalReloc(InputFile &pInput, Relocation &pReloc,
   case llvm::ELF::R_AARCH64_ADR_PREL_PG_HI21:
   case R_AARCH64_ADR_PREL_PG_HI21_NC: {
     std::lock_guard<std::mutex> relocGuard(m_RelocMutex);
+    if (handleScanForNonPreemptibleIFunc(rsym, Obj))
+      return;
     if (relocNeedsDynRel(pReloc)) {
       if (getTarget().symbolNeedsCopyReloc(pReloc, *rsym)) {
         // check if the option -z nocopyreloc is given
@@ -510,10 +508,13 @@ void AArch64Relocator::scanGlobalReloc(InputFile &pInput, Relocation &pReloc,
   case llvm::ELF::R_AARCH64_LD64_GOT_LO12_NC:
   case llvm::ELF::R_AARCH64_LD64_GOTPAGE_LO15: {
     std::lock_guard<std::mutex> relocGuard(m_RelocMutex);
+    if (handleScanForNonPreemptibleIFunc(rsym, Obj))
+      return;
     // Symbol needs GOT entry, reserve entry in .got
     // return if we already create GOT for this symbol
     if (rsym->reserved() & ReserveGOT)
       return;
+
     // if the symbol cannot be fully resolved at link time, then we need a
     // dynamic relocation
     CreateGOT(Obj, pReloc, !config().isCodeStatic(), m_Target,
@@ -686,7 +687,14 @@ Relocator::Result unsupport(Relocation &pReloc, AArch64Relocator &pParent) {
 Relocator::Result abs(Relocation &pReloc, AArch64Relocator &pParent) {
   ResolveInfo *rsym = pReloc.symInfo();
   Relocator::DWord A = pReloc.addend();
-  Relocator::DWord S = pParent.getSymValue(&pReloc);
+  Relocator::DWord S;
+  if (LLVM_UNLIKELY(rsym && rsym->isIFunc() &&
+                    pParent.config().isCodeStatic())) {
+    AArch64PLT *P = pParent.getTarget().findEntryInPLT(rsym);
+    S = P->getAddr(pParent.config().getDiagEngine());
+    // Create GOT slot for rsym with the fragment reference of P
+  } else
+    S = pParent.getSymValue(&pReloc);
   Relocator::Address targetVal = S + A;
 
   bool isAuthAbs = pReloc.type() == llvm::ELF::R_AARCH64_AUTH_ABS64;
@@ -949,13 +957,24 @@ Relocator::Result condbr(Relocation &pReloc, AArch64Relocator &pParent) {
 
 // R_AARCH64_ADR_GOT_PAGE: Page(G(GDAT(S+A))) - Page(P)
 Relocator::Result adr_got_page(Relocation &pReloc, AArch64Relocator &pParent) {
-  DiagnosticEngine *DiagEngine = pParent.config().getDiagEngine();
-  if (!(pReloc.symInfo()->reserved() & Relocator::ReserveGOT)) {
+  LinkerConfig &config = pParent.config();
+  DiagnosticEngine *DiagEngine = config.getDiagEngine();
+  ResolveInfo *symInfo = pReloc.symInfo();
+  bool isStaticIFunc = symInfo->isIFunc() && config.isCodeStatic();
+  if ((!isStaticIFunc && !(symInfo->reserved() & Relocator::ReserveGOT)) ||
+      (isStaticIFunc && !(symInfo->reserved() & Relocator::ReservePLT))) {
     return Relocator::BadReloc;
   }
-
-  Relocator::Address GOT_S =
-      pParent.getTarget().findEntryInGOT(pReloc.symInfo())->getAddr(DiagEngine);
+  Relocator::Address GOT_S;
+  if (LLVM_UNLIKELY(isStaticIFunc)) {
+    AArch64PLT *PLT = pParent.getTarget().findEntryInPLT(symInfo);
+    ASSERT(PLT, "Must not be null!");
+    GOT_S = PLT->getGOT()->getAddr(DiagEngine);
+  } else {
+    GOT_S = pParent.getTarget()
+                .findEntryInGOT(pReloc.symInfo())
+                ->getAddr(DiagEngine);
+  }
   Relocator::DWord A = pReloc.addend();
   Relocator::Address P = pReloc.place(pParent.module());
   Relocator::DWord X =
@@ -969,13 +988,27 @@ Relocator::Result adr_got_page(Relocation &pReloc, AArch64Relocator &pParent) {
 
 // R_AARCH64_LD64_GOT_LO12_NC: G(GDAT(S))
 Relocator::Result ld64_got_lo12(Relocation &pReloc, AArch64Relocator &pParent) {
-  if (!(pReloc.symInfo()->reserved() & Relocator::ReserveGOT)) {
+  ResolveInfo *symInfo = pReloc.symInfo();
+  LinkerConfig &config = pParent.config();
+  bool isStaticIFunc = symInfo->isIFunc() && config.isCodeStatic();
+  if ((!isStaticIFunc &&
+       !(pReloc.symInfo()->reserved() & Relocator::ReserveGOT)) ||
+      (isStaticIFunc && !(symInfo->reserved() & Relocator::ReservePLT))) {
     return Relocator::BadReloc;
   }
 
-  Relocator::Address GOT_S = pParent.getTarget()
-                                 .findEntryInGOT(pReloc.symInfo())
-                                 ->getAddr(pParent.config().getDiagEngine());
+  Relocator::Address GOT_S;
+
+  if (LLVM_UNLIKELY(isStaticIFunc)) {
+    AArch64PLT *PLT = pParent.getTarget().findEntryInPLT(symInfo);
+    ASSERT(PLT, "Must not be null!");
+    GOT_S = PLT->getGOT()->getAddr(pParent.config().getDiagEngine());
+  } else {
+    GOT_S = pParent.getTarget()
+                .findEntryInGOT(pReloc.symInfo())
+                ->getAddr(pParent.config().getDiagEngine());
+  }
+
   Relocator::DWord A = pReloc.addend();
   Relocator::DWord X = helper_get_page_offset(GOT_S + A);
 
@@ -1269,4 +1302,16 @@ Relocator::Result copyInstruction(Relocation &pReloc,
   std::memcpy((void *)&insn, data, AArch64InsnHelpers::InsnSize);
   pReloc.target() = insn;
   return Relocator::OK;
+}
+
+bool AArch64Relocator::handleScanForNonPreemptibleIFunc(ResolveInfo *symInfo,
+                                                        ELFObjectFile *Obj) {
+  if (!symInfo || !symInfo->isIFunc() || !config().isCodeStatic())
+    return false;
+  if (symInfo->reserved() & Relocator::ReservePLT)
+    return true;
+  m_Target.createPLT(Obj, symInfo, /*isIRelative=*/true);
+  m_Target.defineIRelativeRange(*symInfo);
+  symInfo->setReserved(symInfo->reserved() | ReservePLT);
+  return true;
 }
--- a/test/AArch64/linux/IFuncs/IFuncs.test
+++ b/test/AArch64/linux/IFuncs/IFuncs.test
@@ -1,0 +1,51 @@
+#---IFuncs.test--------------------- Exe------------------#
+#BEGIN_COMMENT
+# This test verifies the IFUNC support when building AArch64 static executables.
+#END_COMMENT
+#START_TEST
+RUN: %clang %clangopts --target=aarch64-linux-gnu -o %t1.1.o %p/Inputs/1.c -c
+RUN: %clang %clangopts --target=aarch64-linux-gnu -o %t1.main.1.o %p/Inputs/main.1.c -c
+RUN: %link %linkopts -march aarch64 -o %t1.a.1.out %t1.main.1.o %t1.1.o --image-base 0x1000
+RUN: (%readelf -S %t1.a.1.out; %objdump -d %t1.a.1.out) | %filecheck %s --check-prefix CHECK1
+RUN: %clang %clangopts --target=aarch64-linux-gnu -o %t1.main.2.o %p/Inputs/main.2.c -c
+RUN: %link %linkopts -march aarch64 -o %t1.a.2.out %t1.main.2.o %t1.1.o --image-base 0x1000
+RUN: (%readelf -S %t1.a.2.out; %objdump -d %t1.a.2.out) | %filecheck %s --check-prefix CHECK2
+RUN: %clang %clangopts --target=aarch64-linux-gnu -o %t1.1.no_pic.o %p/Inputs/1.c -c -fno-pic
+RUN: %clang %clangopts --target=aarch64-linux-gnu -o %t1.main.2.no_pic.o %p/Inputs/main.2.c -c -fno-pic
+RUN: %link %linkopts -march aarch64 -o %t1.a.2.no_pic.out %t1.main.2.no_pic.o %t1.1.no_pic.o --image-base 0x1000
+RUN: (%readelf -S %t1.a.2.no_pic.out; %objdump -d %t1.a.2.no_pic.out) | %filecheck %s --check-prefix CHECK2_NOPIC
+RUN: %clang %clangopts --target=aarch64-linux-gnu -o %t1.main.3.o %p/Inputs/main.3.c -c
+RUN: %link %linkopts -march aarch64 -o %t1.a.3.out %t1.main.3.o %t1.1.o --image-base 0x1000
+RUN: (%readelf -S %t1.a.3.out; %objdump -D %t1.a.3.out) | %filecheck %s --check-prefix CHECK3
+RUN: %clang %clangopts --target=aarch64-linux-gnu -o %t1.main.4.o %p/Inputs/main.4.c -c
+RUN: %link %linkopts -march aarch64 -o %t1.a.4.out %t1.main.4.o %t1.1.o --image-base 0x1000
+RUN: (%readelf -S %t1.a.4.out; %objdump -D %t1.a.4.out) | %filecheck %s --check-prefix CHECK4
+#END_TEST
+
+CHECK1: .plt PROGBITS [[#%x,PLT_START:]]
+CHECK1: <main>:
+CHECK1: bl 0x[[#PLT_START+0x20]]
+
+CHECK2: .got.plt PROGBITS {{0+}}[[#%x,GOT_PLT_START:]]
+CHECK2: <main>:
+CHECK2: adrp x8, 0x[[#GOT_PLT_START]]
+CHECK2: ldr	x8, [x8, #0x18]
+
+CHECK2_NOPIC: .plt PROGBITS {{0+}}1020
+CHECK2_NOPIC: <main>:
+CHECK2_NOPIC: adrp x8, 0x1000
+CHECK2_NOPIC: add x8, x8, #0x40
+
+CHECK3: .plt PROGBITS [[#%x,PLT_START:]]
+CHECK3: Disassembly of section .data:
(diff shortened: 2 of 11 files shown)


quic-seaswara:
Thanks for identifying the issue and push a PR.

lib/Target/AArch64/AArch64Relocator.cpp
```diff
@@ -514,6 +514,31 @@ void AArch64Relocator::scanGlobalReloc(InputFile &pInput, Relocation &pReloc,
...
+        m_Target.createPLT(Obj, rsym, /*isIRelative=*/true);
+        rsym->setReserved(rsym->reserved() | ReservePLT);
+        m_Target.defineIRelativeRange(*rsym);
+      }
+      // Create a GOT entry whose value is the PLT entry address.
+      AArch64GOT *G = m_Target.createGOT(GOT::Regular, Obj, rsym);
+      AArch64PLT *P = m_Target.findEntryInPLT(rsym);
+      FragmentRef *PLTRef = make<FragmentRef>(*P, 0);
+      Relocation *r = Relocation::Create(llvm::ELF::R_AARCH64_ABS64, 64,
+                                         make<FragmentRef>(*G, 0), 0);
```
efriedma-quic: This is a weird way to set a GOT.  Usually the value of a GOT entry is computed by AArch64GOT::getContent().  I don't know if using a relocation like this actually causes any issues, but it's inconsistent with all the other code related to the values of GOT entries.

lib/Target/AArch64/AArch64Relocator.cpp
```diff
@@ -514,6 +514,31 @@ void AArch64Relocator::scanGlobalReloc(InputFile &pInput, Relocation &pReloc,
     // return if we already create GOT for this symbol
     if (rsym->reserved() & ReserveGOT)
       return;
+
+    // For IFUNC symbols in static executables, the GOT entry must contain
+    // the PLT entry address rather than the IFUNC resolver address. The PLT
+    // entry correctly indirects through the IRELATIVE-resolved GOT.PLT slot,
+    // so callers that load the function address from the GOT and then call
+    // through it will reach the actual implementation.
```
efriedma-quic: Looking at the output of other linkers, it looks like they match unpatched eld: the relocations end up pointing to the resolver.  I suspect this doesn't cause a problem, normally, because these relocations are supposed to be resolved at startup.

Maybe eld isn't exposing some necessary predefined symbol?

lib/Target/AArch64/AArch64Relocator.cpp
quic-seaswara: We already do this to fix the GOTPLT0/GOTPLT1/.../GOTPLTN entries to the appropriate PLT entries. This allows regular relocation code to patch relocations from GOTPLTN to PLTN entries. This is ok IMO

androm3da: Fixed

androm3da: oh oops should I un-fix it?

lib/Target/AArch64/AArch64Relocator.cpp
```diff
@@ -515,26 +515,23 @@ void AArch64Relocator::scanGlobalReloc(InputFile &pInput, Relocation &pReloc,
...
-      Obj->getGOT()->addRelocation(r);
-      r->modifyRelocationFragmentRef(PLTRef);
+      G->setValueType(GOT::SymbolValue);
+      // Place the IRELATIVE in .rela.plt so the CRT can find it via
+      // __rela_iplt_start / __rela_iplt_end.
+      Relocation &rela_entry = *Obj->getRelaPLT()->createOneReloc();
+      rela_entry.setType(llvm::ELF::R_AARCH64_IRELATIVE);
+      rela_entry.setTargetRef(make<FragmentRef>(*G, 0));
+      rela_entry.setSymInfo(rsym);
+      m_Target.defineIRelativeRange(*rsym);
```
quic-seaswara: GNU linker create a PLT, so should we.

test/AArch64/standalone/IFuncGotAddr/IFuncGotAddr.s
```diff
@@ -20,30 +18,27 @@
...
 # RUN: %readelf -S -W out | %filecheck --check-prefix=SECTIONS %s
 
-## The GOT entry (.got) for myfunc must contain the PLT entry address, not the
-## resolver address.  With --section-start .text=0x1000, the IFUNC resolver
-## lives at 0x1008.  If the linker incorrectly places the resolver address in
-## the GOT, the hex dump would contain 0x1008 in little-endian (08100000
-## 00000000).  With the fix, the GOT instead holds the PLT entry address which
-## is below .text.
+## The regular GOT entry contains the resolver address at link time (0x1008 in
+## LE).  This is correct: the IRELATIVE relocation will fix it up at startup.
```
quic-seaswara: With GNU ld, the .got.plt entry contains the value of the PLT reserved for that symbol.

As Eli points out, this is already happening, that .got.plt 

**eld**
```
$clang-aarch64-x -c ifunc.c && ld.eld ifunc.o && readelf -x .got.plt a.out && readelf -s a.out | grep resolve; llvm-readelf -r a.out | grep RELATIVE

Hex dump of section '.got.plt':
 NOTE: This section has relocations against it, but these have NOT been applied to this dump.
  0x00001000 00000000 00000000 00000000 00000000 ................
  0x00001010 00000000 00000000 ac010000 00000000 ................

    10: 00000000000001ac    12 FUNC    LOCAL  DEFAULT    3 resolve_foo

0000000000001018  0000000000000408 R_AARCH64_IRELATIVE               1ac

```

**bfd** -- got entry does not point to the resolver, and points to the PLT 

```
 /opt/aarch64-gnu-tools/bin/aarch64-unknown-elf-ld.bfd ifunc.o && readelf -x .got.plt a.out && readelf -s a.out | grep resolve
/opt/aarch64-gnu-tools/bin/aarch64-unknown-elf-ld.bfd: warning: cannot find entry symbol _start; defaulting to 0000000000400028

Hex dump of section '.got.plt':
 NOTE: This section has relocations against it, but these have NOT been applied to this dump.
  0x004100d8 18004000 00000000                   ..@.....

     9: 0000000000400064    12 FUNC    LOCAL  DEFAULT    3 resolve_foo
```

lib/Target/AArch64/AArch64Relocator.cpp
quic-seaswara: The linker correctly emits the range of __rela_iplt_start and __rela_iplt_end with IRELATIVE relocs.

lib/Target/AArch64/AArch64Relocator.cpp
efriedma-quic: I'll let Shankar make the call on this.

quic-seaswara: We need to create a PLT and point the .got.plt to the PLT entry to emulate GNU ld IMO.

partaror: >We need to create a PLT and point the .got.plt to the PLT entry to emulate GNU ld IMO.

I have made this change.

lib/Target/AArch64/AArch64Relocator.cpp
partaror: We now create a PLT entry for all ifunc symbols when static linking.

lib/Target/AArch64/AArch64Relocator.cpp
```diff
@@ -686,7 +687,13 @@ Relocator::Result unsupport(Relocation &pReloc, AArch64Relocator &pParent) {
 Relocator::Result abs(Relocation &pReloc, AArch64Relocator &pParent) {
   ResolveInfo *rsym = pReloc.symInfo();
   Relocator::DWord A = pReloc.addend();
-  Relocator::DWord S = pParent.getSymValue(&pReloc);
+  Relocator::DWord S;
+  if (LLVM_UNLIKELY(rsym && rsym->isIFunc() &&
+                    pParent.config().isCodeStatic())) {
+    AArch64PLT *P = pParent.getTarget().findEntryInPLT(rsym);
+    S = P->getAddr(pParent.config().getDiagEngine());
```
quic-seaswara, line 694: can this be simplified, if the sym has a PLT entry then pick the address from the PLT entry ?

lib/Target/AArch64/AArch64Relocator.cpp
```diff
@@ -949,13 +956,22 @@ Relocator::Result condbr(Relocation &pReloc, AArch64Relocator &pParent) {
...
 Relocator::Result adr_got_page(Relocation &pReloc, AArch64Relocator &pParent) {
-  DiagnosticEngine *DiagEngine = pParent.config().getDiagEngine();
-  if (!(pReloc.symInfo()->reserved() & Relocator::ReserveGOT)) {
+  LinkerConfig &config = pParent.config();
+  DiagnosticEngine *DiagEngine = config.getDiagEngine();
+  ResolveInfo *symInfo = pReloc.symInfo();
+  bool isStaticIFunc = symInfo->isIFunc() && config.isCodeStatic();
+  if ((!isStaticIFunc && !(symInfo->reserved() & Relocator::ReserveGOT)) ||
+      (isStaticIFunc && !(symInfo->reserved() & Relocator::ReservePLT))) {
     return Relocator::BadReloc;
```
quic-seaswara, line 966: should this be if (symInfo->isFunc() & !Relocation::ReservePLT) return BadReloc ?

lib/Target/AArch64/AArch64Relocator.cpp
```diff
@@ -969,13 +985,27 @@ Relocator::Result adr_got_page(Relocation &pReloc, AArch64Relocator &pParent) {
...
 Relocator::Result ld64_got_lo12(Relocation &pReloc, AArch64Relocator &pParent) {
-  if (!(pReloc.symInfo()->reserved() & Relocator::ReserveGOT)) {
+  ResolveInfo *symInfo = pReloc.symInfo();
+  LinkerConfig &config = pParent.config();
+  bool isStaticIFunc = symInfo->isIFunc() && config.isCodeStatic();
+  if ((!isStaticIFunc &&
+       !(pReloc.symInfo()->reserved() & Relocator::ReserveGOT)) ||
+      (isStaticIFunc && !(symInfo->reserved() & Relocator::ReservePLT))) {
     return Relocator::BadReloc;
   }
```
quic-seaswara, line 998: Same as previous comment

lib/Target/AArch64/AArch64Relocator.cpp
```diff
@@ -969,13 +985,27 @@ Relocator::Result adr_got_page(Relocation &pReloc, AArch64Relocator &pParent) {
...
-  Relocator::Address GOT_S = pParent.getTarget()
-                                 .findEntryInGOT(pReloc.symInfo())
-                                 ->getAddr(pParent.config().getDiagEngine());
+  Relocator::Address GOT_S;
+
+  if (LLVM_UNLIKELY(isStaticIFunc)) {
+    AArch64PLT *PLT = pParent.getTarget().findEntryInPLT(symInfo);
+    ASSERT(PLT, "Must not be null!");
+    GOT_S = PLT->getGOT()->getAddr(pParent.config().getDiagEngine());
+  } else {
```
quic-seaswara, line 1006: follow up : it might be better to organize the code in such a way to get the address of a symbol, and the wrapper function can return the address of the PLT or GOT as per the ABI.

lib/Target/AArch64/AArch64Relocator.cpp
```diff
@@ -1270,3 +1300,14 @@ Relocator::Result copyInstruction(Relocation &pReloc,
   pReloc.target() = insn;
   return Relocator::OK;
 }
+
+bool AArch64Relocator::handleScanForNonPreemptibleIFunc(ResolveInfo *symInfo, ELFObjectFile *Obj) {
+  if (!symInfo->isIFunc() || !config().isCodeStatic())
+    return false;
+  if (symInfo->reserved() & Relocator::ReservePLT)
+    return true;
+  m_Target.createPLT(Obj, symInfo, /*isIRelative=*/true);
```
quic-seaswara, line 1313: By any chance do we get a PLT and a GOT ? If so what is preferred ?

lib/Target/AArch64/AArch64Relocator.cpp
parth-07: > if the sym has a PLT entry then pick the address from the PLT entry

I am not certain that this will be valid for all the relocations. I am okay to make this change but I worry that it may break something.

lib/Target/AArch64/AArch64Relocator.cpp
parth-07: By ` !Relocation::ReservePLT` do you mean `!(symInfo->reserved() & Relocator::ReservePLT)`? If so, then I think this is exactly what the condition is checking.

lib/Target/AArch64/AArch64Relocator.cpp
parth-07: I agree. There are also a lot more refactoring we need to do as a follow-up. Most of the code that relates with GNU IFunc can be moved to the target-independent part of the linker.

lib/Target/AArch64/AArch64Relocator.cpp
parth-07: I am not sure if I completely understood this question. `createPLT` creates a PLT entry and a .got.plt entry. We do not create a `.got` slot as of now for IRELATIVE symbols.

lib/Target/AArch64/AArch64Relocator.cpp
```diff
@@ -344,13 +345,14 @@ void AArch64Relocator::scanGlobalReloc(InputFile &pInput, Relocation &pReloc,
...
   // rsym - The relocation target symbol
   ResolveInfo *rsym = pReloc.symInfo();
-
   switch (pReloc.type()) {
   case llvm::ELF::R_AARCH64_AUTH_ABS64:
   case llvm::ELF::R_AARCH64_ABS16:
   case llvm::ELF::R_AARCH64_ABS32:
   case llvm::ELF::R_AARCH64_ABS64: {
     std::lock_guard<std::mutex> relocGuard(m_RelocMutex);
+    if (handleScanForNonPreemptibleIFunc(rsym, Obj))
```
quic-akaryaki, line 354: Where the list of relocations come from? What happens if IFUNC symbol is used by relocations of a different type? Can we handle IFUNC in a common path for all relocations?

lib/Target/AArch64/AArch64Relocator.cpp
```diff
@@ -949,13 +957,24 @@ Relocator::Result condbr(Relocation &pReloc, AArch64Relocator &pParent) {
 
 // R_AARCH64_ADR_GOT_PAGE: Page(G(GDAT(S+A))) - Page(P)
 Relocator::Result adr_got_page(Relocation &pReloc, AArch64Relocator &pParent) {
-  DiagnosticEngine *DiagEngine = pParent.config().getDiagEngine();
-  if (!(pReloc.symInfo()->reserved() & Relocator::ReserveGOT)) {
+  LinkerConfig &config = pParent.config();
+  DiagnosticEngine *DiagEngine = config.getDiagEngine();
+  ResolveInfo *symInfo = pReloc.symInfo();
+  bool isStaticIFunc = symInfo->isIFunc() && config.isCodeStatic();
+  if ((!isStaticIFunc && !(symInfo->reserved() & Relocator::ReserveGOT)) ||
```
quic-akaryaki, line 964: This check can be merged with the if below.

quic-akaryaki, line 964: I think this code at least needs an explanation. I believe a PLT cannot exist without a GOT.

lib/Target/AArch64/AArch64Relocator.cpp
```diff
@@ -949,13 +957,24 @@ Relocator::Result condbr(Relocation &pReloc, AArch64Relocator &pParent) {
...
+      (isStaticIFunc && !(symInfo->reserved() & Relocator::ReservePLT))) {
     return Relocator::BadReloc;
   }
-
-  Relocator::Address GOT_S =
-      pParent.getTarget().findEntryInGOT(pReloc.symInfo())->getAddr(DiagEngine);
+  Relocator::Address GOT_S;
+  if (LLVM_UNLIKELY(isStaticIFunc)) {
```
quic-akaryaki, line 969: Similar, we have to have a GOT entry but cannot use one code path? Will findEntryInGOT() not work for some reason?

lib/Target/AArch64/AArch64Relocator.cpp
efriedma-quic, line 1021: Does R_AARCH64_LD64_GOTPAGE_LO15 also need to be fixed?  (I don't think clang will generate it currently, but gcc can...)

test/AArch64/linux/IFuncs/IFuncs.test
```diff
@@ -0,0 +1,51 @@
...
+CHECK3: 2000: {{.*}} .word 0x{{0+}}[[#%x,PLT_START+0x20]]
+
+CHECK4: .plt PROGBITS [[#%x,PLT_START:]]
+CHECK4: .got.plt PROGBITS {{0+}}2008
+CHECK4: <main>:
+CHECK4: adrp x8, 0x2000
+CHECK4: ldr	x8, [x8, #0x20]
+CHECK4: Disassembly of section .data:
+CHECK4: {{0+}}2000 <foogp>:
+CHECK4: 2000: {{.*}} .word 0x{{0+}}[[#%x,PLT_START+0x20]]
```
efriedma-quic, line 51: For the case where there's both an absolute reference and a GOT reference, I think we need two GOT entries: one pointing to the PLT, one pointing to the actual function, so that pointers to the function consistently return the same address.

I'm okay with a FIXME for now, though; pointer equality is unlikely to matter in most cases.

lib/Target/AArch64/AArch64Relocator.cpp
```diff
@@ -969,13 +988,27 @@ Relocator::Result adr_got_page(Relocation &pReloc, AArch64Relocator &pParent) {
...
+      (isStaticIFunc && !(symInfo->reserved() & Relocator::ReservePLT))) {
     return Relocator::BadReloc;
   }
 
-  Relocator::Address GOT_S = pParent.getTarget()
-                                 .findEntryInGOT(pReloc.symInfo())
-                                 ->getAddr(pParent.config().getDiagEngine());
+  Relocator::Address GOT_S;
+
+  if (LLVM_UNLIKELY(isStaticIFunc)) {
```
efriedma-quic, line 1002: LLVM_UNLIKELY should be used very sparingly; it makes code harder to read, and usually doesn't really help.

lib/Target/AArch64/AArch64Relocator.cpp
quic-seaswara: Yes, I think its a valid concern. Let us treat handling IFUNC in a common path for all relocations as a follow up as I suspect it might be a bigger change

lib/Target/AArch64/AArch64Relocator.cpp
quic-seaswara: It appears that regular functions accessed via GOT, is treated to a have a GOT slot vs functions that use IFUNC type use a PLT.

lib/Target/AArch64/AArch64Relocator.cpp
quic-seaswara: Can we handle this as a follow up ?

test/AArch64/linux/IFuncs/IFuncs.test
quic-seaswara: The absolute reference and the GOT reference IMO should point to the same GOT slot.

lib/Target/AArch64/AArch64Relocator.cpp
quic-seaswara: Currently we reserve a PLT and point the symbol to the PLT. Your suggestion is valid and can simplify things, if we recorded the GOTPLT. 

We need to look into this on how we can address.

lib/Target/AArch64/AArch64Relocator.cpp
quic-seaswara: I dont think as Static IFUNC will always have a PLT whereas non IFUNC will use the GOT for handling this relocation. 

May be we can add comments.

quic-seaswara:
Overall I think it looks good with minor tweaks as per review comments. 

We will have this merged and work on follow up comments and make incremental progress. 

We need to start working and cleaning up this code for handling shared libraries, as I feel that is going to get very tricky, but definitely a follow up.

test/AArch64/linux/IFuncs/IFuncs.test
quic-seaswara: we planned to do this in a follow up patch

lib/Target/AArch64/AArch64Relocator.cpp
efriedma-quic: That should be fine, I think?  I don't think it impacts any of the configs we're currently using.

test/AArch64/linux/IFuncs/IFuncs.test
efriedma-quic: To be more clear, consider something like the following:

```
int foo();
int (*foogp)(void) = foo;
int main() {
  int (*foop)(void) = foo;
  return foop == foogp;
}
```

This program should return 1.  Which means the GOT load for computing `foop` should return the address of the PLT... which means it needs to be separate from the GOT entry used by the PLT?  At least, I think that's what lld does.

apazos (approved):
CPULLVM tests are passing.

lib/Target/AArch64/AArch64Relocator.cpp
partaror: > Can we handle IFUNC in a common path for all relocations?

Yes, ideally, if any relocation reference an `ifunc` symbol then we need to create a PLT entry (with a corresponding `.got.plt` entry). However, for some relocations, the behavior needs to be slightly different. For example, if an absolute relocation refers to an `ifunc` symbol, then we also need to also emit a `.got` entry for the symbol and fill it with the address of the `.plt` entry. Currently, I was not sure about all such cases, and thus I decided to refactor it as a follow-up PR. Another important refactoring here would be to move most of the `ifunc` handling logic to target independent parts, so that we can easily enable `ifunc` support for any target.

> Let us treat handling IFUNC in a common path for all relocations as a follow up as I suspect it might be a bigger change

Yes, I will create a GitHub issue to track this.

lib/Target/AArch64/AArch64Relocator.cpp
partaror: > This check can be merged with the if below.

Can you please specify which check you are referring here? Did you mean the `if(LLVM_UNLIKELY(isStaticIFunc))` check? If so, I agree yes it can be part of this if-check but I do not see any benefit in doing so.

partaror: > . I believe a PLT cannot exist without a GOT.

Where does the code suggest that a PLT entry can exist without a corresponding GOT entry? Please note that the condition `symInfo->reserved() & Relocator::ReserveGOT` is false when a symbol has a `.got.plt` entry instead of `.got` entry.

lib/Target/AArch64/AArch64Relocator.cpp
partaror: Adding to the Shankar's comments, in the current framework, we do not store `.got.plt` entries in the map that is used with `findEntryInGOT`.

lib/Target/AArch64/AArch64Relocator.cpp
partaror: I wanted to use `LLVM_UNLIKELY` wherever we have `isStaticIFunc` in an if-branch, because it is indeed an unlikely case in the real world programs. If it is beneficial to use `LLVM_UNLIKELY` at one of these branches, then wouldn't it be beneficial when used with all such branches that can be assumed to be called the similar number of times? 

I agree that an added macro makes code harder to read, but I was not sure about the performance-readability trade-off here. What do you suggest here? Should we never use `LLVM_UNLIKELY`?

lib/Target/AArch64/AArch64Relocator.cpp
partaror: Thank you for mentioning it. I will create a ticket to address it as a follow-up.

test/AArch64/linux/IFuncs/IFuncs.test
partaror: Hi @efriedma-quic, yes, you are right. I planned to add support for this in a follow-up as it will require much more changes.  

>  Which means the GOT load for computing foop should return the address of the PLT... which means it needs to be separate from the GOT entry used by the PLT? At least, I think that's what lld does.

Yes, both lld and bfd does this. Apart from this, lld also modifies the original symbol to the `plt` slot. It is interesting because bfd does not do this. For example, if there was an `ifunc` symbol `foo` that is referenced by an absolute relocation, and the `ifunc` symbol foo has address 0x2004 and its corresponding PLT slot has the address `0x5008`, then lld will change the `ifunc` symbol foo value to `0x5008` and the type `FUNCTION`.

I have created an issue to track this: #913

lib/Target/AArch64/AArch64Relocator.cpp
partaror: I have created #916 to track this.

lib/Target/AArch64/AArch64Relocator.cpp
partaror: I have created #915 to track this.

PR #920, opened 2026-03-06
[NFC] Fix undefined behavior with misaligned write

namikukr-qc:
in emitGroup, there's an integer-sized write that triggers a sanitizer error (writing to an int-sized slot that isn't 4-byte aligned). This patch removes that and uses memcpy instead, to avoid the strict aliasing violation.

The pointer is shifted by Index * sizeof(llvm::ELF::Elf32_Word), which should make sure nothing changes functionally.

### DIFF
--- a/lib/Writers/ELFObjectWriter.cpp
+++ b/lib/Writers/ELFObjectWriter.cpp
@@ -803,15 +803,15 @@ void ELFObjectWriter::emitGroup(ELFSection *S, MemoryRegion &CurRegion) {
         RegionFrag.getRegion().size() / sizeof(llvm::ELF::Elf32_Word);
     // copy the contents
     memcpy(CurRegion.begin() + CurOffset, From, Frag->size());
-    void *GroupData = (void *)(CurRegion.begin() + CurOffset);
+    uint8_t *GroupData = (uint8_t *)(CurRegion.begin() + CurOffset);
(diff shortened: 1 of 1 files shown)


lib/Writers/ELFObjectWriter.cpp
```diff
@@ -810,7 +810,8 @@ void ELFObjectWriter::emitGroup(ELFSection *S, MemoryRegion &CurRegion) {
       if (Si == Se)
         break;
       uint32_t SectionIdx = (*Si)->getOutputELFSection()->getIndex();
-      ((uint32_t *)GroupData)[Index] = SectionIdx;
+      std::memcpy((uint8_t *) GroupData + (Index * sizeof(uint32_t)),
+          &SectionIdx, sizeof(uint32_t));
```
lenary: I think a cleaner fix would be:
- declare `GroupData` as `uint8_t *` so you can avoid the cast
- use `sizeof(llvm::ELF::Elf32_Word)` instead of `sizeof(uint32_t)`

namikukr-qc: Added the changes you suggested, apologies for the late reply.

quic-seaswara (approved):

lenary (approved):
LGTM

PR #921, opened 2026-03-06
Make ARM tests independent of the default target triple

parth-07:
This commit makes a few ARM tests independent of the default target triple. The filecheck CHECKs values may change depending upon the default target triple. For example, if arm-unknown-linux is the default triple, then static links use 0x08048000 as the image-base and loads FileHdr and Phdrs by default. On the other hand, if 'arm-unknown-elf' is the default triple, then static links use 0x0 as the image base and does not load FileHdr and Phdrs.

### DIFF
--- a/test/ARM/standalone/etext/etext.test
+++ b/test/ARM/standalone/etext/etext.test
@@ -1,6 +1,6 @@
 # Checks that the value of etext has been set appropriately.
 RUN: %clang %clangopts -target arm -c %p/Inputs/1.c -o %t1.o
-RUN: %link %linkopts -march arm %t1.o -o %t2.out -z max-page-size=0x1000
+RUN: %link %linkopts -mtriple armelf %t1.o -o %t2.out -z max-page-size=0x1000
 RUN: %readelf -s %t2.out | %grep etext | %filecheck %s
 
 #CHECK:    {{[0-9]+}}: 0000001c     0 NOTYPE  GLOBAL DEFAULT  ABS __etext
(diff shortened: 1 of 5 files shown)


test/ARM/standalone/etext/etext.test
```diff
@@ -1,6 +1,6 @@
 # Checks that the value of etext has been set appropriately.
 RUN: %clang %clangopts -target arm -c %p/Inputs/1.c -o %t1.o
-RUN: %link %linkopts -march arm %t1.o -o %t2.out -z max-page-size=0x1000
+RUN: %link %linkopts -mtriple armelf %t1.o -o %t2.out -z max-page-size=0x1000
```
quic-seaswara, line 3: Can we remove -mtriple ?

partaror: Sadly no, only `-mtriple` option affects the OS property of the target triple. If we use `-m armelf`, then the OS property of the target triple remains the same as of the default target triple.

quic-seaswara: Can we make the test `os` ignorant ?

partaror: We cannot change the link commands for them to be OS-independent because there is no command-line option for disabling default loading of FILEHDR / PHDRS, and changing the CHECKs to work for both the cases was turning out to be difficult.

quic-seaswara (changes requested):

test/ARM/standalone/etext/etext.test
apazos: @jonathonpenix please force launch CPULLVM builder precheckins with this change

jonathonpenix: https://github.com/qualcomm/cpullvm-toolchain/actions/runs/22969308806 should have it

PR #922, opened 2026-03-06
Fix stack-buffer-overflow sanitizer issue in AUIPC_JALR_TO_XQCI.test

parth-07:
This commit fixes stack-buffer-overflow sanitizer issue in AUIPIC_JALR_TO_XQCI.test. The root cause was that the QC_E_J relaxation patch 6 bytes instruction but the 'Instr' parameter of 'RegionFragmentEx::replaceInstruction' that stores the updated instruction is only of 4 bytes.

### DIFF
--- a/include/eld/Fragment/RegionFragmentEx.h
+++ b/include/eld/Fragment/RegionFragmentEx.h
@@ -36,7 +36,7 @@ public:
 
   static bool classof(const RegionFragmentEx *) { return true; }
 
-  bool replaceInstruction(uint32_t Offset, Relocation *Reloc, uint32_t Instr,
+  bool replaceInstruction(uint32_t Offset, Relocation *Reloc, uint8_t *Instr,
                           uint8_t Size);
   void deleteInstruction(uint32_t Offset, uint32_t Size);
   void addRequiredNops(uint32_t Offset, uint32_t NumNopsToAdd);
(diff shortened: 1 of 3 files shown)


include/eld/Fragment/RegionFragmentEx.h
```diff
@@ -36,7 +36,7 @@ class RegionFragmentEx : public Fragment {
 
   static bool classof(const RegionFragmentEx *) { return true; }
 
-  bool replaceInstruction(uint32_t Offset, Relocation *Reloc, uint32_t Instr,
+  bool replaceInstruction(uint32_t Offset, Relocation *Reloc, uint64_t Instr,
```
quic-seaswara: templatized function might be better

lenary: Given this is just doing a memcpy, maybe we can just expose `Instr` as `const char *` to avoid templating or issues with sizing, and pass `&Instr` everywhere we're calling it?

parth-07: @quic-seaswara I like @lenary's suggestion. Please let me know your thoughts on it.

quic-seaswara: sounds good!, const uint8_t *

lenary: Technically a `const uint8_t *` isn't the same as a `const char *` in terms of strict aliasing, BUT you use the former so much in the codebase I'm not going to complain if you keep using it in this interface too.

quic-seaswara: I agree, we should be consistent, for now const char * is ok, but we should clean it up when dealing with ELF data as const uint8_t * as much as possible

PR #923, opened 2026-03-06
Reject invalid absolute relocations in PIE/shared links

### LINKED ISSUE #832: Add support for -z text
-z text support is missing or is broken.

Test 

```
cat > 1.s << \!
.section .text.foo
.global foo
foo:
.word 100
.text
.word foo
!

clang -c 1.s
ld.eld 1.o -z text -shared
ld.lld 1.o -z text -shared
```

lld fails correctly with the error below 

`ld.lld: error: relocation R_ARM_ABS32 cannot be used against symbol 'foo'; recompile with -fPIC`

quic-areg:
Reject invalid absolute relocations in PIE/shared links

ELD currently permits some absolute relocations that should be rejected when producing PIE/shared objects.

ELD would also allow truncated absolute relocations in 64-bit outputs. Now it correctly reports an error.

Symbols fully resolved at link time (eg defsym) should not require a dynamic relocation.

With -z text, reject absolute relocations in PIE/shared links.

With -z notext, split behavior by target bitness:

- 32-bit targets may emit TEXTREL.
- 64-bit targets error on truncated absolute dynamic relocations.

Closes #832.

### DIFF
--- a/lib/Target/AArch64/AArch64Relocator.h
+++ b/lib/Target/AArch64/AArch64Relocator.h
@@ -75,7 +75,7 @@ public:
                              const ELFSection &pSection) override;
 
 private:
-  bool isInvalidReloc(Relocation &pReloc) const;
+  bool isPICRelocTypeSupported(const Relocation &reloc) const override;
   bool isRelocSupported(Relocation &pReloc) const;
   bool relocNeedsDynRel(Relocation &pReloc) const;
 
--- a/lib/Target/GNULDBackend.cpp
+++ b/lib/Target/GNULDBackend.cpp
@@ -4085,17 +4085,32 @@ const LDSymbol *GNULDBackend::getEntrySymbol() const {
   return entrySymbol;
 }
 
+bool GNULDBackend::isTargetReadOnly(const ELFSection &input) const {
+  if (!input.isAlloc() || input.isWritable())
+    return false;
+
+  assert(input.getOutputELFSection());
+  if (input.getOutputELFSection()->isWritable())
+    return false;
+
+  assert(input.getOutputSection());
+  for (const RuleContainer *rule : *input.getOutputSection()) {
+    if (rule->getSection()->isWritable())
+      return false;
+    for (const ELFSection *matched : rule->getMatchedInputSections())
+      if (matched->isWritable())
+        return false;
+  }
+  return true;
+}
+
 void GNULDBackend::checkAndSetHasTextRel(const ELFSection &pSection) {
   if (m_bHasTextRel)
     return;
 
   // if the target section of the dynamic relocation is ALLOCATE but is not
   // writable, than we should set DF_TEXTREL
-  const uint32_t flag = pSection.getFlags();
-  if (0 == (flag & llvm::ELF::SHF_WRITE) && (flag & llvm::ELF::SHF_ALLOC))
-    m_bHasTextRel = true;
-
-  return;
+  m_bHasTextRel = isTargetReadOnly(pSection);
 }
 
 /// sortRelocation - sort the dynamic relocations to let dynamic linker
--- a/lib/Target/Relocator.cpp
+++ b/lib/Target/Relocator.cpp
@@ -284,6 +284,43 @@ uint32_t Relocator::getRelocType(std::string RelocName) {
   return RelocNameMap[RelocName];
 }
 
+bool Relocator::reportNonPICRelocation(const Relocation &reloc) const {
+  const auto *sym = reloc.symInfo();
+  config().raise(Diag::non_pic_relocation)
+      << getName(reloc.type()) << sym->name()
+      << reloc.getSourcePath(config().options());
+  return false;
+}
+
+bool Relocator::checkPICRelocSupported(const Relocation &reloc) const {
+  if (!config().isCodeIndep())
+    return true;
+  if (isPICRelocTypeSupported(reloc))
+    return true;
+  return reportNonPICRelocation(reloc);
+}
+
+bool Relocator::checkDynamicRelocAllowed(const Relocation &reloc,
+                                         const ELFSection &section,
+                                         bool isAbs) const {
+  if (!config().isCodeIndep())
+    return true;
+
+  // -z text (default) rejects dynamic relocations in read-only output sections.
+  if (section.isAlloc() && getTarget().isTargetReadOnly(section) &&
+      !config().options().textRelocsAllowed())
+    return reportNonPICRelocation(reloc);
+
+  // Truncated absolute relocations cannot be safely represented as dynamic
+  // relocations for 64-bit outputs.
+  if (isAbs && config().targets().is64Bits() && getSize(reloc.type()) < 64)
+    return reportNonPICRelocation(reloc);
+
+  if (!isDynamicRelocSupported(reloc))
+    return reportNonPICRelocation(reloc);
+  return true;
+}
+
 bool Relocator::doDeMangle() const {
   return m_Config.options().shouldDemangle();
 }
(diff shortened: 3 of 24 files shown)


lib/Target/X86/x86_64Relocator.cpp
```diff
@@ -262,6 +265,15 @@ void x86_64Relocator::scanLocalReloc(InputFile &pInputFile, Relocation &pReloc,
     }
     return;
   }
+  case llvm::ELF::R_X86_64_32:
+  case llvm::ELF::R_X86_64_32S:
+  case llvm::ELF::R_X86_64_16:
+  case llvm::ELF::R_X86_64_8:
+    if (config().isCodeIndep()) {
```
quic-seaswara: Remove code().isCodeInDep() as its tested in checkDynamicRelocAllowed

lib/Target/GNULDBackend.cpp
```diff
@@ -3883,11 +3883,11 @@ bool GNULDBackend::symbolNeedsDynRel(const ResolveInfo &pSym, bool pSymHasPLT,
...
     return false;
 
-  // An absolute symbol can be resolved directly if it is either local
-  // or we are linking statically. Otherwise it can still be overridden
-  // at runtime.
+  // An absolute symbol can be resolved directly if it is either local,
+  // non-preemptible, or we are linking statically.
   if (pSym.isAbsolute() &&
-      (pSym.binding() == ResolveInfo::Local || config().isCodeStatic()))
+      (pSym.binding() == ResolveInfo::Local || !isSymbolPreemptible(pSym) ||
```
quic-seaswara, line 4026: This fix needs to be part of a different PR.

Please raise a issue and fix accordingly.

lib/Target/Relocator.cpp
```diff
@@ -286,6 +286,46 @@ uint32_t Relocator::getRelocType(std::string RelocName) {
...
+
+  // -z text (default) rejects dynamic relocations in read-only sections.
+  if (section.isAlloc() && isReadOnly &&
+      !config().options().textRelocsAllowed())
+    return reportNonPICRelocation(reloc);
+
+  // Truncated absolute relocations cannot be safely represented as dynamic
+  // relocations for 64-bit outputs.
+  if (isAbs && config().targets().is64Bits() && getSize(reloc.type()) < 64)
+    return reportNonPICRelocation(reloc);
```
quic-seaswara, line 317: I could not follow this. This is not a dynamic relocation if the size of the reloc < 64. 

Are we creating a dynamic relocation for this ?

lib/Target/Relocator.cpp
```diff
@@ -286,6 +286,46 @@ uint32_t Relocator::getRelocType(std::string RelocName) {
...
+
+bool Relocator::checkPICRelocSupported(const Relocation &reloc) {
+  if (!config().isCodeIndep())
+    return true;
+  if (isPICRelocTypeSupported(reloc))
+    return true;
+  return reportNonPICRelocation(reloc);
+}
+
+bool Relocator::checkDynamicRelocAllowed(const Relocation &reloc,
```
quic-seaswara, line 303: Instead of introducing a new function, can we change the function  helper_DynRel_init to error out if we are ending up creating a dynamic relocation from read only sections ?

lib/Target/Relocator.cpp
```diff
@@ -286,6 +286,46 @@ uint32_t Relocator::getRelocType(std::string RelocName) {
...
+                                         const ELFSection &section,
+                                         bool isAbs) const {
+
+  if (!config().isCodeIndep())
+    return true;
+
+  const bool isReadOnly = !(section.getFlags() & llvm::ELF::SHF_WRITE);
+
+  // -z text (default) rejects dynamic relocations in read-only sections.
+  if (section.isAlloc() && isReadOnly &&
```
quic-seaswara: This should probably be checking for output section instead of input section.

```
cat > 1.s << 'S'
.globl foo
.section .foo, "a", @progbits
foo:
  .word bar
  ret

.data
.hidden bar
bar:
  .word 0
S

cat > s.t << \!
SECTIONS {
  .t : {
            *(.foo*)
            *(.data*)
  }
}
!


clang -c 1.s -target riscv32 -fPIC
$ riscv-ld -shared 1.o -o lib1.so -z text -T s.t && llvm-readelf -r lib1.so

Relocation section '.rela.dyn' at offset 0x1050 contains 1 entries:
 Offset     Info    Type                Sym. Value  Symbol's Name + Addend
0000005c  00000003 R_RISCV_RELATIVE                  62

```

the bfd linker does not error here.

quic-seaswara (changes requested):
Please add tests for -z text with -shared links. 

It might be helpful to add tests that will fail with -z text and not with -z notext

lib/Target/AArch64/AArch64Relocator.h
```diff
@@ -76,6 +76,7 @@ class AArch64Relocator : public Relocator {
                              const ELFSection &pSection) override;
 
 private:
+  bool isPICRelocTypeSupported(const Relocation &reloc) const override;
   bool isInvalidReloc(Relocation &pReloc) const;
```
partaror: It seems that this function is no longer used now.

test/Common/standalone/PIEAbsoluteRelocs/pie-abs32-64.s
```diff
@@ -0,0 +1,33 @@
+# REQUIRES: 64BitsArch
+
+## 64-bit targets should reject absolute relocations in pie/shared
```
partaror, line 3: We do not have any link commands in the test with shared libraries...

lib/Target/Relocator.cpp
quic-areg: The point is that we cannot emit truncated absolute relocations for pie/shared links for 64-bit outputs. We currently do, and that's what this code aims to address. See pie-abs32-64.s for a test case: lld and bfd will reject the truncated relocation while today's eld will accept it.

lib/Target/Relocator.cpp
quic-areg: I think it is easier and cleaner to keep legality decisions in a separate function.

lib/Target/AArch64/AArch64Relocator.h
quic-areg: Thanks; I will remove it

test/Common/standalone/PIEAbsoluteRelocs/pie-abs32-64.s
quic-areg: Done

lib/Target/Relocator.cpp
quic-areg: Thanks for catching this. Updated.

lib/Target/GNULDBackend.cpp
quic-areg: I think it should be part of this PR since it is part of making PIC relocation legality correct. Without this change, the existing defsym case (which isn't actually checked by the following readelf line) would incorrectly fail in ABS64RelocationPIE.test with `recompile with -fPIC`.

lib/Target/X86/x86_64Relocator.cpp
quic-areg: Done

lib/Target/GNULDBackend.cpp
quic-seaswara: I agree it is an issue. We may need to cherry pick the change to another release branch if needed.

For this reason, this change needs to be in a separate PR.

quic-areg: I can push it to the release branch when it's needed

quic-seaswara: Bug fixes / features need to be separated. Please push that as a separate PR.

quic-seaswara: Also I am not sure if absolute symbols ever need a dynamic relocation, should it return false for all cases ?

quic-seaswara (approved):

PR #929, opened 2026-03-09
Improve sysroot command-line option documentation

parth-07:
This commit improves sysroot command-line option sphinx documentation
and adds a framework to supplement TableGen documentation of options
for creating the command-line options sphinx documentation.
GenerateOptionsDocsFromTableGen.py script creates the sphinx
documentation for the command-line options by utilizing the TableGen option
files and a command line options supplement documentation directory that
contains one rst file per option for the options' supplement
documentations.

### DIFF
--- a/docs/userguide/CMakeLists.txt
+++ b/docs/userguide/CMakeLists.txt
@@ -20,6 +20,8 @@ if(LLVM_ENABLE_SPHINX)
 
   add_custom_target(eld-linkeroptions-docs)
   set(LINKER_OPTS_DUMP_DIR ${DOCS_BUILD_SOURCE}/LinkerOptionsDump)
+  set(OPTIONS_SUPPLEMENTS_DIR
+      ${CMAKE_CURRENT_SOURCE_DIR}/CommandLineOptionsSupplements)
   add_custom_target(
     eld-linkeroptions-create-json-dump-dir
     COMMAND ${CMAKE_COMMAND} -E make_directory ${LINKER_OPTS_DUMP_DIR})
@@ -50,7 +52,8 @@ if(LLVM_ENABLE_SPHINX)
         python3 ${CMAKE_CURRENT_SOURCE_DIR}/GenerateOptionsDocsFromTblGen.py
         ${LINKER_OPTS_DUMP_DIR}/${target}LinkerOptionsTblGenDump.json -o
         ${DOCS_BUILD_SOURCE}/documentation/options/${target}LinkerOptions.rst
-        ${SKIP_OPTION})
+        ${SKIP_OPTION}
+        -S ${OPTIONS_SUPPLEMENTS_DIR})
 
     add_dependencies(eld-${targetLC}linkeroptions-docs
                      eld-${targetLC}linkeroptions-json-dump)
--- a/docs/userguide/GenerateOptionsDocsFromTblGen.py
+++ b/docs/userguide/GenerateOptionsDocsFromTblGen.py
@@ -1,9 +1,10 @@
-#!/usr/bin/env
+#!/usr/bin/env python3
 
 import argparse
 import sys
 import json
 import re
+import os
 
 
 class Option:
@@ -73,7 +74,7 @@ class Option:
 
         return option_forms
 
-    def generate_docs(self, out):
+    def generate_docs(self, out, supplement=None):
         option_forms = self.get_all_option_forms()
         if not option_forms:
             return
@@ -82,6 +83,16 @@ class Option:
         if self.help_text:
             out.write("\n")
             out.write(self.help_text)
+            out.write("\n")
+        if supplement:
+            out.write("\n")
+            # Indent every line of the supplement by 3 spaces so it sits
+            # inside the .. option:: directive body.
+            indented = "\n".join(
+                "   " + line if line.strip() else line
+                for line in supplement.splitlines()
+            )
+            out.write(indented)
             out.write("\n")
 
     @classmethod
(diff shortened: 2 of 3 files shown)


docs/userguide/GenerateOptionsDocsFromTblGen.py
```diff
@@ -1,9 +1,10 @@
-#!/usr/bin/env
+#!/usr/bin/env python3
 
 import argparse
 import sys
 import json
 import re
+import yaml
```
quic-seaswara: need this in requirements.txt

docs/userguide/CMakeLists.txt
```diff
@@ -20,6 +20,8 @@ if(LLVM_ENABLE_SPHINX)
 
   add_custom_target(eld-linkeroptions-docs)
   set(LINKER_OPTS_DUMP_DIR ${DOCS_BUILD_SOURCE}/LinkerOptionsDump)
+  set(OPTIONS_SUPPLEMENTS_YAML
+      ${CMAKE_CURRENT_SOURCE_DIR}/OptionsSupplements.yaml)
```
quic-seaswara: Can we name the file CommandLineOptions.yaml/LinkerOptions.yaml instead, because this is relevant only for command line options.

Do you think having one yaml file per option might be better ?

Also put this file in a directory ?

docs/userguide/GenerateOptionsDocsFromTblGen.py
```diff
@@ -234,6 +248,16 @@ def create_argparser():
         "--skip",
         help="Options present in the specified json dump file are skipped / not documented",
     )
+    parser.add_argument(
+        "-S",
+        "--supplement",
+        metavar="yaml-file",
```
quic-seaswara: General question, why --yaml, can use we .rst file itself as rst is parseable, and easy to integrate

docs/userguide/GenerateOptionsDocsFromTblGen.py
partaror: I checked regarding this. We already have pyyaml in the requirements.txt

docs/userguide/CMakeLists.txt
partaror: > Can we name the file CommandLineOptions.yaml/LinkerOptions.yaml instead, because this is relevant only for command line options.

Can we suffix the name with 'Supplements' because it contains supplementary documentation, that is added to the one-liner documentation present in TableGen.

> Do you think having one yaml file per option might be better ?

One yaml file per option seems like way too many files to me. I believe that the documentations in the command-line option documentation should be brief, stating all the important information that a user may want to quickly review. For explaining core concepts that require more detailed explanation, we should use a separate page altogether. Please let me know your thoughts on this.

> Also put this file in a directory ?

It is currently in docs/userguide. Should  we place it somewhere else?

docs/userguide/GenerateOptionsDocsFromTblGen.py
partaror: rst file is much more complicated to parse than yaml. Yaml just contains the option to the rst documentation mapping. Why would parsing rst be easier to integrate than yaml here?

quic-areg: You don't really need to parse anything do you? All you have to do is insert one blob of text into another.

quic-seaswara: You will need to add the text for that option that is being processed

parth-07: I have change the framework to use one rst file per option for the option's supplement documentation as we discussed offline.

docs/userguide/GenerateOptionsDocsFromTblGen.py
```diff
@@ -200,14 +211,17 @@ def get_refined_options_info(raw_options_info):
...
     """Generate command-line options documentation"""
+    if supplements is None:
+        supplements = {}
     for key, group_info in groups.items():
         if not group_info.options or key is None:
             continue
         group_info.generate_docs(out)
         for option_info in group_info.options:
-            option_info.generate_docs(out)
+            supplement = supplements.get(option_info.name)
```
quic-seaswara, line 223: The option is --sysroot, where is the -- stripped away in the python code ?

quic-seaswara: Probably it might be useful to say `dashdashoption.rst` or `dash<option>.rst`

partaror: This file parses the JSON file generated by running the below command. `--` is not stripped away, it is never assumed to be part of the option name. We take the option name as specified in tablegen sources.

```
llvm-tblgen
        ${ELD_SOURCE_DIR}/include/eld/Driver/${target}LinkerOptions.td
        -dump-json -I${LLVM_MAIN_INCLUDE_DIR}
        -I${ELD_SOURCE_DIR}/include/eld/Driver -o
        ${LINKER_OPTS_DUMP_DIR}/${target}LinkerOptionsTblGenDump.json
```

Often, we support multiple forms of an option, for example, both `-whole-archive`, and `--whole-archive` are supported. For this, there's only one entry named `whole_archive` in the tablegen, with two values in the array member `Prefixes`. This directly corresponds to the `whole_archive` definition in the tablegen file:

```
def whole_archive : Flag<["-", "--"], "whole-archive">,
                    HelpText<"Force load of all members in a static library">,
                    Group<grp_main>;
```

partaror: > Probably it might be useful to say `dashdashoption.rst` or `dash<option>.rst`

Wouldn't this be ambiguous for the options that support both kind of dashes? Isn't using the same name as the tablegen more convenient as it avoids all ambiguity. We also have the concept of canonical and alias options. Using the non-alias (canonical) option defined in the TableGen for specifying the supplement documentation help further avoids ambiguity.

docs/userguide/CMakeLists.txt
quic-seaswara: one more comment, do we know that if rst is rendered directly, sphinx looks for rst files in the folder anywhere and renders it ? can you pls confirm ?

if thats the case we can make the files named as sysroot.rst.in

partaror: Thank you for this comment. I just confirmed the behavior. Yes, sphinx renders all rst files within the source directory into html, which the end-users can see by visiting `${ROOT}/CommandLineOptionsSupplements/<option>.html`.

I have converted the option supplement documentation `.rst` files to `.rst.in` files as you have suggested.

quic-seaswara (approved):

PR #932, opened 2026-03-10
[hexagon] Add linux emulation

quic-seaswara:
Add hexagonlinux emulation, and enable hexagon linux tests.

This change also removes the special casing for hexagon-linux-link/x86_64-link.


lib/LinkerWrapper/HexagonLinkDriver.cpp
```diff
@@ -207,7 +207,39 @@ bool HexagonLinkDriver::createInputActions(
...
+  std::string emulation = Config.options().getEmulation().str();
+  std::string arch = Config.targets().getArch();
+  // If a specific emulation was requested, apply it now.
+  if (!emulation.empty()) {
+    llvm::Triple TheTriple = Config.targets().triple();
+    std::optional<Triple> OptEmulationTriple =
+        ParseEmulation(emulation, Config.getDiagEngine());
+    // Report invalid emulation error for unknown emulation.
+    if (!OptEmulationTriple) {
+      DiagEngine->raise(eld::Diag::err_invalid_emulation) << emulation << "\n";
```
quic-areg: do we need the "\n"?

quic-seaswara: we can remove it, thanks for the super quick review!

lib/LinkerWrapper/HexagonLinkDriver.cpp
```diff
@@ -207,7 +207,39 @@ bool HexagonLinkDriver::createInputActions(
...
+  return result;
+}
+
+std::optional<Triple>
+HexagonLinkDriver::ParseEmulation(std::string pEmulation,
+                                  eld::DiagnosticEngine *DiagEngine) {
+  std::optional<Triple> result =
+      StringSwitch<std::optional<Triple>>(pEmulation)
+          .Case("hexagonlinux", Triple("hexagon", "", "linux", "gnu"))
+          .Default(Triple("hexagon", "", "", ""));
```
partaror: Why don't we return `nullopt` in `Default`? Here the code is always returning a valid triple even for invalid emulation strings.

test/Hexagon/linux/LinuxTriples/LinuxTriples.test
```diff
@@ -6,7 +6,7 @@
 RUN: %clang %clangopts -c %clangg0opts %p/Inputs/1.c -o %t1.o -fPIC
 RUN: %clang %clangopts -c %clangg0opts %p/Inputs/2.c -o %t2.o -fPIC
 RUN: %link %linkopts -shared %t2.o -o %t2.so
-RUN: %link %linkopts  -mtriple hexagon-unknown-linux-musl -Map %tmusl.map %t1.o %t2.so -o %tmusl.out
+RUN: %link -mtriple hexagon-unknown-linux-musl -Map %tmusl.map %t1.o %t2.so -o %tmusl.out
```
partaror, line 9: Can we also add test for how the emulation is affecting the link, for example, adding CHECKs for start address?

include/eld/Driver/HexagonLinkDriver.h
```diff
@@ -77,6 +77,9 @@ class HexagonLinkDriver : public GnuLdDriver {
 
   static bool isValidEmulation(llvm::StringRef Emulation);
 
+  static std::optional<llvm::Triple>
+  ParseEmulation(std::string pEmulation, eld::DiagnosticEngine *DiagEngine);
```
partaror, line 81: Can we pass the emulation string as const-reference or `llvm::StringRef` to avoid copying?

lib/LinkerWrapper/HexagonLinkDriver.cpp
```diff
@@ -207,7 +207,39 @@ bool HexagonLinkDriver::createInputActions(
 
 template <class T>
 bool HexagonLinkDriver::processTargetOptions(llvm::opt::InputArgList &Args) {
-  return GnuLdDriver::processTargetOptions<T>(Args);
+  bool result = GnuLdDriver::processTargetOptions<T>(Args);
+  std::string emulation = Config.options().getEmulation().str();
+  std::string arch = Config.targets().getArch();
```
partaror: This variable is unused.

test/Hexagon/linux/lit.local.cfg
```diff
@@ -1,4 +1,6 @@
-config.suffixes = ['.test']
+for i, (k, v) in enumerate(config.substitutions):
+      if k == '%linkopts':
+          config.substitutions[i] = (k, '-mhexagonlinux')
```
partaror: Should we append/prepend `-mhexagonlinux` to `%linkopts` instead of completely replacing it?

lib/LinkerWrapper/HexagonLinkDriver.cpp
quic-seaswara: Thanks for this comment, I will try to fix it.

test/Hexagon/linux/LinuxTriples/LinuxTriples.test
quic-seaswara: We have a test already : Map/ImageStartAddress/ImageStartAddress.test

include/eld/Driver/HexagonLinkDriver.h
quic-seaswara: This will be a follow up, as other targets also need to incur this change. Hope thats ok

test/Hexagon/linux/lit.local.cfg
quic-seaswara: I will update this by adding emulation to all tests, and replace emulation with hexagon linux targeting linux.

lib/LinkerWrapper/HexagonLinkDriver.cpp
quic-seaswara: Removed arch

include/eld/Driver/HexagonLinkDriver.h
quic-areg: Isn't it a static function?

quic-seaswara: Each driver has this function.

PR #942, opened 2026-03-16
Align start of tdata

quic-k:
Align start and end of TLS template so that the final TLS size is aligned.

### DIFF
--- a/templates/staticExecutable/static-executable-h2-picolibc.lcs.template
+++ b/templates/staticExecutable/static-executable-h2-picolibc.lcs.template
@@ -103,6 +103,7 @@ SECTIONS
 
   /* Thread Local Storage sections  */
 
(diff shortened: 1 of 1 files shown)


templates/staticExecutable/static-executable-h2-picolibc.lcs.template
```diff
@@ -93,13 +93,13 @@ SECTIONS
   /* TLS Initialized Data */
   .tdata          :
   {
+    . = ALIGN(8);
```
quic-seaswara: You might want to move this before .tdata output section description ? 

. = ALIGN(8);
.tdata : {
}

both are same but if there is no .tdata and the alignment is not 8 byte, the image will incur a size increase of x bytes as .tdata will become NOBITS because of alignment padding.

PR #944, opened 2026-03-17
Recompute layout until symbol values converge

### LINKED ISSUE #943: Incorrect symbol value due to forward reference in linker script
We may observe incorrect symbol values due to forward references in linker script. This can issues as incorrect relaxations. 

Reproducer:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
int foo() { return 1; }
int u = 3;
EOF

cat >script.t <<\EOF
SECTIONS {
  .text : { *(.text*) }
  .data : { *(.data*) }
  .comment : { *(.comment*) }
  a = b + c;
  b = c + 0x100;
  c = 0x1000;
}
EOF


clang -o 1.o -target aarch64-linux-gnu 1.c -c
LDs=(ld.eld ld.lld aarch64-none-linux-gnu-ld.bfd)
SFs=(eld lld bfd)

for i in "${!SFs[@]}"; do
  ${LDs[$i]} -o 1.${SFs[$i]}.out 1.o -T script.t
  llvm-readelf -s 1.${SFs[$i]}.out | grep 'ABS a'
done
```

Output:
```
13: 0000000000001100     0 NOTYPE  GLOBAL DEFAULT   ABS a
    10: 0000000000002100     0 NOTYPE  GLOBAL DEFAULT   ABS a
aarch64-none-linux-gnu-ld.bfd: warning: 1.bfd.out has a LOAD segment with RWX permissions
    13: 0000000000002100     0 NOTYPE  GLOBAL DEFAULT   ABS a
```

eld generated executable has incorrect value for the symbol `a`, `1100` instead of the `2100`.

parth-07:
This commit modifies the layout step to recompute layout until the symbol values converge. The reiteration has an upper-limit of 4. On reaching this limit, a note is emitted and layout reiteration is stopped.

Please note that this reiteration limit is for layout iterations before relaxations take place. After a layout relaxation pass, the reiteration count is reset to 0.

Resolves #943

### DIFF
--- a/include/eld/Target/GNULDBackend.h
+++ b/include/eld/Target/GNULDBackend.h
@@ -772,13 +772,23 @@ public:
 
   struct LayoutSnapshot {
     llvm::DenseMap<const OutputSectionEntry *, SectionAddrs> outSections;
+    std::vector<uint64_t> assignmentValues;
   };
 
-  // Capture current layout snapshot keyed by OutputSectionEntry*.
+  struct DivergenceResult {
+    const OutputSectionEntry *outputSection = nullptr;
+    const Assignment *assignment = nullptr;
+  };
+
   LayoutSnapshot captureLayoutSnapshot() const;
 
-  const OutputSectionEntry *findDivergence(const LayoutSnapshot &Prev,
-                                           const LayoutSnapshot &cur) const;
+  void captureAssignmentsSnapshot(LayoutSnapshot &S) const;
+
+  DivergenceResult findDivergence(const LayoutSnapshot &Prev,
+                                  const LayoutSnapshot &Cur) const;
+
+  const Assignment *findAssignmentDivergence(const LayoutSnapshot &Prev,
+                                             const LayoutSnapshot &Cur) const;
 
   // --- Exclude symbol support
   void markSymbolForRemoval(const ResolveInfo *S);
--- a/lib/Target/GNULDBackend.cpp
+++ b/lib/Target/GNULDBackend.cpp
@@ -3318,25 +3318,65 @@ GNULDBackend::LayoutSnapshot GNULDBackend::captureLayoutSnapshot() const {
     A.lma = Sec->pAddr();
     S.outSections.insert({O, A});
   }
+  captureAssignmentsSnapshot(S);
   return S;
 }
 
-const OutputSectionEntry *
+static bool shouldSkipAssignForLayoutConv(const Assignment *A) {
+  if (A->type() == Assignment::ASSERT || A->type() == Assignment::FILL)
+    return true;
+  if (A->isProvideOrProvideHidden() && !A->isUsed())
+    return true;
+  return false;
+}
+
+void GNULDBackend::captureAssignmentsSnapshot(LayoutSnapshot &S) const {
+  for (const Assignment *A : m_Module.getScript().assignments()) {
+    if (shouldSkipAssignForLayoutConv(A))
+      continue;
+    S.assignmentValues.push_back(A->value());
+  }
+}
+
+const Assignment *
+GNULDBackend::findAssignmentDivergence(const LayoutSnapshot &Prev,
+                                       const LayoutSnapshot &Cur) const {
+  const auto &Assignments = m_Module.getScript().assignments();
+  size_t Idx = 0;
+  for (const Assignment *A : Assignments) {
+    if (shouldSkipAssignForLayoutConv(A))
+      continue;
+    if (Idx >= Prev.assignmentValues.size() ||
+        Idx >= Cur.assignmentValues.size() ||
+        Prev.assignmentValues[Idx] != Cur.assignmentValues[Idx])
+      return A;
+    ++Idx;
+  }
+  return nullptr;
+}
+
+GNULDBackend::DivergenceResult
 GNULDBackend::findDivergence(const LayoutSnapshot &Prev,
-                             const LayoutSnapshot &cur) const {
+                             const LayoutSnapshot &Cur) const {
+  DivergenceResult Result;
   const SectionMap &SM = m_Module.getScript().sectionMap();
   for (auto It = SM.begin(), End = SM.end(); It != End; ++It) {
     const OutputSectionEntry *O = *It;
     auto PrevIt = Prev.outSections.find(O);
-    auto CurIt = cur.outSections.find(O);
-    if (PrevIt == Prev.outSections.end() || CurIt == cur.outSections.end())
-      return O;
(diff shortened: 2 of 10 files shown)


lib/Target/GNULDBackend.cpp
```diff
@@ -3292,25 +3293,68 @@ GNULDBackend::LayoutSnapshot GNULDBackend::captureLayoutSnapshot() const {
...
   return S;
 }
 
-const OutputSectionEntry *
+std::vector<const LDSymbol *> GNULDBackend::collectScriptDefinedSymbols() const {
+  std::vector<const LDSymbol *> Result;
+  const llvm::StringMap<ResolveInfo *> &Globals =
+      m_Module.getNamePool().getGlobals();
+  for (const auto &Entry : Globals) {
+    const ResolveInfo *RI = Entry.second;
```
quic-seaswara: We should not be comparing the symbol values of global symbols, as you can clearly see that local symbols are missing. 

Can we check for fragment address instead ?

lib/Target/GNULDBackend.cpp
```diff
@@ -3292,25 +3293,68 @@ GNULDBackend::LayoutSnapshot GNULDBackend::captureLayoutSnapshot() const {
...
+    ASSERT(RI, "Must not be null!");
```
quic-seaswara: Is this assert needed ?

include/eld/Target/GNULDBackend.h
```diff
@@ -771,13 +771,31 @@ class GNULDBackend {
 
   struct LayoutSnapshot {
     llvm::DenseMap<const OutputSectionEntry *, SectionAddrs> outSections;
+    llvm::DenseMap<const LDSymbol *, uint64_t> symbolValues;
   };
 
-  // Capture current layout snapshot keyed by OutputSectionEntry*.
-  LayoutSnapshot captureLayoutSnapshot() const;
+  struct DivergenceResult {
+    const OutputSectionEntry *section = nullptr;
```
quic-seaswara: section -> output section

lib/Target/GNULDBackend.cpp
```diff
@@ -3292,25 +3293,68 @@ GNULDBackend::LayoutSnapshot GNULDBackend::captureLayoutSnapshot() const {
...
+    if (!RI->outSymbol())
+      continue;
+    const LDSymbol *Sym = RI->outSymbol();
+    if (Sym->scriptDefined())
+      Result.push_back(Sym);
```
quic-seaswara: It would be useful to show the expression in the diagnostic instead of symbol, that will allow users do diagnose the problem better.

lib/Target/GNULDBackend.cpp
```diff
@@ -3292,25 +3293,68 @@ GNULDBackend::LayoutSnapshot GNULDBackend::captureLayoutSnapshot() const {
...
+  }
+  return Result;
+}
+
+void GNULDBackend::captureSymbolsSnapshot(
+    LayoutSnapshot &S,
+    const std::vector<const LDSymbol *> &ScriptDefinedSymbols) const {
+  for (const LDSymbol *Sym : ScriptDefinedSymbols)
+    S.symbolValues.insert({Sym, Sym->value()});
```
quic-seaswara: Does this work for provide as well ? because the symbols are not defined

lib/Target/GNULDBackend.cpp
```diff
@@ -4123,17 +4170,19 @@ bool GNULDBackend::relax() {
...
-      if (changed) {
-        // Emit a note for the first observed changed section.
-        if (const ELFSection *S = changed->getSection())
+      if (diverged.section) {
+        if (const ELFSection *S = diverged.section->getSection())
           config().raise(Diag::note_section_address_not_converging)
               << S->name() << maxIterations;
+      } else if (diverged.symbol) {
+        config().raise(Diag::note_symbol_value_not_converging)
+            << diverged.symbol->name() << maxIterations;
```
quic-seaswara: print the symbol, section information and the input file/ input linker script

lib/Target/GNULDBackend.cpp
parth-07: I thought about it. I believe that just comparing the linker script symbols is enough. If all the linker script symbols and the output section address converge, then all the section-relative symbols will also converge. Please let me know your thoughts on this.

quic-seaswara: interesting observation. Would this capture bumping alignment using dot inside output sections ?

parth-07: >Would this capture bumping alignment using dot inside output sections ?
> It would be useful to show the expression in the diagnostic instead of symbol, that will allow users do diagnose the problem better.

Thank you for these comments. I thought more about it. If we ensure that linker script symbols values converge after layout passes, then for most cases dot assignments within output section would converge as well. However, it is possible that due to corner cases / layout issues, the final value of a symbol is converging but the intermittent value is not. 

It would be much better to ensure that the assignments result converge instead of linker script values converge. This will capture the symbol intermediate values case as well. Additionally, it would make it possible to pin-point the exact expression where non-convergence is happening -- something that we cannot do if we only compare symbol values. Please let me know your thoughts on this.

lib/Target/GNULDBackend.cpp
parth-07: Yes, it works with PROVIDE as well. Provide symbols are handled / included much before the layout starts.

lib/Target/GNULDBackend.cpp
parth-07: I have changed the design to use assignments instead of symbols to check layout convergence based on our offline discussion.

include/eld/Target/GNULDBackend.h
parth-07: Done.

lib/Target/GNULDBackend.cpp
parth-07: Done. Now the diagnostic show the exact assignment that is non-converging.

lib/Target/GNULDBackend.cpp
parth-07: I have added the assignment dump and the input linker script information here.

lib/Target/GNULDBackend.cpp
parth-07: I used the assert as a safety check for nulls. I was not sure if it is really required. This code has been removed as it is not required in the updated design in which we use assignment values instead of symbol values for determining convergence.

PR #945, opened 2026-03-17
[AArch64] Fix pointer mismatch for ifunc fn for adrp-add and abs relocs

### LINKED ISSUE #913: Different address in pointers to the same `ifunc` function
Currently, we have different address in pointers to the same `ifunc` functions when building a pie static executable and there is an absolute reference to the `ifunc` symbol.

Reproducer:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
#include <stdio.h>

char *foo1() {
  return "foo";
}

char *foo2() {
  return "bar";
}

static char *(*resolve_foo (void)) (void) {
  return foo1;
}

char *foo();
EOF

cat >main.c <<\EOF
#include <stdio.h>
char *foo();

char *(*foogp)() = foo;
int main() {
  char *(*foop)() = foo;
  printf("Hello %s, %p, %p\n", foo(), foop, foogp);
  return 0;
}
EOF
rm ./ld
cat >ld <<\EOF
#!/usr/bin/env bash

ld.eld "$@"
EOF

chmod +x ./ld
aarch64-none-linux-gnu-gcc -o 1.eld.out 1.c main.o -O0 -static -B . -Wl,-Map,1.eld.map.txt
aarch64-none-linux-gnu-gcc -o 1.lld.out 1.c main.o -O0 -static -fuse-ld=lld -Wl,-Map,1.lld.map.txt
aarch64-none-linux-gnu-gcc -o 1.bfd.out 1.c main.o -O0 -static  -Wl,-Map,1.bfd.map.txt

qemu-aarch64 ./1.eld.out
qemu-aarch64 ./1.lld.out
qemu-aarch64 ./1.bfd.out
```

Output:
```
+ qemu-aarch64 ./1.eld.out
Hello foo, 0x400654, 0x4002c0
+ qemu-aarch64 ./1.lld.out
Hello foo, 0x289c70, 0x289c70
+ qemu-aarch64 ./1.bfd.out
Hello foo, 0x400280, 0x400280
```

In the eld case, the pointers to the same `ifunc` function holds different addresses.

The root cause for this is as follows:

1) When there is no absolute reference to an `ifunc` symbol, then `R_AARCH64_ADR_GOT_PAGE` relocation returns the `.got.plt` slot of the `ifunc` function. This behavior is consistent in all ld.eld, ld.bfd and ld.lld.

2) However, when there is an absolute reference to `ifunc` symbol and a GOT-relocation referencing the symbol, then both bfd and lld, creates a `.got` slot for the symbol and fill it with the address of the `.plt` slot of the symbol. The GOT-relocation now returns the address of this `.got` slot instead of the `.got.plt` slot. This is done so that all the address-of `ifunc` function expressions return the same address.

In eld, we do not handle the case (2) currently.

parth-07:
This commit fixes pointer inequality in pointers to the same ifunc
symbol when adrp-add relocation pair and abs relocations are used. Let's understand the issue in more detail using the below example:

```cpp
char *foo_impl() { return "foo"; }

static char *(*foo_resolver())(void) { return foo_impl; }

char *foo() __attribute__((ifunc("foo_resolver")));

char *(*foogp)() = foo;

int main() {
  char *(*foop)() = foo;
}
```

`foogp` and `foop` pointers both point to the same function `foo`. However, their values were different. The root cause was improper handling of the underlying relocations when the symbol is of type ifunc.

```
char *(*foogp)() = foo; // R_AARCH64_ABS64
// ...
char *(*foop)() = foo; // R_AARCH64_ADR_GOT_PAGE + R_AARCH64_LD64_GOT_LO12_NC
```

Previously, the linker resolved the R_AARCH64_ABS64 relocation here to the plt slot of foo, let's refer to it as plt[foo], and the GOT-relocations were resolved to the .got.plt slot of foo, let's refer to it as .got.plt[foo]. As a result, foogp stores the address of plt[foo] and foop stores the contents of .got.plt[foo], that is, the address of the resolved function foo. Clearly, the two values are different.

We resolve this issue by creating a got slot for foo when the foo has both an absolute reference and a GOT-reference. The got slot of foo is filled by the linker and stores the address of plt[foo]. With this, both the absolute-reference and got-reference to an ifunc symbol results in the same address. Please note that:

- we do not create .got slot of foo when we only have an absolute reference to foo because .got slot is unnecessary in this case, and
- we do not create .got slot of foo when we only have a GOT-reference because in this case we can directly use .got.plt slot and access the function directly without any indirection penalty or the pointer-inequality bug.

Additionally, this PR adds a developer guide docs/DeveloperDocs/AArch64/IFunc.md
that details the eld behavior for GNU IFunc functionality on AArch64
target.

Resolves #913

### DIFF
--- a/include/eld/SymbolResolver/ResolveInfo.h
+++ b/include/eld/SymbolResolver/ResolveInfo.h
@@ -149,6 +149,18 @@ public:
     return ((ThisBitField & PatchableMask) == PatchableMask);
   }
 
+  void setIFuncDirectRef() { ThisBitField |= IFuncDirectRefFlag; }
+
+  bool hasIFuncDirectRef() const {
+    return ((ThisBitField & IFuncDirectRefMask) == IFuncDirectRefMask);
+  }
+
+  void setIFuncNeedsGOT() { ThisBitField |= IFuncNeedsGOTFlag; }
+
+  bool hasIFuncNeedsGOT() const {
+    return ((ThisBitField & IFuncNeedsGOTMask) == IFuncNeedsGOTMask);
+  }
+
   // -----  observers  ----- //
   bool isNull() const;
 
@@ -314,14 +326,19 @@ private:
   static const uint32_t PreserveOffset = 18;
   static const uint32_t PreserveMask = 1 << PreserveOffset;
 
-  // FIXME: offset 19 can be used here!
+  static const uint32_t IFuncDirectRefOffset = 19;
+  static const uint32_t IFuncDirectRefMask = 1 << IFuncDirectRefOffset;
+
   static const uint32_t PatchableOffset = 20;
   static const uint32_t PatchableMask = 1 << PatchableOffset;
 
+  static const uint32_t IFuncNeedsGOTOffset = 21;
+  static const uint32_t IFuncNeedsGOTMask = 1 << IFuncNeedsGOTOffset;
+
   static const uint32_t InfoMask = 0xF;
 
-  // Bits are from 0-20.
-  static const uint32_t ResolveMask = 0x1FFFFF;
+  // Bits are from 0-21.
+  static const uint32_t ResolveMask = 0x3FFFFF;
 
 public:
   static const uint32_t GlobalFlag = 0 << GlobalOffset;
@@ -343,6 +360,8 @@ public:
   static const uint32_t InbitcodeFlag = 1 << InBitcodeOffset;
   static const uint32_t PreserveFlag = 1 << PreserveOffset;
   static const uint32_t PatchableFlag = 1 << PatchableOffset;
+  static const uint32_t IFuncDirectRefFlag = 1 << IFuncDirectRefOffset;
+  static const uint32_t IFuncNeedsGOTFlag = 1 << IFuncNeedsGOTOffset;
   ResolveInfo();
   ResolveInfo(llvm::StringRef SymbolName);
   ~ResolveInfo();
--- a/lib/Target/AArch64/AArch64LDBackend.cpp
+++ b/lib/Target/AArch64/AArch64LDBackend.cpp
@@ -574,6 +574,33 @@ bool AArch64LDBackend::finalizeTargetSymbols() {
   return true;
 }
 
+bool AArch64LDBackend::finalizeScanRelocations() {
+  if (!config().isCodeStatic())
+    return true;
+
+  ELFObjectFile *Obj = getDynamicSectionHeadersInputFile();
+  if (!Obj)
+    return true;
+
+  for (auto &[symInfo, plt] : m_PLTMap) {
+    if (!symInfo->isIFunc() || !symInfo->hasIFuncDirectRef() ||
+        !symInfo->hasIFuncNeedsGOT())
+      continue;
+
+    AArch64GOT *G = AArch64GOT::Create(Obj->getGOT(), symInfo);
+
+    FragmentRef *PLTFragRef = make<FragmentRef>(*plt, 0);
+    Relocation *r = Relocation::Create(llvm::ELF::R_AARCH64_ABS64, 64,
+                                       make<FragmentRef>(*G, 0), 0);
+    Obj->getGOT()->addRelocation(r);
+    r->modifyRelocationFragmentRef(PLTFragRef);
+
+    recordGOT(symInfo, G);
+    symInfo->setReserved(symInfo->reserved() | Relocator::ReserveGOT);
+  }
+  return true;
+}
+
 void AArch64LDBackend::setupStaticTCBForTLSSupport() {
   if (!config().isCodeStatic())
     return;
--- a/lib/Target/AArch64/AArch64Relocator.cpp
+++ b/lib/Target/AArch64/AArch64Relocator.cpp
@@ -346,14 +346,14 @@ void AArch64Relocator::scanGlobalReloc(InputFile &pInput, Relocation &pReloc,
   ELFObjectFile *Obj = llvm::dyn_cast<ELFObjectFile>(&pInput);
   // rsym - The relocation target symbol
   ResolveInfo *rsym = pReloc.symInfo();
+  if (rsym && rsym->isIFunc() && config().isCodeStatic())
+    return handleScanForNonPreemptibleIFunc(pReloc, Obj);
   switch (pReloc.type()) {
   case llvm::ELF::R_AARCH64_AUTH_ABS64:
   case llvm::ELF::R_AARCH64_ABS16:
   case llvm::ELF::R_AARCH64_ABS32:
   case llvm::ELF::R_AARCH64_ABS64: {
     std::lock_guard<std::mutex> relocGuard(m_RelocMutex);
-    if (handleScanForNonPreemptibleIFunc(rsym, Obj))
-      return;
     bool isAuthAbs = pReloc.type() == llvm::ELF::R_AARCH64_AUTH_ABS64;
     // Absolute relocation type, symbol may needs PLT entry or
     // dynamic relocation entry
@@ -450,8 +450,6 @@ void AArch64Relocator::scanGlobalReloc(InputFile &pInput, Relocation &pReloc,
   case llvm::ELF::R_AARCH64_JUMP26:
   case llvm::ELF::R_AARCH64_CALL26: {
     std::lock_guard<std::mutex> relocGuard(m_RelocMutex);
-    if (handleScanForNonPreemptibleIFunc(rsym, Obj))
-      return;
     // return if we already create plt for this symbol
     if (rsym->reserved() & ReservePLT)
       return;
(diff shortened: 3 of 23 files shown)


lib/Target/AArch64/AArch64LDBackend.cpp
```diff
@@ -574,6 +574,36 @@ bool AArch64LDBackend::finalizeTargetSymbols() {
...
+
+    FragmentRef *PLTFragRef = make<FragmentRef>(*plt, 0);
+    Relocation *r = Relocation::Create(llvm::ELF::R_AARCH64_ABS64, 64,
+                                       make<FragmentRef>(*G, 0), 0);
+    Obj->getGOT()->addRelocation(r);
+    r->modifyRelocationFragmentRef(PLTFragRef);
+
+    recordGOT(symInfo, G);
+    symInfo->setReserved(symInfo->reserved() | Relocator::ReserveGOT);
+  }
```
quic-seaswara, line 600: The GNU linker creates a igot and igot.plt probably it would be useful to follow the pattern for IFUNC symbols.

quic-seaswara (changes requested):

lib/Target/AArch64/AArch64LDBackend.cpp
quic-seaswara: This can simplify the ifunc implementation too, probably we can reserve the GOT and PLT slots and later remove the GOT slots not needed.

parth-07: > The GNU linker creates a igot and igot.plt probably it would be useful to follow the pattern for IFUNC symbols.

This is interesting. I did not observe this during my experimentation of IFunc functionality with gnu ld. Can you please share which GNU version are you using? 

```bash
$ aarch64-none-linux-gnu-ld.bfd --version
GNU ld (Arm GNU Toolchain 15.2.Rel1 (Build arm-15.86)) 2.45.1.20251203
Copyright (C) 2025 Free Software Foundation, Inc.
This program is free software; you may redistribute it under the terms of
the GNU General Public License version 3 or (at your option) a later version.
This program has absolutely no warranty.
$ llvm-readelf -S 1.bfd.out | grep -i .got
  [20] .got              PROGBITS        000000000001ffb8 00ffb8 000030 08  WA  0   0  8
  [21] .got.plt          PROGBITS        000000000001ffe8 00ffe8 000048 08  WA  0   0  8
```

> This can simplify the ifunc implementation too, probably we can reserve the GOT and PLT slots and later remove the GOT slots not needed.

This would be the same even if we use the normal `.got` slot instead of the `.igot` slot, right? I do prefer to have mechanisms in place so that we only create the slots if they are required. Preemptively creating the slots may lead to unnecessary operations, and reverting/removing the slot may silently bring in bugs in the future if we add any new side-effect to the .got slot creation.

lib/Target/AArch64/AArch64Relocator.cpp
```diff
@@ -471,6 +473,8 @@ void AArch64Relocator::scanGlobalReloc(InputFile &pInput, Relocation &pReloc,
   case llvm::ELF::R_AARCH64_ADR_PREL_PG_HI21:
   case R_AARCH64_ADR_PREL_PG_HI21_NC: {
     std::lock_guard<std::mutex> relocGuard(m_RelocMutex);
+    if (rsym->isIFunc() && config().isCodeStatic())
+      rsym->setIFuncAbsRef();
```
efriedma-quic: I'm not sure "AbsRef" really captures the rule here... the issue is really whether the reference is direct, i.e. not through the GOT.

Do we also need to handle R_AARCH64_MOVW_UABS_G0 etc.?

parth-07: Thank you so much for this comment. I went through all (most) of the AArch64 relocations to ensure that we are not missing any other important case. For now, I have limited the scope of this PR to fix pointer mismatch issue for adrp-add / abs relocations. Additionally, I have documented the current eld behavior, along with the future fixes required, in the developer documentation added as part of this PR: https://github.com/parth-07/eld/blob/e19e8ceff47202b8138954093a4c88cd9eb6cde6/docs/DeveloperDocs/AArch64/IFunc.md.

Some important points:

- Yes, we do need to handle `R_AARCH64_MOVW_UABS_G{0, 1, 2, 3}{_NC}` properly. Currently, it has the same pointer inequality issue. However, GNU AArch64 toolchain does not support this relocation for IFunc symbols. However, lld does. @efriedma-quic @quic-seaswara What do you think the eld behavior should be here?

Additionally, GNU AArch64 toolchain does not support many relocations for IFunc symbols. We need to decide eld behavior here. LLD support most, if not all, of the relocations. Some do not work correctly and some do. These relocations are:

- `R_AARCH64_PREL{16, 32, 64}` 
   - In my opinion, we should support this with IFunc in eld
- `R_AARCH64_ADR_PREL_LO21`
   - Currently, we support this in eld
- `R_AARCH64_MOVW_UABS_G{0,1,2,3}{_NC} / R_AARCH64_SABS_G{0,1,2,3}`  
   - In my opinion, we should support this with IFunc in eld
- `LD_PREL_LO19`  
   - In my opinion, it does not make sense for function symbols and hence we should error out
- `LDST{8, 16, 32, 64, 128}_ABS_LO12_NC`  
   - In my opinion, it does not make sense for function symbols and hence we should error out

efriedma-quic: We probably need R_AARCH64_MOVW_UABS_G* to support code compiled with large code model.  Otherwise it probably shouldn't show up, so it's probably fine to bail out for now.

Not sure if LLVM ever generates R_AARCH64_PREL* for ifuncs; probably not.

parth-07:
@quic-seaswara Gentle ping for review.

parth-07:
clang-format check is incorrectly failing here. More information here: https://github.com/qualcomm/eld/pull/1077#issuecomment-4294965261

lib/Target/AArch64/AArch64Relocator.cpp
parth-07: > We probably need R_AARCH64_MOVW_UABS_G* to support code compiled with large code model.

I will add support for MOVW relocations as a follow-up patch.

efriedma-quic (approved):
Can you add an explicit bailout for the relocations which we support, but don't currently work with ifuncs?

Otherwise LGTM

lib/Target/AArch64/AArch64Relocator.cpp
```diff
@@ -1302,14 +1306,58 @@ Relocator::Result copyInstruction(Relocation &pReloc,
...
+  // These relocations are UNSUPPORTED for now!
+  case llvm::ELF::R_AARCH64_GOT_LD_PREL19:
+  case llvm::ELF::R_AARCH64_MOVW_GOTOFF_G0:
+  case llvm::ELF::R_AARCH64_MOVW_GOTOFF_G0_NC:
+  case llvm::ELF::R_AARCH64_MOVW_GOTOFF_G1:
+  case llvm::ELF::R_AARCH64_MOVW_GOTOFF_G1_NC:
+  case llvm::ELF::R_AARCH64_MOVW_GOTOFF_G2:
+  case llvm::ELF::R_AARCH64_MOVW_GOTOFF_G2_NC:
+  case llvm::ELF::R_AARCH64_MOVW_GOTOFF_G3:
+    llvm_unreachable("Unsupported relocations should not reach here!");
```
quic-areg, line 1358: This will crash now instead of error:
```
cat <<\! >1.s
  .text
  .global _start
_start:
  ldr x0, :got:foo
  ret

foo_resolver:
  ret

  .global foo
  .type foo, %gnu_indirect_function
  .set foo, foo_resolver
!
llvm-mc -triple=aarch64-linux-gnu -filetype=obj -o 1.o 1.s
ld.eld -static -o t.out 1.o
```

include/eld/SymbolResolver/ResolveInfo.h
```diff
@@ -314,14 +326,19 @@ class ResolveInfo {
   static const uint32_t PreserveOffset = 18;
   static const uint32_t PreserveMask = 1 << PreserveOffset;
 
-  // FIXME: offset 19 can be used here!
+  static const uint32_t IFuncAbsRefOffset = 19;
```
quic-areg: could this be `IFuncDirectRefOffset` for consistency?

lib/Target/AArch64/AArch64Relocator.cpp
parth-07: I looked into it. Here the root cause was that the function AArch64Relocator::isRelocSupported did not used to return false for unsupported relocations. I have fixed this function and we no longer crash here now.

include/eld/SymbolResolver/ResolveInfo.h
parth-07: Thank you for mentioning this. I have updated it.

parth-07:
> Can you add an explicit bailout for the relocations which we support, but don't currently work with ifuncs?

Thank you for the reviews. I have added warnings for unsupported and invalid relocations. This list of relocations will be improved in the follow-up patches.

quic-areg (approved):

PR #946, opened 2026-03-17
Fix stack-buffer-overflow sanitizer issue in AUIPC_JALR_TO_XQCI.test

parth-07:
This commit fixes stack-buffer-overflow sanitizer issue in AUIPIC_JALR_TO_XQCI.test. The root cause was that the QC_E_J relaxation patch 6 bytes instruction but the 'Instr' parameter of 'RegionFragmentEx::replaceInstruction' that stores the updated instruction is only of 4 bytes.

### DIFF
--- a/lib/Target/RISCV/RISCVLDBackend.cpp
+++ b/lib/Target/RISCV/RISCVLDBackend.cpp
@@ -313,7 +313,7 @@ bool RISCVLDBackend::doRelaxationCall(Relocation *reloc) {
                  ->getInput()
                  ->decoratedPath();
 
-    region->replaceInstruction(offset, reloc, c_j, 2);
+    region->replaceInstruction(offset, reloc, reinterpret_cast<uint8_t *>(&c_j), 2);
     reloc->setTargetData(c_j);
     reloc->setType(llvm::ELF::R_RISCV_RVC_JUMP);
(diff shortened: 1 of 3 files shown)


lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -669,7 +669,9 @@ bool RISCVLDBackend::doRelaxationTLSDESC(Relocation &R, bool Relax) {
     else {
       // Otherwise, the instruction is replaced with a NOP.
       reportMissedRelaxation(RelaxType, *region, offset, 4, Sym.name());
-      region->replaceInstruction(offset, &R, NOP, 4);
+      uint32_t NOPi32 = static_cast<uint32_t>(NOP);
+      region->replaceInstruction(
+          offset, &R, reinterpret_cast<uint8_t *>(&NOPi32), 4);
```
quic-seaswara, line 674: `reinterpret_cast<uint8_t *>(NOP) ? `

parth-07: We cannot do this because `NOP` is an enum value and we need an address. We also cannot use `&NOP` because `NOP` is an enum value and it is converted to a temporary integer in this context and we cannot take address of a temporary.

PR #950, opened 2026-03-18
Expand =/$SYSROOT for INPUT/GROUP script commands

### LINKED ISSUE #927: `=` and `$SYSROOT` in `INPUT`|`GROUP` command does not expand to SYSROOT
eld does not expand `=` and `$SYSROOT` in `INPUT` and `GROUP` commands to sysroot. 

Reproducer:

```bash
#!/usr/bin/env bash

mkdir -p A/lib64
mkdir -p A/script

cat >1.c <<\EOF
int foo() { return 1; }
EOF

cat >script.t <<\EOF
// or $SYSROOT instead of =
INPUT(=/lib64/lib1.so)
EOF

cat >main.c <<\EOF
int main() { return 0; }
EOF

mv lib1.so A/lib64/lib1.so

ld.eld -o 1.eld.out --sysroot=A main.o -T script.t 
ld.lld -o 1.lld.out --sysroot=A main.o -T script.t 
ld.bfd -o 1.bfd.out --sysroot=A main.o -T script.t  
```

parth-07:
This commit improves the script parser to expand
the = and $SYSROOT prefix in INPUT/GROUP inputs.

Resolves #927

### DIFF
--- a/lib/Input/Input.cpp
+++ b/lib/Input/Input.cpp
@@ -85,6 +85,25 @@ bool Input::isPathValid(const std::string &Path) const {
   return true;
 }
 
+std::string Input::expandSysrootMarkers(llvm::StringRef Name,
+                                        const SearchDirs &PSearchDirs,
+                                        DiagnosticEngine &DiagEngine) {
+  llvm::StringRef Suffix;
+  if (Name.starts_with("="))
+    Suffix = Name.substr(1);
+  else if (Name.starts_with("$SYSROOT"))
+    Suffix = Name.substr(strlen("$SYSROOT"));
+  else
+    return Name.str();
+
+  std::string ExpandedPath = Suffix.str();
+  if (PSearchDirs.hasSysRoot())
+    ExpandedPath = (PSearchDirs.sysroot().native() + Suffix).str();
+
+  DiagEngine.raise(Diag::verbose_sysroot_expansion) << Name << ExpandedPath;
+  return ExpandedPath;
+}
+
 /// \return True if path able to be resolved, otherwise false
 bool Input::resolvePath(const LinkerConfig &PConfig) {
   if (ResolvedPath)
@@ -104,9 +123,14 @@ bool Input::resolvePath(const LinkerConfig &PConfig) {
     }
   }
   auto &PSearchDirs = PConfig.directories();
+
+  std::string ExpandedFileName = FileName;
+  if (Type == Input::InputType::Script || Type == Input::InputType::Default)
+    ExpandedFileName = expandSysrootMarkers(FileName, PSearchDirs, *DiagEngine);
+
   switch (Type) {
   default:
-    ResolvedPath = eld::sys::fs::Path(FileName);
+    ResolvedPath = eld::sys::fs::Path(ExpandedFileName);
     break;
   case Input::Internal:
     ResolvedPath = eld::sys::fs::Path(FileName);
@@ -115,11 +139,11 @@ bool Input::resolvePath(const LinkerConfig &PConfig) {
   if (Type == Input::Script) {
     if (shouldPrependSysrootToScriptInput(PConfig)) {
       ResolvedPath = PSearchDirs.sysroot();
-      ResolvedPath->append(FileName);
+      ResolvedPath->append(ExpandedFileName);
     }
     if (!llvm::sys::fs::exists(ResolvedPath->native())) {
-      const sys::fs::Path *P =
-          PSearchDirs.find(FileName, SearchDirs::SearchInputType::Script);
+      const sys::fs::Path *P = PSearchDirs.find(
+          ExpandedFileName, SearchDirs::SearchInputType::Script);
       if (P != nullptr)
         ResolvedPath = *P;
     }
--- a/lib/Input/InputAction.cpp
+++ b/lib/Input/InputAction.cpp
@@ -44,10 +44,15 @@ InputFileAction::InputFileAction(std::string Name,
     : InputAction(K, Printer), Name(Name), I(nullptr) {}
 
 bool InputFileAction::activate(InputBuilder &PBuilder) {
+  const LinkerConfig &Config = PBuilder.getLinkerConfig();
+  std::string ExpandedName = Input::expandSysrootMarkers(
+      Name, Config.directories(), *Config.getDiagEngine());
+
   std::ifstream Is;
-  Is.open(Name.c_str(), std::ifstream::in);
+  Is.open(ExpandedName.c_str(), std::ifstream::in);
   if (!Is.good()) {
-    PBuilder.getDiagEngine()->raise(Diag::fatal_cannot_read_input) << Name;
+    PBuilder.getDiagEngine()->raise(Diag::fatal_cannot_read_input)
+        << ExpandedName;
     return false;
   }
   Is.close();
--- a/lib/ScriptParser/ScriptParser.cpp
+++ b/lib/ScriptParser/ScriptParser.cpp
@@ -745,9 +745,10 @@ void ScriptParser::addFile(StringRef Name) {
   if (Name.consume_front("-l"))
     InputStrTok = ThisScriptFile.createNameSpecToken(Name.str(),
                                                      ThisScriptFile.asNeeded());
-  else
+  else {
     InputStrTok =
         ThisScriptFile.createFileToken(Name.str(), ThisScriptFile.asNeeded());
+  }
   ThisScriptFile.getCurrentStringList()->pushBack(InputStrTok);
 }
 
--- a/test/Common/standalone/linkerscript/SysrootExpansion/Inputs/group_equals.t
+++ b/test/Common/standalone/linkerscript/SysrootExpansion/Inputs/group_equals.t
@@ -1,0 +1,1 @@
+GROUP(=/lib64/lib1.so)
--- a/test/Common/standalone/linkerscript/SysrootExpansion/SysrootExpansion.test
+++ b/test/Common/standalone/linkerscript/SysrootExpansion/SysrootExpansion.test
@@ -1,0 +1,53 @@
+#---SysrootExpansion.test---------------- Linker Script ----------------#
+#BEGIN_COMMENT
+# Validate sysroot expansion for "=" and "$SYSROOT" markers in INPUT/GROUP
+# script commands and command-line object files.
+#END_COMMENT
+#START_TEST
+
+RUN: rm -rf %t.dir && mkdir -p %t.dir/lib64
+RUN: %clang %clangopts -o %t.1.o %p/Inputs/1.c -c -fPIC
+RUN: %clang %clangopts -o %t.main.o %p/Inputs/main.c -c
+
+# Create shared library in sysroot
+RUN: %link %linkopts -o %t.dir/lib64/lib1.so -shared %t.1.o
+
+# Test INPUT with = prefix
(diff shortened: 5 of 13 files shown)


lib/ScriptParser/ScriptParser.cpp
```diff
@@ -739,14 +739,31 @@ void ScriptParser::readAsNeeded() {
...
   if (Name.consume_front("-l"))
     InputStrTok = ThisScriptFile.createNameSpecToken(Name.str(),
                                                      ThisScriptFile.asNeeded());
-  else
+  else {
+    std::string ExpandedName = expandSysrootMarkers(Name);
     InputStrTok =
-        ThisScriptFile.createFileToken(Name.str(), ThisScriptFile.asNeeded());
+        ThisScriptFile.createFileToken(ExpandedName, ThisScriptFile.asNeeded());
+  }
```
quic-seaswara, line 751: This needs to be done in resolvePath 

For example 

ld.bfd --sysroot=/tmp/xxx/ =main.o

I could not find this expansion of $SYSROOT for input files, may be I am not using the right version of ld.

lib/ScriptParser/ScriptParser.cpp
```diff
@@ -739,14 +739,31 @@ void ScriptParser::readAsNeeded() {
   ThisScriptFile.setAsNeeded(false);
 }
 
+std::string ScriptParser::expandSysrootMarkers(StringRef Name) const {
+  StringRef Suffix;
+  if (Name.starts_with("="))
+    Suffix = Name.substr(1);
+  else if (Name.starts_with("$SYSROOT"))
+    Suffix = Name.substr(8); // strlen("$SYSROOT") == 8
```
quic-seaswara: Name.substr(strlen("$SYSROOT")) or create a function to get the suffix.

test/Common/standalone/linkerscript/SysrootExpansion/Inputs/group_equals.t
```diff
@@ -0,0 +1 @@
+GROUP(=/lib64/lib1.so)
```
quic-seaswara, line 1: Can you add GROUP($SYSROOT) as well?

test/Common/standalone/linkerscript/SysrootExpansion/SysrootExpansion.test
```diff
@@ -0,0 +1,23 @@
...
+
+RUN: rm -rf %t.dir && mkdir -p %t.dir/lib64
+RUN: %clang %clangopts -o %t.1.o %p/Inputs/1.c -c -fPIC
+RUN: %clang %clangopts -o %t.main.o %p/Inputs/main.c -c
+RUN: %link %linkopts -o %t.dir/lib64/lib1.so -shared %t.1.o
+RUN: %link %linkopts -o %t.out1 --sysroot=%t.dir -T %p/Inputs/script_equals.t %t.main.o --verbose 2>&1 | %filecheck %s --check-prefix=EQUALS
+RUN: %link %linkopts -o %t.out2 --sysroot=%t.dir -T %p/Inputs/script_sysroot.t %t.main.o --verbose 2>&1 | %filecheck %s --check-prefix=SYSROOT
+RUN: %link %linkopts -o %t.out3 --sysroot=%t.dir -T %p/Inputs/group_equals.t %t.main.o --verbose 2>&1 | %filecheck %s --check-prefix=GROUP
+RUN: %not %link %linkopts -o %t.out4 -T %p/Inputs/script_equals.t %t.main.o 2>&1 | %filecheck %s --check-prefix=NOSYSROOT
+#END_TEST
```
quic-seaswara, line 43: Need verbose output to show how the files are resolved.

Also please annotate map file to show what files are being expanded.

lib/ScriptParser/ScriptParser.cpp
```diff
@@ -739,14 +739,31 @@ void ScriptParser::readAsNeeded() {
...
+  auto &SearchDirs = ThisConfig.directories();
+  if (SearchDirs.hasSysRoot())
+    return (SearchDirs.sysroot().native() + Suffix).str();
+  return Suffix.str();
+}
+
 void ScriptParser::addFile(StringRef Name) {
   StrToken *InputStrTok = nullptr;
   if (Name.consume_front("-l"))
     InputStrTok = ThisScriptFile.createNameSpecToken(Name.str(),
```
quic-seaswara, line 746: What does ld for namespec files ?

quic-seaswara (changes requested):
This is very cool feature!

lib/ScriptParser/ScriptParser.cpp
parth-07: > ld.bfd --sysroot=/tmp/xxx/ =main.o

Thank you for adding this comment. I did not know GNU ld supported this. GNU ld docs does not mention _anywhere_ that `=` and `$SYSROOT` are expanded for filenames specified directly in the link command. In my local experiments with GNU ld 2.38, I am able to reproduce SYSROOT expansion for both `=main.o` and `$SYSROOT/main.o`. I am updating the patch to include this functionality.

parth-07: Done. I have updated the documentation sysroot.rst.in as well.

lib/ScriptParser/ScriptParser.cpp
parth-07: Done.

test/Common/standalone/linkerscript/SysrootExpansion/SysrootExpansion.test
parth-07: I have added the verbose output for the sysroot marker expansion. Can you please help with how we should annotate the sysroot marker expansion in the map-file? Should we add a new section in the map-file with all files that had sysroot markers and their corresponding expansions?

lib/ScriptParser/ScriptParser.cpp
parth-07: I could not find any difference in interpretation for namespec files with explicit sysroot in GNU ld.

test/Common/standalone/linkerscript/SysrootExpansion/Inputs/group_equals.t
parth-07: Done.

test/Common/standalone/linkerscript/SysrootExpansion/SysrootExpansion.test
quic-seaswara: We can use the map file that shows LOAD <file> with

`LOAD <user passed in file> #expanded file information`

quic-seaswara: Can be a follow up

lib/Input/InputAction.cpp
```diff
@@ -44,10 +44,15 @@ InputFileAction::InputFileAction(std::string Name,
     : InputAction(K, Printer), Name(Name), I(nullptr) {}
 
 bool InputFileAction::activate(InputBuilder &PBuilder) {
+  const LinkerConfig &Config = PBuilder.getLinkerConfig();
+  std::string ExpandedName = Input::expandSysrootMarkers(
+      Name, Config.directories(), *Config.getDiagEngine());
```
quic-seaswara, line 49: Please open an issue to support these with reproduce.

lib/Input/Input.cpp
```diff
@@ -103,10 +129,11 @@ bool Input::resolvePath(const LinkerConfig &PConfig) {
   if (Type == Input::Script) {
     if (shouldPrependSysrootToScriptInput(PConfig)) {
       ResolvedPath = PSearchDirs.sysroot();
-      ResolvedPath->append(FileName);
+      ResolvedPath->append(ExpandedFileName);
     }
     if (!llvm::sys::fs::exists(ResolvedPath->native())) {
-      const sys::fs::Path *P = PSearchDirs.find(FileName, Input::Script);
+      const sys::fs::Path *P =
```
quic-seaswara: Do we want to have it const ?

lib/Input/Input.cpp
```diff
@@ -85,16 +85,42 @@ bool Input::isPathValid(const std::string &Path) const {
...
+                                        DiagnosticEngine &DiagEngine) {
+  llvm::StringRef Suffix;
+  if (Name.starts_with("="))
+    Suffix = Name.substr(1);
+  else if (Name.starts_with("$SYSROOT"))
+    Suffix = Name.substr(strlen("$SYSROOT"));
+  else
+    return Name.str();
+
+  std::string ExpandedPath;
```
quic-seaswara: Default ExpandedPath = suffix.str()

lib/Input/Input.cpp
```diff
@@ -85,16 +85,42 @@ bool Input::isPathValid(const std::string &Path) const {
...
 }
 
+std::string Input::expandSysrootMarkers(llvm::StringRef Name,
+                                        const SearchDirs &PSearchDirs,
+                                        DiagnosticEngine &DiagEngine) {
+  llvm::StringRef Suffix;
+  if (Name.starts_with("="))
+    Suffix = Name.substr(1);
+  else if (Name.starts_with("$SYSROOT"))
+    Suffix = Name.substr(strlen("$SYSROOT"));
```
quic-seaswara, line 95: Is $SYSROOT a reserved word ?

What about linker script file patterns ? Does that support = ?

test/Common/standalone/linkerscript/SysrootExpansion/SysrootExpansion.test
parth-07: I will make the map-file change as a follow-up.

parth-07: I have created an issue to track this: https://github.com/qualcomm/eld/issues/971

lib/Input/Input.cpp
parth-07: It has to be `const` `PSearchDirs.find` returns `const sys::fs::Path *`.

lib/Input/Input.cpp
parth-07: Done.

lib/Input/Input.cpp
parth-07: > Is $SYSROOT a reserved word ?

I could not find any GNU documentation about `$SYSROOT` being a reserved word in the linker script.

> What about linker script file patterns ? Does that support = ?

I have verified that `=` and `$SYSROOT` are not expanded when used in the linker script file patterns.

lib/Input/InputAction.cpp
parth-07: I have created an issue to track this: #972

lib/Input/Input.cpp
quic-seaswara: Does it make sense to support this for linker script inputs to be consistent ? can be a follow up ?

quic-seaswara (approved):

lib/Input/Input.cpp
parth-07: I don't think we should support this in linker script input description patterns, unless there is a customer need for it. One key difference is that input description are taken as string patterns and on the other hand the cases where sysroot prefixes are expanded are taken as non-pattern strings.

quic-seaswara (approved):

quic-seaswara (approved):

PR #955, opened 2026-03-19
Switch Zephyr nightly workflow to self-hosted runner

quic-areg:


.github/workflows/zephyr-nightly.yml
```diff
@@ -156,7 +157,7 @@ jobs:
           ./scripts/twister -W -p qemu_riscv32 \
             --inline-logs \
             --timestamps \
-            -j 2 \
+            -j 40 \
```
quic-seaswara: Can we drop -j 40 ?

https://docs.zephyrproject.org/latest/develop/test/twister.html

-j JOBS, --jobs JOBS  Number of jobs for building, defaults to number of CPU
                        threads

It would just use the number of CPU's on the box as the default.

quic-seaswara (approved):

.github/workflows/zephyr-nightly.yml
```diff
@@ -22,30 +23,16 @@ jobs:
         with:
           path: llvm-project/llvm/tools/eld
 
-      - name: Record pre-build entry
```
quic-shaksuma, line 25: Would this not run on self-hosted runner?

PR #964, opened 2026-03-25
Explicity include .gcc_except_table

quic-k:
Implicity it is getting placed above .init_array, we want .ctors and .init_array to be continuous and this section breaks the conitnuity.

### DIFF
--- a/templates/staticExecutable/static-executable-h2-picolibc.lcs.template
+++ b/templates/staticExecutable/static-executable-h2-picolibc.lcs.template
@@ -35,6 +35,9 @@ SECTIONS
 
   /* Need to pre-align so that the symbols come after padding */
   . = ALIGN(8);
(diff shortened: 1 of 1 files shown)


templates/staticExecutable/static-executable-h2-picolibc.lcs.template
```diff
@@ -100,6 +103,10 @@ SECTIONS
     KEEP (*(.eh_frame))
     __eh_frame_end = .;
   }
+  .except_ordered :
```
parth-07: Nitpick: is there a reason to not use `.gcc_except_table` as the output section name here? The output section name would be visible in the output elf file.

quic-k: hm, I picked .except_ordered referring picolibcpp.ld but probably .gcc_except_table would be better

PR #966, opened 2026-03-25
[sanitize] Dont rely on host tools for eld tests

quic-seaswara:


.github/workflows/sanitize.yml
```diff
@@ -11,17 +11,30 @@ permissions:
   contents: write
 jobs:
   build-and-runtest-with-asan-ubsan:
-    runs-on: self-hosted
+    runs-on: ubuntu-latest
```
quic-shaksuma, line 14: didn't we want to run sanitizer builds on self-hosted?

quic-seaswara: Yes, found a way to just build eld that will allow to test sanitizer builds with latest llvm

PR #981, opened 2026-03-30
[RISCV] Add support for the Zcmt Extension

svs-quic:
This patch adds support for the RISCV Zcmt Compressed Table Jumps Extension. 

Table jump allows the linker to replace the following instruction sequences with a `CM.JT` or `CM.JALT` encoding, and an entry in the table:

- 32-bit J calls
- 32-bit JAL ra calls
- 64-bit AUIPC+JR calls to fixed locations
- 64-bit AUIPC+JALR ra calls to fixed locations

More information about the extension can be found [here](https://riscv.github.io/riscv-isa-manual/snapshot/spec/#ext:zcmt).


include/eld/Driver/RISCVLinkerOptions.td
```diff
@@ -58,6 +58,11 @@ def no_relax_tlsdesc
 def riscv_relax : Flag<["--"], "relax">,
                   HelpText<"Enable relaxation (default behavior)">,
                   Group<grp_riscv_linker>;
+
+def relax_tbljal
+    : Flag<["--"], "relax-tbljal">,
+      HelpText<"Enable conversion of call/jump instructions to Zcmt table jump instructions (cm.jt/cm.jalt)">,
+      Group<grp_riscv_linker>;
```
lenary, line 65: Can we have both positive and negative versions of this flag for users? That allows us to change the default in the future (this is what I did for the xqci relaxation flag)

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -135,6 +151,26 @@ void RISCVLDBackend::initPatchSections(ELFObjectFile &InputFile) {
...
+
+  if (TableJumpFragment) {
+    std::string JvtName = "__jvt_base$";
+    LDSymbol *JvtBase =
+        m_Module.getIRBuilder()
+            ->addSymbol<IRBuilder::Force, IRBuilder::Resolve>(
+                m_Module.getInternalInput(Module::InternalInputType::Sections),
+                JvtName, ResolveInfo::NoType, ResolveInfo::Define,
+                ResolveInfo::Global,
+                0x0, // size
```
lenary: I don't know if we should be setting the size here. It might help people trying to debug things, i don't think loaders will look at this information.

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -1004,6 +1134,9 @@ void RISCVLDBackend::mayBeRelax(int relaxation_pass, bool &pFinished) {
     return;
   }
 
+  if (relaxation_pass == RELAXATION_CALL)
+    initTableJump();
```
lenary: Given that `initTableJump()` does a full scan, can we make it explicitly a different phase, before `RELAXATION_CALL`? maybe `RELAXATION_TABLE_JUMP`?

lib/Target/RISCV/RISCVTableJump.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+static void
+addEntry(llvm::DenseMap<const ResolveInfo *, RISCVTableJumpEntry> &M,
+         const ResolveInfo *Sym, int Saved) {
+  auto &E = M[Sym];
+  E.Saved += Saved;
+}
+
+static int64_t getCallDisplace(RISCVLDBackend &Backend, const Relocation &R) {
+  int64_t S = static_cast<int64_t>(Backend.getSymbolValuePLT(R));
```
lenary, line 52: I think we have to jump to the PLT entry if there is one (including e.g. for patching), but it would be good to confirm that this code is right.

The jump table is sort-of like a PLT, but it cannot be lazy and there's no preemption for its addresses.

lib/Target/RISCV/RISCVTableJump.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+    uint32_t Insn = 0;
+    uint8_t Rd = 0;
+    if (R->type() == llvm::ELF::R_RISCV_JAL) {
+      R->targetRef()->memcpy(&Insn, sizeof(Insn), 0);
+      Rd = (Insn >> 7) & 0x1f;
+    } else {
+      // CALL/CALL_PLT: read the following JALR instruction to get rd.
+      uint32_t Jalr = 0;
+      R->targetRef()->memcpy(&Jalr, sizeof(Jalr), 4);
```
lenary: I don't know why this side uses a local `Jalr` uint32 variable, when the other side uses `Insn`. Both could use something local or `Insn`?

lib/Target/RISCV/RISCVTableJump.cpp
```diff
@@ -0,0 +1,207 @@
...
+void RISCVTableJumpFragment::scanTableJumpEntries(ELFSection &Sec) {
+  if (!Sec.isCode())
+    return;
+
+  auto &Relocs = Sec.getRelocations();
+  for (size_t I = 0; I < Relocs.size(); ++I) {
+    const Relocation *R = Relocs[I];
+    if (R->type() != llvm::ELF::R_RISCV_JAL &&
+        R->type() != llvm::ELF::R_RISCV_CALL &&
+        R->type() != llvm::ELF::R_RISCV_CALL_PLT)
```
lenary, line 73: What about `R_RISCV_QC_E_CALL_PLT`? :) A planned follow-up or an oversight?

lib/Target/RISCV/RISCVTableJump.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+static void
+writeEntries(RISCVLDBackend &Backend, uint8_t *Buf,
+             const llvm::DenseMap<const ResolveInfo *, RISCVTableJumpEntry>
+                 &Candidates) {
+  llvm::SmallVector<std::pair<const ResolveInfo *, RISCVTableJumpEntry>, 0>
+      Entries(Candidates.begin(), Candidates.end());
+  llvm::sort(Entries, [](const auto &A, const auto &B) {
+    return A.second.Index < B.second.Index;
+  });
```
lenary, line 243: IIRC, we may want these sorted in a specific order. We should confirm with the architects if there's a performance implication.

lenary (commented):
Some comments here, I'm not sure if any of these are a surprise but maybe they point at the next steps.

lib/Target/RISCV/RISCVLDBackend.cpp
svs-quic: Do you want me to completely remove it? I've set it to the table size for now.

lib/Target/RISCV/RISCVLDBackend.cpp
svs-quic: I've added RELAXATION_TABLE_JUMP before the CALL.

lib/Target/RISCV/RISCVTableJump.cpp
svs-quic: I think I have addressed this in the latest patch

lib/Target/RISCV/RISCVTableJump.cpp
svs-quic: Yeah I plan to add the Xqci changes in a follow-up.

lib/Target/RISCV/RISCVTableJump.cpp
svs-quic: Yeah both can use Insn. I dont know why I did this.

lib/Target/RISCV/RISCVTableJump.cpp
svs-quic: I checked with Albert and he wants it sorted by frequency of the addresses called.

include/eld/Driver/RISCVLinkerOptions.td
svs-quic: I've added the negative version of the flag.

lenary:
I don't think I have any more comments.

svs-quic:
@quic-seaswara @parth-07 I think this patch is now ready for review. @lenary had an initial set of comments that I have addressed.

lib/Target/RISCV/RISCVTableJump.cpp
```diff
@@ -0,0 +1,219 @@
...
+  return S + A - P;
+}
+
+static uint64_t getTableJumpTargetVA(RISCVLDBackend &Backend,
+                                     const ResolveInfo *Sym) {
+  return Backend.getSymbolValuePLT(*Sym);
+}
+
+void RISCVTableJumpFragment::scanTableJumpEntries(ELFSection &Sec) {
+  if (!Sec.isCode())
```
parth-07: I believe we can remove this if-condition or make it an assert because this function is only called if `Sec.isCode()` is true.

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -73,6 +75,20 @@ Relocation::Address RISCVLDBackend::getSymbolValuePLT(const Relocation &R) {
   return getRelocator()->getSymValue(&R);
 }
 
+Relocation::Address RISCVLDBackend::getSymbolValuePLT(const ResolveInfo &Sym) {
+  ResolveInfo *MutableSym = const_cast<ResolveInfo *>(&Sym);
```
parth-07: Instead of doing `const_cast` for removing const-ness, can we directly pass `ResolveInfo` as non-const?

lib/LinkerWrapper/RISCVLinkDriver.cpp
```diff
@@ -139,6 +139,11 @@ RISCVLinkDriver::parseOptions(ArrayRef<const char *> Args,
   if (ArgList.hasArg(OPT_RISCVLinkOptTable::no_relax_tlsdesc))
     Config.options().setRISCVRelaxTLSDESC(false);
 
+  // --relax-tbljal, --no-relax-tbljal (default)
+  Config.options().setRISCVRelaxTbljal(
```
parth-07, line 143: We should report warning/error if we are building a shared library with --relax-tbljal enabled.

lib/Target/RISCV/RISCVTableJump.cpp
```diff
@@ -0,0 +1,219 @@
...
+                     /*Align=*/64, /*Size=*/0),
+      Backend(B) {}
+
+size_t RISCVTableJumpFragment::size() const { return ThisSize; }
+
+eld::Expected<void> RISCVTableJumpFragment::emit(MemoryRegion &Mr, Module &M) {
+  if (size() == 0)
+    return {};
+  uint8_t *Buf = Mr.begin() + getOffset(M.getConfig().getDiagEngine());
+  memset(Buf, 0, size());
```
parth-07, line 39: Do we need to explicitly set the buffer to 0 here? Wouldn't `writeTo` overwrite the complete [buffer, buffer + size() ) area?

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -1005,6 +1152,12 @@ void RISCVLDBackend::mayBeRelax(int relaxation_pass, bool &pFinished) {
     return;
   }
 
+  if (relaxation_pass == RELAXATION_TABLE_JUMP) {
+    initTableJump();
```
parth-07: It is not very intuitive to run `initTableJump` as part of a relaxation pass as it is not doing any relaxation by itself. What do you think about adding a new virtual function `GNULDBackend::preRelaxation()` which the backends can override to do any setup/initialization required for the relaxations?

lib/Target/RISCV/RISCVTableJump.cpp
```diff
@@ -0,0 +1,219 @@
...
+    SavedBoth += KV.second.Saved;
+  }
+  for (auto &KV : CMJALTCandidates)
+    SavedBoth += KV.second.Saved;
+
+  // Using cm.jalt requires padding the cm.jt region to 32 entries.
+  if (!CMJALTCandidates.empty() && SavedBoth < SavedCMJTOnly)
+    CMJALTCandidates.clear();
+
+  if (SavedCMJTOnly <= 0) {
```
parth-07:
```suggestion
  if (SavedCMJTOnly < 0) {
```

I believe this should be less-than instead of less-than-or-equal-to. If `SavedBoth > SavedCMJTOnly` and  `SavedCMJTOnly` is `0`, then we should convert `jalr` -> `cm.jalt`. However, due to the equal-to in the condition here, we are clearing `CMJALTCandidates` as well.

Reproducer:

```bash
#!/usr/bin/env bash

cat >1.s <<\EOF
  .section .text.foo,"ax",%progbits
  .global foo
  .type foo, %function
foo:
  nop
  .size foo, .-foo

  .section .text.bar,"ax",%progbits
  .global bar
  .type bar, %function
bar:
  nop
  .size bar, .-bar

  .section .text,"ax",%progbits
  .global _start
  .type _start, %function
_start:
  .rept 24 
  call foo
  call bar
  .endr
EOF

cat >script.t <<\EOF
SECTIONS {
  foo (0x800000) : {
    *(.text.foo)
  }
  bar (0x900000) : {
    *(.text.bar)
  }
  text (0x0) : {
    *(.text*)
  }
}
EOF


llvm-mc -o 1.o 1.s -triple=riscv32 -filetype=obj -mattr=+relax,+zcmt
ld.eld -o 1.out 1.o --relax-tbljal -T script.t
```

We should relax the call instructions `call foo` and `call bar` to `cm.jalt` but we are not due to the equal-to in the condition.

parth-07 (changes requested):

lib/Target/RISCV/RISCVLDBackend.cpp
svs-quic: I have made the change to pass it as non-const

lib/Target/RISCV/RISCVLDBackend.cpp
svs-quic: Yeah thanks for the suggestion. I have implemented the same.

lib/Target/RISCV/RISCVTableJump.cpp
svs-quic: Hmm yeah I guess we don't. Have removed it.

lib/Target/RISCV/RISCVTableJump.cpp
svs-quic: Thanks I've changed it into an assert.

lib/Target/RISCV/RISCVTableJump.cpp
svs-quic: Thanks for catching this Parth. I have made the change and also added a test for this.

lib/LinkerWrapper/RISCVLinkDriver.cpp
svs-quic: Thanks I have added one now

lib/Target/RISCV/RISCVTableJump.cpp
parth-07: I now think the `memset` here call was serving a purpose. `writeTo` call does not 0-initialize the unused CMJT entries, and thus they may contain random values. 0-initialization would be helpful for the unused CMJT entries.

parth-07 (approved):
LGTM.

lib/Target/RISCV/RISCVTableJump.cpp
svs-quic: Added it back.

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -135,6 +164,26 @@ void RISCVLDBackend::initPatchSections(ELFObjectFile &InputFile) {
 void RISCVLDBackend::initTargetSymbols() {
   if (config().codeGenType() == LinkerConfig::Object)
     return;
+
+  if (TableJumpFragment) {
+    std::string JvtName = "__jvt_base$";
```
quic-seaswara, line 169: Can you give me a pointer to the ABI document on how Zcmt works ?

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -157,6 +206,50 @@ void RISCVLDBackend::initTargetSymbols() {
...
+      continue;
+    for (auto &Rs : ObjFile->getRelocationSections()) {
+      if (Rs->isIgnore() || Rs->isDiscard())
+        continue;
+      ELFSection *Linked = Rs->getLink();
+      if (!Linked || !Linked->isCode())
+        continue;
+      if (!Visited.insert(Linked).second)
+        continue;
+      TableJumpFragment->scanTableJumpEntries(*Linked);
```
quic-seaswara, line 235: can this be handled in RISCVLDBackend::handleRelocation ?

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -321,6 +414,25 @@ bool RISCVLDBackend::doRelaxationCall(Relocation *reloc) {
     return true;
   }
 
+  if (config().options().getRISCVRelaxTbljal() && TableJumpFragment &&
+      TableJumpFragment->size() != 0) {
```
quic-seaswara: remove != 0 ? 

Just TableJumpFragment->size()  is good enough.

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -321,6 +414,25 @@ bool RISCVLDBackend::doRelaxationCall(Relocation *reloc) {
...
+    int EntryIndex = -1;
+    if (rd == 0)
+      EntryIndex = TableJumpFragment->getCMJTEntryIndex(reloc->symInfo());
+    else if (rd == 1)
+      EntryIndex = TableJumpFragment->getCMJALTEntryIndex(reloc->symInfo());
```
quic-seaswara: adding a comment would be useful on what is being done here.

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -321,6 +414,25 @@ bool RISCVLDBackend::doRelaxationCall(Relocation *reloc) {
...
+    if (EntryIndex >= 0) {
```
quic-seaswara, line 444: if (EntryIndex)

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -359,6 +471,39 @@ bool RISCVLDBackend::doRelaxationCall(Relocation *reloc) {
...
+  reloc->targetRef()->memcpy(&Jal, sizeof(Jal), 0);
+  unsigned rd = (Jal >> 7) & 0x1fu;
+
+  int EntryIndex = -1;
+  if (rd == 0)
+    EntryIndex = TableJumpFragment->getCMJTEntryIndex(reloc->symInfo());
+  else if (rd == 1)
+    EntryIndex = TableJumpFragment->getCMJALTEntryIndex(reloc->symInfo());
+  if (EntryIndex < 0)
+    return false;
```
quic-seaswara, line 511: This seems to have duplicate code (see lines above 417)

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -982,14 +1127,16 @@ void RISCVLDBackend::translatePseudoRelocation(Relocation *reloc) {
...
-  RELAXATION_CALL, // Must start at zero
+  RELAXATION_CALL, // Must start at zero.
   RELAXATION_PC,
   RELAXATION_LUI,
   RELAXATION_TLSDESC,
   RELAXATION_ALIGN,
   RELAXATION_PASS_COUNT, // Number of passes
 };
 
+void RISCVLDBackend::preRelaxation() { initTableJump(); }
```
quic-seaswara: Do we need this ? With the change that I mentioned above by moving to handleRelocation you can probably remove this ?

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -1043,6 +1190,11 @@ void RISCVLDBackend::mayBeRelax(int relaxation_pass, bool &pFinished) {
             doRelaxationCall(relocation);
           break;
         }
+        case llvm::ELF::R_RISCV_JAL: {
+          if (nextRelax && relaxation_pass == RELAXATION_CALL)
+            doRelaxationJal(relocation);
```
quic-seaswara, line 1212: Does this work for pcrel calls too ?

lib/Target/RISCV/RISCVRelocationInfo.cpp
```diff
@@ -697,6 +697,16 @@ RISCVRelocationMap RISCVRelocs = {
       /*.VerifyAlignment = */ false,
       /*.Signed = */ true,
       /*.Size = */ 32}},
+    {eld::ELF::riscv::internal::R_RISCV_TBJAL,
+     {/*.Name = */ "R_RISCV_TBJAL",
+      /*.Type = */ eld::ELF::riscv::internal::R_RISCV_TBJAL,
+      /*EncodingType = */ EncTy_None,
```
quic-seaswara, line 705: Why is this EncTy_None ?

quic-seaswara (commented):
some initial comments.

lib/Target/RISCV/RISCVLDBackend.cpp
svs-quic: https://docs.riscv.org/reference/isa/unpriv/zc.html#Zcmt and https://riscv-non-isa.github.io/riscv-elf-psabi-doc/#_table_jump_relaxation should help.

lib/Target/RISCV/RISCVLDBackend.cpp
svs-quic: I dont think we can do this. The jumpTable relaxation depends  on final symbol/PLT addresses and RELAX-paired relocation ordering, which are not stable while relocations are being created in handleRelocation()

lib/Target/RISCV/RISCVLDBackend.cpp
svs-quic: thanks yeah I've removed this.

lib/Target/RISCV/RISCVLDBackend.cpp
svs-quic: i've abstracted this code into a separate function and added a comment.

lib/Target/RISCV/RISCVLDBackend.cpp
svs-quic: EntryIndex can be zero so I think we need the check,

lib/Target/RISCV/RISCVLDBackend.cpp
svs-quic: i've abstracted this code into a separate function and added a comment.

lib/Target/RISCV/RISCVRelocationInfo.cpp
svs-quic: The relaxation pass writes the final encoding making the relocation applyNone so no encoding is needed.

lib/Target/RISCV/RISCVLDBackend.cpp
svs-quic: No the jump table stores absolute addresses and hence the Zcmt instructions dont work with pcrel calls.

lib/LinkerWrapper/RISCVLinkDriver.cpp
```diff
@@ -139,6 +139,16 @@ RISCVLinkDriver::parseOptions(ArrayRef<const char *> Args,
...
+      ArgList.hasFlag(OPT_RISCVLinkOptTable::relax_tbljal,
+                      OPT_RISCVLinkOptTable::no_relax_tbljal, false));
+  if (Config.options().getRISCVRelaxTbljal() &&
+      ArgList.getLastArg(OPT_RISCVLinkOptTable::shared)) {
+    Config.raise(Diag::warn_incompatible_option) << "relax-tbljal" << "shared";
+    Config.options().setRISCVRelaxTbljal(false);
+  }
```
quic-seaswara: can this be a seperate PR ? I feel this needs to be an error if the option is incompatible.

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -114,6 +129,20 @@ void RISCVLDBackend::initTargetSections(ObjectBuilder &pBuilder) {
...
+        Module::InternalInputType::Sections, LDFileFormat::Internal,
+        ".riscv.jvt", llvm::ELF::SHT_PROGBITS, llvm::ELF::SHF_ALLOC,
+        /*Align=*/64);
+    TableJumpFragment =
+        make<RISCVTableJumpFragment>(*this, m_pRISCVTableJumpSection);
+    m_pRISCVTableJumpSection->addFragmentAndUpdateSize(TableJumpFragment);
+    m_pRISCVTableJumpSection->setWanted(true);
+    // Keep the section in the output even if its final size ends up being 0
+    // (matching lld behavior and the upstream tests).
+    m_pRISCVTableJumpSection->setWantedInOutput(true);
```
quic-seaswara: Why emit this if size is 0 ?

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -114,6 +129,20 @@ void RISCVLDBackend::initTargetSections(ObjectBuilder &pBuilder) {
   if (LinkerConfig::Object == config().codeGenType())
     return;
 
+  if (config().options().getRISCVRelaxTbljal()) {
+    m_pRISCVTableJumpSection = m_Module.createInternalSection(
+        Module::InternalInputType::Sections, LDFileFormat::Internal,
```
quic-seaswara: create a seperate internal input file for this category ?

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -114,6 +129,20 @@ void RISCVLDBackend::initTargetSections(ObjectBuilder &pBuilder) {
...
+        ".riscv.jvt", llvm::ELF::SHT_PROGBITS, llvm::ELF::SHF_ALLOC,
+        /*Align=*/64);
```
quic-seaswara, line 136: Why 64 alignment ?

lib/LinkerWrapper/RISCVLinkDriver.cpp
quic-seaswara: -pie support also needs to be tested and verified.

lib/Target/RISCV/RISCVLDBackend.cpp
quic-seaswara: Do we want to emit if the user has set a DISCARD rule

/DISCARD/ : { *(riscv.jvt) }

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -157,6 +207,50 @@ void RISCVLDBackend::initTargetSymbols() {
...
+  // finalizeContents() computes the fragment size, which must be propagated to
+  // the owning section for layout.
+  if (m_pRISCVTableJumpSection)
+    m_pRISCVTableJumpSection->setSize(TableJumpFragment->size());
+
+  // Ensure the output keeps a (possibly empty) .riscv.jvt section when
+  // --relax-tbljal is enabled.
+  if (m_pRISCVTableJumpSection) {
+    m_pRISCVTableJumpSection->setWantedInOutput(true);
+    m_pRISCVTableJumpSection->setWanted(true);
```
quic-seaswara: If the section is not 0 sized, linker will auto emit this section, this can be removed imo.

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -157,6 +207,50 @@ void RISCVLDBackend::initTargetSymbols() {
...
+    if (ELFSection *Out = m_pRISCVTableJumpSection->getOutputELFSection()) {
+      Out->setWantedInOutput(true);
+      Out->setWanted(true);
```
quic-seaswara: Same here.

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -157,6 +207,50 @@ void RISCVLDBackend::initTargetSymbols() {
...
+    }
+  }
+  if (ELFSection *Jvt = m_Module.getSection(".riscv.jvt")) {
+    Jvt->setWantedInOutput(true);
+    Jvt->setWanted(true);
```
quic-seaswara: I dont think the section is added to the output section table at this time.

I would prefer to remove this book keeping of setWanted/setWantedInOutput

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -222,6 +316,28 @@ void RISCVLDBackend::reportMissedRelaxation(StringRef Name,
...
+}
+
+static void applyTableJumpRelaxation(Relocation *Reloc,
+                                     RegionFragmentEx &Region, uint64_t Offset,
+                                     unsigned EntryIndex) {
+  uint16_t TblJump = static_cast<uint16_t>(0xA002 | (EntryIndex << 2));
+  Region.replaceInstruction(Offset, Reloc,
+                            reinterpret_cast<uint8_t *>(&TblJump), 2);
+  Reloc->setTargetData(TblJump);
+  Reloc->setType(eld::ELF::riscv::internal::R_RISCV_TBJAL);
```
quic-seaswara, line 331: Do you need to delete any bytes ?

lib/Target/RISCV/RISCVLDBackend.cpp
quic-seaswara: Can you please add a comment, that explains what is being done here ?

lib/Target/RISCV/RISCVTableJump.cpp
```diff
@@ -0,0 +1,218 @@
...
+#include "eld/SymbolResolver/ResolveInfo.h"
+#include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/Support/Endian.h"
+#include "llvm/Support/MathExtras.h"
+#include <cassert>
+
+using namespace eld;
+
+RISCVTableJumpFragment::RISCVTableJumpFragment(RISCVLDBackend &B, ELFSection *O)
```
quic-seaswara, line 28: Can you please add a reference to the ABI ?

test/RISCV/FromLLD/riscv-tbljal-many-jumps.s
```diff
@@ -0,0 +1,34 @@
+## Test table jump with many targets filling the cm.jt table (32 entries).
+## Verify the .riscv.jvt section size accounts for all entries.
+#
+# RUN: %llvm-mc -filetype=obj -mattr=+relax,+zcmt -asm-macro-max-nesting-depth=33 %s -o %t.o
+#
+## Use --defsym for targets beyond c.j range so they can only be relaxed via table jump.
+# RUN: %link %linkopts %t.o --relax-tbljal --defsym t0=0x100000 --defsym t1=0x100010 --defsym t2=0x100020 --defsym t3=0x100030 --defsym t4=0x100040 --defsym t5=0x100050 --defsym t6=0x100060 --defsym t7=0x100070 --defsym t8=0x100080 --defsym t9=0x100090 --defsym t10=0x1000a0 --defsym t11=0x1000b0 --defsym t12=0x1000c0 --defsym t13=0x1000d0 --defsym t14=0x1000e0 --defsym t15=0x1000f0 --defsym t16=0x100100 --defsym t17=0x100110 --defsym t18=0x100120 --defsym t19=0x100130 --defsym t20=0x100140 --defsym t21=0x100150 --defsym t22=0x100160 --defsym t23=0x100170 --defsym t24=0x100180 --defsym t25=0x100190 --defsym t26=0x1001a0 --defsym t27=0x1001b0 --defsym t28=0x1001c0 --defsym t29=0x1001d0 --defsym t30=0x1001e0 --defsym t31=0x1001f0 -o %t
```
quic-seaswara: Wrap this to 80 columns for readability, separate this by new lines with RUN:

lib/Target/RISCV/RISCVTableJump.h
```diff
@@ -0,0 +1,54 @@
...
+
+// Implements the Zcmt table jump section (.riscv.jvt) used by table jump
+// relaxation (cm.jt/cm.jalt).
+class RISCVTableJumpFragment final : public TargetFragment {
+public:
+  RISCVTableJumpFragment(RISCVLDBackend &B, ELFSection *O);
+
+  size_t size() const override;
+  eld::Expected<void> emit(MemoryRegion &Mr, Module &M) override;
+
```
quic-seaswara, line 33: add a dump function to emit the jump fragment in the text map file for readability.

It will be easy to verify by inspecting entries that are emitted.

quic-seaswara (changes requested):
Can you accompany this by tests that 

* use partial linking (-r)
* --gc-sections
* linker script placements 
* linker script discard 

We need to make sure that the entries are able to be verified using map file outputs.

Also now eld supports build/run tests, that can be used to make sure functionality works end to end. Can we use this as a way to make sure this functionality works ?

lib/Target/RISCV/RISCVLDBackend.cpp
svs-quic: I've created a separate type now.

lib/Target/RISCV/RISCVLDBackend.cpp
svs-quic: it is mandated by the ABI.

"The value in the BASE field must always be aligned on a 64-byte boundary."

lib/Target/RISCV/RISCVLDBackend.cpp
svs-quic: I had initially based this on the lldb implementation but having thought about it I dont think we need it. Don't emit it now,

svs-quic: I dont think we want to emit it (though I can't think of a situation where someone would want to do that). Have made changes to not emit it.

lib/Target/RISCV/RISCVLDBackend.cpp
svs-quic: Thanks have removed it.

lib/Target/RISCV/RISCVLDBackend.cpp
svs-quic: Removed this as well.

lib/Target/RISCV/RISCVLDBackend.cpp
svs-quic: Okay have removed it as well.

lib/Target/RISCV/RISCVLDBackend.cpp
svs-quic: Deletion is handled in relaxDeleteBytes.

lib/Target/RISCV/RISCVLDBackend.cpp
svs-quic: Have added a comment

lib/Target/RISCV/RISCVTableJump.h
svs-quic: Added this and also added tests for the same.

test/RISCV/FromLLD/riscv-tbljal-many-jumps.s
svs-quic: Done

lib/LinkerWrapper/RISCVLinkDriver.cpp
svs-quic: I have removed the shared warning check from this PR and will add it as a follow-up. Also zcmt is incompatible with pie. I will add an error check for this as well in a follow-up. From the ABI:

Condition

The zcmt extension is required, the linker output is not position independent

lib/Target/RISCV/RISCVTableJump.cpp
svs-quic: I've added header comment for the file and included the ABI reference there.

svs-quic (commented):
Thanks for your comments Shankar. I have addressed your comments and also added the tests you had wanted.

I have a local test that I can build and run with Qemu. I have verified that this produces the required instructions and succeeds when run. I will include this in a follow-up patch as well.

svs-quic:
CI failure is in a Hexagon test which seems unrelated.

parth-07:
> CI failure is in a Hexagon test which seems unrelated.

Hi Sudharsan,

The test is failing because of the new TableJump internal input. Can you please update the below CHECK in eld/test/Hexagon/Plugin/TrampolineInfo/TrampolineInfo.test:

```
CHECK: Output section 'BAZ' has trampoline 'trampoline_for_bar_from_.text.baz_{{27|28}}' for target symbol 'bar'
```

to:

```
CHECK: Output section 'BAZ' has trampoline 'trampoline_for_bar_from_.text.baz_{{.*}}' for target symbol 'bar'
```

{{27|28}} were referring to the input ordinal, the input ordinal has now become 29 because of the new TableJump internal input.

svs-quic:
> {{.*}}

Thanks Parth, I have fixed this in https://github.com/qualcomm/eld/pull/1209

quic-seaswara:
LGTM.

lib/LinkerWrapper/RISCVLinkDriver.cpp
```diff
@@ -139,6 +139,11 @@ RISCVLinkDriver::parseOptions(ArrayRef<const char *> Args,
   if (ArgList.hasArg(OPT_RISCVLinkOptTable::no_relax_tlsdesc))
     Config.options().setRISCVRelaxTLSDESC(false);
 
+  // --relax-tbljal, --no-relax-tbljal (default)
+  Config.options().setRISCVRelaxTbljal(
+      ArgList.hasFlag(OPT_RISCVLinkOptTable::relax_tbljal,
+                      OPT_RISCVLinkOptTable::no_relax_tbljal, false));
+
```
apazos, line 146: @svs-quic, This option should not be enabled if PIC mode is enabled (-shared), is there any place that checks for it?

apazos:
@svs-quic, before merging this change we need to enable it in nightly tests to validate it beyond unit tests, have you done that?

quic-seaswara:
> Can you accompany this by tests that
> 
> * use partial linking (-r)
> * --gc-sections
> * linker script placements
> * linker script discard
> 
> We need to make sure that the entries are able to be verified using map file outputs.
> 
> Also now eld supports build/run tests, that can be used to make sure functionality works end to end. Can we use this as a way to make sure this functionality works ?

I am not sure if this comment was looked into .... ?

lib/LinkerWrapper/RISCVLinkDriver.cpp
svs-quic: I had added this in the previous patch but Shankar had wanted it to be a separate PR. Will add it as a follow-up.

https://github.com/qualcomm/eld/pull/981#discussion_r3276582417

svs-quic:
> @svs-quic, before merging this change we need to enable it in nightly tests to validate it beyond unit tests, have you done that?

I am in the process of doing that by making changes in picolibc/musl to initialize the jvt CSR.

svs-quic:
> > Can you accompany this by tests that
> > 
> > * use partial linking (-r)
> > * --gc-sections
> > * linker script placements
> > * linker script discard
> > 
> > We need to make sure that the entries are able to be verified using map file outputs.
> > Also now eld supports build/run tests, that can be used to make sure functionality works end to end. Can we use this as a way to make sure this functionality works ?
> 
> I am not sure if this comment was looked into .... ?

test/RISCV/FromLLD/riscv-tbljal-map.s has the tests you had requested.

As for the end-to-end test, I had the following comment above in https://github.com/qualcomm/eld/pull/981#pullrequestreview-4336903091:

_"I have a local test that I can build and run with Qemu. I have verified that this produces the required instructions and succeeds when run. I will include this in a follow-up patch as well."_

lib/Target/RISCV/RISCVRelocationInfo.cpp
lenary: I wasn't sure this was right - usually we have to apply relocations after relaxation - but in this case, we're not writing an address (that might change) into the instruction, we're writing a table index (that won't change). If a function address changes due to a later relaxation, then we just need to make sure the table contents takes that into account.

I don't know if this case is tested.

lib/Target/RISCV/RISCVLDBackend.cpp
lenary: Setting it to the table size makes sense.

lib/Target/RISCV/RISCVRelocationInfo.cpp
svs-quic: I have added a test case for this, Sam. 

test/RISCV/FromLLD/riscv-tbljal-align.s

lenary (approved):
LGTM

svs-quic:
> > @svs-quic, before merging this change we need to enable it in nightly tests to validate it beyond unit tests, have you done that?
> 
> I am in the process of doing that by making changes in picolibc/musl to initialize the jvt CSR.

@apazos I have verified that the nightly benchmark tests are passing with zcmt enabled. I had enabled this for the rv32/64imac with musl.

svs-quic:
The clang-format failure is from a piece of code not touched by this patch.

```
--- a/lib/Target/RISCV/RISCVRelocationInternal.h
+++ b/lib/Target/RISCV/RISCVRelocationInternal.h
@@ -72,8 +72,8 @@ enum : uint32_t {
   /* 'QUALCOMM' vendor relocations: 192-197 are represented by 320-325 */
   FirstQUALCOMMVendorRelocation = 320,
   LastQUALCOMMVendorRelocation = 325,
-  QUALCOMMVendorRelocationOffset = FirstQUALCOMMVendorRelocation -
-      FirstNonstandardRelocation,
+  QUALCOMMVendorRelocationOffset =
+      FirstQUALCOMMVendorRelocation - FirstNonstandardRelocation,
 
   /* We don't support ANDES and CHERIOT1 vendor relocations */
   ANDESVendorRelocationOffset = 0,
```

apazos (approved):

svs-quic:
@quic-seaswara can you please approve the PR if you do not have any further comments?

quic-seaswara (approved):

PR #985, opened 2026-03-30
[LTO] Do not delete assembly inputs to LTO

### LINKED ISSUE #979: [lto] lto asm override deletes input files
```
cat > 1.c << \!
extern int foo();
int main() { return foo(); }
!

cat > 2.c << \!
extern int bar();
int foo() { return bar(); }
!

cat > 2.s << \!
.text
.global foo
foo:
.word 0
!

clang -c -target hexagon 1.c
clang -c -target hexagon -flto 2.c
ld.eld -mhexagonelf 1.o  2.o  -flto-options=lto-asm-file=2.s 
ls -l 2.s || echo FAIL

```

Running this test shows that input asm files getting deleted.

Mostly caused by change to the LTO machinery.

quic-akaryaki:
Fixes #979.

### DIFF
--- a/test/Common/LTO/LTOOptions/LTOOptions.test
+++ b/test/Common/LTO/LTOOptions/LTOOptions.test
@@ -1,0 +1,8 @@
+UNSUPPORTED: x86
+# TODO: Calling assembler from LTO is not implemented for RISC-V.
+UNSUPPORTED: riscv32, riscv64
+# Check that input assembler files are not deleted.
+RUN: %clang %clangopts -flto -c %p/Inputs/main.c -o %t.main.o
+RUN: %cp %p/Inputs/foo.s %t.foo.s
+RUN: %link %linkopts -mtriple %triple %t.main.o -flto-options=lto-asm-file=%t.foo.s -o %t.out
+RUN: ls %t.foo.s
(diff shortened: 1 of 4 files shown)


test/Common/LTO/LTOOptions/LTOOptions.test
```diff
@@ -0,0 +1,8 @@
+UNSUPPORTED: x86
+# TODO: Calling assembler from LTO is not implemented for RISC-V.
+UNSUPPORTED: riscv32, riscv64
+# Check that input assembler files are not deleted.
+RUN: %clang %clangopts -flto -c %p/Inputs/main.c -o %t.main.o
+RUN: %cp %p/Inputs/foo.s %t.foo.s
+RUN: %link %linkopts -mtriple %triple %t.main.o -flto-options=lto-asm-file=%t.foo.s -o %t.out
```
parth-07, line 7: Do we need to explicitly use %triple?

quic-akaryaki: Oh yeah, see the previous CI run.

quic-seaswara: Can we use emulation instead ?

quic-akaryaki: ltoCallExternalAssembler() only passes -mtriple and -mcpu. Also I thought we wanted to deprecate emulation.

quic-seaswara: it is the other way around, we will deprecate/remove triple and use emulation.

quic-akaryaki: Ok, my bad. ltoCallExternalAssembler() still does not pass emulation or arch.

quic-seaswara: no problem

PR #986, opened 2026-03-31
Fix incorrect "Plugin Error" when plt/got sections are discarded

### LINKED ISSUE #959: Incorrect error by the linker: `Plugin Error : Referenced Chunk {{.*}} last seen from {{.*}} has been deleted!`
Linker emits incorrect `Plugin Error` in some cases even when no plugin is involved. 

Reproducer:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
int a = 10;
int foo() { 
  return a;
}
EOF

cat >script.t <<\EOF
SECTIONS {
.text : { *(.text) }
/DISCARD/ : { *(.got) *(got*) }
}
EOF

clang -target arm-unknown-linux-gnu -o 1.o 1.c -c -fPIC
ld.eld -o 1.out 1.o -T script.t
```

Output:

```
Error: Plugin Error : Referenced Chunk .got from 1.o, last seen from Rule *(.got) #Rule 3, script.t in Output Section /DISCARD/ has been deleted!
Fatal: Linking had errors.
```

parth-07:
This commit fixes the incorrect `Plugin Error` when the real cause is that the .plt/.got/.plt.got sections are required but are explicitly discarded. This patch fixes this issue by reporting the correct error.

Resolves #959

### DIFF
--- a/lib/Target/GNULDBackend.cpp
+++ b/lib/Target/GNULDBackend.cpp
@@ -1623,6 +1623,37 @@ ELFSection *GNULDBackend::getRelaPLT() const {
   return m_DynamicSectionHeadersInputFile->getRelaPLT();
 }
 
+void GNULDBackend::reportErrorIfGOTIsDiscarded(ResolveInfo *R) const {
+  ELFSection *GOT = getGOT();
+  if (GOT && GOT->isIgnore()) {
+    llvm::StringRef SymName = R->name();
+    std::string FileName = R->resolvedOrigin()->getInput()->getName();
+    config().raise(Diag::error_discarded_dynamic_section_required)
+        << SymName << FileName << ".got";
+  }
+}
+
+void GNULDBackend::reportErrorIfPLTIsDiscarded(ResolveInfo *R) const {
+  ELFSection *PLT = getPLT();
+  if (PLT && PLT->isIgnore()) {
+    llvm::StringRef SymName = R->name();
+    std::string FileName = R->resolvedOrigin()->getInput()->getName();
+
+    config().raise(Diag::error_discarded_dynamic_section_required)
+        << SymName << FileName << ".plt";
+  }
+}
+
+void GNULDBackend::reportErrorIfGOTPLTIsDiscarded(ResolveInfo *R) const {
+  ELFSection *GOTPLT = getGOTPLT();
+  if (GOTPLT && GOTPLT->isIgnore()) {
+    llvm::StringRef SymName = R->name();
+    std::string FileName = R->resolvedOrigin()->getInput()->getName();
+    config().raise(Diag::error_discarded_dynamic_section_required)
+        << SymName << FileName << ".got.plt";
+  }
+}
+
 // Patching sections.
 ELFSection *GNULDBackend::getGOTPatch() const {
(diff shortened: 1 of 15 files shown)


lib/Target/GNULDBackend.cpp
```diff
@@ -1623,6 +1623,37 @@ ELFSection *GNULDBackend::getRelaPLT() const {
...
+    std::string FileName = R->resolvedOrigin()->getInput()->getName();
+
+    config().raise(Diag::error_discarded_dynamic_section_required)
+        << SymName << FileName << ".plt";
+  }
+}
+
+void GNULDBackend::reportErrorIfGOTPLTIsDiscarded(ResolveInfo *R) const {
+  ELFSection *GOTPLT = getGOTPLT();
+  if (GOTPLT && GOTPLT->isIgnore()) {
```
quic-seaswara, line 1649: early return ?

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -1599,10 +1599,13 @@ RISCVGOT *RISCVLDBackend::createGOT(GOT::GOTType T, ELFObjectFile *Obj,
     break;
   }
   if (R) {
-    if (GOT)
+    if (GOT) {
+      reportErrorIfGOTIsDiscarded(R);
       recordGOT(R, G);
```
quic-seaswara, line 1604: if (!Diagnose())

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -1599,10 +1599,13 @@ RISCVGOT *RISCVLDBackend::createGOT(GOT::GOTType T, ELFObjectFile *Obj,
...
-    else
+    } else {
+      reportErrorIfGOTPLTIsDiscarded(R);
       recordGOTPLT(R, G);
+    }
```
quic-seaswara, line 1608: can we move this to GNULDBackend ?

lib/Target/RISCV/RISCVLDBackend.cpp
parth-07: We cannot use `!Diagnose` here because no linker error is reported due to the discarded section yet.

lib/Target/RISCV/RISCVLDBackend.cpp
parth-07: This will be a much larger refactoring effort and I believe that it should be handled as a separate task.

lib/Target/GNULDBackend.cpp
parth-07: It is a small function, I don't think that early return would be any beneficial here. I am fine with both early return and the if-block. Please let me know if you strongly prefer early-return in these functions.

lib/Target/RISCV/RISCVLDBackend.cpp
quic-seaswara: Sure

lib/Target/RISCV/RISCVLDBackend.cpp
quic-seaswara: oh ok

PR #998, opened 2026-04-01
[github-workflow] Bundle ld.eld runtime libraries in the nightly tool…

quic-seaswara:
…chain

### DIFF
--- a/.github/workflows/nightly-test.yml
+++ b/.github/workflows/nightly-test.yml
@@ -61,6 +61,15 @@ jobs:
       - name: Configure workflow environment variables
         uses: ./llvm-project/llvm/tools/eld/.github/workflows/ConfigureBuildWorkflowENV
 
+      - name: Restore ccache
+        id: cache-ccache
+        uses: actions/cache/restore@v4
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-test.yml
```diff
@@ -99,6 +110,33 @@ jobs:
...
 
+      - name: Bundle ld.eld runtime libraries
+        shell: bash
+        run: |
+          set -euo pipefail
+          INSTALL_ROOT="${{ github.workspace }}/install-nightly-toolchain"
+          mkdir -p "${INSTALL_ROOT}/lib"
+
+          for lib in libc++.so.1 libc++abi.so.1 libunwind.so.1; do
+            src="$(ldconfig -p | awk -v lib="${lib}" '$1 == lib { print $NF; exit }')"
```
partaror, line 121: Are we copying the dependencies of libc++.so.1 and the others to the `${INSTALL_ROOT}/lib`? If so, I don't think we should do this. Why are we copying their dependencies to the INSTALL_ROOT/lib?

PR #999, opened 2026-04-01
Remove InputTarReader.h from PluginADT.h

quic-shaksuma:


include/eld/PluginAPI/PluginADT.h
```diff
@@ -1250,12 +1250,12 @@ class DLL_A_EXPORT TarFile final {
 private:
   TarFile(std::unique_ptr<eld::MemoryArea> Buf,
           std::vector<std::string> EntryNames,
-          std::vector<eld::InputTarReader::EntryInfo> RegularEntries);
+          std::vector<eld::EntryInfo> RegularEntries);
```
partaror, line 1253: `EntryInfo` definition needs to be visible in this file.

include/eld/Support/InputTarReader.h
```diff
@@ -23,19 +23,21 @@ namespace eld {
 
 class LinkerConfig;
 
+/// EntryInfo has the name,offset and size of the contents of
+/// a tar file and can be used to inspect the tar contents.
+struct EntryInfo {
```
partaror, line 28: Can you please make the name more descriptive? It is not inside any scope now. `EntryInfo` does not describe at all what kind of entry is being referred to here.

PR #1022, opened 2026-04-06
fix: deduplicate DT_NEEDED entries by SONAME

### LINKED ISSUE #50: Shared libraries are not identified by the SO name when emitting `DT_NEEDED` entries
Shared libraries should identified by their SO name when emitting `DT_NEEDED` entries. Currently, if multiple different shared libraries with the same SO name are present in the link, then we emit repeated `DT_NEEDED` entries for the same SO name.

Reproducible example:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
int foo() { return 1; }
EOF

cat >2.c <<\EOF
int bar() { return 3; }
EOF

LDs=(ld.eld ld.lld arm-none-eabi-ld.bfd)
SFs=(eld lld bfd)

clang -o 1.o 1.c -target arm -c -ffunction-sections
clang -o 2.o 2.c -target arm -c -ffunction-sections

for i in "${!SFs[@]}"; do
  ${LDs[$i]} -o lib1.${SFs[$i]}.so 1.o -shared -soname=foo
  ${LDs[$i]} -o lib2.${SFs[$i]}.so 2.o -shared -soname=foo
  ${LDs[$i]} -o a.${SFs[$i]}.out /dev/null lib1.${SFs[$i]}.so lib2.${SFs[$i]}.so 
done

llvm-readelf --dynamic a.eld.out | grep foo
```

Output:

```
  0x00000001 (NEEDED)   Shared library: [foo]
  0x00000001 (NEEDED)   Shared library: [foo]
```

deepakshirkem:
## Problem
Fixes #50 

When multiple shared libraries with the same SONAME are passed to the linker, ELD emits duplicate DT_NEEDED entries for the same SONAME.

## Fix

Deduplicate by SONAME string instead of file pointer using `llvm::StringSet<>`. This matches the behavior of `ld.lld` and `ld.bfd`.

## Testing
Verified manually using the reproducer from issue #50.

### DIFF
--- a/include/eld/Diagnostics/DiagReaders.inc
+++ b/include/eld/Diagnostics/DiagReaders.inc
@@ -54,7 +54,7 @@ DIAG(discard_reloc_section_map, DiagnosticEngine::Note,
 DIAG(input_file_has_zero_size, DiagnosticEngine::Note,
      "Input file %0 has no contents")
 DIAG(string_not_null_terminated, DiagnosticEngine::Error,
-    "%0: string is not null terminated")
+     "%0: string is not null terminated")
 DIAG(invalid_elf_class, DiagnosticEngine::Error,
      "Invalid ELF file %0 for target %1")
 DIAG(fatal_big_endian_target, DiagnosticEngine::Fatal,
@@ -79,7 +79,8 @@ DIAG(error_failed_to_read_thin_archive_mem, DiagnosticEngine::Error,
 DIAG(warn_whole_archive_enabled, DiagnosticEngine::Warning,
      "'whole-archive' enabled for '%0'")
 DIAG(warn_archive_mem_repeated_content, DiagnosticEngine::Warning,
-     "Archive member '%0:%1' at index %2 has same content as archive member '%0:%3' "
+     "Archive member '%0:%1' at index %2 has same content as archive member "
+     "'%0:%3' "
      "at index %4")
 DIAG(verbose_performing_archive_symbol_resolution, DiagnosticEngine::Verbose,
      "Performing symbol resolution for archive '%0'")
@@ -97,7 +98,7 @@ DIAG(error_patch_invalid_symbol, DiagnosticEngine::Error,
 DIAG(warning_abi_differences, DiagnosticEngine::Warning,
      "'%0' has OS/ABI set to %1, previously seen %2")
 DIAG(error_unknown_target_emulation, DiagnosticEngine::Error,
-  "target emulation unknown: -m or at least one .o file required")
+     "target emulation unknown: -m or at least one .o file required")
 DIAG(error_target_not_found, DiagnosticEngine::Error,
      "Target not registered : %0")
 DIAG(error_emulation_failed_for_target, DiagnosticEngine::Error,
@@ -110,7 +111,8 @@ DIAG(error_failed_to_find_driver, DiagnosticEngine::Error,
      "No driver found for target : %0")
 DIAG(fatal_unsupported_bit_code_file, DiagnosticEngine::Fatal,
      "Unsupported architecture %0 in LLVM Bitcode file : %1")
-DIAG(note_empty_archive, DiagnosticEngine::Note,
-     "Empty archive file: '%0'")
+DIAG(note_empty_archive, DiagnosticEngine::Note, "Empty archive file: '%0'")
 DIAG(sframe_read_error, DiagnosticEngine::Error,
      "SFrame Read Error : %0 from file %1")
+DIAG(warn_duplicate_soname, DiagnosticEngine::Warning,
+     "'%0' and '%1' have the same SONAME '%2'")
--- a/include/eld/Diagnostics/DiagReaders.inc
+++ b/include/eld/Diagnostics/DiagReaders.inc
@@ -56,7 +56,7 @@ DIAG(discard_reloc_section_map, DiagnosticEngine::Note,
 DIAG(input_file_has_zero_size, DiagnosticEngine::Note,
      "Input file %0 has no contents")
 DIAG(string_not_null_terminated, DiagnosticEngine::Error,
-    "%0: string is not null terminated")
+     "%0: string is not null terminated")
 DIAG(invalid_elf_class, DiagnosticEngine::Error,
      "Invalid ELF file %0 for target %1")
 DIAG(fatal_big_endian_target, DiagnosticEngine::Fatal,
@@ -81,7 +81,8 @@ DIAG(error_failed_to_read_thin_archive_mem, DiagnosticEngine::Error,
 DIAG(warn_whole_archive_enabled, DiagnosticEngine::Warning,
      "'whole-archive' enabled for '%0'")
 DIAG(warn_archive_mem_repeated_content, DiagnosticEngine::Warning,
-     "Archive member '%0:%1' at index %2 has same content as archive member '%0:%3' "
+     "Archive member '%0:%1' at index %2 has same content as archive member "
+     "'%0:%3' "
      "at index %4")
 DIAG(verbose_performing_archive_symbol_resolution, DiagnosticEngine::Verbose,
      "Performing symbol resolution for archive '%0'")
@@ -94,7 +95,7 @@ DIAG(verbose_reusing_archive_file_info, DiagnosticEngine::Verbose,
 DIAG(warning_abi_differences, DiagnosticEngine::Warning,
      "'%0' has OS/ABI set to %1, previously seen %2")
 DIAG(error_unknown_target_emulation, DiagnosticEngine::Error,
-  "target emulation unknown: -m or at least one .o file required")
+     "target emulation unknown: -m or at least one .o file required")
 DIAG(error_target_not_found, DiagnosticEngine::Error,
      "Target not registered : %0")
 DIAG(error_emulation_failed_for_target, DiagnosticEngine::Error,
@@ -107,7 +108,8 @@ DIAG(error_failed_to_find_driver, DiagnosticEngine::Error,
      "No driver found for target : %0")
 DIAG(fatal_unsupported_bit_code_file, DiagnosticEngine::Fatal,
      "Unsupported architecture %0 in LLVM Bitcode file : %1")
-DIAG(note_empty_archive, DiagnosticEngine::Note,
-     "Empty archive file: '%0'")
+DIAG(note_empty_archive, DiagnosticEngine::Note, "Empty archive file: '%0'")
 DIAG(sframe_read_error, DiagnosticEngine::Error,
      "SFrame Read Error : %0 from file %1")
+DIAG(warn_duplicate_soname, DiagnosticEngine::Warning,
+     "'%0' and '%1' have the same SONAME '%2'")
--- a/lib/Config/LinkerConfig.cpp
+++ b/lib/Config/LinkerConfig.cpp
@@ -233,6 +233,14 @@ bool LinkerConfig::setWarningOption(llvm::StringRef WarnOption) {
     setShowLinkerScriptMemoryWarning(false);
     return true;
   }
+  if (WarnOpt == "dynamic") {
+    setShowDynamicWarnings(true);
+    return true;
+  }
+  if (WarnOpt == "no-dynamic") {
+    setShowDynamicWarnings(false);
+    return true;
+  }
   if (WarnOpt == "bad-dot-assignments") {
     setShowBadDotAssginmentsWarning(true);
     return true;
--- a/lib/Object/ObjectLinker.cpp
+++ b/lib/Object/ObjectLinker.cpp
@@ -3854,12 +3854,21 @@ bool ObjectLinker::readAndProcessInput(Input *Input, bool IsPostLto) {
         ThisConfig.raiseDiagEntry(std::move(ExpRead.error()));
       return false;
     }
-#ifdef ELD_ENABLE_SYMBOL_VERSIONING
     ELFDynObjectFile *DynObjFile = llvm::cast<ELFDynObjectFile>(CurInput);
+#ifdef ELD_ENABLE_SYMBOL_VERSIONING
     if (DynObjFile->hasSymbolVersioningInfo())
       getTargetBackend().setShouldEmitVersioningSections(true);
 #endif
+    std::string SOName = DynObjFile->getSOName();
+    std::string FilePath = CurInput->getInput()->decoratedPath();
     ThisModule->getDynLibraryList().push_back(CurInput);
+    if (!MSonameToFile.count(SOName)) {
+      MSonameToFile[SOName] = FilePath;
+    } else if (MSonameToFile[SOName] != FilePath &&
+               ThisConfig.showDynamicWarnings()) {
+      ThisConfig.raise(Diag::warn_duplicate_soname)
+          << MSonameToFile[SOName] << FilePath << SOName;
+    }
   } else if (CurInput->getKind() == InputFile::GNUArchiveFileKind) {
     eld::RegisterTimer T("Read Archive Files", "Read all Input files",
                          ThisConfig.options().printTimingStats());
--- a/lib/Target/GNULDBackend.cpp
+++ b/lib/Target/GNULDBackend.cpp
@@ -988,7 +988,7 @@ void GNULDBackend::sizeDynamic() {
     std::string fileName = dynObjFile->getInput()->getFileName();
     if (sonameToFile.count(SOName)) {
       // Only warn when different files have the same SONAME
-      if (sonameToFile[SOName] != fileName)
+      if (sonameToFile[SOName] != fileName && config().showDynamicWarnings())
         config().raise(Diag::warn_duplicate_soname)
             << sonameToFile[SOName] << fileName << SOName;
       continue;
--- a/lib/Target/GNULDBackend.cpp
+++ b/lib/Target/GNULDBackend.cpp
@@ -979,18 +979,24 @@ void GNULDBackend::sizeDynamic() {
   }
 
   // add DT_NEEDED
-  std::unordered_set<MemoryArea *> addedLibs;
+  llvm::StringMap<std::string> sonameToFile;
   for (auto &lib : m_Module.getDynLibraryList()) {
-    if (llvm::dyn_cast<ELFFileBase>(lib)->isELFNeeded()) {
-      const ELFDynObjectFile *dynObjFile = llvm::cast<ELFDynObjectFile>(lib);
-      if (addedLibs.count(dynObjFile->getInput()->getMemArea()))
-        continue;
-      addedLibs.insert(dynObjFile->getInput()->getMemArea());
-      std::size_t SONameOffset =
-          FileFormat->addStringToDynStrTab(dynObjFile->getSOName());
-      auto DTEntry = dynamic()->reserveNeedEntry();
-      DTEntry->setValue(llvm::ELF::DT_NEEDED, SONameOffset);
+    if (!llvm::dyn_cast<ELFFileBase>(lib)->isELFNeeded())
+      continue;
+    const ELFDynObjectFile *dynObjFile = llvm::cast<ELFDynObjectFile>(lib);
+    std::string SOName = dynObjFile->getSOName();
+    std::string fileName = dynObjFile->getInput()->getFileName();
+    if (sonameToFile.count(SOName)) {
+      // Only warn when different files have the same SONAME
+      if (sonameToFile[SOName] != fileName)
+        config().raise(Diag::warn_duplicate_soname)
+            << sonameToFile[SOName] << fileName << SOName;
+      continue;
     }
+    sonameToFile[SOName] = fileName;
+    std::size_t SONameOffset = FileFormat->addStringToDynStrTab(SOName);
+    auto DTEntry = dynamic()->reserveNeedEntry();
+    DTEntry->setValue(llvm::ELF::DT_NEEDED, SONameOffset);
   }
 
   // add DT_RUNPATH
--- a/lib/Target/GNULDBackend.cpp
+++ b/lib/Target/GNULDBackend.cpp
@@ -1125,7 +1125,7 @@ void GNULDBackend::sizeDynamic() {
   }
 
   // add DT_NEEDED
-  std::unordered_set<MemoryArea *> addedLibs;
+  llvm::StringMap<std::string> sonameToFile;
   for (auto &lib : m_Module.getDynLibraryList()) {
     if (llvm::dyn_cast<ELFFileBase>(lib)->isELFNeeded()) {
       const ELFDynObjectFile *dynObjFile = llvm::cast<ELFDynObjectFile>(lib);
@@ -1138,6 +1138,10 @@ void GNULDBackend::sizeDynamic() {
       DTEntry->tag = llvm::ELF::DT_NEEDED;
       DTEntry->value = SONameOffset;
     }
+    sonameToFile[SOName] = fileName;
+    std::size_t SONameOffset = FileFormat->addStringToDynStrTab(SOName);
+    auto DTEntry = dynamic()->reserveNeedEntry();
+    DTEntry->setValue(llvm::ELF::DT_NEEDED, SONameOffset);
   }
 
   // add DT_RUNPATH
--- a/lib/Target/GNULDBackend.cpp
+++ b/lib/Target/GNULDBackend.cpp
@@ -1125,23 +1125,18 @@ void GNULDBackend::sizeDynamic() {
   }
 
   // add DT_NEEDED
-  llvm::StringMap<std::string> sonameToFile;
+  llvm::StringSet<> addedSONAMEs;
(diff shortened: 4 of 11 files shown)


lib/Target/GNULDBackend.cpp
```diff
@@ -979,15 +979,15 @@ void GNULDBackend::sizeDynamic() {
...
 
   // add DT_NEEDED
-  std::unordered_set<MemoryArea *> addedLibs;
+  llvm::StringSet<> addedSONAMEs;
   for (auto &lib : m_Module.getDynLibraryList()) {
     if (llvm::dyn_cast<ELFFileBase>(lib)->isELFNeeded()) {
       const ELFDynObjectFile *dynObjFile = llvm::cast<ELFDynObjectFile>(lib);
-      if (addedLibs.count(dynObjFile->getInput()->getMemArea()))
+      std::string SOName = dynObjFile->getSOName();
+      if (!addedSONAMEs.insert(SOName).second)
```
parth-07: We are changing the if-condition without preserving the behavior of the old if-condition. Would it break anything?

parth-07 (changes requested):
We need tests with this patch.

Additionally, we should warn if there are multiple different shared library with the same SO name.

lib/Target/GNULDBackend.cpp
deepakshirkem: Hi @parth-07,

I tested this locally and the behavior is preserved. The old code deduplicated by `MemoryArea` pointer to prevent the same  file from being added twice. The new code deduplicates by SONAME string, 
which is strictly stronger:

So the new check covers everything the old one did, plus fixes the bug.

Screenshot With reproduce step at #50 
Screenshot from 2026-04-06 19-27-28

lib/Target/GNULDBackend.cpp
```diff
@@ -979,15 +979,21 @@ void GNULDBackend::sizeDynamic() {
...
 
   // add DT_NEEDED
-  std::unordered_set<MemoryArea *> addedLibs;
+  llvm::StringSet<> addedSONAMEs;
+  llvm::StringMap<std::string> sonameToFile;
   for (auto &lib : m_Module.getDynLibraryList()) {
     if (llvm::dyn_cast<ELFFileBase>(lib)->isELFNeeded()) {
       const ELFDynObjectFile *dynObjFile = llvm::cast<ELFDynObjectFile>(lib);
-      if (addedLibs.count(dynObjFile->getInput()->getMemArea()))
+      std::string SOName = dynObjFile->getSOName();
```
quic-seaswara: llvm::StringRef ?

lib/Target/GNULDBackend.cpp
```diff
@@ -979,15 +979,21 @@ void GNULDBackend::sizeDynamic() {
...
+      std::string fileName = dynObjFile->getInput()->getFileName();
```
quic-seaswara: can you move filenanme to 990 and use StringRef

lib/Target/GNULDBackend.cpp
```diff
@@ -979,15 +979,21 @@ void GNULDBackend::sizeDynamic() {
   }
 
   // add DT_NEEDED
-  std::unordered_set<MemoryArea *> addedLibs;
+  llvm::StringSet<> addedSONAMEs;
+  llvm::StringMap<std::string> sonameToFile;
   for (auto &lib : m_Module.getDynLibraryList()) {
     if (llvm::dyn_cast<ELFFileBase>(lib)->isELFNeeded()) {
```
quic-seaswara: if (!isELFNeeded())
 continue;

test/Common/standalone/DtNeededSonameDedup/DtNeededSonameDedup.test
```diff
@@ -0,0 +1,17 @@
...
+# Test that DT_NEEDED entries are deduplicated by SONAME.
+# When multiple shared libraries with the same SONAME are linked,
+# only one DT_NEEDED entry should be emitted.
+#END_COMMENT
+#START_TEST
+RUN: %clang %clangopts -fPIC -c %p/Inputs/1.c -o %t1.1.o
+RUN: %clang %clangopts -fPIC -c %p/Inputs/2.c -o %t1.2.o
+RUN: %link %linkopts %t1.1.o -o %t1.lib1.so -shared -soname=foo
+RUN: %link %linkopts %t1.2.o -o %t1.lib2.so -shared -soname=foo
+RUN: %link %linkopts /dev/null %t1.lib1.so %t1.lib2.so -o %t1.out 2>&1 | %filecheck %s --check-prefix=WARN
```
quic-seaswara: why /dev/null ? is it portable to windows ?

test/Common/standalone/DtNeededSonameDedup/DtNeededSonameDedup.test
```diff
@@ -0,0 +1,17 @@
...
+RUN: %readelf --dynamic %t1.out | %filecheck %s --check-prefix=NEEDED
+#WARN: Warning: Multiple shared libraries with the same SONAME "foo" detected
```
quic-seaswara: Does bfd warn ?

include/eld/Diagnostics/DiagReaders.inc
```diff
@@ -113,4 +113,6 @@ DIAG(fatal_unsupported_bit_code_file, DiagnosticEngine::Fatal,
 DIAG(note_empty_archive, DiagnosticEngine::Note,
      "Empty archive file: '%0'")
 DIAG(sframe_read_error, DiagnosticEngine::Error,
-     "SFrame Read Error : %0 from file %1")
\ No newline at end of file
+     "SFrame Read Error : %0 from file %1")
```
quic-seaswara, line 113: remove new line

include/eld/Diagnostics/DiagReaders.inc
```diff
@@ -113,4 +113,6 @@ DIAG(fatal_unsupported_bit_code_file, DiagnosticEngine::Fatal,
...
+DIAG(warn_duplicate_soname, DiagnosticEngine::Warning,
+     "Multiple shared libraries with the same SONAME \"%0\" detected: %1 and %2")
```
quic-seaswara: can you use the same format of message used by Gnu Ld or lld ?

test/Common/standalone/DtNeededSonameDedup/Inputs/2.c
```diff
@@ -0,0 +1 @@
+int bar() { return 3; }
```
quic-seaswara: you can have the same input file to generate two shared libraries ?

quic-seaswara (changes requested):

lib/Target/GNULDBackend.cpp
parth-07: > The old code deduplicated by MemoryArea pointer to prevent the same file from being added twice. The new code deduplicates by SONAME string,
which is strictly stronger:

What happens for libraries without any SONAME? Deduplication using SONAME wouldn't work for such libraries.

test/Common/standalone/DtNeededSonameDedup/DtNeededSonameDedup.test
parth-07: bfd and lld does not warn here, they simply emit the first shared library seen when there are multiple libraries with the same SONAME. I requested @deepakshirkem to emit the warning diagnostic here.

lib/Target/GNULDBackend.cpp
deepakshirkem: Hi @parth-07,  I investigated this concern and the fix handles this case correctly. Every SONAME  library always has a SONAME set before our code runs 

https://github.com/qualcomm/eld/blob/b03021e8fc22d87deef4ca934fe4c81d0799f48f/lib/Readers/DynamicELFReader.cpp#L122-L123


`getFallbackSOName()` returns the resolved file path, which is unique per file. By the time our code runs in `sizeDynamic()`, every library already has a SONAME set either real or fallback. We never see an empty SONAME.

I am still open to your feedback!

lib/Target/GNULDBackend.cpp
deepakshirkem: Hi @quic-seaswara, getSOName() returns std::string by value. Using llvm::StringRef here would result in a dangling reference since the temporary std::string would be destroyed at the end of the full expression.
I did try using llvm::StringRef initially but it resulted in a dangling reference error during compilation.

deepakshirkem:
Hi @parth-07 / @quic-seaswara, I just wanted to follow up on this PR when you get a chance. I understand you have a busy schedule. Just wanted to make sure this didn't get lost!

Thank You ::))

deepakshirkem:
Hi @quic-seaswara / @parth-07, Fixed the misleading indentation issue in GNULDBackend.cpp  and applied clang-format. All related tests pass locally. Please re-run the pipeline when you get a chance.

Thank you ::))

quic-areg:
Change looks correct to me.

Eld will warn when the same library is listed twice:
`ld.eld -o a.out lib1.so lib1.so`

We probably should not warn in this case. The warning is only meaningful when the files differ. GNU ld and lld do not error here.

include/eld/Diagnostics/DiagReaders.inc
```diff
@@ -113,4 +113,6 @@ DIAG(fatal_unsupported_bit_code_file, DiagnosticEngine::Fatal,
 DIAG(note_empty_archive, DiagnosticEngine::Note,
      "Empty archive file: '%0'")
 DIAG(sframe_read_error, DiagnosticEngine::Error,
-     "SFrame Read Error : %0 from file %1")
\ No newline at end of file
+     "SFrame Read Error : %0 from file %1")
+DIAG(warn_duplicate_soname, DiagnosticEngine::Warning,
+     "'%1' and '%2' have the same SONAME '%0'")
```
quic-areg: The placeholders are used out of order. I think it's easier to read if they're in order.

lib/Target/GNULDBackend.cpp
```diff
@@ -979,18 +979,23 @@ void GNULDBackend::sizeDynamic() {
   }
 
   // add DT_NEEDED
-  std::unordered_set<MemoryArea *> addedLibs;
+  llvm::StringSet<> addedSONAMEs;
+  llvm::StringMap<std::string> sonameToFile;
```
quic-areg: The `StringSet<>` and `StringMap<std::string>` look redundant to me. We can just use the StringMap.

deepakshirkem:
Hi @quic-areg, Addressed all three review comments Please re-run the pipeline when you get a chance.

Thank you ::))

lib/Target/GNULDBackend.cpp
deepakshirkem: Removed redundant StringSet, using only  StringMap now

quic-seaswara (approved):

quic-areg:
commit message repeats itself:
```
fix: deduplicate DT_NEEDED entries by SONAME and warn on duplicates

When multiple shared libraries with the same SONAME are passed to the
linker, ELD was emitting duplicate DT_NEEDED entries for the same SONAME.

fix: deduplicate DT_NEEDED entries by SONAME and warn on duplicates

When multiple shared libraries with the same SONAME are passed to the
linker, ELD was emitting duplicate DT_NEEDED entries.
```

deepakshirkem:
Hi @quic-areg  / @quic-seaswara, Sorry for the mistake. I fixed the commit message.  You can run pipeline once you get a chance.

Thank You ::))

lib/Target/GNULDBackend.cpp
```diff
@@ -979,18 +979,24 @@ void GNULDBackend::sizeDynamic() {
...
+    if (!llvm::dyn_cast<ELFFileBase>(lib)->isELFNeeded())
+      continue;
+    const ELFDynObjectFile *dynObjFile = llvm::cast<ELFDynObjectFile>(lib);
+    std::string SOName = dynObjFile->getSOName();
+    std::string fileName = dynObjFile->getInput()->getFileName();
+    if (sonameToFile.count(SOName)) {
+      // Only warn when different files have the same SONAME
+      if (sonameToFile[SOName] != fileName)
+        config().raise(Diag::warn_duplicate_soname)
+            << sonameToFile[SOName] << fileName << SOName;
```
quic-seaswara: Can we move this warning under -Wdynamic ? If so please add a -Wdynamic.

I am ok to treat that as a follow up patch.

quic-seaswara (approved):

quic-seaswara (approved):

quic-seaswara (commented):
Please change the commit message here too.

deepakshirkem:
Hi @quic-seaswara, Updated the commit message. I will make sure future commits follow the correct format.

Thank you ::))

deepakshirkem:
Hi @quic-seaswara,  Added the `-Wdynamic` flag as a follow up patch. The duplicate SONAME warning is now only shown when `-Wdynamic` is explicitly passed. Also updated the test to use `-Wdynamic` flag. Please review when you get a chance.

Thank you ((::

test/Common/standalone/DtNeededSonameDedup/DtNeededSonameDedup.test
```diff
@@ -0,0 +1,17 @@
...
+# Test that DT_NEEDED entries are deduplicated by SONAME.
+# When multiple shared libraries with the same SONAME are linked,
+# only one DT_NEEDED entry should be emitted.
+# Warning is only shown with -Wdynamic flag.
+#END_COMMENT
+#START_TEST
+RUN: %clang %clangopts -fPIC -c %p/Inputs/1.c -o %t1.1.o
+RUN: %link %linkopts %t1.1.o -o %t1.lib1.so -shared -soname=foo
+RUN: %link %linkopts %t1.1.o -o %t1.lib2.so -shared -soname=foo
+RUN: %link %linkopts %t1.lib1.so %t1.lib2.so -o %t1.out -Wdynamic 2>&1 | %filecheck %s --check-prefix=WARN
```
parth-07, line 12: Currently, this patch is de-duplicating the shared libraries based on soname at the time of writing the DT_NEEDED entries. What do you think about de-duplicating based on sonames at the time of reading the inputs? Can you please check whether bfd and lld de-duplicate based on soname during reading the input files or when emitting the DT_NEEDED entries?

parth-07 (changes requested):

test/Common/standalone/DtNeededSonameDedup/DtNeededSonameDedup.test
deepakshirkem: I checked both BFD and LLD implementations BFD performs deduplication at emit time, which is similar to the approach used in this patch and LLD performs deduplication during input reading  If you prefer the LLD-style approach at read time, will move the deduplication logic there as well

lib/Config/LinkerConfig.cpp
```diff
@@ -233,6 +233,14 @@ bool LinkerConfig::setWarningOption(llvm::StringRef WarnOption) {
     setShowLinkerScriptMemoryWarning(false);
     return true;
   }
+  if (WarnOpt == "dynamic") {
```
parth-07, line 236: We need to document this new warning option in the GnuLinkerOptions.td

lib/Target/GNULDBackend.cpp
```diff
@@ -979,18 +979,24 @@ void GNULDBackend::sizeDynamic() {
...
-      addedLibs.insert(dynObjFile->getInput()->getMemArea());
-      std::size_t SONameOffset =
-          FileFormat->addStringToDynStrTab(dynObjFile->getSOName());
-      auto DTEntry = dynamic()->reserveNeedEntry();
-      DTEntry->setValue(llvm::ELF::DT_NEEDED, SONameOffset);
+    if (!llvm::dyn_cast<ELFFileBase>(lib)->isELFNeeded())
+      continue;
+    const ELFDynObjectFile *dynObjFile = llvm::cast<ELFDynObjectFile>(lib);
+    std::string SOName = dynObjFile->getSOName();
+    std::string fileName = dynObjFile->getInput()->getFileName();
```
parth-07: Please use `Input::decoratedPath` for reporting filename in diagnostics.

test/Common/standalone/DtNeededSonameDedup/DtNeededSonameDedup.test
parth-07: It seems that both bfd and lld are doing de-duplication at read-time:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
int foo() {
  return 1;
}
EOF

cat >2.c <<\EOF
int bar() { return 3; }
EOF


cat >main.c <<\EOF
int foo();
int bar();

int main() {
  return foo() + bar();
}
EOF

clang -target x86_64-linux-gnu -o 1.o 1.c -c -fPIC
clang -target x86_64-linux-gnu -o 2.o 2.c -c -fPIC
clang -target x86_64-linux-gnu -o main.o main.c -c 
LDs=(ld.eld ld.lld ld.bfd)
SFs=(eld lld bfd)

for i in "${!SFs[@]}"; do
  ${LDs[$i]} -o lib1.${SFs[$i]}.so 1.o -shared -soname lib12.so.1 -Map lib1.${SFs[$i]}.map.txt
  ${LDs[$i]} -o lib2.${SFs[$i]}.so 2.o -shared -soname lib12.so.1 -Map lib2.${SFs[$i]}.map.txt
  ${LDs[$i]} -o main.${SFs[$i]}.out main.o lib1.${SFs[$i]}.so lib2.${SFs[$i]}.so -Map main.${SFs[$i]}.map.txt 
done
```

deepakshirkem: Sorry, my mistake I used the wrong target while checking. Will update PR shortly.

deepakshirkem:
Hi @quic-areg / @quic-seaswara / @parth-07 ,  This need changes or can it be merged?

test/Common/standalone/DtNeededSonameDedup/DtNeededSonameSameFile.test
```diff
@@ -0,0 +1,15 @@
+#---DtNeededSonameSameFile.test--------------------------- Executable -----------------#
+#BEGIN_COMMENT
+# Test that no warning is emitted when the same shared library
+# is listed twice on the command line, even with -Wdynamic.
+#END_COMMENT
+#START_TEST
+RUN: %clang %clangopts -fPIC -c %p/Inputs/1.c -o %t1.1.o
+RUN: %link %linkopts %t1.1.o -o %t1.lib1.so -shared -soname=foo
+RUN: %link %linkopts %t1.lib1.so %t1.lib1.so -o %t1.out -Wdynamic 2>&1 | (cat; echo "LINK_DONE") | %filecheck %s --check-prefix=NOWARN
+RUN: readelf --dynamic %t1.out 2>&1 | %filecheck %s --check-prefix=NEEDED
```
quic-areg: `%readelf`

test/Common/standalone/DtNeededSonameDedup/DtNeededSonameNoWarn.test
```diff
@@ -0,0 +1,16 @@
...
+#BEGIN_COMMENT
+# Test that no warning is emitted when -Wdynamic is not passed,
+# even when multiple shared libraries with the same SONAME are linked.
+#END_COMMENT
+#START_TEST
+RUN: %clang %clangopts -fPIC -c %p/Inputs/1.c -o %t1.1.o
+RUN: %link %linkopts %t1.1.o -o %t1.lib1.so -shared -soname=foo
+RUN: %link %linkopts %t1.1.o -o %t1.lib2.so -shared -soname=foo
+RUN: %link %linkopts %t1.lib1.so %t1.lib2.so -o %t1.out 2>&1 | (cat; echo "LINK_DONE") | %filecheck %s --check-prefix=NOWARN
+RUN: readelf --dynamic %t1.out 2>&1 | %filecheck %s --check-prefix=NEEDED
```
quic-areg: `%readelf`

quic-areg (commented):
LGTM otherwise

deepakshirkem:
> LGTM otherwise

Hi @quic-areg, addressed the comments. Thank you! ::))

deepakshirkem:
> > LGTM otherwise
> 
> Hi @quic-areg, addressed the comments. Thank you! ::))

Ping @quic-areg / @parth-07 / @quic-seaswara

deepakshirkem:
Hi @quic-areg / @quic-seaswara / @parth-07 , This need changes or can it be merged?

quic-seaswara (approved):

deepakshirkem:
@quic-seaswara Can it be merged?

quic-seaswara:
@parth-07 / @quic-areg / @Steven6798  can you guys review this ?

Steven6798:
@deepakshirkem can you mark any conversation that has been addressed as resolved? That will make reviewing this easier.

deepakshirkem:
@Steven6798, I have marked all conversations that I have addressed  as resolved. Some conversations from @parth-07 are  still open as I am waiting for his confirmation that  the changes are acceptable.

Steven6798:
> Steven Ramirez Rosa (@Steven6798), I have marked all conversations that I have addressed as resolved. Some conversations from Parth (@parth-07) are still open as I am waiting for his confirmation that the changes are acceptable.

Thanks. Look forward to resolving all conversations. Also, I see some formatting changes unrelated to your changes (DiagReaders.inc‎), remove those. Other than that, it looks good to me.

lib/Object/ObjectLinker.cpp
```diff
@@ -3854,12 +3854,21 @@ bool ObjectLinker::readAndProcessInput(Input *Input, bool IsPostLto) {
...
 #endif
-    ThisModule->getDynLibraryList().push_back(CurInput);
+    std::string SOName = DynObjFile->getSOName();
+    std::string FilePath = CurInput->getInput()->decoratedPath();
+    if (!MSonameToFile.count(SOName)) {
+      MSonameToFile[SOName] = FilePath;
+      ThisModule->getDynLibraryList().push_back(CurInput);
+    } else if (MSonameToFile[SOName] != FilePath &&
+               ThisConfig.showDynamicWarnings()) {
+      ThisConfig.raise(Diag::warn_duplicate_soname)
```
parth-07, line 3869: Why aren't we adding the adding the shared library to the `ThisModule->getDynLibraryList()`? The issue is only about de-duplicating the DT_NEEDED entries. The linker still needs to read all the files.

lib/Target/GNULDBackend.cpp
parth-07: How is `addedLibs` getting used here when you have removed the definition which was at line 1128?

lib/Target/GNULDBackend.cpp
```diff
@@ -1138,6 +1137,10 @@ void GNULDBackend::sizeDynamic() {
       DTEntry->tag = llvm::ELF::DT_NEEDED;
       DTEntry->value = SONameOffset;
     }
+    sonameToFile[SOName] = fileName;
+    std::size_t SONameOffset = FileFormat->addStringToDynStrTab(SOName);
+    auto DTEntry = dynamic()->reserveNeedEntry();
+    DTEntry->setValue(llvm::ELF::DT_NEEDED, SONameOffset);
```
parth-07: What is this code doing? Where is `filename` and `SOName` defined?

parth-07 (changes requested):

lib/Object/ObjectLinker.cpp
deepakshirkem: Fixed! The shared library is now always added to getDynLibraryList() so the linker reads all files.
The SONAME tracking is only used for deduplication of DT_NEEDED entries and warning purposes.

lib/Target/GNULDBackend.cpp
deepakshirkem: Fixed!

lib/Target/GNULDBackend.cpp
deepakshirkem: Fixed!

lib/Target/GNULDBackend.cpp
```diff
@@ -1125,19 +1125,18 @@ void GNULDBackend::sizeDynamic() {
   }
 
   // add DT_NEEDED
-  std::unordered_set<MemoryArea *> addedLibs;
```
parth-07, line 1128: Why did we remove the addedLibs? Aren't the addedLibs and SONames solving different things?

deepakshirkem: Yes, they solve both cases. Now addedSONAMEs takes care of both: the same file appearing twice gets deduplicated because it has the same SONAME, and different files with the same SONAME are also deduplicated. Since we moved ObjectLinker.cpp to always add all libraries to getDynLibraryList(), the addedSONAMEs logic in GNULDBackend.cpp now handles both cases.

deepakshirkem:
@parth-07 Any Feedback

deepakshirkem:
Hi @parth-07 Does this need changes or can it be merged?

deepakshirkem:
@parth-07 Ping.

PR #1034, opened 2026-04-07
[zephyr] test board qemu_riscv32_xip

quic-areg:


quic-areg:
run https://github.com/qualcomm/eld/actions/runs/24272131051/job/70879090149?pr=1034

.github/workflows/zephyr-nightly.yml
```diff
@@ -1,7 +1,8 @@
 name: Zephyr Nightly with ELD
 
 on:
-  # pull_request: {} # Uncomment only to test this WF file update.
+  # FIXME: comment
+  pull_request: {} # Uncomment only to test this WF file update.
```
quic-areg:
```suggestion
  # pull_request: {} # Uncomment only to test this WF file update.
```

quic-areg:
run: https://github.com/qualcomm/eld/actions/runs/26592304710/job/78354004738?pr=1034

.github/workflows/zephyr-nightly.yml
```diff
@@ -149,8 +152,12 @@ jobs:
         run: |
           cd zephyr
           source zephyr-env.sh
-          set -o pipefail
-          timeout -k 1s 5s west build -t run 2>&1 | tee run_output.log || true
+          # No pipe to tee: qemu_riscv32_xip uses -machine sifive_e which has
+          # no QEMU exit mechanism, so QEMU outlives `timeout` and would keep
+          # tee blocked. pkill below frees the CPU for the next step.
+          timeout -k 1s 5s west build -t run > run_output.log 2>&1 || true
+          pkill -KILL -f "qemu-system.*build/zephyr/zephyr.elf" 2>/dev/null || true
```
quic-seaswara, line 159: Is this an issue in upstream zephyr ?

I think there is something similar going on here like this https://github.com/zephyrproject-rtos/zephyr/issues/54330

@jonathonpenix can you pls review this too ?

quic-seaswara (approved):

.github/workflows/zephyr-nightly.yml
quic-areg: qemu_riscv32_xip happens to not exit on halt . That is the expected behavior of its QEMU machine model, so I don't think there's an upstream bug to fix here. Without this change, the hello_world step would hang because QEMU never exits.

quic-seaswara: I was expecting upstream zephyr testing to handle timeouts. 

Does upstream use a similar mechanism to build and kill them using pkill ?

jonathonpenix: I'm not that familiar with how this works, but I guess two thoughts:
1. If you manually/locally build and run the hello_world sample for this board (`west build...`) the usual `Ctrl a + x` to exit QEMU works fine like other boards
2. If you run this sample standalone through twister (add `xip` to the sample.yaml tags then I used this command: `west twister -c -W -p qemu_riscv32_xip -T samples/hello_world/ -x=CONFIG_LLVM_USE_LLD=y -x=CONFIG_COMPILER_RT_RTLIB=y --force-toolchain`) the test builds/runs/terminates normally as far as I can tell

So I'm assuming there is a standard way to do this (might have to dig through twister)?

jonathonpenix: Just quick asking Claude about how twister is terminating this (which, I didn't verify too closely so do with that what you will), sounds like twister is doing similar SIGTERM shenanigans under the hood.

So I guess this seems reasonable enough to me at a glance?

quic-seaswara: will merge it for now.

PR #1038, opened 2026-04-08
[package] package only what is needed

quic-seaswara:
use LLVM_DISTRIBUTION_COMPONENTS


quic-seaswara:
Reduces the size from 900MB to 200MB 

Example 

https://github.com/qualcomm/eld/actions/runs/24212682289

PR #1051, opened 2026-04-14
Fix: use actual content size for LMA overlap check

### LINKED ISSUE #621: Load address range overlap due to incorrect LMA assignment when VMA/LMA are not congruent wrt alignment
eld reports load address range overlap error due to incorrect LMA assignmetn when VMA / LMA are not congruent with respect to the section alignment. 

Reproducer:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
extern int u1;
int foo() { return 1; }
int bar() { return 3; }
int baz() { return 5; }
EOF

cat >script.t <<\EOF
SECTIONS {
  u1 = 0x1001;
  foo u1 : {
    *(.text.foo)
  }
  u2 = u1 + SIZEOF(foo);
  bar u2 : {
    *(.text.bar)
  }
  u3 = u2 + SIZEOF(bar);
  baz u3 : {
    *(.text.baz)
  }
}
EOF

LDs=(ld.eld)
SFs=(eld)

for i in "${!SFs[@]}"; do
done
```

**Output**
```
Error: section foo load address range overlaps with bar
         >>> foo range is [0x1010, 0x1029]
         >>> bar range is [0x1020, 0x102F]
Fatal: Linking had errors.
```

The error here is wrong because eld can create a valid image by correctly assigning the LMAs.

deepakshirkem:
## Problem
Fixes #621

ELD incorrectly emits a load address range overlap error when VMA/LMA are not congruent with respect to section alignment.

## Fix
Introduced `getContentSize()` which calculates actual content size  by summing only non-`FillFragment` sizes, excluding alignment padding.
The LMA overlap check now uses `getContentSize()` instead of  `sec->size()` to correctly detect real overlaps

## Testing
Added a lit test that verifies ELD does not emit a false load address  range overlap error when VMA/LMA are non-congruent due to alignment.

### DIFF
--- a/lib/Target/GNULDBackend.cpp
+++ b/lib/Target/GNULDBackend.cpp
@@ -893,8 +893,8 @@ void GNULDBackend::sizeDynNamePools() {
 
   if (GNUVerNeedSection) {
     if (DP->traceSymbolVersioning())
-      config().raise(Diag::trace_creating_symbol_versioning_fragment) <<
-          GNUVerNeedSection->name();
+      config().raise(Diag::trace_creating_symbol_versioning_fragment)
+          << GNUVerNeedSection->name();
     GNUVerNeedFragment *F = make<GNUVerNeedFragment>(GNUVerNeedSection);
     bool is32Bits = config().targets().is32Bits();
     if (is32Bits)
@@ -3557,8 +3557,16 @@ bool GNULDBackend::postLayout() {
   // ranges in the file.
   std::vector<SectionOffset> vmas;
   for (ELFSection *sec : outputSections) {
-    if (sec->size() > 0 && (sec->isAlloc() && !sec->isTLS()))
+    if (sec->size() > 0 && (sec->isAlloc() && !sec->isTLS())) {
+      // Warn if section address is not a multiple of aligment
+      // Same behaviour as LLD
+      if (sec->addr() % sec->getAddrAlign() != 0 &&
+          config().showLinkerScriptWarnings())
+        config().raise(Diag::warn_address_not_aligned)
+            << utility::toHex(static_cast<uint64_t>(sec->addr())) << sec->name()
+            << sec->getAddrAlign();
       vmas.push_back({sec, sec->addr()});
+    }
   }
   checkOverlap("virtual address", vmas, true);
 
@@ -3569,8 +3577,9 @@ bool GNULDBackend::postLayout() {
   std::vector<SectionOffset> lmas;
   for (ELFSection *sec : outputSections) {
     if (sec->size() > 0 && (sec->isAlloc() && !sec->isTLS()) &&
-        !sec->isNoBits())
+        !sec->isNoBits()) {
       lmas.push_back({sec, sec->pAddr()});
+    }
   }
   checkOverlap("load address", lmas, false);
   return true;
@@ -3580,7 +3589,6 @@ std::string GNULDBackend::rangeToString(uint64_t addr, uint64_t len) {
   return "[0x" + llvm::utohexstr(addr) + ", 0x" +
          llvm::utohexstr(addr + len - 1) + "]";
 }
-
 // Check whether sections overlap for a specific address range (file offsets,
 // load and virtual addresses).
 void GNULDBackend::checkOverlap(llvm::StringRef name,
@@ -3828,8 +3836,7 @@ GNULDBackend::postProcessing(llvm::FileOutputBuffer &pOutput) {
                          m_Module.getConfig().options().printTimingStats());
     MemoryRegion region =
         getFileOutputRegion(pOutput, 0, pOutput.getBufferSize());
-    eld::Expected<void> expEmit =
-        m_pSFrameFragment->emit(region, getModule());
+    eld::Expected<void> expEmit = m_pSFrameFragment->emit(region, getModule());
     ELDEXP_RETURN_DIAGENTRY_IF_ERROR(expEmit);
   }
   {
@@ -5576,7 +5583,8 @@ void GNULDBackend::assignOutputVersionIDs() {
       continue;
     if (!DynObjFile->hasSymbolVersioningInfo())
       continue;
-    uint16_t InputVersionID = DynObjFile->getSymbolVersionID(sym->getSymbolIndex());
+    uint16_t InputVersionID =
+        DynObjFile->getSymbolVersionID(sym->getSymbolIndex());
     if (InputVersionID == llvm::ELF::VER_NDX_LOCAL ||
         InputVersionID == llvm::ELF::VER_NDX_GLOBAL)
       continue;
--- a/include/eld/Diagnostics/DiagLDScript.inc
+++ b/include/eld/Diagnostics/DiagLDScript.inc
@@ -28,8 +28,7 @@ DIAG(promoting_bss_to_progbits, DiagnosticEngine::Note,
      "Section %0(%1) is before %2(%3), Promoting %0 to PROGBITS")
 DIAG(undefined_symbol_in_linker_script, DiagnosticEngine::Fatal,
      "%0: Error: undefined symbol '%1' referenced in expression")
-DIAG(error_sect_invalid, DiagnosticEngine::Error,
-     "%0: Invalid section '%1'")
+DIAG(error_sect_invalid, DiagnosticEngine::Error, "%0: Invalid section '%1'")
 DIAG(warn_section_no_segment, DiagnosticEngine::Warning,
      "Section %0 does not have segment assignment in linker script.")
 DIAG(fatal_segment_not_defined_ldscript, DiagnosticEngine::Fatal,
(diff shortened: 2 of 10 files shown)


deepakshirkem:
Hi  @quic-seaswara  / @parth-07, I have addressed the copyright issue by adding the license header  to the test input file. Could you please re-run the pipeline when  you get a chance?

Thank you ::(

deepakshirkem:
Hi @parth-07, Apologies for the CI failure. The failing test  `AtAddressPlacementStartStop` is ARM-specific and uses `.ARM.exidx` sections which are not available on my local RISCV64 setup.

I ran all overlap-related tests locally and they pass:
- LMAOverlapAlignment (our new test)
- OverlappingSectionsError  (existing test)  
- overlapping-sections.s  (existing test)
I will investigate the ARM-specific case and fix.

Thank you  ::));

deepakshirkem:
Hi @quic-seaswara / @parth-07, I have addressed the ARM test error. Could you please re-run the pipeline when you get a chance?

Thank you :))

lib/Target/GNULDBackend.cpp
```diff
@@ -3574,7 +3574,24 @@ bool GNULDBackend::postLayout() {
...
   }
-  checkOverlap("load address", lmas, false);
+
+  llvm::sort(lmas, [](const SectionOffset &a, const SectionOffset &b) {
+      return a.offset < b.offset;
+  });
+  for (size_t i = 1; i < lmas.size(); i++) {
+    SectionOffset &a = lmas[i - 1];
+    SectionOffset &b = lmas[i];
+    if (b.offset >= a.offset + getContentSize(a.sec))
```
quic-seaswara: why getContentSize ? the size of the section is good enough to use

lib/Target/GNULDBackend.cpp
```diff
@@ -3574,7 +3574,24 @@ bool GNULDBackend::postLayout() {
         !sec->isNoBits())
       lmas.push_back({sec, sec->pAddr()});
   }
-  checkOverlap("load address", lmas, false);
+
+  llvm::sort(lmas, [](const SectionOffset &a, const SectionOffset &b) {
+      return a.offset < b.offset;
+  });
+  for (size_t i = 1; i < lmas.size(); i++) {
```
quic-seaswara: I dont know what this is trying to do, we already support examining sections that overlap for lma. Can the checks be added at line 3577.

quic-seaswara (changes requested):

lib/Target/GNULDBackend.cpp
deepakshirkem: Hi @quic-seaswara, The reason for the custom loop is that `sec->size()` includes  FillFragment alignment padding which causes false overlap errors when VMA/LMA are not congruent with respect to section alignment.

For example in issue #621:
  foo: LMA=0x1004  sec->size()=0xb (includes 0x3 padding)
  bar: LMA=0x100c

  With sec->size(): 0x100c < 0x1004 + 0xb = 0x100f → false overlap! 
  With getContentSize(): 0x100c >= 0x1004 + 0x8 = 0x100c → no overlap! 

To use the existing checkOverlap() infrastructure at line 3577,  we would need to either:
1. Add a size field to SectionOffset and pass getContentSize()  when building the lmas vector
2. Modify checkOverlap() to accept a custom size function

I implemented approach 1 (custom loop with getContentSize) but I am also open to modifying checkOverlap() to accept a custom size function  and use it at line 3577, which would be cleaner and consistent with the existing infrastructure.

deepakshirkem:
Hi @quic-seaswara, I have refactored the fix to avoid adding a new function. The content size calculation is now inlined directly in the lmas population loop and uses the existing checkOverlap() infrastructure 
at line 3577. Please take a look and let me know if any further changes are needed.

Thank you ::((

lib/Target/GNULDBackend.cpp
quic-seaswara: Why is bar's LMA not 0x1004+0xb ?

quic-seaswara (changes requested):

lib/Target/GNULDBackend.cpp
deepakshirkem: Hi @quic-seaswara, 
bar's LMA is 0x100c because the linker script places bar at  u2 = u1 + SIZEOF(foo) = 0x1001 + 0xb = 0x100c.
If you think the correct fix should instead be in the LMA assignment  making bar's LMA start at 0x1004 + 0xb = 0x100f  (right after foo's LMA end) i am open to that approach too.

lib/Target/GNULDBackend.cpp
```diff
@@ -3571,8 +3571,23 @@ bool GNULDBackend::postLayout() {
...
+            continue;
+          for (Fragment *f : inputSec->getFragmentList())
+            if (!FillFragment::classof(f))
+              contentSize += f->size();
+        }
+        if (contentSize > 0 && contentSize <= size)
+          size = contentSize;
+      }
+      lmas.push_back({sec, sec->pAddr(), size});
+    }
```
quic-seaswara, line 3582: The root issue is that the linker is aligning the LMA seperately from the VMA. 

You can check the LMA that was assigned by bfd and eld here.

You can use ld.eld --no-check-sections to understand what is going on fyi.

To see the LMA,  you can objdump -h.

I prefer that we dont do anything else except that a warning that lld emits would be something that will be useful for eld.

deepakshirkem:
Hi @quic-seaswara , I have updated the fix based on suggestion: All overlap related tests pass locally. Please re-run the pipeline when you get a chance.

Thank you ::((

lib/Target/GNULDBackend.cpp
```diff
@@ -3559,8 +3559,16 @@ bool GNULDBackend::postLayout() {
   // ranges in the file.
   std::vector<SectionOffset> vmas;
   for (ELFSection *sec : outputSections) {
-    if (sec->size() > 0 && (sec->isAlloc() && !sec->isTLS()))
+    if (sec->size() > 0 && (sec->isAlloc() && !sec->isTLS())) {
+      // Warn if section address is not a multiple of aligment
+      // Same behaviour as LLD
+      if (sec->addr() % sec->getAddrAlign() != 0)
```
quic-seaswara: We might want to add this under -Wlinker-script

deepakshirkem:
Hi @quic-seaswara, The warning is now controlled by -Wlinker-script flag: Please re-run the pipeline when you get a chance!

Thank You ::))

lib/Target/GNULDBackend.cpp
```diff
@@ -3571,8 +3579,13 @@ bool GNULDBackend::postLayout() {
   std::vector<SectionOffset> lmas;
   for (ELFSection *sec : outputSections) {
     if (sec->size() > 0 && (sec->isAlloc() && !sec->isTLS()) &&
-        !sec->isNoBits())
+        !sec->isNoBits()) {
+      // Skip LMA overlap check when VMA is not aligned
+      // In this case LMA may differ from VMA due to alignment
+      if (sec->addr() % sec->getAddrAlign() != 0)
+        continue;
```
quic-seaswara: I think we are just hiding a bug in the linker by doing this. 

Look into why the LMA stays aligned when VMA is not aligned. We can keep LMA/VMA independent, but I am not sure if we do the former, how many tests would break.

deepakshirkem: Hi @quic-seaswara, I will investigate why LMA gets aligned separately when VMA is not aligned, and try to fix.

Thank You ::))

deepakshirkem:
Hi @quic-seaswara, I found the cause in `CreateProgramHeaders.hpp `where LMA was being aligned separately from VMA even when no AT() was specified.

The fix adds a condition to all three branches to only  align LMA when VMA is also aligned, ensuring LMA = VMA 
when the user does not explicitly specify AT().

I have tested this locally and all overlap-related test cases pass. Please let me know if this is the correct approach or if further investigation is needed.

Thank  You ::((

quic-seaswara (commented):
Please update createScriptProgramHeaders as well.

deepakshirkem:
Hi @quic-seaswara,  Updated createScriptProgramHeaders.

quic-seaswara:
> Please update createScriptProgramHeaders as well.

Need a test for the change as well. 

We require a test for every change unless its NFC. Just FYI.

quic-seaswara (changes requested):

quic-seaswara:
Also please clang-format your change

deepakshirkem:
Hi @quic-seaswara,  Addressed suggested changes Please review

quic-seaswara:
Move the test files to test/Common/standalone/linkerscript/

deepakshirkem:
Hi @quic-seaswara,  Moved both test directories to test/Common/standalone/linkerscript/

Thank You ::))

quic-seaswara:
Fix the clang-format issue and your patch will land!

Thanks for all the work!

deepakshirkem:
Hi @quic-seaswara, Fixed the clang-format issue.

Thank You ::))

quic-seaswara (approved):

PR #1052, opened 2026-04-14
Correctly compute .gnu.hash hashes for versioned symbols

parth-07:
This commit fixes the computation of .gnu.hash hash for versioned symbols. We were incorrectly using the versioned name for computing this hash. The .gnu.hash hash for versioned symbol must be computed using the unversioned name.

Reference: DT_GNU_HASH section in https://sourceware.org/gnu-gabi/program-loading-and-dynamic-linking.txt

Additionally, this commit also fixes an out-of-bound index in DynamicELFReader::readVerNeedSection.

Resolves #1033

### DIFF
--- a/lib/Fragment/GNUHashFragment.cpp
+++ b/lib/Fragment/GNUHashFragment.cpp
@@ -161,8 +161,14 @@ template <class ELFT> void GNUHashFragment<ELFT>::sortSymbols() {
   if (Mid == DynamicSymbols.end())
     return;
   for (auto I = Mid, E = DynamicSymbols.end(); I != E; ++I) {
-    Symbols.push_back(make<SymbolData>(*I, I - DynamicSymbols.begin(),
-                                       hashGnu(llvm::StringRef((*I)->name()))));
+#ifdef ELD_ENABLE_SYMBOL_VERSIONING
+    SymbolData *symData = make<SymbolData>(
+        *I, I - DynamicSymbols.begin(), hashGnu((*I)->getNonVersionedName()));
+#else
+    SymbolData *symData = make<SymbolData>(
+        *I, I - DynamicSymbols.begin(), hashGnu(llvm::StringRef((*I)->name())));
+#endif
+    Symbols.push_back(symData);
   }
 
   unsigned NBuckets = calcNBuckets(Symbols.size());
--- a/lib/Readers/DynamicELFReader.cpp
+++ b/lib/Readers/DynamicELFReader.cpp
@@ -274,7 +274,7 @@ eld::Expected<void> DynamicELFReader<ELFT>::readVerNeedSection() {
           reinterpret_cast<const typename ELFT::Vernaux *>(VernAuxBuf);
       uint16_t VersionIdentifier =
           CurVernAux->vna_other & llvm::ELF::VERSYM_VERSION;
-      if (VersionIdentifier > VerNeeds.size())
+      if (VersionIdentifier >= VerNeeds.size())
         VerNeeds.resize(VersionIdentifier + 1);
       VerNeeds[VersionIdentifier] = CurVernAux->vna_name;
       VernAuxBuf += CurVernAux->vna_next;
--- a/test/x86_64/linux/SymbolVersioning/UserFlowTests/Inputs/bar.c
+++ b/test/x86_64/linux/SymbolVersioning/UserFlowTests/Inputs/bar.c
@@ -1,0 +1,15 @@
+#include <stdio.h>
+
+/* Tell the linker: when this TU calls bar(), bind it to bar@FOO_1.0,
+   NOT to the default bar@@FOO_2.0.  This is the canonical way for a
+   library to pin itself to a specific symbol version from a dependency. */
+__asm__(".symver bar, bar@FOO_1.0");
+
+/* Forward-declare bar with the version we just pinned. */
+extern void bar(void);
+
+/* bar_wrapper is libbar's public API -- it delegates to the old bar. */
+void bar_wrapper(void) {
+    puts("libbar: bar_wrapper -> calling bar@FOO_1.0 from libfoo:");
+    bar();
+}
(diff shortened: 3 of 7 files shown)


lib/Fragment/GNUHashFragment.cpp
```diff
@@ -161,8 +161,14 @@ template <class ELFT> void GNUHashFragment<ELFT>::sortSymbols() {
   if (Mid == DynamicSymbols.end())
     return;
   for (auto I = Mid, E = DynamicSymbols.end(); I != E; ++I) {
-    Symbols.push_back(make<SymbolData>(*I, I - DynamicSymbols.begin(),
-                                       hashGnu(llvm::StringRef((*I)->name()))));
+#ifdef ELD_ENABLE_SYMBOL_VERSIONING
+    SymbolData *symData = make<SymbolData>(
+        *I, I - DynamicSymbols.begin(), hashGnu((*I)->getNonVersionedName()));
```
quic-seaswara, line 166: Is it possible to have two dynamic symbol entries to have the same name ?

lib/Fragment/GNUHashFragment.cpp
```diff
@@ -161,8 +161,14 @@ template <class ELFT> void GNUHashFragment<ELFT>::sortSymbols() {
...
+#else
+    SymbolData *symData = make<SymbolData>(
+        *I, I - DynamicSymbols.begin(), hashGnu(llvm::StringRef((*I)->name())));
+#endif
```
quic-seaswara, line 170: We need to consider sysv hash

lib/Target/GNULDBackend.cpp
```diff
@@ -5528,7 +5528,7 @@ void GNULDBackend::assignOutputVersionIDs() {
   // foo@@VER, assign the version index based on version node names.
   for (std::size_t i = 1, e = DynamicSymbols.size(); i < e; ++i) {
     ResolveInfo *R = DynamicSymbols[i];
-    if (llvm::isa<ELFDynObjectFile>(R->resolvedOrigin()))
+    if (llvm::isa<ELFDynObjectFile>(R->resolvedOrigin()) || R->isUndef())
```
quic-seaswara: is this intended ?

quic-seaswara (changes requested):

lib/Fragment/GNUHashFragment.cpp
parth-07: We already correctly compute sysv hash.

lib/Fragment/GNUHashFragment.cpp
parth-07: Yes.

lib/Target/GNULDBackend.cpp
parth-07: We need this change, but I guess it is not required for this issue. I have removed it from this PR and I will open a separate patch for it.

lib/Fragment/GNUHashFragment.cpp
quic-seaswara: Please add the test to use sysv hash and get it to run.

lib/Fragment/GNUHashFragment.cpp
quic-seaswara: Isn't this wrong behavior, two dynamic symbols having the same name, will cause the loader to choose only one.

quic-seaswara (changes requested):

lib/Fragment/GNUHashFragment.cpp
parth-07: test/x86_64/linux/SymbolVersioning/UserFlowTests/UserFlowTests.test verifies both sysv hash and gnu hash.

lib/Fragment/GNUHashFragment.cpp
parth-07: No, this is not a wrong behavior. Loader checks both the symbol name and the version name.

parth-07: Both lld and bfd only stores the unversioned name in the .dynsym table.

quic-seaswara: We need to have a test for this. If you want to make this a follow up, I am fine with that too.

lib/Readers/DynamicELFReader.cpp
```diff
@@ -274,7 +274,7 @@ eld::Expected<void> DynamicELFReader<ELFT>::readVerNeedSection() {
           reinterpret_cast<const typename ELFT::Vernaux *>(VernAuxBuf);
       uint16_t VersionIdentifier =
           CurVernAux->vna_other & llvm::ELF::VERSYM_VERSION;
-      if (VersionIdentifier > VerNeeds.size())
+      if (VersionIdentifier >= VerNeeds.size())
```
quic-seaswara, line 277: Why >= ?

lib/Fragment/GNUHashFragment.cpp
parth-07: I am not sure I understand. We already have several tests where we print the dynamic symbol table. If we were storing the version name in the dynamic symbol table itself, then the readelf would show the version twice per symbol.

lib/Readers/DynamicELFReader.cpp
parth-07: This is because at line 278, we are accessing `VerNeeds[VersionIdentifier]`. If `VersionIdentifier` is the same as `VerNeeds.size()` then `VerNeeds[VersionIdentifier]` will be an out-of-bound access.

I have mentioned this in the commit message as well:

> Additionally, this commit also fixes an out-of-bound index in DynamicELFReader::readVerNeedSection.

lib/Fragment/GNUHashFragment.cpp
quic-seaswara: We never have dynsym entries that use the same symbol name. If symbol versioning with stripping of `@` creates duplicate entries, we need a test.

quic-seaswara (approved):

PR #1053, opened 2026-04-14
[build] Fix Build with BUILD_SHARED_LIBS=On

lenary:
llvm::WriteBitcodeToFile is in the LLVMBitWriter library. This needs to explicitly be included in the link libraries when building with `BUILD_SHARED_LIBS=On`, or there will be a link failure.

### DIFF
--- a/lib/Object/CMakeLists.txt
+++ b/lib/Object/CMakeLists.txt
@@ -22,4 +22,5 @@ target_link_libraries(
           ELDSupport
           ELDTarget
           LLVMSupport
+          LLVMBitWriter
           LLVMLTO)


quic-seaswara (approved):

quic-seaswara:
Is this good to merge ? 

probably we need to have a nightly workflow to build with shared libs on, that can be used to figure out if there is a breakage.

lenary:
> Is this good to merge ?

Yes, i was just finishing a local rebase/rebuild before I marked it as ready to review/merge.

> 
> probably we need to have a nightly workflow to build with shared libs on, that can be used to figure out if there is a breakage.

I wouldn't bother. BUILD_SHARED_LIBS=On is supposed to only be for developers (rather than for distribution), so it seems a waste of resources as long as you're willing to take manual fixes.

I don't know if using `LLVM_BUILD_LLVM_DYLIB=On` would catch or hit the same issues. This is an option intended for distributors. CI time could be put on this, but it does seem to me to be lower priority than many other things we could spend CI time on.

PR #1062, opened 2026-04-16
Produce diagnostic for segment assignments when no PHDRS present.

### LINKED ISSUE #348: Diagnose segment assignments when PHDRS isn't present
**Description**
eld and lld both don't seem to produce a diagnostic if sections are assigned a segment when PHDRS aren't being used. bfd seems to issue an error here and I can't see a non-erroneous usage for this, so it seems like it might be nice to issue an error here as well.

**Reproducer**
```
cat > foo.c << '!'
void foo(void) { return; }
!
cat > script.t << '!'
SECTIONS {
  .text : { *(.text) } :missing_phdr
}
!

clang --target=riscv32-unknown-elf -c foo.c
# eld and lld don't issue any diagnostics here
ld.eld -melf32lriscv -T script.t foo.o
ld.lld -T script.t foo.o
# riscv32-unknown-elf-ld.bfd: section `.text' assigned to non-existent phdr `missing_phdr'
riscv32-unknown-elf-ld.bfd -T script.t foo.o
```

quic-shaksuma:
Produce a diagnostic if sections are assigned to a segment when PHDRS are not being used in the linker script.

Fixes #348

### DIFF
--- a/lib/Script/OutputSectDesc.cpp
+++ b/lib/Script/OutputSectDesc.cpp
@@ -44,8 +44,6 @@ OutputSectDesc::OutputSectDesc(const std::string &PName)
   OutputSectDescEpilog.FillExpression = nullptr;
 }
 
-
-
 void OutputSectDesc::dump(llvm::raw_ostream &Outs) const {
   Outs << Name << "\t";
 
@@ -285,6 +283,15 @@ eld::Expected<void> OutputSectDesc::activate(Module &CurModule) {
   // Assignment in an output section
   OutputSectCmds Assignments;
   const LinkerScript &Script = CurModule.getLinkerScript();
+
+  if (OutputSectDescEpilog.hasPhdrs() && !Script.phdrsSpecified()) {
+    for (const auto &PhdrNameToken : OutputSectDescEpilog.phdrs()->tokens()) {
+      return std::make_unique<plugin::DiagnosticEntry>(
+          plugin::DiagnosticEntry(Diag::error_phdrs_not_specified_ldscript,
+                                  {getContext(), Name, PhdrNameToken->name()}));
+    }
+  }
+
   if (OutputSectDescEpilog.OutputSectionMemoryRegion) {
     eld::Expected<eld::ScriptMemoryRegion *> MemRegion = Script.getMemoryRegion(
         OutputSectDescEpilog.OutputSectionMemoryRegion->name(), getContext());
--- a/include/eld/Diagnostics/DiagLDScript.inc
+++ b/include/eld/Diagnostics/DiagLDScript.inc
@@ -33,6 +33,8 @@ DIAG(warn_section_no_segment, DiagnosticEngine::Warning,
      "Section %0 does not have segment assignment in linker script.")
 DIAG(fatal_segment_not_defined_ldscript, DiagnosticEngine::Fatal,
      "%0: Segment %1 is not defined in linker script.")
+DIAG(error_phdrs_not_specified_ldscript, DiagnosticEngine::Error,
+     "%0: Section %1 assigned to segment %2 but PHDRS not specified.")
 DIAG(assert_failed, DiagnosticEngine::Error, "Assertion failed %0")
 DIAG(error_printcmd, DiagnosticEngine::Error, "%0: PRINT: %1")
 DIAG(linker_script_uses_phdrs_no_sections, DiagnosticEngine::Error,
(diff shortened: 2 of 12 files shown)


lib/Target/CreateProgramHeaders.hpp
```diff
@@ -28,6 +28,18 @@ bool GNULDBackend::createProgramHdrs() {
...
+    for (auto &OutSection : script.sectionMap()) {
+      if (!OutSection->epilog().hasPhdrs())
+        continue;
+      for (const auto &PhdrNameToken : *OutSection->epilog().phdrs()) {
+        config().raise(Diag::fatal_segment_not_defined_ldscript)
+            << OutSection->name() << PhdrNameToken->name();
+        return false;
+      }
+    }
+  }
```
quic-seaswara: This needs to be in `createSegmentsFromLinkerScript`

quic-seaswara (changes requested):

lib/Target/CreateProgramHeaders.hpp
quic-shaksuma: Move it out of CreateProgramHeaders.hpp

lib/Target/GNULDBackend.cpp
```diff
@@ -2269,13 +2269,31 @@ void GNULDBackend::evaluateAssignmentsAtEndOfOutputSection(
   }
 }
 
+bool GNULDBackend::checkSegmentAssignmentWhenNoPHDRS() {
+  LinkerScript &ldscript = m_Module.getScript();
+  if (!ldscript.phdrsSpecified()) {
+    for (auto &OutSection : ldscript.sectionMap()) {
+      if (!OutSection->epilog().hasPhdrs())
+        continue;
```
quic-seaswara: Why can't this be detected during parsing ? or when the OutputSectDesc is activated.

quic-seaswara (changes requested):

lib/Script/OutputSectDesc.cpp
```diff
@@ -285,6 +283,15 @@ eld::Expected<void> OutputSectDesc::activate(Module &CurModule) {
   // Assignment in an output section
   OutputSectCmds Assignments;
   const LinkerScript &Script = CurModule.getLinkerScript();
+
+  if (OutputSectDescEpilog.hasPhdrs() && !Script.phdrsSpecified()) {
+    for (const auto &PhdrNameToken : OutputSectDescEpilog.phdrs()->tokens()) {
+      return std::make_unique<plugin::DiagnosticEntry>(
+          plugin::DiagnosticEntry(Diag::fatal_segment_not_defined_ldscript,
```
quic-seaswara: print the linker script file

quic-shaksuma: Done

quic-seaswara (approved):

lib/Target/GNULDBackend.cpp
jonathonpenix: Sorry for the very very belated drive-by comment, I'm finally clearing out my github notifications and saw this closed an old issue filed (thanks Shakti for working on this!).

I'm not familiar with what it means for the OutputSectDesc to be activated but FWIW it seems like this works a bit differently in bfd/lld (I guess they handle it "later"? or, by the time they've seen or processed all input linker scripts?) than eld. Something like the test below passes for bfd/lld but eld errors:

```
cat > test.c << '!'
int data = 1;
int bss;
int main() {}
!

cat > script.t << '!'
SECTIONS {
  .text : { *(.text) } :HDR
  .data : { *(.data) } :HDR
  .bss : { *(.bss) } :HDR
}
!

cat > phdrs.t << '!'
PHDRS {
  HDR PT_LOAD;
}
!

clang --target=riscv32-unknown-elf test.c -c

# bfd/lld seem to be fine with it
riscv32-unknown-elf-ld.bfd test.o -T script.t -T phdrs.t
ld.lld test.o -T script.t -T phdrs.t

# eld errors. You can flip the order of the scripts to make it work, not sure if this is
# intended to be a requirement though.
# Error: script.t: Section .text assigned to segment HDR but PHDRS not specified.
# Fatal: Linking had errors.
ld.eld test.o -T script.t -T phdrs.t
```

Not sure if or how much this matters in practice. bfd doesn't seem to handle ex: `MEMORY` very consistently in this regard either, so might not be worth dealing with.

PR #1070, opened 2026-04-17
[ELD] Update ASCIZ syntax and map-file output

quic-seaswara:
Change linker-script ASCIZ parsing from parenthesized form to quoted form (ASCIZ "...").

Refresh docs and tests to match the new syntax and map expectations.


quic-areg:
We don't handle escape sequences like: `\n` `\r` `\t` `\101` `\\`

Any linker script using ASCIZ with escape sequences will be 1 byte longer and contain a literal backslash with eld compared to GNU ld.

quic-seaswara:
> We don't handle escape sequences like: `\n` `\r` `\t` `\101` `\\`
> 
> Any linker script using ASCIZ with escape sequences will be 1 byte longer and contain a literal backslash with eld compared to GNU ld.

Thanks for letting me know about this. Will make a follow up patch.

PR #1072, opened 2026-04-17
Add -Wdeprecated to show deprecated warnings

### LINKED ISSUE #336: Improve error messages from assignment expressions
```
$ cat syms.sym
ADDITIONAL_INPUT=0x600df00d;
```

```
$ ld.eld -T syms.sym
Error: syms.sym:1: unknown directive: ADDITIONAL_INPUT=0x600df00d
>>> ADDITIONAL_INPUT=0x600df00d;
>>> ^
Fatal: Linker script syms.sym has errors, Please check
Fatal: Linking had errors.
```

quic-shaksuma:
Add diagnostics for assignment expression having improper white-spaces around the assignment operator.
Add -Wdeprecated to show the warnings of these kind.
Update the documentation to include the exception for MEMORY command.

Fixes  #336


lib/ScriptParser/ScriptParser.cpp
```diff
@@ -471,13 +470,63 @@ bool ScriptParser::isValidSymbolName(StringRef S) {
...
+}
+
+bool ScriptParser::isAssignmentOperator(StringRef Op) const {
+  for (StringRef AssignmentOp : AssignmentOps)
+    if (AssignmentOp == Op)
+      return true;
+  return false;
+}
+
+bool ScriptParser::isValidAssignment(llvm::StringRef Op) {
```
parth-07, line 496: Can we please rename the function to something like `reportInvalidAssignment`? The name `isValidAssignment` does not indicate that this function has a side affect and it also does not return false for invalid assignments as the name suggests it should.

parth-07 (changes requested):
Thank you for this patch. There are a few cases that does not work quite right with this change.

1)
```bash
#!/usr/bin/env bash

cat >script.t <<\EOF
u =v;
PROVIDE(v =11);
EOF

ld.eld -m armelf -o a.out /dev/null -T script.t -Wdeprecated
ld.bfd -o a.bfd.out /dev/null -T script.t
```

We only want to warn/error when `=` does not have a space with the LHS symbol.

2)
```bash
#!/usr/bin/env bash

cat >script.t <<\EOF
a = 0;
b = 11;
a += b; // +=
EOF

ld.eld -m armelf -o a.out /dev/null -T script.t -Wdeprecated
ld.bfd -o a.bfd.out /dev/null -T script.t
```

This incorrectly reports the warning:
```
Warning: script.t:3: missing whitespace around assignment operator '+=' in assignment expression is deprecated
>>> a += b; // +=
>>> 
```

3) 
```bash
#!/usr/bin/env bash

cat >script.t <<\EOF
SECTIONS {
  u=v : {
    *(.text)
  }
}
EOF

ld.eld -m armelf -o a.out /dev/null -T script.t -Wdeprecated
ld.bfd -o a.bfd.out /dev/null -T script.t
```

This linker script works fine with GNU ld and the current eld but fails with this patch. 

If we are okay with not allowing `=` in symbol names and section names, then can you please take a look at this patch: https://github.com/qualcomm/eld/pull/295. It parses the section and symbol names in `LexState::Expr` if they contain `=`. We can report the warning easily as well by checking if the `op` token is immediately after the LHS symbol token. This PR avoids many of the complication such as `getOpWithoutSpace` and splitting the token based on operators.

PR #1074, opened 2026-04-17
[linker-script] ScriptParser: support ASCIZ escapes and reject hex es…

### LINKED ISSUE #1073: ASCIZ : We don't handle escape sequences like: `\n` `\r` `\t` `\101` `\\`
We don't handle escape sequences like: `\n` `\r` `\t` `\101` `\\`

Any linker script using ASCIZ with escape sequences will be 1 byte longer and contain a literal backslash with eld compared to GNU ld.

_Originally posted by @quic-areg in https://github.com/qualcomm/eld/issues/1070#issuecomment-4271350062_

quic-seaswara:
Add ASCIZ string parsing for linker script output section data.

Supported escapes: \n, \r, \t, and 1-3 digit octal escapes (up to 0xFF).

Reject hex escapes (\x/\X) with a clear parser error as unsupported.

Resolves #1073

### DIFF
--- a/docs/userguide/documentation/linker_script.rst
+++ b/docs/userguide/documentation/linker_script.rst
@@ -1092,8 +1092,9 @@ ELD supports the following output section data commands:
      - Emit eight bytes (64 bits) as a signed value.
    * - ``ASCIZ "string"``
      - Emit a null-terminated ASCII string. The string literal (enclosed in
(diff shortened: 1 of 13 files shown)


quic-areg:
Why are these being merged without review? 

`\\` doesn't work.

PR #1075, opened 2026-04-18
String list iterator range

quic-seaswara:


docs/userguide/documentation/options/options.rst
```diff
@@ -124,6 +124,14 @@ useful for debugging and enforcing correctness:
 - ``-Wlinker-script-memory`` enables warnings for suspicious ``MEMORY`` setups
   (for example, zero-sized regions or regions that end up unused).
 
+PIE options
```
parth-07, line 127: We do not need to explicitly document the command-line options. The command-line option documentation is automatically generated using the TableGen files. Currently here we have duplicate --pie/--no-pie documentation: https://qualcomm.github.io/eld/documentation/options/options.html#

docs/userguide/documentation/options/options.rst
```diff
@@ -207,6 +215,9 @@ The following ``-z`` keywords are supported by ELD:
   from any other data. This is a no-op when a linker script with a ``SECTIONS``
   command is used.
 
+``-z separate-loadable-segments``
```
parth-07, line 218: We have duplicate docs for `-z separate-loadable-segments`: https://qualcomm.github.io/eld/documentation/options/options.html#.

quic-seaswara: We need to plan to remove that handling from HelpText in the python generated documentation,  because that is intended for command line usage vs documentation which is much more in detail. 

@parth-07  can you please raise a PR to not eat HelpText in the generated documentation?

docs/userguide/documentation/options/options.rst
quic-seaswara: I will remove the --pie and --no-pie options separately documented. Thanks for noticing that.

docs/userguide/documentation/options/options.rst
parth-07: > because that is intended for command line usage vs documentation which is much more in detail

We have added the docs/userguide/CommandLineOptionsSupplements framework to extend the documentation from HelpText. The same framework can be extended to replace the docs as well, instead of extending it. This being said, for most options, taking docs from HelpText is a good default. This allows the docs to always stay-upto date.

quic-seaswara: I would like to get the documentation generation completely moved to the supplement framework. 

May be if we can do some heuristics and use the HelpText if the HelpText is one line. If HelpText is more than a line, then we can add a warning/error and get the developer to add the documentation in the supplement.

parth-07: > If HelpText is more than a line, then we can add a warning/error and get the developer to add the documentation in the supplement.

I don't think that more than one line in HelpText is always bad for the command-line option docs. For things such as `--trace` documentation, taking multi-line documentation from `HelpText` definitely seems useful to me.

I think we should take this case-by-case basis. As a good default, taking HelpText is useful because it helps to keep the docs upto-date.

parth-07: Additionally, we will need to convert the rst files in `docs/userguide/CommandLineOptionsSupplements` to JSON, or add a JSON configuration intermediate that selects between replacing / extending the HelpText documentation. Currently, we unconditionally take the rst file to extend the HelpText.

(Not necessarily JSON configuration, could be any configuration file that allows us to pass replace/extend information to the command-line docs generator framework)

quic-seaswara: > > If HelpText is more than a line, then we can add a warning/error and get the developer to add the documentation in the supplement.
> 
> I don't think that more than one line in HelpText is always bad for the command-line option docs. For things such as `--trace` documentation, taking multi-line documentation from `HelpText` definitely seems useful to me.
> 
> I think we should take this case-by-case basis. As a good default, taking HelpText is useful because it helps to keep the docs upto-date.

HelpText is for command line or probably even man pages, not for sphinx doc.

PR #1077, opened 2026-04-18
Add LLVM clang-format policy checks and helper script

quic-seaswara:
- Add repository .clang-format based on LLVM style

- Add PR workflow to enforce clang-format on changed C/C++ files

- Add etc/bash/eld_clang_format_helpers.sh with check/fix helpers

### DIFF
--- a/.github/workflows/clang-format-pr.yml
+++ b/.github/workflows/clang-format-pr.yml
@@ -1,0 +1,76 @@
+name: Clang Format Check
+
+on:
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  clang-format:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout PR
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: Install clang-format
+        run: |
+          set -euo pipefail
+          sudo apt-get update
+          sudo apt-get install -y wget gpg lsb-release
+          wget -qO- https://apt.llvm.org/llvm-snapshot.gpg.key \
+            | gpg --dearmor \
+            | sudo tee /usr/share/keyrings/llvm-archive-keyring.gpg >/dev/null
+          UBUNTU_CODENAME="$(lsb_release -cs)"
+          echo "deb [signed-by=/usr/share/keyrings/llvm-archive-keyring.gpg] http://apt.llvm.org/${UBUNTU_CODENAME}/ llvm-toolchain-${UBUNTU_CODENAME} main" \
+            | sudo tee /etc/apt/sources.list.d/llvm.list >/dev/null
+          sudo apt-get update
+          sudo apt-get install -y clang-format
+          clang-format --version
+
+      - name: Check formatting of changed files
+        shell: bash
+        run: |
+          set -euo pipefail
+          GREEN=$'\033[0;32m'
+          RED=$'\033[0;31m'
+          RESET=$'\033[0m'
+          CLANG_FORMAT_BIN=clang-format
+
+          BASE_REF="${{ github.base_ref }}"
+          git fetch origin "${BASE_REF}" --depth=1
+          "$CLANG_FORMAT_BIN" --version
+
+          mapfile -t FILES < <(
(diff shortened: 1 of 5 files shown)


parth-07:
@quic-seaswara The clang-format checker added by this PR check all the changed files in the current PR instead of the changes of the current PR. This causes false positive when the modified files original content do not follow clang-format style.

.github/workflows/clang-format-pr.yml
```diff
@@ -0,0 +1,76 @@
...
+        run: |
+          set -euo pipefail
+          sudo apt-get update
+          sudo apt-get install -y wget gpg lsb-release
+          wget -qO- https://apt.llvm.org/llvm-snapshot.gpg.key \
+            | gpg --dearmor \
+            | sudo tee /usr/share/keyrings/llvm-archive-keyring.gpg >/dev/null
+          UBUNTU_CODENAME="$(lsb_release -cs)"
+          echo "deb [signed-by=/usr/share/keyrings/llvm-archive-keyring.gpg] http://apt.llvm.org/${UBUNTU_CODENAME}/ llvm-toolchain-${UBUNTU_CODENAME} main" \
+            | sudo tee /etc/apt/sources.list.d/llvm.list >/dev/null
```
parth-07, line 28: Why do we need to do all this? Wouldn't `sudo apt-get install -y clang-format` work without these?

quic-seaswara: We are using the latest clang-format. apt-get install installs clang-format 18 vs this one install clang-format-23. 

clang-format keeps evolving and we should try to move to the latest rules as much as possible.

This is another reason why we are moving to clang formatting code changes for the PR. Those formatting in a lot of cases are minimal. 

I encourage we move in this direction and get all the code eventually touched by the code clang-formatted.

That said, clang-tidy checks are disabled and we need to move to the direction of using guidelines followed by LLVM for all new code we write. This will be something I plan to gate new changes.

parth-07: > I encourage we move in this direction and get all the code eventually touched by the code clang-formatted.

Is it possible to just test for the changed code in the clang-format checker using the mechanism used by `git clang-format`? If not, then we should have a separate patch for formatting the entire codebase and add that commit to `.git-blame-ignore-revs`.

quic-seaswara:
> @quic-seaswara The clang-format checker added by this PR check all the changed files in the current PR instead of the changes of the current PR. This causes false positive when the modified files original content do not follow clang-format style.

Yes this was the intent. In a year from now we can make clang-format run just on the diff. 

We are conveniently using whatever clang-format appears to be on our machines. At times this has become error prone. We now have a mix of clang formatted code and non clang formatted code in the repo.

The clang-format changes should be very minimal and we have helper scripts to handle it without much pain. 

Please create another commit in the PR to clang-format code, that is outside of your committed changes.

parth-07:
> Please create another commit in the PR to clang-format code, that is outside of your committed changes.

This will cause issues with git blame because if we follow this way then we would not be able to take advantage of `.git-blame-ignore-revs`.

What do you think about this?

> then we should have a separate patch for formatting the entire codebase and add that commit to .git-blame-ignore-revs.

.github/workflows/clang-format-pr.yml
quic-seaswara: Yes in the long run. Note clang-format keeps evolving.

quic-seaswara:
> > Please create another commit in the PR to clang-format code, that is outside of your committed changes.
> 
> This will cause issues with git blame because if we follow this way then we would not be able to take advantage of `.git-blame-ignore-revs`.
> 
> What do you think about this?
> 
> > then we should have a separate patch for formatting the entire codebase and add that commit to .git-blame-ignore-revs.

I see what you are saying, do you all use git blame ?

parth-07:
> I see what you are saying, do you all use git blame ?

git blame is definitely useful. For formatting the patch changes, I use `git clang-format HEAD~`. It only updates formatting of the code that is modified in the latest commit.

PR #1088, opened 2026-04-22
Fix: Emit diagnostic when LMA is unaligned

### LINKED ISSUE #444: eld should emit a diagnostic when LMA is unaligned
Currently eld does not emit a diagnostic if the LMA is unaligned 

`
cat > 1.t << \!
SECTIONS {
  .foo (0x208) : AT(0x201) { *(.text.foo) }
}
!

cat > 1.c << \!
int foo() { return 0; }
!

hexagon-clang -c 1.c  -ffunction-sections
hexagon-link 1.o -T 1.t -Wlinker-script 
`

shauryarane05:
Currently eld does not warn when a section's load memory address (LMA) is not a multiple of its alignment, like when AT() specifies an unaligned address.Added a warning in GNULDBackend::postLayout() to catch when the LMA isn’t properly aligned , mirroring the existing VMA-alignment check added in e09c562f.

Fixes #444

### DIFF
--- a/lib/Target/GNULDBackend.cpp
+++ b/lib/Target/GNULDBackend.cpp
@@ -3576,9 +3576,14 @@ bool GNULDBackend::postLayout() {
   // in the output file.
   std::vector<SectionOffset> lmas;
   for (ELFSection *sec : outputSections) {
-    if (sec->size() > 0 && (sec->isAlloc() && !sec->isTLS()) &&
-        !sec->isNoBits()) {
-      lmas.push_back({sec, sec->pAddr()});
+    if (sec->size() > 0 && (sec->isAlloc() && !sec->isTLS())) {
+      if (sec->pAddr() % sec->getAddrAlign() != 0 &&
+          config().showLinkerScriptWarnings())
+        config().raise(Diag::warn_lma_not_aligned)
+            << utility::toHex(static_cast<uint64_t>(sec->pAddr()))
+            << sec->name() << sec->getAddrAlign();
+      if (!sec->isNoBits())
+        lmas.push_back({sec, sec->pAddr()});
     }
   }
   checkOverlap("load address", lmas, false);
--- a/lib/Target/GNULDBackend.cpp
+++ b/lib/Target/GNULDBackend.cpp
@@ -3363,9 +3363,14 @@ bool GNULDBackend::postLayout() {
   // in the output file.
   std::vector<SectionOffset> lmas;
   for (ELFSection *sec : outputSections) {
-    if (sec->size() > 0 && (sec->isAlloc() && !sec->isTLS()) &&
-        !sec->isNoBits()) {
-      lmas.push_back({sec, sec->pAddr()});
+    if (sec->size() > 0 && (sec->isAlloc() && !sec->isTLS())) {
+      if (sec->pAddr() % sec->getAddrAlign() != 0 &&
+          config().showLinkerScriptWarnings())
+        config().raise(Diag::warn_lma_not_aligned)
+            << utility::toHex(static_cast<uint64_t>(sec->pAddr()))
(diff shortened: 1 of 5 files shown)


shauryarane05:
fixed a clang-format issue flagged by CI (cosmetic line-break only, no logic change)

test/Common/standalone/linkerscript/LMAUnaligned/Inputs/1.c
```diff
@@ -0,0 +1,6 @@
+//===- 1.c ----------------------------------------------------------------===//
```
quic-areg: License can be removed; they aren't needed for test files.

test/Common/standalone/linkerscript/LMAUnaligned/LMAUnaligned.test
```diff
@@ -0,0 +1,10 @@
+#---LMAUnaligned.test--------------------------- Executable -----------------#
+#BEGIN_COMMENT
+# Test that ELD emits a warning diagnostic when the LMA specified via AT()
+# is not a multiple of the section alignment. (issue #444)
```
quic-areg: I think the `(issue #444)` can be removed.

lib/Target/GNULDBackend.cpp
```diff
@@ -3578,6 +3578,11 @@ bool GNULDBackend::postLayout() {
   for (ELFSection *sec : outputSections) {
     if (sec->size() > 0 && (sec->isAlloc() && !sec->isTLS()) &&
         !sec->isNoBits()) {
```
quic-areg: Do we want this warning for `NOBITS` sections too? The VMA loop above (line 3559) warns on `NOBITS`. It also does not skip sections with zero alignment. I think the two cases should probably match.

test/Common/standalone/linkerscript/LMAUnaligned/Inputs/1.c
shauryarane05: done! will keep that in mind

test/Common/standalone/linkerscript/LMAUnaligned/LMAUnaligned.test
shauryarane05: removed

lib/Target/GNULDBackend.cpp
shauryarane05: Good point, thanks! Updated the LMA loop to mirror the VMA one and dropped !isNoBits() and the getAddrAlign() > 0 guard from the warning path. NOBITS is still skipped for the overlap push since it has no physical content.

quic-seaswara (approved):

shauryarane05:
Does this need changes or can it be merged?

quic-areg (approved):

quic-seaswara:
@shauryarane05 can you resolve the conflicts and update the PR ?

quic-seaswara:
can you push the PR after resolving the conflicts ?

shauryarane05:
Sorry the pr slipped my mind , resolved the merge conflicts

jonathonpenix:
As a higher level question, is the LMA not being aligned actually problematic in general? I get that something like `AT(0x201)` is often (likely?) user error, but it doesn't seem like there's any general requirement that the LMA be aligned in a particular way. My understanding is GNU ld, for example, will often not align LMAs by default.

Along the same lines, there's ex: `ALIGN_WITH_INPUT` where (IIUC) the point is that the LMA should get the same amount of padding as the VMA, which can create sections where the LMA isn't aligned (but IIUC that's desirable to be able to bulk memcpy from LMA to VMA, etc.). Are people going to start getting warnings in those cases (is that expected/desirable?)?

PR #1095, opened 2026-04-22
Fix: avoid creating unnecessary .got.plt section

deepakshirkem:
### Problem
Fixes #1030

ELD unnecessarily creates `.got.plt` section when only `.got`  section is required for a static executable.

### Fix
The `.got.plt` section is only needed when dynamic linking is involved. Added `isCodeDynamic()` check before creating `GOTPLT0`  entry in `createGOT()` across all targets:
- AArch64
- ARM
- RISCV
- Hexagon
- x86_64

### Testing
Added test `UnnecessaryGOTPLT` that verifies `.got.plt` is not  created for static executables that only require `.got`.

cc @quic-seaswara @parth-07

### DIFF
--- a/lib/Target/AArch64/AArch64LDBackend.cpp
+++ b/lib/Target/AArch64/AArch64LDBackend.cpp
@@ -721,8 +721,9 @@ AArch64GOT *AArch64LDBackend::createGOT(GOT::GOTType T,
                         config().options().traceSymbol(*R)) ||
                        m_Module.getPrinter()->traceDynamicLinking()))
     config().raise(Diag::create_got_entry) << R->name();
-  // If we are creating a GOT, always create a .got.plt.
-  if (!getGOTPLT()->hasFragments()) {
+  // Only create .got.plt when dynamic linking is involved
+  if (!getGOTPLT()->hasFragments() &&
+   (config().isCodeDynamic() || T == GOT::GOTPLT0)) {
     // TODO: This should be GOT0, not GOTPLT0.
     LDSymbol *Dynamic = m_Module.getNamePool().findSymbol("_DYNAMIC");
     AArch64GOTPLT0::Create(getGOTPLT(),
--- a/lib/Target/AArch64/AArch64LDBackend.cpp
+++ b/lib/Target/AArch64/AArch64LDBackend.cpp
@@ -722,8 +722,9 @@ AArch64GOT *AArch64LDBackend::createGOT(GOT::GOTType T,
                        m_Module.getPrinter()->traceDynamicLinking()))
     config().raise(Diag::create_got_entry)
         << GOT::getGOTTypeAsStr(T) << R->name();
-  // If we are creating a GOT, always create a .got.plt.
-  if (!getGOTPLT()->hasFragments()) {
+  // Only create .got.plt when dynamic linking is involved
+  if (!getGOTPLT()->hasFragments() &&
+      (config().isCodeDynamic() || T == GOT::GOTPLT0)) {
     // TODO: This should be GOT0, not GOTPLT0.
     LDSymbol *Dynamic = m_Module.getNamePool().findSymbol("_DYNAMIC");
     AArch64GOTPLT0::Create(getGOTPLT(),
--- a/test/Common/standalone/UnnecessaryGOTPLT/Inputs/1.c
+++ b/test/Common/standalone/UnnecessaryGOTPLT/Inputs/1.c
@@ -1,0 +1,2 @@
+int u = 1;
+int foo() { return u; }
--- a/lib/Target/ARM/ARMLDBackend.cpp
+++ b/lib/Target/ARM/ARMLDBackend.cpp
@@ -149,7 +149,7 @@ void ARMGNULDBackend::initTargetSymbols() {
   if (m_Module.getScript().linkerScriptHasSectionsCommand())
     return;
 
-  const NamePool& NP = m_Module.getNamePool();
+  const NamePool &NP = m_Module.getNamePool();
   SymbolName = "__exidx_start";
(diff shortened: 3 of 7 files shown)


test/Common/standalone/UnnecessaryGOTPLT/Inputs/1.c
```diff
@@ -0,0 +1,7 @@
+//===- 1.c ----------------------------------------------------------------===//
+// Part of the eld Project, under the BSD License
+// See https://github.com/qualcomm/eld/LICENSE.txt for license information.
+// SPDX-License-Identifier: BSD-3-Clause
+//===----------------------------------------------------------------------===//
```
quic-seaswara: Remove this license from test file.

lib/Target/AArch64/AArch64LDBackend.cpp
```diff
@@ -694,8 +694,8 @@ AArch64GOT *AArch64LDBackend::createGOT(GOT::GOTType T,
                         config().options().traceSymbol(*R)) ||
                        m_Module.getPrinter()->traceDynamicLinking()))
     config().raise(Diag::create_got_entry) << R->name();
-  // If we are creating a GOT, always create a .got.plt.
-  if (!getGOTPLT()->hasFragments()) {
+  // Only create .got.plt when dynamic linking is involved
+  if (!getGOTPLT()->hasFragments() && config().isCodeDynamic()) {
```
quic-seaswara: What about if you have a IFUNC ?

quic-seaswara (changes requested):

quic-areg:
A few existing `IFUNC` tests should fail (segfault) with this patch like Shankar mentioned. See `AArch64/standalone/IFunc/IFunc.test ` for example.

I feel that the test is too narrow. It could also cover static PIE executables and include a a positive check that dynamic/shared/pie still create `.got.plt` for example.

I also suspect that the test as written might not fail even if we did still emit an unnecessary `.got.plt` because of the way FileCheck works: 
```
#CHECK: .got
#CHECK-NOT: .got.plt
```
A `.got.plt` section would satisfy the first check, and the CHECK-NOT would start searching from `.plt` making the test pass unexpectedly. I think a `--implicit-check-not=.got.plt` should fix this, if that's what is happening.

For ARM and Hexagon, `_GLOBAL_OFFSET_TABLE_` is defined by the first `.got.plt` fragment, and this patch might leave that symbol undefined in the static-exec case.

We duplicate code across backends with this change. Is there a way to share it?

lib/Target/AArch64/AArch64LDBackend.cpp
```diff
@@ -694,8 +694,8 @@ AArch64GOT *AArch64LDBackend::createGOT(GOT::GOTType T,
...
     // TODO: This should be GOT0, not GOTPLT0.
```
quic-areg, line 728: The true fix, which might be outside the scope of this PR, would be to address this `TODO`:  `createGOT()` creates the GOT's reserved slot as `GOTPLT0` in `.got.plt` instead of as a `GOT0` in `.got`. It looks like RISCV has already partially done the fix.

deepakshirkem: Hi @quic-areg , Thank you for the insight! I am currently investigating  the IFUNC case to make sure the fix handles it correctly.

Thank You ::))

deepakshirkem:
Hi @quic-areg @quic-seaswara,  I have updated the fix to correctly handle the IFUNC case. The condition now creates GOTPLT0 when:
1. Dynamic linking is involved (isCodeDynamic())
2. OR when explicitly requested for GOTPLT0   (T == GOT::GOTPLT0), which handles the IFUNC case  in static executables.

I have tested the following cases locally:
  - Static executable → no .got.plt 
  - IFUNC in static executable → .got.plt created   
  - Dynamic linking → .got.plt created 

Could you please run the pipeline when you get a chance? Some test cases are unsupported in my local environment 
so I am unable to verify all targets locally.

Thank you ::))

deepakshirkem:
Hi @parth-07 / @quic-seaswara / @quic-areg , I just wanted to follow up on this PR when you get a chance. Just wanted to make sure this didn't get lost!

Thank You ::

deepakshirkem:
Hi [Parth (@parth-07)](https://github.com/parth-07) / [Shankar Easwaran (@quic-seaswara)](https://github.com/quic-seaswara) / [quic-areg](https://github.com/quic-areg) , I just wanted to follow up on this PR when you get a chance. Just wanted to make sure this didn't get lost!

deepakshirkem:
> A few existing `IFUNC` tests should fail (segfault) with this patch like Shankar mentioned. See `AArch64/standalone/IFunc/IFunc.test ` for example.
> 
> I feel that the test is too narrow. It could also cover static PIE executables and include a a positive check that dynamic/shared/pie still create `.got.plt` for example.
> 
> I also suspect that the test as written might not fail even if we did still emit an unnecessary `.got.plt` because of the way FileCheck works:
> 
> ```
> #CHECK: .got
> #CHECK-NOT: .got.plt
> ```
> 
> A `.got.plt` section would satisfy the first check, and the CHECK-NOT would start searching from `.plt` making the test pass unexpectedly. I think a `--implicit-check-not=.got.plt` should fix this, if that's what is happening.
> 
> For ARM and Hexagon, `_GLOBAL_OFFSET_TABLE_` is defined by the first `.got.plt` fragment, and this patch might leave that symbol undefined in the static-exec case.
> 
> We duplicate code across backends with this change. Is there a way to share it?

Closing this PR. I believe @quic-areg is right, and the fix belongs in a different place. I'll work on a follow-up PR with the proper fix.

Issue #747, opened 2026-01-27
support reproduce with namespec


quic-seaswara:
Add support for reproduce with namespec support

```
cat > 1.c << \!
int foo() { return 0; }
!

cat > 2.c << \!
int bar() { return 0; }
!

rm -f lib2.* && clang -c -target x86_64 -c 2.c
ld.eld -shared 2.o -o lib2.so
clang -c -target x86_64 -c 1.c
ld.eld -Bdynamic 1.o -L. -l:lib2.so -o a --reproduce a.tar
```

**namespec notes**

If namespec is of  the form :filename, ld will search the library path for a file called filename, otherwise it will search the library path for a file called libnamespec.a

Running the reproduce produces this message :

```
 (rm -rf reproduce.a;tar xf a.tar) && cd reproduce.a && bash -x response.txt
+ ld.eld -Bdynamic Object/1.o.6201977920901693556 lib2.so -o a --mapping-file=mapping.ini
Fatal: cannot read file 'lib2.so'
Fatal: Linking had errors.
```

quic-areg:
fixed in 77aeff9614b595e8d3b4227cd128c52cd6c48f44

PR #1101, opened 2026-04-24
[reproduce] Fix response file for namespec-resolved libraries

### LINKED ISSUE #731: reproduce option tarball does not work as expected with shared objects
```
cat > 1.c << \!
int foo() { return 0; }
!

cat > 2.c << \!
int bar() { return 0; }
!

rm -f lib2.* && clang -c -target hexagon -c 2.c
ld.eld -shared 2.o -o lib2.so
clang -c 1.c
ld.eld -Bdynamic 1.o -L. -l2 -o a --reproduce a.tar
```

Extracting the tarball and running the response file, produces an error

```
ld.eld -Bdynamic Object/1.o.6060996770729379307 lib2.so -o a --mapping-file=mapping.ini
Fatal: cannot read file 'lib2.so'
Fatal: Linking had errors.
```

quic-areg:
Libraries passed as -l<name> or -l:<filename> were emitted as a bare basename rather than their tarball path, so relinking from the extracted tarball failed.

Fixes: #731, #747


lib/LinkerWrapper/GnuLdDriver.cpp
```diff
@@ -1653,7 +1653,13 @@ bool GnuLdDriver::processReproduceOption(
           auto ipt = action->getInput();
           if (!ipt)
             return false;
-          os << outputTar->rewritePath(ipt->getName()) << ' ';
+          // Use the mapped path as the rewrite key. For a namespec input
+          // e.g. "-l2", use "2" instead of "lib2.so". This will match the
```
quic-seaswara, line 1657: map file needs to be updated too. follow up is ok.

quic-seaswara (approved):

lib/LinkerWrapper/GnuLdDriver.cpp
quic-areg: Updated how?

quic-seaswara:
```
cat > 1.c << \!
int foo() { return 0; }
!

cat > 2.c << \!
int bar() { return 0; }
!

rm -f lib2.* && clang -c -target hexagon -c 2.c
ld.eld -shared 2.o -o lib2.so
clang -c 1.c
ld.eld -Bdynamic 1.o -L /tmp/ -l2 -o a --reproduce a.tar
```

This is what the map file shows with this change

Linker Script and memory map
LOAD 1.o(Object/1.o.12657934096141885456)[x86_64]
LOAD 2(2)[x86_64]

Recommend that 2(2) be expanded to 

-l2 (SharedLibrary/2.x)

PR #1102, opened 2026-04-25
[memory-region] Error when section VMA does not fit in region

### LINKED ISSUE #327: eld does not give error if the specified vma/lma does not fit into the specified memory regions
eld does not give error if the specified vma/lma does not fit into the specified memory regions.

Reproducer:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
int foo() { return 1; }
int bar() { return 3; }
EOF

cat >script.t <<\EOF
MEMORY {
  RAM : ORIGIN = 0x1000, LENGTH = 0x1000
  ROM : ORIGIN = 0x5000, LENGTH = 0x1000
}

SECTIONS {
  .foo (0x2500) : { *(.text.foo) } >RAM
  .bar : { *(.text.bar) } >RAM
}
EOF

LDs=("ld.eld -m aarch64elf")
SFs=(eld)

clang -o 1.o -target aarch64-unknown-elf 1.c -c -ffunction-sections
for i in "${!SFs[@]}"; do
  ${LDs[$i]} -o 1.${SFs[$i]}.o 1.o -T script.t -Map 1.${SFs[$i]}.map.txt
done
```

deepakshirkem:
### Problem
Fixes #327

ELD was silently linking when a section's VMA was placedoutside its assigned memory region bounds:
```
MEMORY {
RAM : ORIGIN = 0x1000, LENGTH = 0x1000
}
SECTIONS {
.foo (0x2500) : { *(.text.foo) } >RAM  ← outside RAM!
}
```
### Fix
Added check in `postLayout()` to verify each section's VMA fits within its assigned memory region bounds, matching LLD
behavior exactly:

`Error: section '.foo' will not fit in region 'RAM':overflowed by 1288 bytes`

### Testing
Added test `MemoryRegionVMACheck` that verifies ELD emits a clear error when a section VMA does not fit in its
assigned memory region.

cc @parth-07  @quic-seaswara

### DIFF
--- a/lib/Object/ScriptMemoryRegion.cpp
+++ b/lib/Object/ScriptMemoryRegion.cpp
@@ -42,8 +42,16 @@ void ScriptMemoryRegion::addOutputSection(const OutputSectionEntry *O) {
     FirstOutputSectionExceededLimit = O;
   }
 }
+void ScriptMemoryRegion::verifyMemoryUsage(LinkerConfig &Config,
+                                           const OutputSectionEntry &out) {
+  if (!out.prolog().hasVMA())
+    return;
+  uint64_t vma = out.prolog().vma().result();
+  if (!containsVMA(vma))
+    Config.raise(Diag::error_section_not_in_region)
+        << out.name() << getName() << utility::toHex(vma)
+        << out.prolog().vma().getContext();
 
-void ScriptMemoryRegion::verifyMemoryUsage(LinkerConfig &Config) {
   auto ExpLen = getLength();
   if (ExpLen && !ExpLen.value() && Config.showLinkerScriptWarnings()) {
     MemorySpec *Spec = MMemoryDesc->getMemorySpec();
--- a/lib/Target/GNULDBackend.cpp
+++ b/lib/Target/GNULDBackend.cpp
@@ -5119,8 +5119,11 @@ void GNULDBackend::clearMemoryRegions() {
 }
 
 void GNULDBackend::verifyMemoryRegions() {
-  for (auto &M : m_Module.getLinkerScript().getMemoryRegions())
-    M->verifyMemoryUsage(config());
+  for (auto *out : m_Module.getLinkerScript().sectionMap()) {
+    if (!out->epilog().hasRegion())
+      continue;
+    out->epilog().region().verifyMemoryUsage(config(), *out);
+  }
 }
 
 bool GNULDBackend::assignMemoryRegions() {
--- a/test/Common/standalone/linkerscript/MemoryRegionVMACheck/MemoryRegionVMACheck.test
+++ b/test/Common/standalone/linkerscript/MemoryRegionVMACheck/MemoryRegionVMACheck.test
@@ -1,0 +1,10 @@
+#---MemoryRegionVMACheck.test----------------------- Executable -----------------#
+#BEGIN_COMMENT
+# Test that ELD errors when a section VMA does not fit in its assigned
+# memory region, matching BFD and LLD behavior. 
+#END_COMMENT
+#START_TEST
+RUN: %clang %clangg0opts -o %t1.1.o %p/Inputs/1.c -c -ffunction-sections
+RUN: %not %link %linkopts -o %t1.1.out %t1.1.o -T %p/Inputs/script.t 2>&1 | %filecheck %s
+#END_TEST
+CHECK: address {{0x[0-9a-f]+}} of section '.foo' is not within region 'RAM'
--- a/include/eld/Diagnostics/DiagLDScript.inc
+++ b/include/eld/Diagnostics/DiagLDScript.inc
@@ -76,6 +76,8 @@ DIAG(error_memory_region_exceeded_limit, DiagnosticEngine::Error,
      "0x%2 bytes")
 DIAG(error_memory_region_empty, DiagnosticEngine::Error,
      "Missing name for memory region, name cannot be empty")
+DIAG(error_section_not_in_region, DiagnosticEngine::Error,
+     "address 0x%2 of section '%0' is not within region '%1' (%3)")
 DIAG(error_duplicate_memory_region, DiagnosticEngine::Error,
      "Duplicate memory region %0")
 DIAG(verbose_verified_add_memory_region, DiagnosticEngine::Verbose,
--- a/include/eld/Object/ScriptMemoryRegion.h
+++ b/include/eld/Object/ScriptMemoryRegion.h
@@ -33,7 +33,7 @@ public:
 
   void addOutputSectionLMA(const OutputSectionEntry *O);
 
-  void verifyMemoryUsage(LinkerConfig &Config);
+  void verifyMemoryUsage(LinkerConfig &Config, const OutputSectionEntry &out);
 
   eld::Expected<uint64_t> getOrigin() const;
 
--- a/test/Common/standalone/linkerscript/MemoryRegionVMACheck/Inputs/1.c
+++ b/test/Common/standalone/linkerscript/MemoryRegionVMACheck/Inputs/1.c
@@ -1,0 +1,2 @@
+int foo() { return 1; }
+int bar() { return 3; }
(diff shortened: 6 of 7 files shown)


test/Common/standalone/linkerscript/MemoryRegionVMACheck/MemoryRegionVMACheck.test
```diff
@@ -0,0 +1,10 @@
+#---MemoryRegionVMACheck.test----------------------- Executable -----------------#
+#BEGIN_COMMENT
+# Test that ELD emits an error when a section VMA does not fit
+#END_COMMENT
+#START_TEST
+RUN: %clang %clangg0opts -o %t1.1.o %p/Inputs/1.c -c -ffunction-sections
+RUN: %not %link %linkopts -o %t1.1.out %t1.1.o -T %p/Inputs/script.t 2>&1 | %filecheck %s
+#CHECK: section '.foo' will not fit in region 'RAM'
```
quic-seaswara: There is no error with bfd linker.

lib/Target/GNULDBackend.cpp
```diff
@@ -3570,6 +3570,22 @@ bool GNULDBackend::postLayout() {
...
+        auto &region = out->epilog().region();
+        uint64_t secEnd = sec->addr() + sec->size();
+        uint64_t regionEnd =
+            region.getOrigin().value() + region.getLength().value();
+        if (secEnd > regionEnd)
+          config().raise(Diag::error_section_not_in_region)
+              << sec->name() << region.getName() << (secEnd - regionEnd);
+      }
+    }
+  }
```
quic-seaswara: With bfd, any explicit address specified, is not assigned to the memory region. 

Probably the user needs to be notified, that this output section is not part of the memory region as a note, and also a linker script warning using -Wlinker-script that this section is outside the memory region bounds.

In any case these might be needed to be moved to ScriptMemoryRegion to flag.

quic-seaswara (changes requested):

lib/Target/GNULDBackend.cpp
deepakshirkem: Hi @quic-seaswara,  My initial thought was also to add this check in ScriptMemoryRegion or verifyMemoryRegions(), but  during debugging I found that final section addresses are not yet assigned at that point (they are still 0x0).That's why I moved the check to postLayout() where addresses are fully assigned. That's why I moved the check to postLayout() where addresses are fully assigned.

I will address suggested comments those are making more sense.

Thank You ::((

deepakshirkem:
Hi @quic-seaswara, 
Updated the fix as per your feedback.Moved check to ScriptMemoryRegion::verifyMemoryUsage() as you suggested Changed from error to warning under -Wlinker-script matching BFD behavior Warning runs only once after layout is complete. Please review when you get a chance.

Thank you ::((

deepakshirkem:
Hi @quic-seaswara / @quic-areg,  Please take another look at these changes when you get a chance?

lib/Object/ScriptMemoryRegion.cpp
```diff
@@ -43,7 +43,23 @@ void ScriptMemoryRegion::addOutputSection(const OutputSectionEntry *O) {
...
+      if (!out->epilog().hasRegion())
+        continue;
+      if (&out->epilog().region() != this)
+        continue;
+      if (!out->prolog().hasVMA())
+        continue;
+      uint64_t vma = out->prolog().vma().result();
+      if (!containsVMA(vma))
+        Config.raise(Diag::warn_section_not_in_region)
+            << out->name() << getName() << utility::toHex(vma);
```
quic-seaswara: Need the linker script file also in the warning message.

lib/Object/ScriptMemoryRegion.cpp
```diff
@@ -43,7 +43,23 @@ void ScriptMemoryRegion::addOutputSection(const OutputSectionEntry *O) {
...
+                                           const SectionMap &SectMap) {
+
+  if (Config.showLinkerScriptWarnings()) {
+    for (auto *out : SectMap) {
+      if (!out->epilog().hasRegion())
+        continue;
+      if (&out->epilog().region() != this)
+        continue;
+      if (!out->prolog().hasVMA())
+        continue;
```
quic-seaswara: You might want to move this condition to line 51 ?

lib/Object/ScriptMemoryRegion.cpp
```diff
@@ -43,7 +43,23 @@ void ScriptMemoryRegion::addOutputSection(const OutputSectionEntry *O) {
...
 
-void ScriptMemoryRegion::verifyMemoryUsage(LinkerConfig &Config) {
+void ScriptMemoryRegion::verifyMemoryUsage(LinkerConfig &Config,
+                                           const SectionMap &SectMap) {
+
+  if (Config.showLinkerScriptWarnings()) {
+    for (auto *out : SectMap) {
+      if (!out->epilog().hasRegion())
+        continue;
+      if (&out->epilog().region() != this)
```
quic-seaswara: Instead of passing the sectionMap can verifyMemoryRegions run on all output sections that have memory region and call verifyMemoryUsage ?  The code can be greatly simplified I think. 

Note : the original loop will not change,

test/Common/standalone/linkerscript/MemoryRegionVMACheck/MemoryRegionVMACheck.test
```diff
@@ -0,0 +1,13 @@
+#---MemoryRegionVMACheck.test----------------------- Executable -----------------#
+#BEGIN_COMMENT
+# Test that ELD emits a warning when a section VMA does not fit
+# in its assigned memory region, matching BFD behavior.
+# Warning is shown under -Wlinker-script flag. (issue #327)
+#END_COMMENT
+#START_TEST
+RUN: %clang %clangg0opts -o %t1.1.o %p/Inputs/1.c -c -ffunction-sections
+RUN: %link %linkopts -o %t1.1.out %t1.1.o -T %p/Inputs/script.t -Wlinker-script 2>&1 | %filecheck %s
+RUN: %link %linkopts -o %t1.2.out %t1.1.o -T %p/Inputs/script.t 2>&1 | %filecheck %s --check-prefix=NOWARN --allow-empty
```
quic-seaswara: Why --allow-empty ?

test/Common/standalone/linkerscript/MemoryRegionVMACheck/MemoryRegionVMACheck.test
```diff
@@ -0,0 +1,13 @@
+#---MemoryRegionVMACheck.test----------------------- Executable -----------------#
+#BEGIN_COMMENT
+# Test that ELD emits a warning when a section VMA does not fit
+# in its assigned memory region, matching BFD behavior.
```
quic-seaswara: Does bfd emit a warning by default ?

deepakshirkem:
Hi @quic-seaswara, I apologize for the misunderstanding. I have not addressed these comments carefully enough. 
> There is no error with bfd linker.

> With bfd, any explicit address specified, is not assigned to the memory region.
> 
> Probably the user needs to be notified, that this output section is not part of the memory region as a note, and also a linker script warning using -Wlinker-script that this section is outside the memory region bounds.
> 
> In any case these might be needed to be moved to ScriptMemoryRegion to flag.



I have re-verified the behavior both BFD and LLD error out for this test case. If you give me the go-ahead, I will implement this as an error to match BFD and LLD behavior.

deepakshirkem:
Hi @quic-seaswara, please review the latest changes. I have matched the BFD behavior:

BFD & LLD:
Screenshot from 2026-05-09 01-39-37

ELD:

Screenshot from 2026-05-09 02-56-47



Thank You ::))

deepakshirkem:
Hi @quic-seaswara, Fixed the build error. Please re-run the pipeline one more time?

deepakshirkem:
@quic-seaswara Ping :(

PR #1108, opened 2026-04-28
[fragment] Fix incorrect address when dot assignment > UINT32_MAX

### LINKED ISSUE #386: Incorrect address when the padding due to dot assignment is greater than UINT_MAX
The linker assign incorrect addresses if the padding due to dot assignment is greater than UINT_MAX. 

Reproducer:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
int u = 1;
int v = 2;

int foo() { return 1; }
EOF

cat >script.t <<\EOF
SECTIONS {
  .data 0x1000 : {
    . = . + 0x400000000;
    *(.data*)
  }
}
EOF

LDs=(ld.eld)
SFs=(eld)

clang -o 1.o -target riscv64 1.c -c -fdata-sections -ffunction-sections
for i in "${!SFs[@]}"; do
done
llvm-readelf -s 1.eld.out | grep 'OBJECT.*u'
```

**Output**
12: 0000000000001000     4 OBJECT  GLOBAL DEFAULT     2 u

**Expected output**
12: 0000000400001000     4 OBJECT  GLOBAL DEFAULT     2 u

deepakshirkem:
### Problem
Fixes #386

ELD assigns incorrect addresses when the padding due to dot assignment is greater than `UINT32_MAX`.
When a linker script uses:  `. = . + 0x400000000;`

The fragment offset was truncated to `0x0` because `Fragment::UnalignedOffset` was stored as `uint32_t` which cannot hold values greater than `0xFFFFFFFF`.  

### Fix
Changed `Fragment::UnalignedOffset` and related offset fields from `uint32_t` to `uint64_t` to correctly handle 64-bit offsets on 64-bit targets.

### Testing
Added test `DotAssignmentOverflow` that verifies ELD correctly assigns addresses when dot assignment value
exceeds `UINT32_MAX`.

cc @quic-seaswara @parth-07 @quic-areg

### DIFF
--- a/docs/DeveloperDocs/AArch64/IFunc.md
+++ b/docs/DeveloperDocs/AArch64/IFunc.md
@@ -1,0 +1,214 @@
+# GNU IFunc support with AArch64
+
+GNU IFunc functionality enables a developer to provide multiple implementations
+of a function and a resolver function that selects which implementation to use at runtime.
+It is typically used for selecting the most optimized implementation for a given CPU / runtime.
+The resolver function is called once at the start of the application runtime and then the
+selected implementation is fixed.
+
+Example:
+
+```c
+__attribute__((ifunc("foo_resolver")))
+int foo(int);
+
+int foo_impl(int u) {
+  return u;
+}
+
+int (*foo_resolver())(int) {
+  return foo_impl;
+}
+
+int bar(int u) {
+  int v = foo(u); // Calls the function that is returned by foo_resolver
+}
+```
+
+## Static linking
+
+eld generates a PLT slot for each ifunc symbol. Each PLT slot has a corresponding
+GOTPLT slot. This model is very similar to how preemptible dynamic symbols are handled.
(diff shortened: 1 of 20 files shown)


test/Common/standalone/linkerscript/DotAssignmentOverflow/Inputs/script.t
```diff
@@ -0,0 +1,6 @@
+SECTIONS {
+  .data 0x1000 : {
+    . = . + 0x400000000;
```
quic-seaswara: I dont think anybody would even do this on 32bit systems or 64bit.

We should try to error out.

quic-seaswara (changes requested):

test/Common/standalone/linkerscript/DotAssignmentOverflow/Inputs/script.t
deepakshirkem: Yes, Noted.

PR #1113, opened 2026-04-29
Improve checking of invalid symbol versions

### LINKED ISSUE #1094: .symver to an unknown version with no version script is silently accepted
```sh
cat > 1.c << \!
int real_foo(void) { return 1; }
__asm__(".symver real_foo, foo@DOES_NOT_EXIST");
!

clang -c -fPIC -target x86_64-linux-gnu 1.c

ld.eld -shared -o eld.so  1.o
ld.lld -shared -o lld.so  1.o
ld.bfd -shared -o bfd.so  1.o
```
```diff
- ld.eld -shared -o eld.so 1.o
ld.lld -shared -o lld.so 1.o
+ ld.lld: error: 1.o: symbol foo@DOES_NOT_EXIST has undefined version DOES_NOT_EXIST
ld.bfd -shared -o gld.so 1.o
+ld.bfd: gld.so: version node not found for symbol foo@DOES_NOT_EXIST
+ld.bfd: failed to set dynamic section sizes: bad value
```

eld seems to only check this when a version script is present. Without it, the symbol is silently exported unversioned.

parth-07:
This commit improves checking of invalid symbol versions. In particular, we now report errors when:
- A defined symbol contains an unknown version.
- An undefined symbol contains a version. Please note that there is no valid case where an undefined symbol can have a symbol version.

Resolves #1094

### DIFF
--- a/include/eld/Diagnostics/DiagSymbolVersioning.inc
+++ b/include/eld/Diagnostics/DiagSymbolVersioning.inc
@@ -20,6 +20,8 @@ DIAG(trace_clear_export_due_to_local_scope, DiagnosticEngine::Trace,
      "Clearing export for symbol '%0' due to local scope")
 DIAG(error_missing_version_node, DiagnosticEngine::Error,
      "%0: symbol %1 has undefined version %2")
+DIAG(error_versioned_undef_symbol, DiagnosticEngine::Error,
+     "%0: undefined symbol '%1' cannot have version")
 DIAG(trace_assign_output_version_ids, DiagnosticEngine::Trace,
      "Assigning version IDs to output symbols")
 DIAG(trace_adding_verneed_entry, DiagnosticEngine::Trace,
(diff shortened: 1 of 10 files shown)


quic-areg:
The error message `undefined symbol 'foo@VER' cannot have version` repeats the version since the name already contains `@VER`. I think these error messages are better:

  lld: `undefined symbol: foo@VER`; gold: `symbol foo has undefined version VER`

parth-07:
> The error message `undefined symbol 'foo@VER' cannot have version` repeats the version since the name already contains `@VER`. I think these error messages are better:
> 
> lld: `undefined symbol: foo@VER`; gold: `symbol foo has undefined version VER`

Thank you for the suggestion. You mean the other diagnostic, `%0: symbol %1 has undefined version %2`, right?
The `undefined symbol 'foo@VER' cannot have version` was not repeating the version name. Though I was incorrectly passing the redundant version name with the `config.raise()` which was getting ignored.

I have modified the `%0: symbol foo@UNKNOWN_VERSION has undefined version UNKNOWN_VERSION` diagnostic to `%0: symbol foo has undefined version UNKNOWN_VERSION`.

Steven6798 (approved):

quic-seaswara (approved):

PR #1116, opened 2026-04-29
[linker-script] Error when dot assignment exceeds UINT32_MAX

### LINKED ISSUE #386: Incorrect address when the padding due to dot assignment is greater than UINT_MAX

deepakshirkem:
### Problem
Fixes #386

ELD assigns incorrect addresses when the padding due to dot assignment is greater than `UINT32_MAX`. When a linker script uses: `. = . + 0x400000000;` 

### Fix
Added validation in `Assignment::assign()` to error out when dot assignment creates padding that exceeds `UINT32_MAX`
as such large padding values are not meaningful.

### Testing
Added test `DotAssignmentOverflow` that verifies ELD emits a clear error when dot assignment value exceeds `UINT32_MAX`.

cc @quic-seaswara @parth-07


parth-07:
Hi @deepakshirkem,

Thank you for the pull-request. The original issue does not state that we should error-out if the padding due to dot assignment is greater than UINT32_MAX, instead it is about the symbol values being incorrect when there is a huge padding. Erroring out when there is a huge padding is a workaround and not a fix for this issue. The huge padding in the original issue can be replaced by any other huge-content that takes up space in the image layout and we will still see the same issue.

deepakshirkem:
Hi @parth-07, 
I have implemented the root cause fix in PR #1108 by changing uint32_t → uint64_t for Fragment::UnalignedOffset to correctly handle 64-bit offsets. Could you please confirm if this approach is correct? If so, I can implement that fix into this PR.

lib/Script/Assignment.cpp
```diff
@@ -213,6 +212,18 @@ bool Assignment::assign(Module &CurModule, const ELFSection *Section) {
   if (!Result)
     return false;
   ExpressionValue = *Result;
+  // Check if dot assignment creates padding > UINT32_MAX
+  if (isDot() && Section) {
+    LDSymbol *DotSym = CurModule.getNamePool().findSymbol(".");
+    if (DotSym && ExpressionValue > DotSym->value()) {
+      uint64_t Padding = ExpressionValue - DotSym->value();
```
parth-07: This will not report the error if ExpressionValue < DotSym->value() but the difference between the two is still greater than UINT32_MAX.

parth-07 (commented):
> I have implemented the root cause fix in PR https://github.com/qualcomm/eld/pull/1108 by changing uint32_t → uint64_t for Fragment::UnalignedOffset

PR #1108 has the right approach but we cannot implement that because of the memory overhead that it may cause. In a large build, there are a LOT of fragments. Increasing the offset size from 4 bytes to 8 bytes does not offer much benefits to justify the memory overhead. @quic-seaswara Do you think we should report error if the dot assignment causes padding greater than UINT32_MAX as this PR is doing?

lib/Script/Assignment.cpp
deepakshirkem: I will reproduce this issue and update the check.

deepakshirkem: Hi @parth-07, Your correct. I updated the patch and added both test cases backward and forward. Please review once you get a chance. Thank You ::))

quic-seaswara:
> Hi @deepakshirkem,
> 
> Thank you for the pull-request. The original issue does not state that we should error-out if the padding due to dot assignment is greater than UINT32_MAX, instead it is about the symbol values being incorrect when there is a huge padding. Erroring out when there is a huge padding is a workaround and not a fix for this issue. The huge padding in the original issue can be replaced by any other huge-content that takes up space in the image layout and we will still see the same issue.

Having a huge content is much more painful to create than just adding a padding expression.

lib/Script/Assignment.cpp
```diff
@@ -213,6 +212,20 @@ bool Assignment::assign(Module &CurModule, const ELFSection *Section) {
...
     return false;
   ExpressionValue = *Result;
+  // Check if dot assignment creates padding > UINT32_MAX
+  if (isDot() && Section) {
+    LDSymbol *DotSym = CurModule.getNamePool().findSymbol(".");
+    if (DotSym) {
+      uint64_t CurDot = DotSym->value();
+      uint64_t Diff = (ExpressionValue >= CurDot) ? ExpressionValue - CurDot
+                                                  : CurDot - ExpressionValue;
+      if (Diff > std::numeric_limits<uint32_t>::max()) {
```
quic-seaswara, line 222: Yes this is perfectly valid if you are on 64bit.

lib/Script/Assignment.cpp
```diff
@@ -213,6 +212,20 @@ bool Assignment::assign(Module &CurModule, const ELFSection *Section) {
...
+        CurModule.getConfig().raise(Diag::error_dot_assignment_overflow)
+            << getContext() << utility::toHex(Diff);
```
quic-seaswara: Do we want to warn instead of error for 64bit targets under -Wlinker-script category ? 

As a contrary to this fill, any expression that evaluates to a 64bit value on a 32bit system is error prone. The internal representation is 64bit but it is definitely a open area for bugs.

quic-seaswara (changes requested):

lib/Script/Assignment.cpp
deepakshirkem: Hi @quic-seaswara, I updated the patch take a look. 

Worried about this multiples warning does this works or we have to add any flag.
Screenshot from 2026-05-15 02-11-15

deepakshirkem: Hi @quic-seaswara, I think we can skip the warning for 64-bit targets entirely.

PR #1118, opened 2026-04-30
Add support for 'extern "C++"' in version scripts

### LINKED ISSUE #1117: `extern "C++"` does not work in version scripts
Version script correctly parses the `extern "C++"` directives but it does not affect any behavior.

parth-07:
This commit adds support for C++ linkage (`extern "C++"`) in the version scripts. Before this change, we were correctly parsing the `extern "C++"` directive but were not changing the logic of symbol matching with version patterns to use C++ linkage names.

Resolves #1117

### DIFF
--- a/lib/Script/ScriptSymbol.cpp
+++ b/lib/Script/ScriptSymbol.cpp
@@ -50,3 +50,14 @@ bool ScriptSymbol::matched(const ResolveInfo &Sym) const {
   }
   return false;
 }
+
+bool ScriptSymbol::matched(const ResolveInfo &RI,
+                           llvm::StringRef demangledName) const {
+  uint64_t Hash = llvm::hash_combine(demangledName);
+  if ((hasHash() && hashValue() == Hash) ||
+      WildcardPattern::matched(demangledName)) {
+    addResolveInfoToContainer(&RI);
+    return true;
+  }
+  return false;
+}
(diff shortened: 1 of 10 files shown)


lib/Script/ScriptSymbol.cpp
```diff
@@ -50,3 +50,14 @@ bool ScriptSymbol::matched(const ResolveInfo &Sym) const {
   }
   return false;
 }
+
+bool ScriptSymbol::matched(const ResolveInfo &RI,
+                           llvm::StringRef demangledName) const {
+  uint64_t Hash = llvm::hash_combine(demangledName);
+  if ((hasHash() && hashValue() == Hash) ||
+      WildcardPattern::matched(demangledName)) {
+    addResolveInfoToContainer(&RI);
```
quic-seaswara, line 59: Please update the map file, by using the SymbolContainer. 

Can be a follow up.

lib/Script/ScriptSymbol.cpp
```diff
@@ -50,3 +50,14 @@ bool ScriptSymbol::matched(const ResolveInfo &Sym) const {
   }
   return false;
 }
+
+bool ScriptSymbol::matched(const ResolveInfo &RI,
+                           llvm::StringRef demangledName) const {
```
quic-seaswara, line 55: Do we want to change deMangledName to const std::string & ?

quic-seaswara (approved):

PR #1123, opened 2026-05-02
[X86_64] Emit _GLOBAL_OFFSET_TABLE_ symbol for x86-64

### LINKED ISSUE #570: `_GLOBAL_OFFSET_TABLE_` symbol is not emitted for x86-64
We do not emit `_GLOBAL_OFFSET_TABLE_` symbol for x86-64 target. This symbol is required for enabling GOT-relative addressing that is used for large code model.

Reproducer:

```bash
#!/usr/bin/env bash 

cat >1.c <<\EOF
extern int _GLOBAL_OFFSET_TABLE_;

int *v = &_GLOBAL_OFFSET_TABLE_;

int a;

int foo() { return a; }
EOF


clang -o 1.o -target x86_64-pc-linux-gnu 1.c -c -ffunction-sections -fPIC 
LDs=(ld.eld)
SFs=(eld)

for i in "${!SFs[@]}"; do
  ${LDs[$i]} -o lib1.${SFs[$i]}.so 1.o -shared
  ${LDs[$i]} -o 1.${SFs[$i]}.out 1.o
done
```

eld errors out with the below error, instead of implicitly providing the `_GLOBAL_OFFSET_TABLE_` symbol.

```
Error: 1.o(.data+0x0): undefined reference to `_GLOBAL_OFFSET_TABLE_'
```

deepakshirkem:
## Problem

ELD does not emit the _GLOBAL_OFFSET_TABLE_ symbol for x86-64, causing an undefined reference error when object files reference this symbol. This affects code compiled with -fPIC or using the large code model.
```
Error: 1.o(.data+0x0): undefined reference to `_GLOBAL_OFFSET_TABLE_'
Fatal: Linking had errors.
```
## Fix

Add _GLOBAL_OFFSET_TABLE_ symbol support to the x86-64 backend following the same pattern used by other backends:

- `initTargetSymbols():` Register the symbol as AsReferred so it is only defined if actually referenced by input files.

- `defineGOTSymbol():` Bind the symbol to the first fragment of .got.plt so it points to the GOT base address.

- `finalizeScanRelocations(): `Call defineGOTSymbol() after all relocations have been scanned and .got.plt is populated.

## Testing
Added test/X86/standalone/GlobalOffsetTable/GlobalOffsetTable.test that verifies _GLOBAL_OFFSET_TABLE_ is emitted and points to .got.plt.

Fixes #570.

cc @quic-seaswara


lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -83,6 +83,17 @@ void x86_64LDBackend::initDynamicSections(ELFObjectFile &InputFile) {
 void x86_64LDBackend::initTargetSymbols() {
   if (config().codeGenType() == LinkerConfig::Object)
     return;
+  auto GOTSymName = "_GLOBAL_OFFSET_TABLE_";
```
quic-seaswara: Remove auto and qualify with appropriate type

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -83,6 +83,17 @@ void x86_64LDBackend::initDynamicSections(ELFObjectFile &InputFile) {
...
+  m_pGOTSymbol =
+      m_Module.getIRBuilder()
+          ->addSymbol<IRBuilder::AsReferred, IRBuilder::Resolve>(
+              m_Module.getInternalInput(Module::Script), GOTSymName,
+              ResolveInfo::Object, ResolveInfo::Define, ResolveInfo::Local,
+              0x0, // size
+              0x0, // value
+              FragmentRef::null(), ResolveInfo::Hidden);
+  if (m_pGOTSymbol)
+    m_pGOTSymbol->setShouldIgnore(false);
```
quic-seaswara: Can you check the behavior of bfd and lld that we are doing the same ?

test/Common/standalone/GlobalOffsetTable/GlobalOffsetTable_X86.test
```diff
@@ -0,0 +1,11 @@
+#----------GlobalOffsetTable.test--------------------- Executable ------#
+#BEGIN_COMMENT
```
quic-seaswara: Does this test make sense to be part of Common ? 

Can you add a test with linker script with GOT/GOT.PLT discarded and see the behavior ?

test/X86/standalone/GlobalOffsetTable/Inputs/1.c
```diff
@@ -0,0 +1,7 @@
+extern int _GLOBAL_OFFSET_TABLE_;
+
+int *v = &_GLOBAL_OFFSET_TABLE_;
+
+int a;
```
quic-seaswara: initialize variable

quic-seaswara (changes requested):

test/Common/standalone/GlobalOffsetTable/GlobalOffsetTable_X86.test
deepakshirkem: Moved to `Common/standalone/GlobalOffsetTable/`
Added `Common/standalone/GlobalOffsetTableDiscard/`

lib/Target/X86/x86_64LDBackend.cpp
deepakshirkem: Used `const std::string`  added comment.

deepakshirkem:
Hi @quic-seaswara, Please go one more time.

deepakshirkem:
Hi @parth-07, Please review these changes when you get a chance? Also, if possible, could you re-run the pipeline?

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -83,6 +83,18 @@ void x86_64LDBackend::initDynamicSections(ELFObjectFile &InputFile) {
 void x86_64LDBackend::initTargetSymbols() {
   if (config().codeGenType() == LinkerConfig::Object)
     return;
+  // addSymbol() takes std::string; use explicit type over auto.
```
parth-07: Can you please remove this comment. I don't think this comment is very useful.

parth-07 (commented):
What should be the value of `_GLOBAL_OFFSET_TABLE_` when there is no `.got.plt` section?

Example:
```
#!/usr/bin/env bash

cat >1.c <<\EOF
extern int _GLOBAL_OFFSET_TABLE_;
int foo() { return 1; }
int bar() { return 3; }
int *u = &_GLOBAL_OFFSET_TABLE_;
EOF

LDs=(ld.eld ld.lld ld.bfd)
SFs=(eld lld bfd)

for i in "${!SFs[@]}"; do
  ${LDs[$i]} -o lib1.${SFs[$i]}.so 1.o -shared
done
```

For this example, eld shared library has `_GLOBAL_OFFSET_TABLE_` defined to 0 whereas lld and bfd have valid `GLOBAL_OFFSET_TABLE_`. Can you please look into it? 

Additionally, we need to document the behavior for the properties(`local`, `default` vs `hidden` visibility, `NOTYPE` vs `OBJECT` ...) of the `_GLOBAL_OFFSET_TABLE_` symbol.

deepakshirkem:
Hi @parth-07, Please go one more time.

test/Common/standalone/GlobalOffsetTableShared/GlobalOffsetTableShared.test
```diff
@@ -0,0 +1,13 @@
...
+# when building a shared library with no PLT relocations.
+# ELD was incorrectly emitting value 0x0 in this case.
+#END_COMMENT
+#START_TEST
+REQUIRES: x86
+RUN: %clang %clangopts -c %p/Inputs/1.c -o %t1.o -fPIC -ffunction-sections
+RUN: %link %linkopts %t1.o -o %t1.so -shared 2>&1
+RUN: %readelf -s %t1.so | %filecheck %s
+#END_TEST
+CHECK: {{[1-9][0-9a-f]+}}{{.*}}OBJECT{{.*}}LOCAL{{.*}}HIDDEN{{.*}}_GLOBAL_OFFSET_TABLE_
```
quic-seaswara: it would be great if the test can be improved to point to the section as intended.

quic-seaswara (changes requested):

test/Common/standalone/GlobalOffsetTableShared/GlobalOffsetTableShared.test
deepakshirkem: Noted !

deepakshirkem:
Hi @quic-seaswara, test has been updated. Please take a look?

quic-seaswara (approved):

quic-seaswara:
> What should be the value of `_GLOBAL_OFFSET_TABLE_` when there is no `.got.plt` section?
> 
> Example:
> 
> ```
> #!/usr/bin/env bash
> 
> cat >1.c <<\EOF
> extern int _GLOBAL_OFFSET_TABLE_;
> int foo() { return 1; }
> int bar() { return 3; }
> int *u = &_GLOBAL_OFFSET_TABLE_;
> EOF
> 
> LDs=(ld.eld ld.lld ld.bfd)
> SFs=(eld lld bfd)
> 
> clang -o 1.o -target x86_64-linux-gnu 1.c -c -fPIC
> 
> for i in "${!SFs[@]}"; do
>   ${LDs[$i]} -o lib1.${SFs[$i]}.so 1.o -shared
> done
> ```
> 
> For this example, eld shared library has `_GLOBAL_OFFSET_TABLE_` defined to 0 whereas lld and bfd have valid `GLOBAL_OFFSET_TABLE_`. Can you please look into it?
> 
> Additionally, we need to document the behavior for the properties(`local`, `default` vs `hidden` visibility, `NOTYPE` vs `OBJECT` ...) of the `_GLOBAL_OFFSET_TABLE_` symbol.

@deepakshirkem can you add a test for this case ?

deepakshirkem:
Hi @quic-seaswara, Can you please re-run the pipeline. I forgot to add specific target on that test.

test/Common/standalone/GlobalOffsetTableDiscard/Inputs/1.c
deepakshirkem, line 1: > What should be the value of `_GLOBAL_OFFSET_TABLE_` when there is no `.got.plt` section?
> 
> Example:
> 
> ```
> #!/usr/bin/env bash
> 
> cat >1.c <<\EOF
> extern int _GLOBAL_OFFSET_TABLE_;
> int foo() { return 1; }
> int bar() { return 3; }
> int *u = &_GLOBAL_OFFSET_TABLE_;
> EOF
> 
> LDs=(ld.eld ld.lld ld.bfd)
> SFs=(eld lld bfd)
> 
> clang -o 1.o -target x86_64-linux-gnu 1.c -c -fPIC
> 
> for i in "${!SFs[@]}"; do
>   ${LDs[$i]} -o lib1.${SFs[$i]}.so 1.o -shared
> done
> ```
> 
> For this example, eld shared library has `_GLOBAL_OFFSET_TABLE_` defined to 0 whereas lld and bfd have valid `GLOBAL_OFFSET_TABLE_`. Can you please look into it?
> 
> Additionally, we need to document the behavior for the properties(`local`, `default` vs `hidden` visibility, `NOTYPE` vs `OBJECT` ...) of the `_GLOBAL_OFFSET_TABLE_` symbol.

@deepakshirkem can you add a test for this case ? 

Added here !

test/Common/standalone/GlobalOffsetTableShared/GlobalOffsetTableShared.test
```diff
@@ -0,0 +1,14 @@
+#---GlobalOffsetTableShared.test----------- Shared Library ---------------#
+#BEGIN_COMMENT
+# Verify that _GLOBAL_OFFSET_TABLE_ is emitted with a valid address
+# pointing to .got.plt when building a shared library with no PLT
+# relocations. ELD was incorrectly emitting value 0x0 in this case.
+#END_COMMENT
+#START_TEST
+REQUIRES: x86
```
quic-seaswara: move REQUIRES to the top.

test/Common/standalone/GlobalOffsetTableShared/GlobalOffsetTableShared_X86.test
```diff
@@ -0,0 +1,14 @@
+#---GlobalOffsetTableShared.test----------- Shared Library ---------------#
+#BEGIN_COMMENT
+# Verify that _GLOBAL_OFFSET_TABLE_ is emitted with a valid address
+# pointing to .got.plt when building a shared library with no PLT
+# relocations. ELD was incorrectly emitting value 0x0 in this case.
+#END_COMMENT
+#START_TEST
+REQUIRES: x86
+RUN: %clang %clangopts -c %p/Inputs/1.c -o %t1.o -fPIC -ffunction-sections
+RUN: %link %linkopts %t1.o -o %t1.so -shared 2>&1
```
quic-seaswara, line 10: What if it is a non shared library link, can we also test for -pie  ? 

There is no .got.plt so what does _GLOBAL_OFFSET_TABLE_ point to ?

quic-seaswara (changes requested):

quic-areg (changes requested):
The tests should live under `test/X86`, not `test/Common`

test/Common/standalone/GlobalOffsetTableShared/GlobalOffsetTableShared_X86.test
deepakshirkem: Hi @quic-seaswara,  Thank you. I tested both the -pie and shared both are correctly point .got.plt. Adding new test for -pie.

deepakshirkem:
Hi @quic-areg,  Moving all test in test/X86

deepakshirkem:
Hi @quic-seaswara / @quic-areg, Please take another look when you get a chance? Thank you.

test/X86/standalone/GlobalOffsetTableShared/GlobalOffsetTableShared.test
```diff
@@ -0,0 +1,14 @@
+REQUIRES: x86
```
quic-seaswara: Remove REQUIRES: x86 and move it to Common ?

deepakshirkem: Hi @quic-seaswara, should I move only GlobalOffsetTableShared to Common, or all the tests we created? Also, if I remove REQUIRES: x86, this test will fail in CI because the fix is currently only implemented in x86_64LDBackend.cpp.

deepakshirkem:
Hi @quic-areg / @quic-seaswara ,  Can you go one more time. Please re-run the pipeline whenever you get chance.

test/X86/standalone/GlobalOffsetTable/GlobalOffsetTable.test
```diff
@@ -0,0 +1,13 @@
+REQUIRES: x86
```
parth-07: We don't need this for tests within the x86 subdirectory.

test/Common/standalone/GlobalOffsetTableDiscard/GlobalOffsetTableDiscard_X86.test
```diff
@@ -0,0 +1,13 @@
...
+#BEGIN_COMMENT
+# Verify that ELD errors out when _GLOBAL_OFFSET_TABLE_ is referenced
+# but .got.plt section is discarded via linker script.
+# Consistent with GNU ld behavior.
+#END_COMMENT
+#START_TEST
+RUN: %clang %clangopts -c %p/Inputs/1.c -o %t1.o -fPIC -ffunction-sections
+RUN: %not %link %linkopts -T %p/Inputs/discard.t %t1.o -o %t1.out 2>&1 | %filecheck %s
+#END_TEST
+CHECK: symbol '_GLOBAL_OFFSET_TABLE_' referenced
```
parth-07, line 12: What is the error message emitted by eld in this case? Can you please add that as part of the test?

test/X86/standalone/GlobalOffsetTable/GlobalOffsetTable.test
```diff
@@ -0,0 +1,13 @@
...
+#BEGIN_COMMENT
+# Test that _GLOBAL_OFFSET_TABLE_ symbol is emitted for x86-64 and
+# points to .got.plt section.
+#END_COMMENT
+#START_TEST
+RUN: %clang %clangopts -c %p/Inputs/1.c -o %t1.o -fPIC -ffunction-sections
+RUN: %link %linkopts %t1.o -o %t1.out 2>&1
+RUN: %readelf -S -W -s %t1.out 2>&1 | %filecheck %s
+#END_TEST
+CHECK: .got.plt PROGBITS [[ADDR:[xa-f0-9]+]]
```
parth-07: Can you please add a check in the test that makes sure that the ADDR is a non-zero value?

parth-07 (changes requested):

test/X86/standalone/GlobalOffsetTable/GlobalOffsetTable.test
deepakshirkem: Thank You very much !

deepakshirkem:
Hi @parth-07, Please go one more time.

deepakshirkem:
Hi @parth-07 / @quic-areg / @quic-seaswara, if you get a chance today to review these changes, please take a look.

deepakshirkem:
Hi @quic-seaswara, Can you please run this pipeline one more time.

deepakshirkem:
Hi @quic-seaswara, The tests in `test/x86/standalone/` are being run by the ARM CI and failing since they are x86-specific. Please let me know the correct location for this test? That would be really helpful.

parth-07 (changes requested):
Why are we using a new directory `test/X86` instead of using the already existing directory `test/x86_64`?

deepakshirkem:
Hi @quic-seaswara / @parth-07 / @quic-areg, Please review one more time and re- run the pipeline.

deepakshirkem:
Hi @quic-seaswara, Check the updated commit? I have removed the duplicate code. Please run the pipeline? I want to confirm the test works correctly with Hexagon and AArch64.

quic-seaswara:
> Hi @quic-seaswara, Check the updated commit? I have removed the duplicate code. Please run the pipeline? I want to confirm the test works correctly with Hexagon and AArch64.

Before you submit the code, you can locally run ninja check-eld on your machine to verify than waiting for us to trigger the runs. Let me know if that would work for you.

deepakshirkem:
Yes, Will work. Thank You ::((

test/Common/standalone/GlobalOffsetTable/GlobalOffsetTable_Hexagon.test
```diff
@@ -0,0 +1,13 @@
...
+#BEGIN_COMMENT
+# Test that _GLOBAL_OFFSET_TABLE_ symbol is emitted for Hexagon and
+# points to .got.plt section with a non-zero address.
+#END_COMMENT
+#START_TEST
+RUN: %clang %clangopts -c %p/Inputs/1.c -o %t1.o -fPIC -ffunction-sections
+RUN: %link %linkopts %t1.o -o %t1.out 2>&1
+RUN: %readelf -S -W -s %t1.out 2>&1 | %filecheck %s
+#END_TEST
+CHECK: .got.plt PROGBITS [[ADDR:0*[1-9a-f][0-9a-f]*]]
```
parth-07: Can we have a common test for all architectures instead of duplicating test per architecture?

One way to do this is to define a lit substitution `GlobalOffsetTableSection`, and then pass that substitution to
FileCheck using `-DGlobalOffsetTableSection=%GlobalOffsetTableSection`, then this section can be referred in the `CHECK`s as `[[GlobalOffsetTableSection]]`

parth-07 (commented):
We should remove the older tests (such as test/RISCV/standalone/GlobalOffsetTable/GlobalOffsetTable.test) which are redundant/duplicate now.

deepakshirkem:
Hi @quic-areg / @quic-seaswara / @parth-07 , "Resolved merge conflict in lib/Target/RISCV/RISCVLDBackend.cpp — kept both our defineGlobalOffsetTableSymbol() call and the upstream __jvt_base$ addition. Please review

deepakshirkem:
@quic-seaswara This PR has grown to 37 files which makes it harder to review. Would you prefer I split this into a patch series  one patch per target  so each change can be reviewed and verified independently?

PR #1127, opened 2026-05-04
ARM: support R_ARM_THM_JUMP19 relocation in THM stubs

### LINKED ISSUE #1005: `R_ARM_THM_JUMP19` relocation fails for static targets
When trying to build a static lib for ARM targets involving thumb instructions, like Cortex-M0plus / Cortex-M33, ELD fails providing reason:

```
/work/obj/bin/ld.eld -m armelf -T link.ld --entry=main -static test.o main.o -o out_eld.elf
Error: applying relocation `51', conditional branch to PLT in THUMB-2 not supported yet
Error: Relocation error when applying relocation `R_ARM_THM_JUMP19' for symbol `plain_entry' referred from test.o[.text.caller] symbol defined in test.o[.text.target]
```

Reproductible example:

`test.S`

```asm
.arch armv8-m.main
.syntax unified
.thumb

@ section 1: caller
@ my_func is declared with .thumb_func, so its symbol value has LSB=1 (T=1).
@ It performs a conditional branch to plain_entry in a different section.

.section .text.caller, "ax", %progbits
.thumb_func
.global my_func
.type   my_func, %function
my_func:
    cmp  r0, #0
    beq  plain_entry        @ → R_ARM_THM_JUMP19, target T=0
    movs r0, #1
    bx   lr

@section 2: target (plain label)
@ different section, so relocation generated
.section .text.target, "ax", %progbits
.global plain_entry
plain_entry:
    movs r0, #0
    bx   lr
```

`main.c`

```C
extern int my_func(int x);

int main(void) {
    return my_func(1);
}
```

`link.ld`

```ld
MEMORY {
    FLASH (rx)  : ORIGIN = 0x10000000, LENGTH = 2M
    RAM   (rwx) : ORIGIN = 0x20000000, LENGTH = 256K
}

SECTIONS {
    .text 0x10000000 : {
        *(.text*)
    } > FLASH

    .data : { *(.data*) } > RAM AT > FLASH
    .bss  : { *(.bss*) *(COMMON) } > RAM
}
```

Commands:

```bash
clang --target=arm-none-eabi -mcpu=cortex-m0plus -mfloat-abi=soft -c test.S -o test.o
clang --target=arm-none-eabi -mcpu=cortex-m0plus -mfloat-abi=soft -c main.c -o main.o
ld.lld -m armelf -T link.ld --entry=main -static test.o main.o -o out_lld.elf # This does not fail
ld.eld -m armelf -T link.ld --entry=main -static test.o main.o -o out_eld.elf
```

sai18022001:
Fixes [#1005](https://github.com/qualcomm/eld/issues/1005)

**Problem**

`R_ARM_THM_JUMP19` (conditional branch) failed for static ARM targets like Cortex-M0plus / Cortex-M33 with:

```
Error: applying relocation `51', conditional branch to PLT in THUMB-2 not supported yet
```

This happened because the stub factory did not recognize `R_ARM_THM_JUMP19`, and the relocator unconditionally rejected T=0 targets even for non-PLT symbols (plain Thumb labels without`.thumb_func`).

**Changes**

- Added `R_ARM_THM_JUMP19` to `THMToTHMStub` and `THMToARMStub` using correct 21-bit range validation (`isInt<21>`)
- Implemented a PLT support check during the scanning phase (`scanGlobalReloc`)
- Updated `ARMRelocator.cpp` to allow `T=0` targets for `JUMP19` when the symbol is a non-PLT Thumb label

**Test**

Added tests in `test/ARM/standalone/THMJump19Veneer/` :
- `thm-jump19-basic.s`, `thm-jump19-veneer.s`, `thm-jump19-plt-diag.s`


lib/Target/ARM/THMToTHMStub.cpp
```diff
@@ -107,7 +108,8 @@ bool THMToTHMStub::isRelocInRange(const Relocation *pReloc,
   Offset = pTargetValue + addend - pReloc->place(Module);
   switch (pReloc->type()) {
   case llvm::ELF::R_ARM_THM_JUMP24:
-  case llvm::ELF::R_ARM_THM_CALL: {
+  case llvm::ELF::R_ARM_THM_CALL:
+  case llvm::ELF::R_ARM_THM_JUMP19: {
     if (m_Target->isJ1J2BranchEncoding())
       return llvm::isInt<THM2_MAX_BRANCH_BITS>(Offset);
```
quic-areg, line 114: `JUMP19` is 21 bits, not 23 or 25.

lib/Target/ARM/THMToARMStub.cpp
```diff
@@ -104,7 +105,8 @@ bool THMToARMStub::isRelocInRange(const class Relocation *pReloc,
   Offset = pTargetValue + addend - pReloc->place(Module);
   switch (pReloc->type()) {
   case llvm::ELF::R_ARM_THM_JUMP24:
-  case llvm::ELF::R_ARM_THM_CALL: {
+  case llvm::ELF::R_ARM_THM_CALL:
+  case llvm::ELF::R_ARM_THM_JUMP19: {
     if (m_Target->isJ1J2BranchEncoding())
       return llvm::isInt<THM2_MAX_BRANCH_BITS>(Offset);
```
quic-areg, line 110: same here

test/ARM/standalone/THMJump19Veneer/THMJump19Veneer.test
```diff
@@ -0,0 +1,13 @@
+#---THMJump19Veneer.test--------- Executable --------------------------#
```
quic-areg: We should test the veneer case (both thm-thm and thm-arm) too. I think this test does not produce veneers since everything appears to be in range.

test/ARM/standalone/THMJump19Veneer/Inputs/main.c
```diff
@@ -0,0 +1,5 @@
+extern int my_func(int x);
```
quic-areg: The bug will reproduce without this file. You can do `--entry=my_func`. I think link.ld can also be dropped since this particular issue should reproduce without a linker script, but you will probably need one when testing the stub cases. Then this test could be a single .s file (test.s) with no `Inputs` dir.

lib/Target/ARM/ARMRelocator.cpp
```diff
@@ -1101,9 +1101,13 @@ Relocator::Result thm_jump19(Relocation &pReloc, ARMRelocator &pParent) {
...
 
   if (0x0 == T) {
-    // FIXME: conditional branch to PLT in THUMB-2 not supported yet
-    DiagEngine->raise(Diag::unsupport_cond_branch_reloc) << (int)pReloc.type();
-    return Relocator::BadReloc;
+    // If the target is a PLT entry, conditional branch to PLT is not
+    // supported yet. For non-PLT targets (plain labels in Thumb code
+    // without .thumb_func), proceed with the relocation.
+    if (pReloc.symInfo()->reserved() & Relocator::ReservePLT) {
+      DiagEngine->raise(Diag::unsupport_cond_branch_reloc) << (int)pReloc.type();
```
quic-seaswara: Probably this needs to be tested at the time PLT is reserved ?

The diagnostics will need to always point to the input object file, section and symbol.

Please add a test for this.

lib/Target/ARM/THMToTHMStub.cpp
```diff
@@ -92,7 +92,8 @@ bool THMToTHMStub::isNeeded(const Relocation *pReloc, int64_t pTargetValue,
                             Module &Module) const {
   int64_t Offset = 0;
   // This stub cannot be used for ARM target.
-  if ((pTargetValue & 0x1) == 0)
+  if ((pTargetValue & 0x1) == 0 &&
+       pReloc->type() != llvm::ELF::R_ARM_THM_JUMP19)
```
quic-seaswara: What if R_ARM_THM_JUMP24 ?

test/ARM/standalone/THMJump19Veneer/THMJump19Veneer.test
```diff
@@ -0,0 +1,13 @@
...
+# This tests that R_ARM_THM_JUMP19 relocation (conditional branch beq)
+# is supported for static ARM targets (e.g. Cortex-M0plus / Cortex-M33).
+# Regression test for https://github.com/qualcomm/eld/issues/1005
+#END_COMMENT
+#START_TEST
+RUN: %clang %clangopts %p/Inputs/test.S -o %t.test.o -c --target=arm-none-eabi -mcpu=cortex-m0plus -mfloat-abi=soft
+RUN: %clang %clangopts %p/Inputs/main.c -o %t.main.o -c --target=arm-none-eabi -mcpu=cortex-m0plus -mfloat-abi=soft
+RUN: %link %linkopts %t.test.o %t.main.o -o %t.out -T %p/Inputs/link.ld -m armelf --entry=main -static 2>&1
+RUN: %readelf -s %t.out | %filecheck %s
+#CHECK: plain_entry
```
quic-seaswara: Need test for veneers to make sure proper veneers are generated

lib/Target/ARM/THMToTHMStub.cpp
sai18022001: Will fix `R_ARM_THM_JUMP19` to use `isInt<21>` instead of reusing `THM2_MAX_BRANCH_BITS`

lib/Target/ARM/THMToARMStub.cpp
sai18022001: Same fix for this too.

test/ARM/standalone/THMJump19Veneer/THMJump19Veneer.test
sai18022001: Understood, will add test cases that force veneer creation by placing caller and target out of range, covering both THMtoTHM and THMtoARM cases.

test/ARM/standalone/THMJump19Veneer/Inputs/main.c
sai18022001: Good point, will simplify to a single `.s file` with `--entry=my_func` and drop `main.c` and `link.ld` for the basic case.

sai18022001: After looking at existing tests in the repo, they all use an `Inputs/` directory for source files. I'll follow the same convention and keep the `Inputs/` dir, but will simplify to a single `.s file` with `--entry=my_func`, dropping `main.c`. The linker script will still be needed to force out-of-range for the veneer test cases.

quic-seaswara (changes requested):

test/ARM/standalone/THMJump19Veneer/Inputs/main.c
quic-areg: There are many examples of single-input assembly tests in eld, such as `ARM/standalone/Relocs/arm-adr-addend.s` for example.

sai18022001: Thanks for the example. I've adopted the single .s file pattern with `RUN:` and `CHECK:` as inline comments. However, I tried using `--section-start` to place sections far apart (to force out-of-range for veneer creation) but it doesn't seem to be honored by ELD without a linker script , both sections end up placed sequentially regardless. So I'll need an `Inputs/ dir` for the linker script for the veneer test cases. 

Do you have a preferred way to force out-of-range section placement without a linker script in ELD?

lib/Target/ARM/THMToTHMStub.cpp
sai18022001: I've updated `THMToTHMStub::isNeeded` to also exempt `R_ARM_THM_JUMP24` alongside `R_ARM_THM_JUMP19` when `T=0`:
```
if ((pTargetValue & 0x1) == 0 &&
     pReloc->type() != llvm::ELF::R_ARM_THM_JUMP19 &&
     pReloc->type() != llvm::ELF::R_ARM_THM_JUMP24)
  return false;
  ```

File: test/ARM/standalone/THMJump19Veneer/THMJump19Veneer.test (hunk 7 above)
Comment (sai18022001) [reply test/ARM/standalone/THMJump19Veneer/THMJump19Veneer.test]: Still working on this with @quic-areg. I'm trying to force out-of-range section placement to trigger veneer creation, but `--section-start` doesn't seem to be honored by ELD without a linker script. Will add the veneer test cases once I figure out the right approach for section placement.

Comment (quic-areg) [reply test/ARM/standalone/THMJump19Veneer/THMJump19Veneer.test]: What I had in mind is to have two tests: one testing the original case mentioned in the ticket (which you already have, and does not need a linker script) and another testing veneers, which will most likely need a linker script (I mentioned this in a separate comment). 

> ` --section-start` doesn't seem to be honored by ELD without a linker script

Do you have an example? It should work without one AFAIK.

File: test/ARM/standalone/THMJump19Veneer/Inputs/main.c (hunk 4 above)
Comment (quic-areg) [reply test/ARM/standalone/THMJump19Veneer/Inputs/main.c]: The veneer/stub case should probably be a separate test that will most likely need a linker script.

File: test/ARM/standalone/THMJump19Veneer/THMJump19Veneer.test (hunk 7 above)
Comment (sai18022001) [reply test/ARM/standalone/THMJump19Veneer/THMJump19Veneer.test]: @quic-seaswara @quic-areg 
Thanks for clarifying! I tested it but the map shows `.text.target` placed at `0xa` (sequentially after `.text.caller`) instead of `0x200000`:
```
.text.caller    0x0     0xa
.text.target    0xa     0x4
```
Command used:
```
ld.eld test.o --entry=my_func -static --section-start=.text.caller=0x0 --section-start=.text.target=0x200000 Map=test.map
```

Am I using it incorrectly, or is there something else I should try?

quic-seaswara: > @quic-seaswara @quic-areg Thanks for clarifying! I tested it but the map shows `.text.target` placed at `0xa` (sequentially after `.text.caller`) instead of `0x200000`:
> 
> ```
> .text.caller    0x0     0xa
> .text.target    0xa     0x4
> ```
> 
> Command used:
> 
> ```
> ld.eld test.o --entry=my_func -static --section-start=.text.caller=0x0 --section-start=.text.target=0x200000 Map=test.map
> ```
> 
> Am I using it incorrectly, or is there something else I should try?

Yes, you are using incorrectly, the --section-start linker command line option will work only on output sections, not input sections.

PR #1130, opened 2026-05-04
[RISCV] Add GNU IFunc functionality support in RISCV static links

### LINKED ISSUE #1137: IFunc functionality does not work for RISCV static links
IFunc functionality does not work for RISCV static links.

parth-07:
This commit adds GNU IFunc functionality support in RISCV static links.

GNU IFunc functionality allows a developer to provide multiple
implementations for a function and resolver fucntion that is called at
runtime to decide which implementation should be used.

GNU IFunc functionality is handled similarly to preemptible-functions in
dynamic linking. Each GNU IFunc function has a PLT slot and a GOTPLT
slot. A call/branch instruction to an IFunc function, jumps to the PLT
slot of the function. However, unlike general pre-emptible functions,
the runtime resolves ifunc functions eagerly (at startup) by iterating
over the R_RISCV_IRELATIVE relocations and calling the resolver function
for each relocation.

A GNU IFunc function additionally has a GOT slot as well if there is a
direct reference to the if IFunc function. It is required to ensure
pointer equality in cases when we have both GOT and direct
references to an ifunc symbol. For example:

Case 1) No direct reference
```
.L0
  auipc a0, %got_pcrel_hi(foo)
  ld a0, %pcrel_lo(.L0)(a0)
  jalr ra, x0
```
In this case, 'auicpc + ld' will directly resolve to GOTPLT[foo] to
avoid paying the PLT-indirection penalty.

Case 2) With direct reference
```
; A direct reference!
foo_gp:
  .quad foo

; A GOT-reference, same as before
.L0
  auipc a0, %got_pcrel_hi(foo)
  ld a0, %pcrel_lo(.L0)(a0)
  jalr ra, x0
```
In this case, if the GOT-reference resolves to GOTPLT[foo], then there
will be pointer inequality issue. Both foo_gp and a0 here stores the
address of foo, but they have different values. foo_gp has the address
of PLT[foo] and a0 has the address stored in GOTPLT[foo].

To resolve this, 'auipc + ld' is resolved to GOT[foo] and the GOT[foo]
stores the address of PLT[foo]. With this, both foo_gp and a0 have PLT[foo]
address now. Hence, the pointer inequality issue is resolved.

The developer documentation docs/DeveloperDocs/IFunc.md provides more
information about for this feature and implementation and details the
behavior of different relocations for this feature.

Additionally, this patch changes enables HelloWorldCXX test for riscv
and hence the test is no longer marked XFAIL for riscv.

Resolves #1137

### DIFF
--- a/docs/DeveloperDocs/IFunc.md
+++ b/docs/DeveloperDocs/IFunc.md
@@ -35,24 +35,24 @@ calls the corresponding resolver function to fill the GOTPLT slot.
 For the case of static executables, libc plays the role of runtime and takes on a task that is typically
 unusual for a libc -- resolve symbols and patch the binary at runtime with the resolution information.
 
-eld emits `R_AARCH64_IRELATIVE` relocations in `.rela.plt` section, and `__rela_iplt_start` and
+eld emits `R_<TARGET>_IRELATIVE` relocations in `.rela.plt` section, and `__rela_iplt_start` and
 `__rela_iplt_end` symbols that store the start and end addresses of `.rela.plt` section.
 
-The tiny-loader in the libc process the `R_AARCH64_IRELATIVE` relocations by
+The tiny-loader in the libc process the `R_<TARGET>_IRELATIVE` relocations by
 iterating over the [`__rela_iplt_start`, `__rela_iplt_end`) range.
 
 ```
 Relocation section '.rela.plt' at offset 0x190 contains 6 entries:
     Offset             Info             Type               Symbol's Value
-0000000000490000  0000000000000408 R_AARCH64_IRELATIVE               4358c0
-0000000000490008  0000000000000408 R_AARCH64_IRELATIVE               40db20
+0000000000490000  0000000000000408 R_<TARGET>_IRELATIVE               4358c0
+0000000000490008  0000000000000408 R_<TARGET>_IRELATIVE               40db20
 ```
 
 The symbol value of `IRELATIVE` relocations contains the IFunc resolver address, and
 the relocation offset points to the GOTPLT slot for the IFunc symbol. For each relocation,
 the tiny-loader in the libc calls the IFunc resolver and stores the result in the GOTPLT slot.
 
-Note that there is no lazy binding here.
+**Note that there is no lazy binding here.**
 
 ### Direct reference to an IFunc symbol
 
@@ -60,11 +60,11 @@ Direct references to an IFunc symbol are resolved to the PLT slot of the IFunc s
 
 ```
 // global variable!
-// R_AARCH64_ABS64
+// Absolute relocation, for example, `R_AARCH64_ABS64`, `R_RISCV64_64`.
 int (*foo_gp)(int) = foo;
 ```
 
-eld will resolve `R_AARCH64_ABS64` relocation to the address of PLT[foo].
+eld will resolve the absolute relocation to the address of PLT[foo].
 
 ### GOT references to an IFunc symbol
 
@@ -130,7 +130,8 @@ into the following categories:
 - Control flow relocations
 - GOT-related instruction relocations
 - Absolute / PC-relative address-forming relocations
-- Absolute / PC-relative load/store relocations.
+- Absolute / PC-relative load/store relocations
+- General address forming relocations
 - General computation relocation
 
 The relocations which are not supported by GNU for IFunc symbols are annotated with
@@ -144,27 +145,32 @@ Resolves to PLT[IFuncSymbol]. Sets HasDirectReference[IFuncSymbol] to true.
 
 - R_AARCH64_ABS{16, 32, 64}
 
-Not handled currently.
+- R_RISCV_{32, 64}
 
-[!IMPORTANT]
-They should be resolved to PLT[IFuncSymbol] as well!
+
+The below relocations are not handled currently by eld.
+They should be resolved to PLT[IFuncSymbol] as well.
 
 - R_AARCH64_PREL{16, 32, 64} (NotSupportedInGNULDForIFunc)
 - R_AARCH64_PLT32 (UNSUPPORTED)
 
 #### GOT-related data relocations
 
-- GOTREL{32, 64} (UNSUPPORTED)
-- GOTPCREL32 (UNSUPPORTED)
+- R_AARCH64_GOTREL{32, 64} (UNSUPPORTED)
+- R_AARCH64_GOTPCREL32 (UNSUPPORTED)
 
 #### Control flow relocations
 
 Resolves to PLT[IFuncSymbol].
 
-- TSTBR14
-- CONDBR19
-- JUMP26
-- CALL26
+- R_AARCH64_TSTBR14
+- R_AARCH64_CONDBR19
+- R_AARCH64_JUMP26
+- R_AARCH64_CALL26
+
+- R_RISCV_CALL
+- R_RISCV_CALL_PLT
+- R_RISCV_JAL
 
 #### GOT-related instruction relocations
 
@@ -178,22 +184,30 @@ IFuncSymbol; otherwise uses GOT[IFuncSymbol].
 - AUTH-ABI GOT relocations (UNSUPPORTED)
 - R_AARCH64_MOVW_GOTOFF_G{0,1}{_NC} (UNSUPPORTED)
 
+- R_RISCV_GOT_HI20
+
 #### Absolute / PC-relative address-forming relocations
+
+Resolves to PLT[IFuncSymbol]. Sets HasDirectReference[IFuncSymbol] to true.
 
 - R_AARCH64_ADR_PREL_LO21 (NotSupportedInGNULDForIFunc)
 - R_AARCH64_ADR_PREL_PG_HI21{_NC}
 
-Resolves to PLT[IFuncSymbol]. Sets HasDirectReference[IFuncSymbol] to true.
+- R_RISCV_HI20
+- R_RISCV_PCREL_HI20
+- R_RISCV_LO12_I
+- R_RISCV_LO12_S
+
+The below relocations are resolved to the IFunc symbol:
 
 - R_AARCH64_MOVW_UABS_G{0,1,2,3}{_NC} (NotSupportedInGNULDForIFunc)
 - R_AARCH64_SABS_G{0,1,2,3} (NotSupportedInGNULDForIFunc)
 - MOVW_PREL_G{0, 1, 2, 3}{_NC} (UNSUPPORTED)
 
-Resolves to IFuncSymbol.
-
 [!IMPORTANT]
 FIXME: We should perhaps resolve these relocations to the PLT[IFuncSymbol]
 instead of the IFuncSymbol for ensuring pointer equality.
+
 
 #### Absolute / PC-relative load/store relocations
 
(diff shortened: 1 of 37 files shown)


lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -1529,13 +1539,79 @@ void RISCVLDBackend::defineGOTSymbol(Fragment &pFrag) {
...
+    }
+    m_pIRelativeStart->setShouldIgnore(false);
+    SymbolName = "__rela_iplt_end";
+    m_pIRelativeEnd =
+        m_Module.getIRBuilder()
+            ->addSymbol<IRBuilder::Force, IRBuilder::Resolve>(
+                m_Module.getInternalInput(Module::Script), SymbolName,
+                ResolveInfo::Type::NoType, ResolveInfo::Define,
+                ResolveInfo::Binding::Local,
+                pSym.size(), // size
```
lenary: Claude is suggesting this size is wrong, and it should be 0, rather than the size of the symbol that triggered this code.

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -1529,13 +1539,79 @@ void RISCVLDBackend::defineGOTSymbol(Fragment &pFrag) {
...
+
+  bool is32Bits = config().targets().is32Bits();
+
+  for (auto &[symInfo, plt] : m_PLTMap) {
+    if (!symInfo->isIFunc() || !symInfo->hasIFuncDirectRef() ||
+        !symInfo->hasIFuncNeedsGOT())
+      continue;
+
+    // ELFObjectFile *PLTSlotObjFile =
+    //     llvm::cast<ELFObjectFile>(plt->getOwningSection()->getInputFile());
```
lenary: Can we just remove this?

docs/DeveloperDocs/IFunc.md
```diff
@@ -144,27 +150,40 @@ Resolves to PLT[IFuncSymbol]. Sets HasDirectReference[IFuncSymbol] to true.
...
-[!IMPORTANT]
-They should be resolved to PLT[IFuncSymbol] as well!
+
+The below relocations are not currently supported by eld.
+For IFunc symbols, they should be resolved to the PLT[IFuncSymbol]
+as well.
 
 - R_AARCH64_PREL{16, 32, 64} (NotSupportedInGNULDForIFunc)
-- R_AARCH64_PLT32 (UNSUPPORTED)
+- R_AARCH64_PLT32
```
lenary, line 160: the RISC-V version is called `R_RISCV_PLT32` - though I think it is unsupported by eld.

include/eld/Diagnostics/DiagDynamicLink.inc
```diff
@@ -11,8 +11,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-DIAG(create_got_entry, DiagnosticEngine::Note,
-     "Created GOT Entry for Symbol %0")
+DIAG(create_got_entry, DiagnosticEngine::Note, "Created %0 Entry for Symbol %1")
```
quic-seaswara, line 14: input file is missing

include/eld/Diagnostics/DiagDynamicLink.inc
```diff
@@ -11,8 +11,7 @@
...
 DIAG(create_plt_entry, DiagnosticEngine::Note,
      "Created PLT Entry for Symbol %0")
```
quic-seaswara, line 16: Do you need this diag ? can we use the create_got_entry diag since %0

lib/Fragment/GOT.cpp
```diff
@@ -34,3 +35,25 @@ const std::string GOT::name() const {
     return "";
   return (llvm::Twine("GOT entry for ") + ThisSymInfo->name()).str();
 }
+
+llvm::StringRef GOT::getGOTTypeAsStr(GOTType T) {
```
quic-seaswara, line 39: const function ?

quic-seaswara (changes requested):

docs/DeveloperDocs/IFunc.md
```diff
@@ -144,27 +150,40 @@ Resolves to PLT[IFuncSymbol]. Sets HasDirectReference[IFuncSymbol] to true.
...
+- R_AARCH64_CONDBR19
+- R_AARCH64_JUMP26
+- R_AARCH64_CALL26
+
+- R_RISCV_CALL
+- R_RISCV_CALL_PLT
+- R_RISCV_JAL
+- R_RISCV_BRANCH
+- R_RISCV_RVC_BRANCH
+- R_RISCV_RVC_JUMP
```
lenary, line 186: These four are a bit more nuanced.

Broadly, these four, when present in user code, can never point at a PLT, because these are used for in-function jumps, and assume that all registers are live over the jump. The PLT entry clobbers some quantity of temporary registers, so these cannot jump to them.

However, with linker relaxation, a call may be changed to a R_RISCV_JAL, which then would be allowed to resolve to a PLT entry, because the call assumes that arguments and temporary registers are not live when execution resumes at the next instruction.

I don't believe AArch64 has this issue, because the registers used by its PLT entry are always reserved, iirc.

include/eld/Diagnostics/DiagDynamicLink.inc
parth-07: We do not have input file in any of the create GOT/PLT entry diagnostics. It is also difficult to add the input files in these diagnostics because there may be many files that have undefined references to a global symbol for which we are creating the PLT/GOT entry. Additionally, this commit is only improving the diagnostic by adding the GOT type.

include/eld/Diagnostics/DiagDynamicLink.inc
parth-07: Yes, we need this diagnostic to differentiate whether a `.got` GOT slot is getting created or a `.got.plt` GOT slot.

docs/DeveloperDocs/IFunc.md
```diff
@@ -178,37 +197,64 @@ IFuncSymbol; otherwise uses GOT[IFuncSymbol].
...
-FIXME: We should perhaps resolve these relocations to the PLT[IFuncSymbol]
-instead of the IFuncSymbol for ensuring pointer equality.
+- MOVW_PREL_G{0, 1, 2, 3}{_NC}
 
 #### Absolute / PC-relative load/store relocations
 
 - LD_PREL_LO19 (NotSupportedInGNULDForIFunc)
 - LDST{8, 16, 32, 64, 128}_ABS_LO12_NC (NotSupportedInGNULDForIFunc)
 
+- RISCV has no relocations specific to load / store relocations.
```
lenary: This statement is a bit strange, and hard to interpret. Maybe just refer to the next section here.

lib/Fragment/GOT.cpp
parth-07: It is a static function. It cannot be const.

docs/DeveloperDocs/IFunc.md
```diff
@@ -178,37 +197,64 @@ IFuncSymbol; otherwise uses GOT[IFuncSymbol].
...
+- R_RISCV_PCREL_LO12_S
+
+- There is no relevant AArch64 relocation here.
+
+#### Arithmetic relocations used for forming address
 
 - R_AARCH64_ADD_ABS_LO12_NC
 
+- RISCV does not have arithmetic-specific relocations that is
+  typically used to form address.
```
lenary: It does, it's `R_RISCV_PCREL_LO12_I` (used by loads or `addi`), but you've already covered it above as it can be either GOT or non-GOT.

include/eld/Diagnostics/DiagDynamicLink.inc
quic-seaswara: Sure, we create a GOT and reuse it for the same symbol. May be a follow up. 

It would be useful to identify the relocation thats causing GOT to be created.

lenary (commented):
Overall looks fine to me. Some specific comments but nothing major.

include/eld/Diagnostics/DiagDynamicLink.inc
parth-07: > It would be useful to identify the relocation thats causing GOT to be created.

Agreed! But I believe this should be treated as a different task altogether.

lib/Target/RISCV/RISCVLDBackend.cpp
parth-07: Thank you for catching it. I have fixed the symbol size here and updated test to explicitly check for the `0` size for these symbols.

docs/DeveloperDocs/IFunc.md
parth-07: Thank you. I have updated the IFunc.md file with this relocation as well.

docs/DeveloperDocs/IFunc.md
parth-07: Thank you for mentioning it. I realized that the subcategory `Arithmetic relocations used for forming address` was not adding any value and can/should be merged with the other address-forming relocation category. I have removed this category altogether.

docs/DeveloperDocs/IFunc.md
parth-07: Agreed! I wanted to indicate that the RISCV does not have any relocations specific to load/store instructions, but it is not adding any value and seems a bit confusing. I have removed this line.

docs/DeveloperDocs/IFunc.md
parth-07: This is interesting. Thank you for explaining it. I think the issue that you have highlighted is broader than the IFunc symbol support. Currently, if an input source has both `jal foo` and `call foo`, then the linker should create a PLT entry for `foo` due to the `call` instruction, but `jal foo` should refer to `foo` directly (non-PLT foo). However, currently we unconditionally prefer PLT entry for a symbol if it is available for all the control-flow related relocations. I will explore this issue in detail as a separate follow-up task.

lib/Target/RISCV/RISCVLDBackend.cpp
parth-07: I have un-commented this code and properly used the `PLTSlotObjFile`. Here I should have used `PLTSlotObjFile` as the input file for the `GOT` slot to be consistent with eld conventions.

docs/DeveloperDocs/IFunc.md
lenary: > I think the issue that you have highlighted is broader than the IFunc symbol support.

Yeah, almost certainly.

I think the reality is that compilers won't leave a `jal` in an input file for a call, unless the user has written inline assembly. Given the `jal` comes from explicit user code, they should know what they're doing anyway :)

lenary (approved):
LGTM

docs/DeveloperDocs/IFunc.md
quic-seaswara: > This is interesting. Thank you for explaining it. I think the issue that you have highlighted is broader than the IFunc symbol support. Currently, if an input source has both `jal foo` and `call foo`, then the linker should create a PLT entry for `foo` due to the `call` instruction, but `jal foo` should refer to `foo` directly (non-PLT foo). However, currently we unconditionally prefer PLT entry for a symbol if it is available for all the control-flow related relocations. I will explore this issue in detail as a separate follow-up task.

Can you create a ticket on this ?

quic-seaswara (approved):

PR #1136, opened 2026-05-04
Added file compatibility check to ELFDynObjParser::parseFile function

### LINKED ISSUE #1097: linker crashes when a wrong arch shared library is passed in the link line
```
cat > 1.c << \!
int foo() { return 0; }
!

cat > 2.c << \!
int bar() { return 0; }
!

clang -c 2.c -fPIC -target x86_64
ld.eld -shared 2.o -o lib2.so
clang -c 1.c -fPIC -target hexagon
ld.eld -Bdynamic 1.o -L. -l2
```

Expected error 

`lib2.so is not compatible with 1.o`

Steven6798:
Detect incompatible input files in ELFDynObjParser

Add a compatibility check when parsing dynamic objects so that
incompatible files are rejected early with a clear error message
instead of causing a crash later in the linking process.

Update tests and documentation to cover dynamic linking and
incompatible object handling.

Fixes #1097


test/AArch64/standalone/WrongArchObject/WrongArchInput.test
```diff
@@ -3,6 +3,8 @@ RUN: %clang %clangopts -target aarch64 -c %p/Inputs/2.c -o %t1.o
 RUN: %not %link %linkopts -march aarch64 %t.1.o %t1.o -o %t2.out 2>&1 | %filecheck %s
 RUN: %ar cr %aropts %t1.a %t.1.o
 RUN: %not %link %linkopts -march aarch64 %t1.o --whole-archive %t1.a -o %t3.out 2>&1 | %filecheck %s -check-prefix=ARCHIVE
+RUN: %not %link   -Bdynamic %t1.o -L%p/Inputs %p/Inputs/lib2.so 2>&1 | %filecheck %s  -check-prefix=DYNAMIC
 
 CHECK: Fatal: cannot recognize the format of file {{[` -\(\)_A-Za-z0-9.\\/:]+}}1.o
 ARCHIVE: Fatal: cannot recognize the format of file {{[` -\(\)_A-Za-z0-9.\\/:]+}}.a
+DYNAMIC: Fatal: cannot recognize the format of file {{[` -\(\)_A-Za-z0-9.\\/:]+}}lib2.so
```
quic-seaswara: * a better message is to infer the architecture of the file and say it is not compatible with the current link. 

* Remove Fatal messages and make this an error. 

Fatal should be used only for use cases when the error is not recoverable to unwind the stack.

* Is there a way to reduce the binaries in this PR.

lib/Readers/ELFDynObjParser.cpp
```diff
@@ -32,6 +32,11 @@ eld::Expected<bool> ELFDynObjParser::parseFile(InputFile &inputFile) {
 
   std::unique_ptr<ELFReaderBase> ELFReader = std::move(expReader.value());
 
+  auto expCompatibility = ELFReader->isCompatible();
+  ELDEXP_RETURN_DIAGENTRY_IF_ERROR(expCompatibility);
+  if (!expCompatibility.value())
+    return false;
+
```
quic-seaswara, line 39: Can you add a test to do 

ld.eld -Bdynamic <sharedobject> ?

quic-seaswara (changes requested):

lib/Readers/ELFDynObjParser.cpp
Steven6798: The update i did for the tests was to add ld.eld -Bdynamic, so this is already done.

test/AArch64/standalone/WrongArchObject/WrongArchInput.test
Steven6798: Sounds good. Thanks for the feedback!

parth-07 (commented):
Do we need to check-in the shared libraries `lib2.so` in the codebase? Can we instead build the shared libraries as part of the test?

test/Common/standalone/WrongArchObject/Inputs/1.yaml
parth-07, line 1: Is there any particular reason for using yaml2obj for this test?

Steven6798: I noticed that RISCV and x86 did not have the existing WrongArchInput tests that other targets had, so I added them. That file was already included in the test; I just added the things required to test the dynamic parser. If this needs to be changed let me know.

Steven6798:
> Do we need to check-in the shared libraries `lib2.so` in the codebase? Can we instead build the shared libraries as part of the test?

The shared library would have to be compiled for another target. I don't think it is a good idea to build them since the test would need that additional target to run. I could look into how does the yaml code works to build the lib2.so object file without having to include the binary or require the additional target. Let me know what approach you think is best.

lib/Readers/ELFDynObjParser.cpp
```diff
@@ -32,6 +32,11 @@ eld::Expected<bool> ELFDynObjParser::parseFile(InputFile &inputFile) {
 
   std::unique_ptr<ELFReaderBase> ELFReader = std::move(expReader.value());
 
+  auto expCompatibility = ELFReader->isCompatible();
+  ELDEXP_RETURN_DIAGENTRY_IF_ERROR(expCompatibility);
+  if (!expCompatibility.value())
```
quic-areg, line 37: Is this check redundant with the `ELDEXP_RETURN_DIAGENTRY_IF_ERROR(expCompatibility);` above?

quic-areg (commented):
I don't think any `.so` files need to be checked in with this patch. You should be able to produce an arch mismatch by using YAML and `EM_NONE`: 

```sh
$ cat 1.yaml
--- !ELF
FileHeader:
  Class:           ELFCLASS32
  Data:            ELFDATA2LSB
  Type:            [[TYPE=ET_REL]]     
  Machine:         EM_NONE              
  ...

$ yaml2obj 1.yaml                  -o tmp.o   
$ yaml2obj 1.yaml -DTYPE=ET_DYN    -o tmp.so 
$ ld.eld -m elf_x86_64 -Bdynamic 1-x64.o ./tmp.so -o out
Fatal: cannot recognize the format of file `./tmp.so'. ...
```
Then, I think that this could be all one test under `Common` using `%emulation` and the target-specific ones (the new ones added in this PR and the existing ones under `AArch64/ARM/Hexagon/{linux,standalone}/WrongArchObject`) can be deleted.

Steven6798:
> I don't think any `.so` files need to be checked in with this patch. You should be able to produce an arch mismatch by using YAML and `EM_NONE`:
> 
> ```shell
> $ cat 1.yaml
> --- !ELF
> FileHeader:
>   Class:           ELFCLASS32
>   Data:            ELFDATA2LSB
>   Type:            [[TYPE=ET_REL]]     
>   Machine:         EM_NONE              
>   ...
> 
> $ yaml2obj 1.yaml                  -o tmp.o   
> $ yaml2obj 1.yaml -DTYPE=ET_DYN    -o tmp.so 
> $ ld.eld -m elf_x86_64 -Bdynamic 1-x64.o ./tmp.so -o out
> Fatal: cannot recognize the format of file `./tmp.so'. ...
> ```
> 
> Then, I think that this could be all one test under `Common` using `%emulation` and the target-specific ones (the new ones added in this PR and the existing ones under `AArch64/ARM/Hexagon/{linux,standalone}/WrongArchObject`) can be deleted.

That's a great idea! I'll add this change to the next commit I have planned for this PR.

lib/Readers/ELFDynObjParser.cpp
Steven6798: I checked with the debugger and yes, it looks like it is redundant. Thanks!

Steven6798: After looking into it more, the check is indeed needed. The isCompatible function can return false via the checkFlag function. I was able to test it.

parth-07:
@Steven6798 @quic-areg Would it be better to simply have a very simple file with contents `int foo() { return 1; }` and then use it to build shared library instead of using `yaml2obj`? Is there any benefit of using yaml2obj here? My concern is that yaml2obj unnecessarily requires much more code than a basic C file.

Steven6798:
> @Steven6798 @quic-areg Would it be better to simply have a very simple file with contents `int foo() { return 1; }` and then use it to build shared library instead of using `yaml2obj`? Is there any benefit of using yaml2obj here? My concern is that yaml2obj unnecessarily requires much more code than a basic C file.

Using the simple C file requires another target to be built to create the shared library. I'm not sure if there are other tests that work that way, but I don't think this is a good idea to require multiple targets to be built to run a test unless it is absolutely necessary.

parth-07:
> Using the simple C file requires another target to be built to create the shared library. I'm not sure if there are other tests that work that way, but I don't think this is a good idea to require multiple targets to be built to run a test unless it is absolutely necessary.

Okay, I see. You are right. Let's proceed with yaml2obj.

quic-areg (approved):

quic-shaksuma (approved):

quic-seaswara:
@Steven6798 can you please squash the commits, and add a good commit message with a description of the change.

Also update the userguide if there is a way to document this behavior.

parth-07 (commented):
Can you please update the commit title to be in imperative-style? (Useful reference: https://cbea.ms/git-commit)

quic-seaswara (approved):

PR #1139, opened 2026-05-06
[RISCV] Add support for R_RISCV_QC_ACCESS_16 and R_RISCV_QC_ACCESS_32

lenary:
These QUALCOMM vendor relocations mark 16-bit compressed and 32-bit load/store instructions as candidates for relaxation from a QC_E_LI + Load/Store sequence.

This Change:
- Extends the internal QUALCOMM relocation range to cover the new relocations
- registers them with applyNone (marker semantics, no instruction changes),
- and add tests covering loads, stores, and compressed instructions.

---

This PR goes with https://github.com/llvm/llvm-project/pull/188671 and https://github.com/quic/riscv-elf-psabi-quic-extensions/pull/63 - the latter is necessary to understand how these relocations are intended to work.

This was written with the assistance of AI.


parth-07 (approved):

quic-seaswara (approved):

lenary:
I have tried to unblock the LLVM changes - I've split them into assembly-only changes and the codegen changes will follow-up afterwards. Once they land this should be able to.

Issue #960, opened 2026-03-23
Remove non standard options and difficult to maintain in eld


quic-seaswara:
- --disable-bss-conversion

Workaround: Use separate segments for sections that have different progbits

GNU linker script does bss conversion as well; ELD is compatible with GNU linker

Notes: progbits – marks sections that occupy space in the file but not in memory

- --enable-bss-mixing

Allows mixing bss and progbits sections together in one segment

Related to bss conversion

GNU linker script does not support bss mixing; ELD is compatible with GNU linker

- --compact

Workaround: Merge segments into one output section

No other linker supports it; ELD is being made compatible with all other linkers

- __attribute__((section([section@address))]

No workaround

GNU linker script does not support setting VA;  

- RegionTable support (“.region_table” keyword in linker script)

No workaround

Contains a table of region addresses and sizes to zero initialize

GNU linker script does not support it

PR #1140, opened 2026-05-06
Remove AtTable support and AtAddress tests

quic-seaswara:
This is part of the patch series that removes difficult to maintain options in eld and increase compatibility with community linkers.

part of #960


quic-areg:
There's still dead code in `ObjectLinker` and `ObjectBuilder` regarding At: 
```cpp

      if ((R && R->isEntry()) ||
          ELFSect->name().find("@") != llvm::StringRef::npos) {
```
```cpp
  if (I->name().find("@") != llvm::StringRef::npos)
    return true;
```

quic-seaswara:
> There's still dead code in `ObjectLinker` and `ObjectBuilder` regarding At:
> 
> ```c++
>       if ((R && R->isEntry()) ||
>           ELFSect->name().find("@") != llvm::StringRef::npos) {
> ```
> 
> ```c++
>   if (I->name().find("@") != llvm::StringRef::npos)
>     return true;
> ```

Thanks for catching this!. Will push a PR.

PR #1142, opened 2026-05-07
[RISCV] Implement linker relaxation for R_RISCV_QC_ACCESS_16 and R_RISCV_QC_ACCESS_32

lenary:
Relaxes qc.e.li + load/store pairs annotated with R_RISCV_QC_ACCESS_32 (32-bit
instruction) or R_RISCV_QC_ACCESS_16 (16-bit compressed instruction). Four
relaxation variants are applied in priority order:

  1. GP-std:   qc.e.li + ls → GP-relative 32-bit load/store (saves 6/4 bytes)
  2. Abs-std:  qc.e.li + ls → x0-based 32-bit load/store (saves 6/4 bytes)
  3. GP-xqci:  qc.e.li + ls → 6-byte qc.e.l*/qc.e.s* (saves 4/2 bytes)
  4. Abs-xqci: qc.e.li + ls → 6-byte qc.e.l*/qc.e.s* (saves 4/2 bytes)

All variants require --relax-xqci. The ACCESS relocation must carry
R_RISCV_RELAX for pair relaxation to occur; without it the qc.e.li falls
back to standalone relaxation. The ACCESS relocation must be exactly 6 bytes
after the qc.e.li.

New internal relocation types R_RISCV_QC_{ABS,GPREL}26_{I,S} are added for
the xqcilo immediate field, along with encoding types EncTy_QC_EI/EncTy_QC_ES
and their encode/clear helpers.

ELFSection::hasFollowing(R, Type) is added to check whether a relocation is
immediately followed by one of a given type in the section's relocation list.

--

Stacked on #1139

Assisted by Claude.


lenary:
My TODOs:
- [x] Review tests more closely, to ensure they are maintainable.
- [x] Check why `R_RISCV_TPREL_S` didn't have an entry in `lib/Target/RISCV/RISCVRelocationInfo.cpp` before -- it did, Claude had duplicated the entry, now deleted.

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -645,6 +645,338 @@ bool RISCVLDBackend::doRelaxationQCELi(Relocation *reloc, Relocator::DWord G) {
...
+  RegionFragmentEx *region =
+      llvm::cast<RegionFragmentEx>(QCELiReloc->targetRef()->frag());
+  uint64_t qceli_offset = QCELiReloc->targetRef()->offset();
+
+  Relocator::Address S = getSymbolValuePLT(*QCELiReloc);
+  Relocator::DWord A = QCELiReloc->addend();
+  Relocator::Address Value = S + A;
+  // The effective address includes the offset from the access instruction;
+  // all range checks operate on this adjusted value.
+  Relocator::Address AdjValue = Value + (Relocator::DWord)access.offset;
```
parth-07, line 1049: What do you think about changing `AdjValue` type to `int64_t`? `access.offset` is a signed value, and with `AdjValue` as signed, we perhaps won't need `signedAddress(...)` calls.

lib/Target/RISCV/RISCVRelocationHelper.h
```diff
@@ -56,6 +61,35 @@ enum Reg {
   return op | (rd << 7) | (imm << 12);
 }
 
+// QC.EI type instruction
+[[maybe_unused]] uint64_t qceitype(uint32_t xf3, uint32_t xf2, uint32_t rd,
+                                   uint32_t rs1, int64_t imm) {
+  uint32_t uimm = (uint32_t)(imm & 0x3FFFFFFu);
```
parth-07: `uimm` is always casted to 64-bit unsigned before use. Is there any particular reason for not making its type as `uint64_t`? The same comment for `uimm` in the `qcetype` function.

lib/Target/RISCV/RISCVLDBackend.h
```diff
@@ -207,6 +207,15 @@ class RISCVLDBackend : public GNULDBackend {
   // Get the value of the symbol, using the PLT slot if one exists.
   Relocation::Address getSymbolValuePLT(const Relocation &R);
 
+  // Interpret a raw address as a signed value at the target's pointer width.
+  // On a 32-bit target, truncate to 32 bits (applying the same wraparound the
+  // processor would) and then sign-extend for range checks; on a 64-bit target
+  // the address is already the right width.
+  int64_t signedAddress(uint64_t addr) const {
```
quic-seaswara: getSignedAddress ?

lib/Target/RISCV/RISCVLDBackend.h
```diff
@@ -232,6 +241,29 @@ class RISCVLDBackend : public GNULDBackend {
...
+                              Relocation::DWord G);
+
+  // Decoded load/store instruction info for QC ACCESS relaxation.
+  struct QCAccess {
+    // Loads come first so isLoad() is a cheap comparison.
+    enum class Operation { Lb, Lbu, Lh, Lhu, Lw, Sb, Sh, Sw };
+    Operation op;
+    uint32_t size;  // access instruction size (in bytes)
+    unsigned reg;   // destination (load) or source data (store) register
+    int64_t offset; // immediate offset in the original access instruction
```
quic-seaswara: initialize all these values to be 0 or have an explicit constructor ?

lib/Target/RISCV/RISCVRelocationCompute.cpp
```diff
@@ -22,6 +22,9 @@ bool checkRange(uint64_t Value, bool IsSigned, EncodingType Type) {
   case EncTy_QC_EAI:
   case EncTy_QC_EJ:
     return llvm::isInt<32>(static_cast<int64_t>(Value));
+  case EncTy_QC_EI:
+  case EncTy_QC_ES:
+    return llvm::isInt<26>(static_cast<int64_t>(Value));
```
quic-seaswara, line 27: The mask has 22 bits and the comparison is against 26 bits ? Is this correct ? 

Do we have a test that checks for overflows ?

lib/Target/RISCV/RISCVRelocator.cpp
```diff
@@ -1097,6 +1105,40 @@ RISCVRelocator::Result applyGPRel(Relocation &pReloc, RISCVLDBackend &Backend,
...
+                                        RelocationDescription &pRelocDesc) {
+  if (RelocDescs.count(pReloc.type()) == 0)
+    return RISCVRelocator::Unsupport;
+  int64_t S = Backend.getSymbolValuePLT(pReloc);
+  int64_t A = pReloc.addend();
+  int64_t G = 0;
+  LDSymbol *gpSym =
+      Backend.getModule().getNamePool().findSymbol("__global_pointer$");
+  if (gpSym)
+    G = gpSym->value();
```
quic-seaswara, line 1148: What if there is no gpsym ?

lib/Target/RISCV/RISCVLDBackend.cpp
lenary: The intention of `signedAddress` is to get the right address wraparound on riscv32. I don't think making this a `int64_t` would remove or avoid the use of `signedAddress(...)`? Maybe I'm wrong.

lib/Target/RISCV/RISCVRelocator.cpp
lenary: Then this internal relocation will never be created during relaxation. This is the same logic as `applyGPRel`, but this version has a different range and no hi/lo complexity.

lib/Target/RISCV/RISCVRelocationCompute.cpp
lenary: Yes, the mask is for the bits not in the immediate, and this check is for the bits in the immediate. 22+26 = 48, and these are 48-bit instructions.

These are used only in internal relocations, which are the result of relaxation. If they would have overflowed, then the relaxation should not have happened.

We have checks that the relaxations are not done if they are out of range.

lib/Target/RISCV/RISCVLDBackend.h
lenary: I'm not sure how I feel about this.

On the one hand, C++ will default initialise all these fields to 0 already, without us having to be explicit.

On the other, not all of these fields have out-of-band values to represent "uninitialised" (e.g. `Operation`), and I think it's somewhat better to have it clear that we cannot create "invalid" values of this struct which need to be checked later - all the values need to have come out of looking at the instruction sequence, and if the sequence is wrong enough that we can't get that info, we shouldn't be calling the function that needs this info. We do, however, have partial initialisation of this in the code - we can initialise some fields before others.

Maybe this all doesn't matter. This is a struct so we can name these four related values being passed between private functions, so maybe it doesn't matter what we do.

lib/Target/RISCV/RISCVRelocationHelper.h
lenary: Sure, I can simplify all of this.

lib/Target/RISCV/RISCVLDBackend.h
lenary: Sure

lib/Target/RISCV/RISCVLDBackend.cpp
parth-07: I thought more about this. I believe that we should not use the `signedAddress` function here. It truncates the value early which can result in missed range overflow diagnostics. For example, if the `adjValue` is `0x1 0000 0000`, then `signedAddress(adjValue)` will be `0x0`, and the checks such as `llvm::isInt<26>getSignedAddress(AdjValue))` will incorrectly return true.

lib/Target/RISCV/RISCVLDBackend.h
parth-07: > On the one hand, C++ will default initialise all these fields to 0 already, without us having to be explicit.

C++ can be weird at times. C++ will initialize the member fields to 0 when value-initialization, direct-initialization and aggregate-initialization is used, but C++ will not do any initialization for the member fields when default-initialization is used and in this case the members may have garbage values.

lib/Target/RISCV/RISCVLDBackend.cpp
parth-07: Thanks @lenary for the discussion. I was incorrect here. We do need to use signedAddress wrap-around here to preserve the behavior of the code before the relaxation. Relocation overflow/truncation diagnostics should not be applied here.

lib/Target/RISCV/RISCVLDBackend.h
lenary: Ok, I've decided to add a `Unknown` access type to the enum, so the whole thing can be default/zero initialized using `QCAccess access = {}`, and then we can check whether the access is actually valid. Given how this is written, I've felt it safe to put the `access.isValid()` checks into asserts rather than into "real" code, as the functions should always return before doing anything with an invalid access info. Hopefully this alleviates concerns.

quic-seaswara (approved):

lenary:
@parth-07 please can you look and maybe approve? I think I need another approval before I can merge.

parth-07 (approved):
LGTM.

PR #1148, opened 2026-05-07
Fix ELFObjectWriter::emitRelocation function

parth-07:
This commit fixes the parameter types of ELFObjectWriter::emitRelocation function to use 64-bit values for r_offset and r_addend fields.

### DIFF
--- a/lib/Writers/ELFObjectWriter.cpp
+++ b/lib/Writers/ELFObjectWriter.cpp
@@ -843,20 +843,27 @@ bool ELFObjectWriter::shouldEmitReloc(const Relocation *R) const {
                                               /*IgnoreUnknown=*/true);
 }
 
-/// emitRelocation - write data to the ELF32_Rel entry
+/// emitRelocation - write data to the ELFX_Rel entry
+/// Please note that it is okay to use 64-bit types for
+/// ELFX_Rel because 64-bit signed type can represent
+/// all the values of 32-bit signed type and the conversion
+/// between 32-bit and 64-bit is well-defined as long as the
+/// value fits in the 32-bit signed range. Similar reasoning
+/// is also valid for the conversion between 32-bit and 64-bit
+/// unsigned values.
 template <typename ELFT>
 void ELFObjectWriter::emitRelocation(typename ELFT::Rel &pRel,
                                      Relocation::Type pType, uint32_t pSymIdx,
-                                     uint32_t pOffset) const {
+                                     uint64_t pOffset) const {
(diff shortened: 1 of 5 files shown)


lib/Writers/ELFObjectWriter.cpp
```diff
@@ -894,13 +901,13 @@ template uint64_t
...
-    uint32_t pOffset) const;
+    uint64_t pOffset) const;
 template void ELFObjectWriter::emitRelocation<llvm::object::ELF32LE>(
     llvm::object::ELF32LE::Rela &pRela, Relocation::Type pType,
-    uint32_t pSymIdx, uint32_t pOffset, int32_t Addend) const;
+    uint32_t pSymIdx, uint64_t pOffset, int64_t Addend) const;
 template void ELFObjectWriter::emitRelocation<llvm::object::ELF64LE>(
     llvm::object::ELF64LE::Rela &pRela, Relocation::Type pType,
-    uint32_t pSymIdx, uint32_t pOffset, int32_t Addend) const;
+    uint32_t pSymIdx, uint64_t pOffset, int64_t Addend) const;
```
quic-seaswara, line 913: Why does 32bit use int64_t and uint64_t for Addend/pOffset ?

quic-seaswara (changes requested):

test/Common/standalone/64BitRela/64BitRela.test
```diff
@@ -0,0 +1,13 @@
+UNSUPPORTED: arm, riscv32, hexagon
```
quic-areg: `UNSUPPORTED: 32BitsArch`

quic-areg (approved):

lib/Writers/ELFObjectWriter.cpp
parth-07: I have explained in the comment at line 847 (https://github.com/qualcomm/eld/pull/1148/changes#diff-a7e6ebf4bc20b779030a84bfddbf06efca9f98514d81bebdbc6041d971584d1fR847) that it is harmless to use uint64_t/int64_t types for uint32_t/int32_t here. We anyways store the original values as 64-bit in many cases.

test/Common/standalone/64BitRela/64BitRela.test
parth-07: This is cool! I did not know about it. Thanks. I have updated the test to use it.

quic-seaswara (approved):

PR #1151, opened 2026-05-08
Make version script pattern matching GNU-compatible

### LINKED ISSUE #1132: Version script symbol descriptions matching needs to be improved
Currently, eld follows sequential order for matching version script symbol descriptions with the symbols. This produces wrong result in many cases and is not what lld and bfd does. 

Reproducer:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
int foo() { return 1; }
int bar() { return 3; }
EOF

cat >vs.t <<\EOF
{
  global:
    *;
  local:
    bar;
};
EOF

LDs=(ld.eld ld.lld ld.bfd)
SFs=(eld lld bfd)

for i in "${!SFs[@]}"; do
  ${LDs[$i]} -o lib1.${SFs[$i]}.so 1.o --version-script vs.t -shared
done
```

In this reproducer example, the symbol `bar` is local in the shared libraries created by lld and bfd, however, it is global and a part of the dynamic symbol table for the eld shared library.

We need to match version script symbol description in phases:
1) First match all the symbol descriptions with no wildcard characters
2) Then match the symbol description that has wildcard characters but is not just `*`.
3) Match `*` symbol descriptions.

parth-07:
This commit makes version script pattern matching GNU-compatible. Uptil now, eld version script pattern matching was very intuitive, and straightforward. Iterate the version nodes and blocks in order and match them with the global symbols, and the first match wins. However, GNU version script pattern matching are a little complex. The updated version script pattern matching follows the below rules:

1) Phase 1: Iterate the version nodes and blocks in order, match symbols
   with the non-wildcard patterns. The first match wins.
2) Phase 2: Iterate the version nodes and blocks in order, match
   unmatched symbols with the non-* wildcard patterns. The last match wins.
3) Phase 3: Iterate the version nodes and blocks in order, match
   unmatched symbols with the '*' wildcard pattern. The last match wins.

Resolves #1132

### DIFF
--- a/include/eld/Diagnostics/DiagDynamicLink.inc
+++ b/include/eld/Diagnostics/DiagDynamicLink.inc
@@ -18,3 +18,5 @@ DIAG(error_discarded_dynamic_section_required, DiagnosticEngine::Error,
      "symbol '%0' referenced in file '%1' depends upon discarded %2 section")
 DIAG(trace_set_soname, DiagnosticEngine::Trace,
      "Setting '%0' as the soname")
+DIAG(warn_version_script_reassign, DiagnosticEngine::Warning,
+     "%0: attempt to reassign symbol '%1' of version '%2' to version '%3'")
--- a/include/eld/Config/LinkerConfig.h
+++ b/include/eld/Config/LinkerConfig.h
@@ -85,6 +85,7 @@ public:
     std::optional<bool> EnableWholeArchiveWarnings;
     std::optional<bool> EnableCommandLineWarnings;
     std::optional<bool> EnableOSABIWarnings;
+    std::optional<bool> EnableVersionScriptWarnings;
   };
 
   struct MappingFileInfo {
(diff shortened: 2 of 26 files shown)


include/eld/Diagnostics/DiagDynamicLink.inc
```diff
@@ -19,3 +19,5 @@ DIAG(error_discarded_dynamic_section_required, DiagnosticEngine::Error,
      "symbol '%0' referenced in file '%1' depends upon discarded %2 section")
 DIAG(trace_set_soname, DiagnosticEngine::Trace,
      "Setting '%0' as the soname")
+DIAG(warn_version_script_reassign, DiagnosticEngine::Warning,
+     "attempt to reassign symbol '%0' of version '%1' to version '%2'")
```
quic-seaswara: add the version script file for users to know where is the origin of the version script assignment

quic-seaswara: Do we need to default warn or predicate by using -Wversion-script ?

quic-seaswara (approved):

include/eld/Diagnostics/DiagDynamicLink.inc
parth-07: I have added the version script origin in the diagnostic.

parth-07: I have added `-Wversion-script` flag.

quic-seaswara:
Let us go ahead and merge this.

quic-seaswara:
We would need the map file to be updated with version script information.

parth-07:
> We would need the map file to be updated with version script information.

Can I please add the map-file support as a follow-up patch? I believe that the map-file support should be an independent change.

parth-07:
@quic-seaswara can this PR be merged?

Steven6798 (approved):

quic-seaswara (approved):

PR #1153, opened 2026-05-10
test: add test for anonymous version definition diagnostic

### LINKED ISSUE #1124: There is no test for the error `anonymous version definition is used in combination with other version definitions`
We do not have a test that covers the error diagnostic: `anonymous version definition is used in combination with other version definitions`. All diagnostics must be covered by a test.

sai18022001:
Fixes #1124 

Adds a lit test covering the error diagnostic emitted when an anonymous 
version definition is used in combination with other named version 
definitions in a version script.

Problem : 
The diagnostic `anonymous version definition is used in combination with 
other version definitions` (emitted in `ScriptParser::readVersionScriptCommand()`) 
had no test coverage.

Solution : 
Added a new lit test at `test/Common/standalone/VersionScriptAnonymousMixed/` 
with the following files:

- `Inputs/1.c` — minimal C source file to produce an object file for linking
- `Inputs/vs_mixed` — version script mixing a named and anonymous version node to trigger the diagnostic
- `VersionScriptAnonymousMixed.test` — compiles `1.c` into an object file, 
  links it as a shared library with `vs_mixed` via `--version-script=`, 
  expects the link to fail (`%not`), and verifies the error message using 
  `FileCheck`

Testing : 
Tested on AArch64, ARM, RISCV, RISCV64, Hexagon, x86 - > all pass.

### DIFF
--- a/test/Common/standalone/VersionScriptAnonymousMixed/Inputs/vs_mixed
+++ b/test/Common/standalone/VersionScriptAnonymousMixed/Inputs/vs_mixed
@@ -1,0 +1,6 @@
+VER_1.0 {
+  global: foo;
+};
+{
+  global: bar;
+};
--- a/test/Common/standalone/VersionScriptAnonymousMixed/VersionScriptAnonymousMixed.test
+++ b/test/Common/standalone/VersionScriptAnonymousMixed/VersionScriptAnonymousMixed.test
@@ -1,0 +1,13 @@
+#---VersionScriptAnonymousMixed.test--------------------- SharedLibrary,VS------------------#
+#BEGIN_COMMENT
+# Test that an error is emitted when an anonymous version definition
+# is used in combination with other named version definitions
+# in a version script.
+#END_COMMENT
+
+RUN: %clang %clangopts -c -fpic %p/Inputs/1.c -o %t1.o
+RUN: %not %link %linkopts -shared -o %t1.so %t1.o --version-script=%p/Inputs/vs_mixed 2>&1 | %filecheck %s --check-prefix=NAMED
+RUN: %not %link %linkopts -shared -o %t1.so %t1.o --version-script=%p/Inputs/vs_anon_first 2>&1 | %filecheck %s --check-prefix=ANON
+
+NAMED: Error: {{.*}}: anonymous version definition is used in combination with other version definitions
+ANON: Error: {{.*}}: EOF expected, but got VER_1.0
--- a/test/Common/standalone/VersionScriptAnonymousMixed/Inputs/1.c
+++ b/test/Common/standalone/VersionScriptAnonymousMixed/Inputs/1.c
@@ -1,0 +1,2 @@
+void foo() {}
+void bar() {}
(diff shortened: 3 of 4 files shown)


test/Common/standalone/VersionScriptAnonymousMixed/VersionScriptAnonymousMixed.test
```diff
@@ -0,0 +1,14 @@
+UNSUPPORTED: x86
```
parth-07: Is it required to explicitly disable the x86 architecture here?

test/Common/standalone/VersionScriptAnonymousMixed/VersionScriptAnonymousMixed.test
```diff
@@ -0,0 +1,14 @@
...
+#---VersionScriptAnonymousMixed.test--------------------- SharedLibrary,VS------------------#
+#BEGIN_COMMENT
+# Test that an error is emitted when an anonymous version definition
+# is used in combination with other named version definitions
+# in a version script.
+# See: https://github.com/qualcomm/eld/issues/1124
```
parth-07:
```suggestion

```

test/Common/standalone/VersionScriptAnonymousMixed/Inputs/vs_mixed
```diff
@@ -0,0 +1,6 @@
+VER_1.0 {
```
parth-07, line 1: Can you please also add a test where anonymous version is first and then we have named version node?

sai18022001: @parth-07 
The anonymous-first case produces a different error: `EOF expected, but got VER_1.0` rather than the `anonymous version definition is used in combination with other version definitions` . 
Should I add a separate CHECK for that error ?

test/Common/standalone/VersionScriptAnonymousMixed/VersionScriptAnonymousMixed.test
sai18022001: It's not required actually

parth-07: In that case, can you please remove it?

test/Common/standalone/VersionScriptAnonymousMixed/Inputs/vs_mixed
parth-07: Yes, please add a separate check for that.

sai18022001:
Hi @parth-07 , All review comments have been addressed. Requesting a re-review when you get a chance.

parth-07 (approved):
LGTM.

sai18022001:
The CI failure is in `lib/Core/LinkerScript.cpp:191 (use of undeclared identifier 'xxHash64')` which is unrelated to this PR, only added test files with no C++ changes. 
@parth-07 Can you please check and confirm ?

sai18022001:
@parth-07 Rebased onto latest main. Could you approve the workflows to re-run CI?

parth-07 (approved):
LGTM.

Steven6798 (approved):

PR #1156, opened 2026-05-11
Fix ELFObjectWriter::emitRelocation function

### LINKED ISSUE #1149: Incorrect addend in the R_*_RELATIVE relocation when the addend is greater than INT32_MAX
We generate incorrect addent in the R_*_RELATIVE rela relocation when the addend value is greater than INT32_MAX.

Reproducer:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
static int u = 11;
int *v = &u;

int foo() {
  return 1;
}
EOF

clang -o 1.o -target aarch64-unknown-elf 1.c -c -fPIC
ld.eld -o lib1.so 1.o -shared --section-start=.data=0x80000000
llvm-readelf -r lib1.so
```

Output:

```
Relocation section '.rela.dyn' at offset 0x1c8 contains 1 entries:
0000000080000008  0000000000000403 R_AARCH64_RELATIVE                ffffffff80000000
```

The addend correct value is `80000000`.

parth-07:
This commit fixes the parameter types of ELFObjectWriter::emitRelocation function to use 64-bit values for r_offset and r_addend fields.

Resolves #1149

### DIFF
--- a/include/eld/Writers/ELFObjectWriter.h
+++ b/include/eld/Writers/ELFObjectWriter.h
@@ -102,12 +102,12 @@ private:
   // Emit SHT_REL/SHT_RELA relocations
   template <typename ELFT>
   void emitRelocation(typename ELFT::Rel &PRel, Relocation::Type CurType,
-                      uint32_t CurSymIdx, uint32_t CurOffset) const;
+                      uint32_t CurSymIdx, uint64_t CurOffset) const;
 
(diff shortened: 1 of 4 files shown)


quic-seaswara:
@apazos please approve this patch for merge.

quic-seaswara (approved):

apazos (approved):

PR #1157, opened 2026-05-11
Add run test framework for ARM, AArch64, RISCV32 and RISCV64

parth-07:
This commit adds the run test framework for ARM, AARch64, RISCV32,
RISCV64 and x86-64. Additionally, it modifies the ci.yaml such that the tools
requirements to enable run tests for these targets is available.

The added HelloWorld and HelloWorldCXX test are marked XFAIL for the riscv target
because of eld bugs. QEMU is crashing for static hello world program likely
because of missing IFunc support and linker is unable to find a shared library that is
indeed present in the sysroot for the dynamic hello world program.

The HelloWorld and HelloWorldCXX test are marked UNSUPPORTED for x86-64
because this test is passing on release builds but is causing a
linker crash on builds with assertions enabled. The crash is happening
when reading the compressed sections. It cannot be marked as XFAIL
because the test will pass on the CI builder, and will cause the
'check-eld' step to fail.


.github/workflows/ci.yml
```diff
@@ -106,7 +107,18 @@ jobs:
...
+            -DELD_ARM_EMULATOR="qemu-arm" \
+            -DELD_AARCH64_C_COMPILER=/local/mnt/workspace/aarch64-unknown-linux-gnu/bin/aarch64-unknown-linux-gnu-cc \
+            -DELD_AARCH64_SYSROOT=/local/mnt/workspace/aarch64-unknown-linux-gnu/aarch64-unknown-linux-gnu/sysroot \
+            -DELD_AARCH64_EMULATOR="qemu-aarch64" \
+            -DELD_RISCV32_C_COMPILER=/local/mnt/workspace/riscv32-unknown-linux-gnu/bin/riscv32-unknown-linux-gnu-cc \
+            -DELD_RISCV32_SYSROOT=/local/mnt/workspace/riscv32-unknown-linux-gnu/riscv32-unknown-linux-gnu/sysroot \
+            -DELD_RISCV32_EMULATOR="qemu-riscv32" \
+            -DELD_RISCV64_C_COMPILER=/local/mnt/workspace/riscv64-unknown-linux-gnu/bin/riscv64-unknown-linux-gnu-cc \
+            -DELD_RISCV64_SYSROOT=/local/mnt/workspace/riscv64-unknown-linux-gnu/riscv64-unknown-linux-gnu/sysroot \
+            -DELD_RISCV64_EMULATOR="qemu-riscv64" \
```
quic-seaswara: Why these variables, can we have a cmake module that will find the tools and the sysroot in /local/mnt/workspace/

We can have a flag BUILD_AND_RUN_TESTS:BOOL=TRUE, which will allow this cmake module to be sourced and variables added to the tree.

parth-07: I believe that we should not hardcode any paths in the CMake as that is not scalable and will make it difficult for outside users to enable these tests. We explicitly need to specify C/CXX compiler, sysroot and emulator because it is possible that the cross-toolchain that is used by the user has different naming/path conventions than what we have in clang-cross GitHub releases.

test/Common/RunTests/HelloWorld/HelloWorld.test
```diff
@@ -0,0 +1,18 @@
+REQUIRES: run_tests
```
quic-seaswara: We require run_tests feature per target, because the cmake variables are seperated.

test/Common/RunTests/HelloWorld/HelloWorld.test
```diff
@@ -0,0 +1,18 @@
...
+UNSUPPORTED: riscv32, riscv64
```
quic-seaswara: Why unsupported for riscv32 / riscv64 ?

.github/workflows/ci.yml
```diff
@@ -106,7 +107,18 @@ jobs:
             -DLLVM_CCACHE_DIR:STRING="$CCACHE_DIR" \
             -DELD_ENABLE_SYMBOL_VERSIONING=ON \
             -DLLVM_ENABLE_WERROR:BOOL=ON \
-            ../llvm-project/llvm
+            -DELD_ARM_C_COMPILER=/local/mnt/workspace/arm-unknown-linux-gnueabi/bin/arm-unknown-linux-gnueabi-cc \
+            -DELD_ARM_SYSROOT=/local/mnt/workspace/arm-unknown-linux-gnueabi/arm-unknown-linux-gnueabi/sysroot \
+            -DELD_ARM_EMULATOR="qemu-arm" \
+            -DELD_AARCH64_C_COMPILER=/local/mnt/workspace/aarch64-unknown-linux-gnu/bin/aarch64-unknown-linux-gnu-cc \
+            -DELD_AARCH64_SYSROOT=/local/mnt/workspace/aarch64-unknown-linux-gnu/aarch64-unknown-linux-gnu/sysroot \
+            -DELD_AARCH64_EMULATOR="qemu-aarch64" \
```
quic-seaswara: SYSTEM_EMULATOR and USER_EMULATOR ?

.github/workflows/ci.yml
quic-seaswara: We should have some defaults for emulators instead of specifying per target.

.github/workflows/ci.yml
```diff
@@ -123,10 +135,23 @@ jobs:
         working-directory: pr-${{ env.PR_NUMBER }}/obj
         run: ccache --show-stats --dir=$CCACHE_DIR
 
+      - name: Get clang-cross toolchains
+        if: steps.file-check.outputs.skip_build == 'false'
+        uses: ./.github/workflows/clang-cross-download.yml
```
quic-seaswara: Can we have a variant of clang-cross-download.yml for ci's ? This may be something that we will need in nightlies eventually and these lines get duplicated.

.github/workflows/ci.yml
```diff
@@ -123,10 +135,23 @@ jobs:
...
+
       - name: Run tests
         if: steps.file-check.outputs.skip_build == 'false'
         working-directory: pr-${{ env.PR_NUMBER }}/obj
-        run: ninja check-eld
+        run: |
+          ls /opt/qemu
+          ls /opt/qemu/bin
+          export PATH="$PATH:/opt/qemu/bin"
+          ninja check-eld
```
quic-seaswara, line 159: can we have a target check-eld-run to run these tests, this will be useful to measure how many tests we are able to build and run, and measure our qa improvements.

CMakeLists.txt
```diff
@@ -266,6 +266,46 @@ add_subdirectory(tools)
...
+# Cross-compilation toolchain for RISCV64 build/run tests
+set(ELD_RISCV64_C_COMPILER "" CACHE FILEPATH
+    "C compiler for RISCV64 run tests (e.g., riscv64-linux-gnu-gcc)")
+set(ELD_RISCV64_CXX_COMPILER "" CACHE FILEPATH
+    "C++ compiler for RISCV64 run tests (e.g., riscv64-linux-gnu-g++)")
+set(ELD_RISCV64_SYSROOT "" CACHE PATH
+    "Sysroot for RISCV64 run tests")
+set(ELD_RISCV64_EMULATOR "" CACHE STRING
+    "Emulator for RISCV64 run tests (e.g., qemu-riscv64)")
+
```
quic-seaswara, line 272: It might be useful to have some defaults

.github/workflows/ci.yml
```diff
@@ -106,7 +107,18 @@ jobs:
...
+            -DELD_ARM_SYSROOT=/local/mnt/workspace/arm-unknown-linux-gnueabi/arm-unknown-linux-gnueabi/sysroot \
+            -DELD_ARM_EMULATOR="qemu-arm" \
+            -DELD_AARCH64_C_COMPILER=/local/mnt/workspace/aarch64-unknown-linux-gnu/bin/aarch64-unknown-linux-gnu-cc \
+            -DELD_AARCH64_SYSROOT=/local/mnt/workspace/aarch64-unknown-linux-gnu/aarch64-unknown-linux-gnu/sysroot \
+            -DELD_AARCH64_EMULATOR="qemu-aarch64" \
+            -DELD_RISCV32_C_COMPILER=/local/mnt/workspace/riscv32-unknown-linux-gnu/bin/riscv32-unknown-linux-gnu-cc \
+            -DELD_RISCV32_SYSROOT=/local/mnt/workspace/riscv32-unknown-linux-gnu/riscv32-unknown-linux-gnu/sysroot \
+            -DELD_RISCV32_EMULATOR="qemu-riscv32" \
+            -DELD_RISCV64_C_COMPILER=/local/mnt/workspace/riscv64-unknown-linux-gnu/bin/riscv64-unknown-linux-gnu-cc \
+            -DELD_RISCV64_SYSROOT=/local/mnt/workspace/riscv64-unknown-linux-gnu/riscv64-unknown-linux-gnu/sysroot \
```
quic-seaswara: We probably want to check the sysroot to check if we are able to compile / build before running the build ?

CMakeLists.txt
```diff
@@ -266,6 +266,46 @@ add_subdirectory(tools)
...
+set(ELD_ARM_SYSROOT "" CACHE PATH
+    "Sysroot for ARM run tests")
+set(ELD_ARM_EMULATOR "" CACHE STRING
+    "Emulator for ARM run tests (e.g., qemu-arm)")
+
+# Cross-compilation toolchain for RISCV32 build/run tests
+set(ELD_RISCV32_C_COMPILER "" CACHE FILEPATH
+    "C compiler for RISCV32 run tests (e.g., riscv32-linux-gnu-gcc)")
+set(ELD_RISCV32_CXX_COMPILER "" CACHE FILEPATH
+    "C++ compiler for RISCV32 run tests (e.g., riscv32-linux-gnu-g++)")
```
quic-seaswara: can we use clang instead of gcc for riscv32 and riscv64 ?

quic-seaswara (changes requested):

test/Common/RunTests/HelloWorld/HelloWorld.test
parth-07: It is due to eld bugs in riscv32/riscv64. I have added below in the commit message and the PR description:

> The added HelloWorld test is disabled for the riscv target as of now
because qemu is crashing for static hello world program and unable to
find a shared library that is present in the sysroot for the dynamic hello
world program. These issues will be addressed in follow-up patches.

test/Common/RunTests/HelloWorld/HelloWorld.test
parth-07: I believe that we do not need this per-target. We run tests with target-specific lit configurations. For a particular lit configuration, run_tests is only added as a feature if the cross-compiler and emulator was specified in the CMake for that target. 

For example, if we specify the cross-compiler and emulator for arm and aarch64, then `run_tests` will be added as a feature for llvm-lit-arm and llvm-lit-aarch64, but it will not be added as a feature for llvm-lit-riscv and llvm-lit-riscv64.

.github/workflows/ci.yml
parth-07: Why do we need SYSTEM_EMULATOR? Can we add the system emulator support when the need arises for it?

.github/workflows/ci.yml
parth-07: I agree we can have defaults for emulators. Should I change it in this patch?

.github/workflows/ci.yml
parth-07: > Can we have a variant of clang-cross-download.yml for ci's ?

We already have .github/workflows/clang-cross-schedule.yml, right? Here the key issue is that we have two github self-hosted machines, and currently one machine contains the clang-cross toolchains but the other does not.

And currently I am not able to properly use clang-cross-download.yml in this workflow. `uses` expects strings without any variables, and we clone eld in `pr-{{ env.PR_NUMBER }}/llvm-project/....` -- it uses a variable.

.github/workflows/ci.yml
parth-07: I like this idea, but I believe that it should be handled as a separate task.

.github/workflows/ci.yml
parth-07: I agree it might be helpful, but I feel that it is much more work than worth it. If the sysroot or cross-compiler are incorrect, then the `HelloWorld` test will fail and we will know that something is wrong. Please let me know your thoughts on it.

CMakeLists.txt
parth-07: I have changed the names here to use clang instead of gcc.

CMakeLists.txt
parth-07: I have added defaults for all the emulator options.

.github/workflows/ci.yml
parth-07: For now, I have made sure that both the self-hosted runners have the clang-cross toolchains by explicitly running the clang-cross-schedule workflow multiple times...

.github/workflows/ci.yml
quic-seaswara: Sure, we can. But I feel it is a small change. Can we integrate this in this patch ?

.github/workflows/ci.yml
quic-seaswara: Sure

.github/workflows/ci.yml
quic-seaswara: Sure

.github/workflows/ci.yml
quic-seaswara: bareminimum we should check if those paths are valid :)

.github/workflows/ci.yml
parth-07: The key issue is that we won't be using it anywhere. We will have CMake / lit variables that would have no current use.

.github/workflows/ci.yml
quic-seaswara: I would prefer the paths point to the bin folder, and lit.cfg can pick up either `cc/clang/clang++`.  

Use -DELD_RISCV_COMPILER=/local/mnt/workspace/riscv64-unknown-linux-gnu/ instead of full path to the compiler ? 

Thoughts ?

quic-seaswara:
Probably complete the list with adding X86 so that we have a stable sysroot to test with.

.github/workflows/ci.yml
parth-07: I have added the checks to ensure that the paths are valid.

parth-07: > I would prefer the paths point to the bin folder, and lit.cfg can pick up either cc/clang/clang++

How do we know the compiler name? Cross-compiler toolchains typically contain a wrapper over the compiler with the right flags. Also, what happens if someone want to use GNU or some other compiler for verifying?

parth-07:
> Probably complete the list with adding X86 so that we have a stable sysroot to test with.

I have added x86-64 cross-toolchain as well.

.github/workflows/ci.yml
```diff
@@ -106,7 +107,16 @@ jobs:
...
+            -DELD_ARM_C_COMPILER=/local/mnt/workspace/arm-unknown-linux-gnueabi/bin/arm-unknown-linux-gnueabi-cc \
+            -DELD_ARM_SYSROOT=/local/mnt/workspace/arm-unknown-linux-gnueabi/arm-unknown-linux-gnueabi/sysroot \
+            -DELD_AARCH64_C_COMPILER=/local/mnt/workspace/aarch64-unknown-linux-gnu/bin/aarch64-unknown-linux-gnu-cc \
+            -DELD_AARCH64_SYSROOT=/local/mnt/workspace/aarch64-unknown-linux-gnu/aarch64-unknown-linux-gnu/sysroot \
+            -DELD_RISCV32_C_COMPILER=/local/mnt/workspace/riscv32-unknown-linux-gnu/bin/riscv32-unknown-linux-gnu-cc \
+            -DELD_RISCV32_SYSROOT=/local/mnt/workspace/riscv32-unknown-linux-gnu/riscv32-unknown-linux-gnu/sysroot \
+            -DELD_RISCV64_C_COMPILER=/local/mnt/workspace/riscv64-unknown-linux-gnu/bin/riscv64-unknown-linux-gnu-cc \
+            -DELD_RISCV64_SYSROOT=/local/mnt/workspace/riscv64-unknown-linux-gnu/riscv64-unknown-linux-gnu/sysroot \
+            -DELD_X86_64_C_COMPILER=/local/mnt/workspace/x86_64-unknown-linux-gnu//bin/x86_64-unknown-linux-gnu-clang \
+            -DELD_X86_64_SYSROOT=/local/mnt/workspace/x86_64-unknown-linux-gnu/x86_64-unknown-linux-gnu/sysroot \
```
quic-seaswara: Update with CXX compiler path also here ? 

Also update README

CMakeLists.txt
```diff
@@ -266,6 +266,68 @@ add_subdirectory(tools)
 add_subdirectory(docs)
 add_subdirectory(etc)
 
+# Cross-compilation toolchain for AArch64 build/run tests
```
quic-areg: This could be simplified with something like:

```cmake
set(ELD_RUN_TEST_ARCHS AARCH64 ARM RISCV32 RISCV64 X86_64)
set(ELD_AARCH64_DEFAULT_EMULATOR "qemu-aarch64")
set(ELD_ARM_DEFAULT_EMULATOR     "qemu-arm")
set(ELD_RISCV32_DEFAULT_EMULATOR "qemu-riscv32")
set(ELD_RISCV64_DEFAULT_EMULATOR "qemu-riscv64")
set(ELD_X86_64_DEFAULT_EMULATOR  "qemu-x86_64")

foreach(_arch IN LISTS ELD_RUN_TEST_ARCHS)
  set(ELD_${_arch}_C_COMPILER   "" CACHE FILEPATH "C compiler for ${_arch} run tests")
  set(ELD_${_arch}_CXX_COMPILER "" CACHE FILEPATH "C++ compiler for ${_arch} run tests")
  set(ELD_${_arch}_SYSROOT      "" CACHE PATH     "Sysroot for ${_arch} run tests")
  set(ELD_${_arch}_EMULATOR
      "${ELD_${_arch}_DEFAULT_EMULATOR}"
      CACHE STRING "Emulator for ${_arch} run tests")
  foreach(_var C_COMPILER CXX_COMPILER SYSROOT)
    set(_full "ELD_${_arch}_${_var}")
    if(${_full} AND NOT EXISTS "${${_full}}")
      message(FATAL_ERROR "${_full} is set to '${${_full}}' but does not exist")
    endif()
  endforeach()
endforeach()
```

test/lit.cfg
```diff
@@ -327,6 +332,11 @@ if config.test_target == 'X86':
     if hasattr(config,'eld_option') and config.eld_option != 'default':
         linkopts = '--thread-count 4 ' + config.eld_option
     config.available_features.add('x86')
+    # Run tests support for X86_64
```
quic-areg: can we set them all at once?

```py
run_test_arches = {
    'X86':     'x86_64',
    'ARM':     'arm',
    'AArch64': 'aarch64',
    'RISCV':   'riscv32',
    'RISCV64': 'riscv64',
}
arch = run_test_arches.get(config.test_target)
if arch:
    run_cc   = getattr(config, f'{arch}_c_compiler',   '')
    run_cxx  = getattr(config, f'{arch}_cxx_compiler', '')
    sysroot  = getattr(config, f'{arch}_sysroot',      '')
    emulator = getattr(config, f'{arch}_emulator',     '')
```

test/Common/RunTests/HelloWorld/HelloWorld.test
quic-areg: XFAIL instead?

CMakeLists.txt
parth-07: Thank you. I have simplified the code using this suggestion.

test/lit.cfg
parth-07: Thank you. I have simplified the code using this suggestion.

.github/workflows/ci.yml
parth-07: I have added CXX compiler here and a hello world C++ test.

I have also updated the README with these new CMake variables information.

test/Common/RunTests/HelloWorld/HelloWorld.test
parth-07: Done.

.github/workflows/ci.yml
```diff
@@ -106,7 +107,21 @@ jobs:
...
+            -DELD_AARCH64_SYSROOT=/local/mnt/workspace/aarch64-unknown-linux-gnu/aarch64-unknown-linux-gnu/sysroot \
+            -DELD_RISCV32_C_COMPILER=/local/mnt/workspace/riscv32-unknown-linux-gnu/bin/riscv32-unknown-linux-gnu-cc \
+            -DELD_RISCV32_CXX_COMPILER=/local/mnt/workspace/riscv32-unknown-linux-gnu/bin/riscv32-unknown-linux-gnu-c++ \
+            -DELD_RISCV32_SYSROOT=/local/mnt/workspace/riscv32-unknown-linux-gnu/riscv32-unknown-linux-gnu/sysroot \
+            -DELD_RISCV64_C_COMPILER=/local/mnt/workspace/riscv64-unknown-linux-gnu/bin/riscv64-unknown-linux-gnu-cc \
+            -DELD_RISCV64_CXX_COMPILER=/local/mnt/workspace/riscv64-unknown-linux-gnu/bin/riscv64-unknown-linux-gnu-c++ \
+            -DELD_RISCV64_SYSROOT=/local/mnt/workspace/riscv64-unknown-linux-gnu/riscv64-unknown-linux-gnu/sysroot \
+            -DELD_X86_64_C_COMPILER=/local/mnt/workspace/x86_64-unknown-linux-gnu//bin/x86_64-unknown-linux-gnu-cc \
+            -DELD_X86_64_CXX_COMPILER=/local/mnt/workspace/x86_64-unknown-linux-gnu//bin/x86_64-unknown-linux-gnu-c++ \
+            -DELD_X86_64_SYSROOT=/local/mnt/workspace/x86_64-unknown-linux-gnu/x86_64-unknown-linux-gnu/sysroot \
```
quic-seaswara: We should use clang instead of gcc here, for compatibility reasons that we have not tested with gcc.

.github/workflows/ci.yml
```diff
@@ -106,7 +107,21 @@ jobs:
             -DLLVM_CCACHE_DIR:STRING="$CCACHE_DIR" \
             -DELD_ENABLE_SYMBOL_VERSIONING=ON \
             -DLLVM_ENABLE_WERROR:BOOL=ON \
-            ../llvm-project/llvm
+            -DELD_ARM_C_COMPILER=/local/mnt/workspace/arm-unknown-linux-gnueabi/bin/arm-unknown-linux-gnueabi-cc \
+            -DELD_ARM_CXX_COMPILER=/local/mnt/workspace/arm-unknown-linux-gnueabi/bin/arm-unknown-linux-gnueabi-c++ \
```
quic-seaswara: I am worried that we are able to test only one compiler and one sysroot at all times, is that the intent ? 

We would want to use the musl sysroot to begin with instead of glibc sysroot.

test/Common/RunTests/HelloWorld/HelloWorld.test
```diff
@@ -0,0 +1,19 @@
...
+# This test requires the run_tests feature which is enabled when the target's
+# C compiler, sysroot, and emulator are all configured.
+#END_COMMENT
+#START_TEST
+RUN: %run_cc -o %t1.o %p/Inputs/HelloWorld.c -c -fno-pic
+RUN: %run_cc -o %t.out %t1.o -static
+RUN: %run %t.out | %filecheck %s
+RUN: %run_cc -o %t1.pic.o %p/Inputs/HelloWorld.c -c -fPIC
+RUN: %run_cc -o %t.dyn.out %t1.pic.o -Wl,-dy
+RUN: %run %t.dyn.out | %filecheck %s
```
quic-seaswara, line 16: You need QEMU_LD_PREFIX for qemu to locate the dynamic shared libraries.

quic-seaswara, line 16: Can we add QEMU_LD_PREFIX that can default to sysroot ?

test/Common/RunTests/HelloWorld/HelloWorld.test
```diff
@@ -0,0 +1,19 @@
+REQUIRES: run_test
+XFAIL: riscv32, riscv64
+UNSUPPORTED: x86
```
quic-seaswara, line 3: Let us use musl for now and remove XFAIL and unsupported.

README.md
```diff
@@ -118,6 +118,36 @@ This command runs only the AliasSymbolExportDynamic test using the ARM default c
...
+
+The `run_test` lit feature is enabled when the C compiler, sysroot, and emulator are all configured for a target. The `run_cxx_test` feature additionally requires the C++ compiler.
+
+#### Example: Configuring for AArch64
+
+```bash
+cmake -G Ninja \
+  -DELD_AARCH64_C_COMPILER=/usr/bin/aarch64-linux-gnu-gcc \
+  -DELD_AARCH64_CXX_COMPILER=/usr/bin/aarch64-linux-gnu-g++ \
+  -DELD_AARCH64_SYSROOT=/usr/aarch64-linux-gnu \
```
quic-seaswara: Is this path valid ? /usr/aarch64-linux-gnu ?

quic-seaswara (changes requested):

.github/workflows/ci.yml
parth-07: We are using clang only. `gnu` here refers to the GNU libc. `-cc` and `-c++` are symlinks to the `x86_64-unknown-linux-gnu-clang` and `x86_64-unknown-linux-gnu-clang++` respectively.

.github/workflows/ci.yml
parth-07: > I am worried that we are able to test only one compiler and one sysroot at all times, is that the intent ?

Yes, that is the current intent. I think it is an overkill to have a framework that allows testing with multiple toolchains for the same architecture. If we really need this, then we can plan to extend the framework in the future to support this. 

> We would want to use the musl sysroot to begin with instead of glibc sysroot.

As we discussed before, the primary motivation for using glibc-based toolchain is so that we can properly test features such as GNU IFunc and symbol versioning.

test/Common/RunTests/HelloWorld/HelloWorld.test
parth-07: Why do we need `QEMU_LD_PREFIX`? `%run` expansion contains `-L <sysroot>`.

parth-07: `%run` contains `-L <sysroot>`. I believe that command-line option is a better approach than the environment variable here because it has less chances of having issues due to shell.

test/Common/RunTests/HelloWorld/HelloWorld.test
parth-07: We need to use GNU libc to properly test for GNU IFunc and symbol versioning.

README.md
parth-07: This is an example in the README.

test/Common/RunTests/HelloWorld/HelloWorld.test
quic-seaswara: Ideally yes, for now we can selectively choose gnu sysroot or musl sysroot, but primarily musl sysroot to begin with to debug issues easily.

parth-07: We currently need GNU libc support for testing IFunc and symbol versioning, and I have already put in the efforts for enabling the support with GNU libc. Can we add support for musl libc build and run tests later?

quic-seaswara: Ok

test/Common/RunTests/HelloWorld/HelloWorld.test
quic-seaswara: Ok

quic-seaswara (approved):

quic-seaswara:
I prefer we use environment variables instead of cmake variables since it affects a small subset. quickly changing the environment and running is so much easy. 

We do need a blanket option in cmake to enable/disable run tests completely though.

CMakeLists.txt
```diff
@@ -266,6 +266,34 @@ add_subdirectory(tools)
...
 
+set(ELD_AARCH64_DEFAULT_EMULATOR "qemu-aarch64")
+set(ELD_ARM_DEFAULT_EMULATOR     "qemu-arm")
+set(ELD_RISCV32_DEFAULT_EMULATOR "qemu-riscv32")
+set(ELD_RISCV64_DEFAULT_EMULATOR "qemu-riscv64")
+set(ELD_X86_64_DEFAULT_EMULATOR  "qemu-x86_64")
+
+# Validate cross-compilation toolchain paths (if specified, must exist)
+foreach(_arch AARCH64 ARM RISCV32 RISCV64 X86_64)
+  set(ELD_{_arch}_C_COMPILER "" CACHE FILEPATH
```
quic-areg: ${_arch}

CMakeLists.txt
```diff
@@ -266,6 +266,34 @@ add_subdirectory(tools)
...
+      "C compiler for ${_arch} run tests")
+  set(ELD_${_arch}_CXX_COMPILER "" CACHE FILEPATH
+      "C++ compiler for ${_arch} run tests")
+  set(ELD_${_arch}_SYSROOT "" CACHE FILEPATH
```
quic-areg: should this be `PATH` instead of `FILEPATH`?

test/Common/RunTests/HelloWorld/HelloWorld.test
```diff
@@ -0,0 +1,19 @@
...
+#---HelloWorld.test--------------------- Executable------------------#
+#BEGIN_COMMENT
+# Test that we can build and run a simple executable using the run test infrastructure.
+# This test requires the run_tests feature which is enabled when the target's
```
quic-areg: run_test feature

.github/workflows/ci.yml
```diff
@@ -106,7 +107,21 @@ jobs:
...
+            -DELD_ARM_SYSROOT=/local/mnt/workspace/arm-unknown-linux-gnueabi/arm-unknown-linux-gnueabi/sysroot \
```
quic-areg: can we test these sysroots at cmake time with something like `execute_process` and-`print-file-name=crt1.o` instead of at test time? It should only be an extra few lines in the existing foreach cmake loop.

parth-07: We can do this, but I think it is an overkill. HelloWorld test will fail if something basic is wrong and we will know.

quic-seaswara: I agree with @parth-07. Building directly with clang is scalable, and these tests would be portable across sysroots(musl/glibc/picolibc)

quic-areg: Sure. Just a note, `crt1.o` is libc-agnostic.

test/Common/RunTests/HelloWorld/HelloWorld.test
parth-07: Thanks for catching this. I have fixed it.

CMakeLists.txt
parth-07: I have removed all of CMake changes, as functionality is now moved to use environment variables instead of CMake.

CMakeLists.txt
parth-07: Yes, you are right. However, all of the CMake code has now been removed, and we now use environment variables for the same purpose.

parth-07:
> I prefer we use environment variables instead of cmake variables since it affects a small subset. quickly changing the environment and running is so much easy.

Thank you for the suggestion. I have modified the solution to use environment variables instead of CMake variables to determine the cross-tools to use.

> We do need a blanket option in cmake to enable/disable run tests completely though.

I have added a blanket option 'ELD_ENABLE_RUN_TESTS' for enabling/disabling run tests.

quic-seaswara (approved):

cmake/modules/LITModules.cmake
```diff
@@ -63,6 +63,10 @@ if(ELD_ENABLE_SYMBOL_VERSIONING)
   set(ELD_ENABLE_SYMBOL_VERSIONING_TESTS "true")
 endif()
 
+if (ELD_ENABLE_RUN_TESTS)
+  set(ELD_ENABLE_RUN_TESTS "true")
```
quic-seaswara, line 67: Why do you need to set this again ?

quic-seaswara (approved):

.github/workflows/ci.yml
```diff
@@ -106,7 +107,7 @@ jobs:
             -DLLVM_CCACHE_DIR:STRING="$CCACHE_DIR" \
             -DELD_ENABLE_SYMBOL_VERSIONING=ON \
             -DLLVM_ENABLE_WERROR:BOOL=ON \
-            ../llvm-project/llvm
+            -DELD_ENABLE_RUN_TESTS=On
```
quic-seaswara: ELD_ENABLE_RUN_TESTS:BOOL=ON ?

quic-seaswara (changes requested):
README update is missing

.github/workflows/ci.yml
parth-07: I can add bool, but I don't think it is wrong/incorrect to not have BOOL here.

cmake/modules/LITModules.cmake
parth-07: Because we use this value in lit.cfg. If we directly use CMake variable, then the value can be any of On, true, 1, and whatever else the CMake considers a truth value. Here we canonicalize this value to "true" to make it easy to use in lit.cfg.

.github/workflows/ci.yml
quic-seaswara: Let us change it to bool for consistency.

cmake/modules/LITModules.cmake
quic-seaswara: You can just add a entry in lit.site.cfg.in with the cmake variable

parth-07: This patch already adds the entry for ELD_ENABLE_RUN_TESTS in lit.site.cfg.in. The issue is that the value of this variable can change for different CMake configurations. CMake considers many values as truth values.

.github/workflows/ci.yml
parth-07: I have changed it to `-DELD_ENABLE_RUN_TESTS:BOOL=On`.

parth-07:
> README update is missing

Thank you for mentioning it. I have updated the README.

quic-seaswara (approved):

PR #1158, opened 2026-05-11
[linker-script] Find namespec libraries in sysroot standard dirs

### LINKED ISSUE #819: eld does not find namespec libraries in the standard directories within the sysroot
eld does not find namespec libraries (such as `-lfoo`) in the standard directories under the sysroot such as, `/lib`, `/lib64`, `/usr/lib`, and so on.

Reproducer:

```bash
#!/usr/bin/env bash

rm -rf A
mkdir -p A/lib64

cat >1.c <<\EOF
int foo() { return 1; }
EOF

cat >script.t <<EOF
GROUP(-l1)
EOF

cat >main.c <<\EOF
int main() { return 0; }
EOF

mv lib1.so A/lib64/lib1.so
mv script.t A/script.t
ld.eld --sysroot=A A/script.t main.o 
ld.bfd --sysroot=A A/script.t main.o 
```

bfd is able to locate the lib1.so in `A/lib64/lib1.so` but eld fails to do so.

deepakshirkem:
## Problem

ELD fails to find namespec libraries (e.g. `GROUP(-lfoo)`) in standard directories under the sysroot such as `/lib`, `/lib64`, `/usr/lib`, and so on.

## Testing

Added `test/Common/standalone/SysrootNamespec` — verifies that `GROUP(-lfoo)` finds a library placed in the sysroot `lib64` directory. Test passes on all targets.

## Output

Screenshot from 2026-05-12 00-29-07

Fixes #819.


deepakshirkem:
Hi @parth-07 / @quic-areg / @quic-seaswara, please review this PR. During investigation I found two root causes `ELDEmulateELF` was missing standard sysroot dirs . When no `-m` flag is passed, `initEmulator` is deferred until the first ELF object is read. Since linker scripts are processed before objects, `ELDEmulateELF` never ran before namespec search happened.
 I referred to commit c960b2c ("Set emulation mode based on  input files") by @quic-seaswara and followed the same pattern of using  `isBackendInitialized()` + `getMachine()` + `initializeTarget()`. The  fix scans ahead in the input list for the first ELF object when a  script is encountered, initializes the target early, and then processes the script with sysroot dirs already set.

Please let me know if the approach is correct or if there's a better way  to handle this.

lib/Object/ObjectLinker.cpp
```diff
@@ -252,6 +252,33 @@ bool ObjectLinker::readLinkerScript(InputFile *Input) {
...
+    if (IF->getKind() != InputFile::ELFObjFileKind)
+      continue;
+    eld::Expected<uint16_t> Machine = getRelocObjParser()->getMachine(*IF);
+    if (!Machine)
+      continue;
+    llvm::cast<ObjectFile>(IF)->setMachine(*Machine);
+    initializeTarget(IF);
+    return;
+  }
+}
```
quic-seaswara, line 277: Why this change ?

lib/Target/ELFEmulation.cpp
```diff
@@ -24,11 +24,14 @@ bool eld::ELDEmulateELF(LinkerScript &pScript, LinkerConfig &pConfig) {
...
-    pConfig.directories().insert("=/usr/lib");
+  static const char *StandardDirs[] = {
+      "=/lib",       "=/lib64",         "=/usr/lib",
+      "=/usr/lib64", "=/usr/local/lib", "=/usr/local/lib64",
+  };
+  for (const char *Dir : StandardDirs) {
+    auto EldDir = ELDDirectory(Dir, Sysroot);
+    if (llvm::sys::fs::is_directory(EldDir.name()))
+      pConfig.directories().insert(Dir);
+  }
```
quic-seaswara, line 36: -nostdlib should not add these search directories I think.

test/Common/standalone/SysrootNamespec/SysrootNamespec.test
```diff
@@ -0,0 +1,13 @@
...
+#BEGIN_COMMENT
+# This test checks that namespec libraries (e.g. GROUP(-lfoo)) are found in
+# standard directories under the sysroot such as /lib64.
+#END_COMMENT
+#START_TEST
+RUN: rm -rf %t.sysroot
+RUN: mkdir -p %t.sysroot/lib64
+RUN: %clang %clangopts -o %t1.o %p/Inputs/1.c -c
+RUN: %clang %clangopts -o %t.lib1.o %p/Inputs/lib1.c -c
+RUN: %ar cr %aropts %t.sysroot/lib64/lib1.a %t.lib1.o
```
quic-seaswara: What about INPUT ?

quic-seaswara (changes requested):

test/Common/standalone/SysrootNamespec/SysrootNamespec.test
deepakshirkem: Yes its working also added test case.

lib/Target/ELFEmulation.cpp
deepakshirkem: Yes your right its working added test cases for this case also.

lib/Object/ObjectLinker.cpp
deepakshirkem: Hi @quic-seaswara, I am also not fully happy about adding this function, and my theory might be wrong. What I observed while reproducing #819 is that when we run the command l`d.eld --sysroot=A A/script.t main.o,` the code reads A/script.t before the target is initialized. While reading it, the process fails, which is why the code path you mentioned in the discussion is not getting triggered.

That is why I added this function. The function first tries to identify the target using the object file, similar to what I mentioned in the description.

I am not fully aware of the complete codebase yet, but I tried to find a place where we can get the target before reading any object files. It is possible that my intuition is wrong, and if we can determine the target earlier, then this function may not be required.

Your views @quic-areg, @partaror?

quic-seaswara: > Hi @quic-seaswara, I am also not fully happy about adding this function, and my theory might be wrong. What I observed while reproducing #819 is that when we run the command l`d.eld --sysroot=A A/script.t main.o,` the code reads A/script.t before the target is initialized. While reading it, the process fails, which is why the code path you mentioned in the discussion is not getting triggered.
> 
> That is why I added this function. The function first tries to identify the target using the object file, similar to what I mentioned in the description.
> 
> I am not fully aware of the complete codebase yet, but I tried to find a place where we can get the target before reading any object files. It is possible that my intuition is wrong, and if we can determine the target earlier, then this function may not be required.
> 
> Your views @quic-areg, @partaror?



> Hi @quic-seaswara, I am also not fully happy about adding this function, and my theory might be wrong. What I observed while reproducing #819 is that when we run the command l`d.eld --sysroot=A A/script.t main.o,` the code reads A/script.t before the target is initialized. While reading it, the process fails, which is why the code path you mentioned in the discussion is not getting triggered.
> 
> That is why I added this function. The function first tries to identify the target using the object file, similar to what I mentioned in the description.
> 
> I am not fully aware of the complete codebase yet, but I tried to find a place where we can get the target before reading any object files. It is possible that my intuition is wrong, and if we can determine the target earlier, then this function may not be required.
> 
> Your views @quic-areg, @partaror?

is the path expansion target specific ?

deepakshirkem: Yes. In the  `ELFEmulation.cpp`, the sysroot search paths (/lib, /lib64, /usr/lib, /usr/lib64) are the same for all targets in ELD

lib/Target/ELFEmulation.cpp
```diff
@@ -24,11 +24,14 @@ bool eld::ELDEmulateELF(LinkerScript &pScript, LinkerConfig &pConfig) {
...
                      : "";
-  auto LibPath = ELDDirectory("=/lib", Sysroot);
-  auto UsrLibPath = ELDDirectory("=/usr/lib", Sysroot);
-  if (llvm::sys::fs::is_directory(LibPath.name()))
-    pConfig.directories().insert("=/lib");
-  if (llvm::sys::fs::is_directory(UsrLibPath.name()))
-    pConfig.directories().insert("=/usr/lib");
+  static const char *StandardDirs[] = {
+      "=/lib",       "=/lib64",         "=/usr/lib",
+      "=/usr/lib64", "=/usr/local/lib", "=/usr/local/lib64",
```
parth-07: Should we search `/lib64` subdirectories on 32-bit architectures? Can you please check the behavior for different targets with GNU ld?

lib/Object/ObjectLinker.cpp
parth-07: I would like to improve and simplify this function. Additionally, it should be called before `readInputs` instead of in `readInputs`.

lib/Target/ELFEmulation.cpp
deepakshirkem: Done.

parth-07: Can you please update this comment thread with your findings of subdirectories searched for sysroot with GNU ld?

deepakshirkem: Verified GNU ld behavior by running `ld --verbose | grep SEARCH_DIR` on available cross-linkers.  Anyone can reproduce:
```
# x86_64
ld --verbose | grep SEARCH_DIR

# AArch64
aarch64-linux-gnu-ld --verbose | grep SEARCH_DIR
```
Output: 
Screenshot from 2026-06-08 22-55-27

Also confirmed from GNU binutils source `ld/emulparams/elf_x86_64.sh — LIBPATH_SUFFIX=64` is set explicitly only for `x86_64*-linux*`, not for other 64-bit targets like AArch64.

Finding: lib64 is not searched on 32-bit architectures. Updated fix uses is64Bits() so only 64-bit targets get lib64 dirs, 32-bit targets (ARM, Hexagon, RISCV32) do not.

lib/Object/ObjectLinker.cpp
parth-07: ping @deepakshirkem

lib/Target/ELFEmulation.cpp
parth-07: @deepakshirkem Your implementation is not matching the finding shared by you. GNU LD does not add */lib64 subdirectories to SEARCH_DIR whereas your PR is unconditionally adding */lib64 subdirectory to search paths. Additionally, GNU ld adds subdirectories such as `/usr/lib/aarch64-linux-gnu` and `/usr/lib/x86_64-linux-gnu` that is not getting added by this PR. Can you please look into which subdirectories should be added for which target in detail?

parth-07 (changes requested):

lib/Target/ELFEmulation.cpp
deepakshirkem: Hi @parth-07, Just wanted to confirm that I understood your feedback correctly.

* Move the `lib64` directory handling out of `ELDEmulateELF` and into the target-specific emulation files.
* Add target-specific multiarch library directories in each target's emulation file, following GNU ld behavior.

Does this direction look correct, or am I missing anything?

deepakshirkem:
>@deepakshirkem Your implementation is not matching the finding shared by you. GNU LD does not add */lib64 subdirectories to SEARCH_DIR whereas your PR is unconditionally adding */lib64 subdirectory to search paths. Additionally, GNU ld adds subdirectories such as /usr/lib/aarch64-linux-gnu and /usr/lib/x86_64-linux-gnu that is not getting added by this PR. Can you please look into which subdirectories should be added for which target in detail?



Hi @parth-07, updated the fix based on your feedback. Verified search directories using `ld --verbose | grep SEARCH_DIR` on available cross-linkers:

| Dir | x86_64 | AArch64 | RISCV64 | ARM32 | Hexagon |
|---|---|---|---|---|---|
| `=/lib` | ✅ | ✅ | ✅ | ✅ | N/A |
| `=/usr/lib` | ✅ | ✅ | ✅ | ✅ | N/A |
| `=/usr/local/lib` | ✅ | ✅ | ✅ | ✅ | N/A |
| `=/lib64` | ✅ | ❌ | ✅ | ❌ | N/A |
| `=/usr/lib64` | ✅ | ❌ | ✅ | ❌ | N/A |
| `=/usr/local/lib64` | ✅ | ❌ | ✅ | ❌ | N/A |
| `=/lib/<target>` | `x86_64-linux-gnu` | `aarch64-linux-gnu` | `riscv64-linux-gnu` | `arm-linux-gnueabihf` | N/A |
| `=/usr/lib/<target>` | `x86_64-linux-gnu` | `aarch64-linux-gnu` | `riscv64-linux-gnu` | `arm-linux-gnueabihf` | N/A |
| `=/usr/local/lib/<target>` | `x86_64-linux-gnu` | `aarch64-linux-gnu` | `riscv64-linux-gnu` | `arm-linux-gnueabihf` | N/A |

Common dirs (`=/lib`, `=/usr/lib`, `=/usr/local/lib`) stay in `ELDEmulateELF`. Target-specific dirs moved to each target's emulation file. Hexagon is Qualcomm-specific and not in GNU ld  left unchanged.

Please confirm if this direction is correct.

deepakshirkem:
Hi @quic-seaswara / @quic-areg, it would really help me if you could review this PR when you get a chance.

lib/Target/X86/x86_64Emulation.cpp
```diff
@@ -24,6 +25,29 @@ static bool ELDEmulatex86_64ELF(LinkerScript &pScript, LinkerConfig &pConfig) {
...
+      "=/lib64",
+      "=/usr/lib64",
+      "=/usr/x86_64-linux-gnu/lib64",
+      "=/usr/x86_64-linux-gnu/lib",
+  };
+  for (const char *Dir : X86_64Dirs) {
+    auto EldDir = ELDDirectory(Dir, Sysroot);
+    if (llvm::sys::fs::is_directory(EldDir.name()))
+      pConfig.directories().insert(Dir);
+  }
```
quic-seaswara, line 50: These paths are very GNU specific. Did you check how `lld` works with this ?

lib/Object/ObjectLinker.cpp
quic-seaswara: > Why this change ?

Please remove this change, and its increasing complexity. 

It would be helpful to figure out if  there is a way to fix when sniffing the object files.

quic-seaswara (changes requested):

lib/Object/ObjectLinker.cpp
deepakshirkem: Hi @quic-seaswara, Sorry, I think I ended up making this more complicated than it needs to be. I tried to find a way to fix this without adding `tryInitializeTargetFromInputs()`, but the main issue is that we don't have target information available until we reach an ELF object file.

> It would be helpful to figure out if there is a way to fix when sniffing the object files.

I also looked into fixing it within the existing sniffing code. However, from what I could see, that would require changing the current input processing order, which may affect existing input handling behavior.

lib/Target/X86/x86_64Emulation.cpp
deepakshirkem: Hi @quic-seaswara, checked lld source  it has no built-in default search paths at all, only -L options are used. So these paths are indeed GNU ld specific. Should we remove the multiarch dirs and only keep the basic standard dirs, or match GNU ld behavior as @parth-07 suggested?

PR #1165, opened 2026-05-11
Make output relocation section order deterministic

### LINKED ISSUE #1164: Build Failure: ELD EmitRelocs test fails due to strict FileCheck pattern.
We are experiencing a test failure in our build:

https://github.com/qualcomm/cpullvm-toolchain/actions/runs/25692635649/job/75432769978

The ELD test AArch64/standalone/EmitRelocs/EmitRelocs.test fails with a FileCheck error.

Correct me if I'm wrong but I think the test should be updated to allow the additional text.

quic-areg:
The order of output relocation sections depended on LLVM's DenseMap iteration order, which can shift across LLVM revisions and between assert/non-assert builds. Switch to a MapVector to make the order deterministic.

Fixes #1164.

### DIFF
--- a/lib/Object/ObjectLinker.cpp
+++ b/lib/Object/ObjectLinker.cpp
@@ -1981,7 +1981,7 @@ void ObjectLinker::createRelocationSections() {
   // Use a MapVector so that output relocation sections are created in a
   // deterministic order.
   llvm::MapVector<SectionKey, uint32_t,
-                  llvm::DenseMap<SectionKey, unsigned, SectionKeyInfo>>
+                  llvm::DenseMap<SectionKey, uint32_t, SectionKeyInfo>>
       OutputRelocCount;
   llvm::DenseMap<SectionKey, uint32_t, SectionKeyInfo> OutputRelocAlignment;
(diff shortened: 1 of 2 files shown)


lib/Object/ObjectLinker.cpp
```diff
@@ -1976,8 +1977,12 @@ void ObjectLinker::createRelocationSections() {
     }
   };
 
-  // apply all relocations of all inputs
-  llvm::DenseMap<SectionKey, uint32_t, SectionKeyInfo> OutputRelocCount;
+  // Apply all relocations of all inputs.
+  // Use a MapVector so that output relocation sections are created in a
+  // deterministic order.
+  llvm::MapVector<SectionKey, uint32_t,
+                  llvm::DenseMap<SectionKey, unsigned, SectionKeyInfo>>
```
parth-07:
```suggestion
                  llvm::DenseMap<SectionKey, uint32_t, SectionKeyInfo>>
```

Nitpick, while typically uint32_t and unsigned are the same, I am worried that it is possible that there can be exceptions to this as it is not mandated by the standard.

quic-seaswara (approved):

PR #1167, opened 2026-05-12
Add support for --no-warn-mismatch for ELF

### LINKED ISSUE #1214: Add support for --no-warn-mismatch for ELF information
Check for compatibility between object files using standard ELF header information.

Steven6798:
Allow users to suppress warnings about incompatible or mismatched input files using the --no-warn-mismatch option. This change disables mismatch diagnostics while preserving normal link behavior and outcomes.

Fix: #1214


docs/userguide/documentation/linker_faq.rst
```diff
@@ -1323,7 +1323,8 @@ If any object file differs, the linker cannot interpret it correctly and will
 produce this error.
 
 **How to verify:** You can inspect the object file using by using tools like
-llvm-readelf.
+llvm-readelf. You can enable/disable this message by using the
+*--[no-]warn-mismatch* flag
```
quic-areg:
```suggestion
*--[no-]warn-mismatch* flag.
```

lib/Readers/ELFReader.cpp
```diff
@@ -423,21 +423,29 @@ eld::Expected<bool> ELFReader<ELFT>::isCompatible() const {
   const InputFile &inputFile = this->m_InputFile;
   LinkerConfig &config = this->m_Module.getConfig();
 
-  if (!checkMachine())
-    return std::make_unique<plugin::DiagnosticEntry>(plugin::DiagnosticEntry(
-        Diag::err_unrecognized_input_file,
-        {inputFile.getInput()->getResolvedPath().native(),
-         config.targets().getArch()}));
+  if (!checkMachine()) {
```
parth-07: We do not need to call `checkMachine()` if `!config.options().noWarnMismatch()` is true.

lib/Readers/ELFReader.cpp
```diff
@@ -423,21 +423,29 @@ eld::Expected<bool> ELFReader<ELFT>::isCompatible() const {
...
   eld::Expected<bool> expCheckFlags = checkFlags();
   ELDEXP_RETURN_DIAGENTRY_IF_ERROR(expCheckFlags);
   if (!expCheckFlags.value())
     return false;
 
-  if (!checkClass())
-    return std::make_unique<plugin::DiagnosticEntry>(plugin::DiagnosticEntry(
-        Diag::invalid_elf_class,
-        {inputFile.getInput()->decoratedPath(), config.targets().getArch()}));
+  if (!checkClass()) {
```
parth-07: We do not need to call `checkClass()` if `!config.options().noWArnMismatch()` is true.

lib/Target/RISCV/RISCVInfo.cpp
```diff
@@ -48,32 +48,37 @@ bool RISCVInfo::isCompatible(uint64_t pFlag, const std::string &pFile) const {
   assert(m_OutputFlag);
   if (isABIFlagSet(pFlag, llvm::ELF::EF_RISCV_FLOAT_ABI_SOFT) ^
       isABIFlagSet(*m_OutputFlag, llvm::ELF::EF_RISCV_FLOAT_ABI_SOFT)) {
-    m_Config.raise(Diag::incompatible_architecture_versions)
-        << flagString(pFlag) << pFile << flagString(*m_OutputFlag);
+    if (!m_Config.options().noWarnMismatch())
+      m_Config.raise(Diag::incompatible_architecture_versions)
+          << flagString(pFlag) << pFile << flagString(*m_OutputFlag);
     return false;
```
parth-07, line 53: We are still returning false, just without the error message. Will the link fail without any error message if `--no-warn-mismatch` is specified? Why is a no-warn option hiding errors?

parth-07 (changes requested):

Steven6798:
This PR will address file mismatches using header information. Architecture specific changes will be handled in another issue/PR.

Steven6798:
While updating the tests I noticed that the dynamic version of the test crashes in the readDynamic function in the DynamicELFReader.cpp file. I think this is unrelated to the issue at hand so I will raise an issue about it.

Steven6798:
I updated the RISCV FloatABI and Emulation tests since they were expecting --warn-mismatch to be enabled.

quic-areg (commented):
For aarch64, arm, and riscv, the default behavior (with no flag at all) silently accepts any object: 
```
# before
ld.eld -m aarch64linux a.arm.o       # cannot recognize the format ... (exit 1)
ld.eld -m elf32lriscv  a.rv64.o          # Error: Invalid ELF file ...           (exit 1)

# after
ld.eld -m aarch64linux a.arm.o    # succeeds
ld.eld -m elf32lriscv  a.rv64.o     # succeeds (32-bit ELF output with rv64 contents merged)
```
This is really confusing imo and unsafe. Ideally, we would match GNU ld where `--no-warn-mismatch` is opt-in. This would remove the 3-state enum to a simple boolean and fix the aarch64/arm/riscv default regression.

GNU ld's no-warn-mismatch also differs from this patch in that it does not allow ELFs of different classes to be mixed, while we do. I think we should follow the GNU linker and not relax the `checkClass()` check. It is also probably unsafe: a 64-bit object in a 32-bit link gets parsed by the 32-bit parser.

Steven6798:
> For aarch64, arm, and riscv, the default behavior (with no flag at all) silently accepts any object:

I think it is best that the team meets up next week to discuss this since there are multiple opinions about how to address this issue. I'll schedule everything with Shankar. Thanks for the feedback.

Steven6798:
This PR only addresses --no-warn-mismatch and how it handled ELF information conflicts. It follows GNU's ld. Keep the feedback focus on this, as anything else will need to be handled in another issue/PR.

quic-seaswara (approved):

quic-areg:
I was mistaken on GNU ld not allowing different ELF classes to be mixed with its `--no-warn-mismatch`, please disregard that. But the other part of my comment still applies: 

```
# before
ld.eld -m aarch64linux a.arm.o       # cannot recognize the format ... (exit 1)
ld.eld -m elf32lriscv  a.rv64.o          # Error: Invalid ELF file ...           (exit 1)

# after
ld.eld -m aarch64linux a.arm.o    # succeeds
ld.eld -m elf32lriscv  a.rv64.o     # succeeds
```

`--no-warn-mismatch` should be opt-in and not default for this case.

Steven6798:
> `--no-warn-mismatch` should be opt-in and not default for this case.

Each architecture defines the default value differently if no flag is passed. This should be handled in another issue/PR. I can look into it as soon as I'm done with --no-warn-mismatch.

Steven6798:
I updated the test ton include dynamic linking. Is this ready to merge?

test/RISCV/standalone/32bit/EFlags/Map/FloatABI/FloatABI.test
```diff
@@ -5,7 +5,7 @@
 #END_COMMENT
 RUN: %yaml2obj %p/Inputs/float.yaml -o %t.o
 RUN: %link %linkopts %t.o -o %t1.out -M 2>&1 | %filecheck --check-prefix=FLOATABI %s
-RUN: %not %link -m elf64lriscv -march riscv64 %t.o -o %t1.out -M 2>&1 | %filecheck --check-prefix=ERROR %s
+RUN: %not %link -m elf64lriscv -march riscv64 %t.o -o %t1.out -M --warn-mismatch 2>&1 | %filecheck --check-prefix=ERROR %s
```
parth-07, line 8: I believe that the default behavior should be of `--warn-mismatch`. Both lld and bfd by default report error when files of different architectures are used. Is there any case where linking files of different architectures can produce a valid output?

parth-07 (changes requested):

test/RISCV/standalone/32bit/EFlags/Map/FloatABI/FloatABI.test
Steven6798: The default behavior will be addressed in another issue/PR. Each architecture currently has their own default behavior defined in initializeAttributes. This will change.

parth-07: Okay, I see. Thanks for the info.

parth-07 (approved):

PR #1176, opened 2026-05-13
[RISCV] Fix standalone test failures

svs-quic:
https://github.com/llvm/llvm-project/pull/193962 caused the output of the programs to change leading to function size differences.


parth-07 (approved):
Thank you for the fix.

PR #1178, opened 2026-05-13
[NFC] simplify emit relocs support

quic-seaswara:
Modularize and add helper functions to support emit relocs.

### DIFF
--- a/include/eld/Object/ObjectLinker.h
+++ b/include/eld/Object/ObjectLinker.h
@@ -17,12 +17,14 @@
 #include "eld/Support/MappingFile.h"
 #include "eld/Support/Path.h"
 #include "eld/Target/Relocator.h"
+#include "llvm/ADT/DenseMap.h"
 #include "llvm/ADT/StringRef.h"
 #include "llvm/ADT/StringSet.h"
 #include "llvm/Support/CodeGen.h"
 #include "llvm/Support/DataTypes.h"
 #include "llvm/Support/FileOutputBuffer.h"
 #include <memory>
+#include <unordered_map>
 #include <vector>
 
 namespace llvm {
@@ -186,7 +188,7 @@ public:
   bool prelayout();
 
   // Create relocation sections for emit relocs.
-  void createRelocationSections();
+  void createEmitRelocationSections();
 
   /// layout - linearly layout all output sections and reserve some space
   /// for GOT/PLT
@@ -506,6 +508,26 @@ private:
   // Initialize the target machine when sniffing object files
   bool initializeTarget(InputFile *I);
 
+  uint32_t getEmitRelocSectionType() const;
+
+  ELFSection *getEmitRelocOutputTargetSection(Relocation *Reloc) const;
+
+  void markRelocationOutputTargetSectionWanted(Relocation *Reloc) const;
+
+  ELFSection *createEmitRelocSection(ELFSection *OutputTargetSect,
+                                     uint32_t RelocSectionType,
+                                     uint32_t MaxAlignment,
+                                     uint32_t RelocCount);
+
+  ELFSection *findEmitRelocOutputSection(Relocation *Reloc,
+                                         uint32_t RelocSectionType) const;
+
+  void initializeEmitRelocCounts(
+      llvm::DenseMap<ELFSection *, unsigned> &RelocCount) const;
+
+  void finalizeEmitRelocSectionSizes(
+      const llvm::DenseMap<ELFSection *, unsigned> &RelocCount);
+
 private:
   LinkerConfig &ThisConfig;
   Module *ThisModule = nullptr;
@@ -523,6 +545,8 @@ private:
   BitcodeReader *MPBitcodeReader = nullptr;
   ObjectReader *symDefReader = nullptr;
   llvm::StringSet<> MDynlistExports;
+  // Maps each output section to its emit-relocs output relocation section.
+  llvm::DenseMap<ELFSection *, ELFSection *> EmitRelocSectionMap;
 
   // Is this the second phase of normalize for LTO
   bool MPostLtoPhase = false;
--- a/lib/Object/ObjectLinker.cpp
+++ b/lib/Object/ObjectLinker.cpp
@@ -1949,42 +1949,96 @@ size_t ObjectLinker::getRelocSectSize(uint32_t Type, uint32_t Count) {
                       : getTargetBackend().getRelEntrySize());
 }
 
-void ObjectLinker::createRelocationSections() {
+uint32_t ObjectLinker::getEmitRelocSectionType() const {
+  return getTargetBackend().getRelocator()->relocType();
+}
 
+ELFSection *
+ObjectLinker::getEmitRelocOutputTargetSection(Relocation *Reloc) const {
+  if (!Reloc || !Reloc->targetRef() || !Reloc->targetRef()->frag())
+    return nullptr;
+
+  ELFSection *OutputTargetSect = Reloc->targetRef()->frag()->getOwningSection();
+  if (OutputTargetSect && OutputTargetSect->getOutputSection())
+    OutputTargetSect = OutputTargetSect->getOutputSection()->getSection();
+  return OutputTargetSect;
+}
+
+// The target relocation section will be seen in the output only if not
+// discarded (linker script) or ignored (garbage collected)
+void ObjectLinker::markRelocationOutputTargetSectionWanted(
+    Relocation *Reloc) const {
+  ELFSection *TargetSection = Reloc->targetSection();
+  if (!TargetSection || TargetSection->isDiscard() || TargetSection->isIgnore())
+    return;
+
+  if (ELFSection *TargetOutputSection = TargetSection->getOutputELFSection())
+    TargetOutputSection->setWantedInOutput();
+}
+
+ELFSection *ObjectLinker::createEmitRelocSection(ELFSection *OutputTargetSect,
+                                                 uint32_t RelocSectionType,
+                                                 uint32_t MaxAlignment,
+                                                 uint32_t RelocCount) {
+  if (!OutputTargetSect)
+    return nullptr;
+
+  std::string RelocSectionName = getTargetBackend().getOutputRelocSectName(
+      OutputTargetSect->name().str(), RelocSectionType);
+  ELFSection *OutputRelocSect = ThisModule->createOutputSection(
+      RelocSectionName, LDFileFormat::Relocation, RelocSectionType /* Kind */,
+      0x0, MaxAlignment);
+  OutputRelocSect->setEntSize(RelocSectionType == llvm::ELF::SHT_RELA
+                                  ? getTargetBackend().getRelaEntrySize()
+                                  : getTargetBackend().getRelEntrySize());
+  OutputTargetSect->setWantedInOutput(true);
+  OutputRelocSect->setLink(OutputTargetSect);
+  OutputRelocSect->setSize(getRelocSectSize(RelocSectionType, RelocCount));
+  EmitRelocSectionMap[OutputTargetSect] = OutputRelocSect;
+  return OutputRelocSect;
+}
+
+ELFSection *
+ObjectLinker::findEmitRelocOutputSection(Relocation *Reloc,
+                                         uint32_t RelocSectionType) const {
+  ELFSection *TargetOutputSect = getEmitRelocOutputTargetSection(Reloc);
+  if (!TargetOutputSect)
+    return nullptr;
+
+  auto It = EmitRelocSectionMap.find(TargetOutputSect);
+  if (It != EmitRelocSectionMap.end())
+    return It->second;
+
+  return ThisModule->getSection(getTargetBackend().getOutputRelocSectName(
+      TargetOutputSect->name().str(), RelocSectionType));
+}
+
+void ObjectLinker::initializeEmitRelocCounts(
+    llvm::DenseMap<ELFSection *, unsigned> &RelocCount) const {
+  for (const auto &SectMap : EmitRelocSectionMap)
+    RelocCount[SectMap.second] = 0;
+}
+
+void ObjectLinker::finalizeEmitRelocSectionSizes(
+    const llvm::DenseMap<ELFSection *, unsigned> &RelocCount) {
+  for (const auto &Kv : RelocCount)
+    Kv.first->setSize(getRelocSectSize(Kv.first->getType(), Kv.second));
+}
+
+void ObjectLinker::createEmitRelocationSections() {
   if (!ThisConfig.options().emitRelocs())
     return;
 
-  // SectionName, contentPermissions, filepath
-  struct SectionKey {
-    SectionKey() : Name(""), Type(-1) {}
-
-    SectionKey(llvm::StringRef Name, int32_t Type) : Name(Name), Type(Type) {}
-
-    // Data members
-    StringRef Name;
-    int32_t Type;
-  };
-
-  struct SectionKeyInfo {
-    static SectionKey getEmptyKey() { return SectionKey(); }
-    static SectionKey getTombstoneKey() { return SectionKey(); }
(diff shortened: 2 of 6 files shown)


lib/Object/ObjectLinker.cpp
```diff
@@ -2379,8 +2408,11 @@ bool ObjectLinker::relocation(bool EmitRelocs) {
     // set relocation target symbol to the output section symbol's
     // resolveInfo
     R->setSymInfo(SymInfo);
-    OutputSect->addRelocation(R);
-    RelocCount[OutputSect] += 1;
+    {
+      std::lock_guard<std::mutex> Guard(EmitRelocsMutex);
```
quic-areg, line 2412: This is not `NFC`

include/eld/Object/ObjectLinker.h
```diff
@@ -506,6 +507,27 @@ class ObjectLinker {
   // Initialize the target machine when sniffing object files
   bool initializeTarget(InputFile *I);
 
+  // --emit-relocs support
```
quic-areg: I don't think this comment is useful

include/eld/Object/ObjectLinker.h
```diff
@@ -506,6 +507,27 @@ class ObjectLinker {
...
+
+  ELFSection *findEmitRelocOutputSection(Relocation *Reloc,
+                                         uint32_t RelocSectionType) const;
+
+  void initializeEmitRelocCounts(
+      std::unordered_map<ELFSection *, unsigned> &RelocCount) const;
+
+  void finalizeEmitRelocSectionSizes(
+      const std::unordered_map<ELFSection *, unsigned> &RelocCount);
+
```
quic-areg, line 530: The new helpers should be `private`

include/eld/Object/ObjectLinker.h
```diff
@@ -506,6 +507,27 @@ class ObjectLinker {
   // Initialize the target machine when sniffing object files
   bool initializeTarget(InputFile *I);
 
+  // --emit-relocs support
+  uint32_t getEmitRelocSectionType() const;
+
+  ELFSection *getEmitRelocOutputTargetSection(Relocation *Reloc) const;
+
+  void markRelocationTargetOutputSectionWanted(Relocation *Reloc) const;
```
quic-areg: `target` is used in opposite ways in these two helpers. Can we rename these functions?

include/eld/Object/ObjectLinker.h
```diff
@@ -506,6 +507,27 @@ class ObjectLinker {
...
+  ELFSection *createEmitRelocSection(ELFSection *OutputTargetSect,
+                                     uint32_t RelocSectionType,
+                                     uint32_t MaxAlignment,
+                                     uint32_t RelocCount);
+
+  ELFSection *findEmitRelocOutputSection(Relocation *Reloc,
+                                         uint32_t RelocSectionType) const;
+
+  void initializeEmitRelocCounts(
+      std::unordered_map<ELFSection *, unsigned> &RelocCount) const;
```
quic-areg: DenseMap?

lib/Object/ObjectLinker.cpp
```diff
@@ -1949,42 +1949,96 @@ size_t ObjectLinker::getRelocSectSize(uint32_t Type, uint32_t Count) {
                       : getTargetBackend().getRelEntrySize());
 }
 
-void ObjectLinker::createRelocationSections() {
+uint32_t ObjectLinker::getEmitRelocSectionType() const {
+  return getTargetBackend().getRelocator()->relocType();
```
parth-07, line 1953: `EmitRelocSectionType` is the same as the relocation type. Can we use the underlying function `getReloctor()::relocType()` instead of adding a new function? If a helper function is beneficial here, then can we please rename it as `getRelocSectionType` so that it is clear that there is no different type for emit relocation sections.

lib/Object/ObjectLinker.cpp
```diff
@@ -1949,42 +1949,96 @@ size_t ObjectLinker::getRelocSectSize(uint32_t Type, uint32_t Count) {
...
+}
+
+void ObjectLinker::finalizeEmitRelocSectionSizes(
+    const std::unordered_map<ELFSection *, unsigned> &RelocCount) {
+  for (const auto &Kv : RelocCount)
+    Kv.first->setSize(getRelocSectSize(Kv.first->getType(), Kv.second));
+}
+
+void ObjectLinker::createRelocationSections() {
+  EmitRelocSectionMap.clear();
```
parth-07: Why do we need to clear it here?

lib/Object/ObjectLinker.cpp
```diff
@@ -1949,42 +1949,96 @@ size_t ObjectLinker::getRelocSectSize(uint32_t Type, uint32_t Count) {
...
+    RelocCount[SectMap.second] = 0;
+}
+
+void ObjectLinker::finalizeEmitRelocSectionSizes(
+    const std::unordered_map<ELFSection *, unsigned> &RelocCount) {
+  for (const auto &Kv : RelocCount)
+    Kv.first->setSize(getRelocSectSize(Kv.first->getType(), Kv.second));
+}
+
+void ObjectLinker::createRelocationSections() {
```
parth-07: Can we please rename this function to `createEmitRelocationSections` so that it is consistent with the other new functions (such as createEmitRelocSection and finalizeEmitRelocSectionSizes) and it is clear that it is only useful for emit relocation functionality.

lib/Object/ObjectLinker.cpp
```diff
@@ -1949,42 +1949,96 @@ size_t ObjectLinker::getRelocSectSize(uint32_t Type, uint32_t Count) {
...
+  ELFSection *OutputTargetSect = Reloc->targetRef()->frag()->getOwningSection();
+  if (OutputTargetSect && OutputTargetSect->getOutputSection())
+    OutputTargetSect = OutputTargetSect->getOutputSection()->getSection();
+  return OutputTargetSect;
+}
+
+void ObjectLinker::markRelocationTargetOutputSectionWanted(
+    Relocation *Reloc) const {
+  ELFSection *TargetSection = Reloc->targetSection();
+  if (!TargetSection || TargetSection->isDiscard() || TargetSection->isIgnore())
```
parth-07, line 1972: Do we need all these checks here?

lib/Object/ObjectLinker.cpp
```diff
@@ -1949,42 +1949,96 @@ size_t ObjectLinker::getRelocSectSize(uint32_t Type, uint32_t Count) {
...
+  if (!Reloc || !Reloc->targetRef() || !Reloc->targetRef()->frag())
+    return nullptr;
+
+  ELFSection *OutputTargetSect = Reloc->targetRef()->frag()->getOwningSection();
+  if (OutputTargetSect && OutputTargetSect->getOutputSection())
+    OutputTargetSect = OutputTargetSect->getOutputSection()->getSection();
+  return OutputTargetSect;
+}
+
+void ObjectLinker::markRelocationTargetOutputSectionWanted(
```
parth-07: This API is only conditionally marking the relocation target output section as wanted. The name should indicate that it has more checks/conditions within it.

lib/Object/ObjectLinker.cpp
```diff
@@ -1949,42 +1949,96 @@ size_t ObjectLinker::getRelocSectSize(uint32_t Type, uint32_t Count) {
...
+ELFSection *ObjectLinker::createEmitRelocSection(ELFSection *OutputTargetSect,
+                                                 uint32_t RelocSectionType,
+                                                 uint32_t MaxAlignment,
+                                                 uint32_t RelocCount) {
+  if (!OutputTargetSect)
+    return nullptr;
+
+  std::string RelocSectionName = getTargetBackend().getOutputRelocSectName(
+      OutputTargetSect->name().str(), RelocSectionType);
+  ELFSection *OutputRelocSect = ThisModule->createOutputSection(
```
parth-07, line 1988: Would the emit relocation output sections take part in rule-matching?

lib/Object/ObjectLinker.cpp
parth-07, line 2412: Why didn't we need mutex here until now? I am slightly concerned that this mutex might cause some performance issues.

lib/Object/ObjectLinker.cpp
quic-seaswara: I kept this for use cases to support relocations such as R_HEX_COPY and relocations that get translated on RISC-V due to relaxation. We would need this function.

lib/Object/ObjectLinker.cpp
quic-seaswara: Sure. We can remove it.

lib/Object/ObjectLinker.cpp
quic-seaswara: Sure. sounds good.

lib/Object/ObjectLinker.cpp
quic-seaswara: Yes, to handle garbage collection. We have tests that cover these.

lib/Object/ObjectLinker.cpp
quic-seaswara: I can add a comment instead of changing the name. Is that ok ?

lib/Object/ObjectLinker.cpp
quic-seaswara: No. We currently do not, and i think its a overkill.

lib/Object/ObjectLinker.cpp
quic-seaswara: ApplyRelocs is multithreaded. I agree there is a performance consideration to be handled. We cannot avoid a lock here with the current design.

include/eld/Object/ObjectLinker.h
quic-seaswara: Why DenseMap ? Any strong reason ?

include/eld/Object/ObjectLinker.h
quic-seaswara: They are private.

include/eld/Object/ObjectLinker.h
quic-seaswara: Sure, we can remove.

lib/Object/ObjectLinker.cpp
quic-seaswara: I will remove NFC from the commit message. thanks for pointing this!

include/eld/Object/ObjectLinker.h
quic-areg: > When both C++ and the LLVM support libraries provide similar functionality, and there isn’t a specific reason to favor the C++ implementation, it is generally preferable to use the LLVM library. For example, llvm::DenseMap should almost always be used instead of std::map or std::unordered_map, and llvm::SmallVector should usually be used instead of std::vector.

from https://llvm.org/docs/CodingStandards.html

include/eld/Object/ObjectLinker.h
quic-seaswara: sure

test/Common/standalone/EmitRelocsAggregation/EmitRelocsAggregation.test
```diff
@@ -0,0 +1,27 @@
...
+RUN: %link %linkopts %t.test.o %t.bar.o -T %p/Inputs/script.t --emit-relocs -o %t.out
+
+RUN: %readelf -r %t.out | %filecheck %s --check-prefix=RELOCS
+RUN: %readelf -W -S %t.out | %filecheck %s --check-prefix=SECTIONS
+
+RELOCS: Relocation section '.rel{{a?}}CODE' at offset
+RELOCS-NOT: Relocation section '.rel{{a?}}\.text'
+RELOCS-NOT: Relocation section '.rel{{a?}}mycode_1'
+RELOCS-NOT: Relocation section '.rel{{a?}}mycode_2'
+RELOCS-NOT: Relocation section '.rel{{a?}}CODE'
```
quic-areg, line 19: I don't think the CHECK-NOTs here work as intended. For example, if `.rel.text` appeared before `.relCODE`, the `NOT` would never catch it. 

I think it should look like this:

```
RELOCS-NOT: Relocation section '.rel{{a?}}\.text'                                                                                                                                                                                     
RELOCS-NOT: Relocation section '.rel{{a?}}mycode_1'
RELOCS-NOT: Relocation section '.rel{{a?}}mycode_2'                                                                                                                                                                                   
RELOCS:     Relocation section '.rel{{a?}}CODE' at offset       
RELOCS-NOT: Relocation section '.rel{{a?}}CODE'    
```
Same for the block below.

quic-seaswara: The way we traverse creating relocation sections is by how the compiler orders it. So mycode_1 and mycode_2 are ordered before.   There should be no '.rel{{a?}}\.text'.

lib/Object/ObjectLinker.cpp
quic-seaswara: Done

lib/Object/ObjectLinker.cpp
quic-seaswara: Added comments in the function

include/eld/Object/ObjectLinker.h
quic-seaswara: Sure done

test/Common/standalone/EmitRelocsAggregation/EmitRelocsAggregation.test
quic-areg: Right. My point is about FileCheck semantics: a `CHECK-NOT` only applies in between positive `CHECK`s. If there were a regression that introduced one of these sections that the test does not expect, the CHECK-NOTs will not catch them.

If there were a regression like the following where `.relamycode_2` becomes its own output section, the current test as written would not catch it, even though that is likely the intent:

```
Relocation section '.relamycode_2' at offset 0x1080 contains 1 entries:                                                                                                                                                               
0000000000000000  0000000400000101  R_AARCH64_ABS64        0000000000000000 z + 0                                                                                                                                                     
                                                                                                                                                                                                                                      
Relocation section '.relaCODE' at offset 0x10a0 contains 4 entries:                                                                                                                                                                   
0000000000000000  0000000200000113  R_AARCH64_CALL26       0000000000000000 bar + 0                                                                                                                                                   
000000000000000c  0000000300000113  R_AARCH64_CALL26       0000000000000000 bar2 + 0                                                                                                                                                  
0000000000000010  0000000500000101  R_AARCH64_ABS64        0000000000000000 x + 0                                                                                                                                                     
000000000000001c  0000000400000101  R_AARCH64_ABS64        0000000000000000 y + 0              
```

quic-seaswara: mycode_2 will never be a seperate output section because the test is driven by a linker script. 

`CODE : { *(.text mycode*) }`

quic-areg:
```
# This checks that --emit-relocs creates a single relocation section for the
# linker-script output text section and does not create per-input text
# relocation sections
```

The point of the test as described is to check that `mycode_2` and other input sections never become separate output sections. If such a regression were to happen, this test won't catch it even though it is intended to. If that check is superfluous like you described, then what is the point of the test?

quic-seaswara: We are just checking that the linker is not creating these relocation sections for some reason. Its ok to have this check imo.

quic-areg: I agree. My point is that if we do have this check then it should work. It currently does not work.

quic-seaswara: Sure.

quic-areg (approved):

PR #1182, opened 2026-05-14
ci: prune ccache entries older than 2h in workflows

quic-seaswara:
Add ccache eviction with --evict-older-than 7200s in workflows that persist cache and in CI after build.

The CI run ccache has increased close to 50G.


quic-seaswara:
After eviction 

  Cache size (GiB):  0.4
  Hits:             4800 / 4844 (99.09%)

quic-shaksuma (commented):
How did we arrive at 7200s?

PR #1185, opened 2026-05-15
Command line options improvements

### LINKED ISSUE #1162: Command line options improvements
## Background

The POSIX/GNU convention for command-line option prefixes is:
- **Single dash (`-`)** — reserved for single-letter options (e.g. `-r`, `-e`, `-l`)
- **Double dash (`--`)** — required for word/multi-character options (e.g. `--shared`, `--sysroot`, `--gc-sections`)

GNU `ld` and LLVM `lld` both follow this convention. `lld`'s `Options.td` even documents it explicitly with the `FF<>` / `JJ<>` helper classes (double-dash only) vs `F<>` / `J<>` (both dashes, used only for legacy compatibility with a small set of options).

---

## Problem

`eld`'s `GnuLinkerOptions.td` currently accepts a single dash for many multi-character (word) options, in three distinct ways:

### 1. `Flag<["-"], "word">` — single-dash-only for word options

These options are defined with only a single-dash prefix, which is incorrect for word-length names:

| Line | Option |
|------|--------|
| 297 | `-static` |
| 300 | `-dynamic` |
| 303 | `-shared` |
| 317 | `-Bstatic` |
| 325 | `-non_shared` |
| 329 | `-Bshareable` |
| 332 | `-Bdynamic` |
| 339 | `-call_shared` |
| 394 | `-nostdlib` |
| 503 | `-Bsymbolic` |
| 507 | `-Bsymbolic-functions` |
| 512 | `-Bgroup` |
| 516 | `-fPIC` |

### 2. `Flag<["-", "--"], "word">` — both dashes accepted for word options

These options accept both `-word` and `--word`. The single-dash form should be removed (or kept only as a deprecated alias where backward compatibility is strictly required):

| Line | Option |
|------|--------|
| 270 | `unique-output-sections` |
| 306 | `pie` |
| 309 | `no-pie` |
| 387 | `whole-archive` |
| 391 | `no-whole-archive` |
| 397 | `emit-relocs` |
| 437 | `export-dynamic` |
| 441 | `force-dynamic` |
| 456 | `no-dynamic-linker` |
| 534 | `start-group` |
| 547 | `end-group` |
| 574 | `allow-shlib-undefined` |
| 591 | `no-undefined` |
| 740 | `cref` |
| 750 | `emit-timing-stats-in-output` |
| 760 | `print-timing-stats` |
| 799 | `opt-record-file` |
| 855 | `gc-sections` |
| 858 | `no-gc-sections` |
| 861 | `print-gc-sections` |
| 1076 | `thin-archive-rule-matching-compatibility` |
| 1089 | `flto` |
| 1097 | `flto-use-as` |
| 1145 | `save-temps` |
| 1289 | `threads` |
| 1297 | `no-threads` |
| 1311 | `symdef` |
| 1353 | `allow-incompatible-section-mix` |

### 3. `smDash*` multiclasses — single-dash aliases generated for word options

The `smDash`, `smDashWithOpt`, `smDashTwoWithOpt`, and `smDashOnlyEq` multiclasses all generate single-dash variants alongside double-dash variants for word options. This means options like `--sysroot`, `--rpath`, `--defsym`, `--wrap`, `--hash-style`, `--dynamic-linker`, etc. are
silently accepted as `-sysroot`, `-rpath`, `-defsym`, etc.

Affected `defm` sites include:

| Line | Option |
|------|--------|
| 138 | `sysroot` |
| 381 | `init` |
| 384 | `fini` |
| 424 | `dynamic-linker` |
| 428 | `rpath` |
| 491 | `hash-style` |
| 585 | `defsym` |
| 653 | `wrap` |
| 656 | `portable` |
| 718 | `trace-section` |
| 722 | `trace-reloc` |
| 732 | `verify-options` |
| 796 | `color` |
| 959 | `reproduce` |
| 964 | `reproduce-compressed` |
| 970 | `reproduce-on-fail` |
| 976 | `mapping-file` |
| 980 | `dump-mapping-file` |

---

## Expected Behavior

- Word options should only be accepted with `--` (double dash).
- Single-letter options (`-r`, `-e`, `-o`, `-l`, `-L`, `-T`, `-m`, etc.) should continue to use `-` (single dash).
- A small set of options that GNU `ld` itself accepts with a single dash for historical reasons (e.g. `-shared`, `-static`, `-Bsymbolic`, `-rpath`) may be retained as **explicit deprecated aliases** using `Alias<>`, clearly documented as compatibility-only, rather than being the
primary definition.

---

## Suggested Fix

1. Change `Flag<["-"], "word">` definitions for word options to `Flag<["--"], "word">`, adding a single-dash `Alias<>` only where GNU `ld` compatibility strictly requires it.
2. Change `Flag<["-", "--"], "word">` to `Flag<["--"], "word">` throughout, with explicit single-dash aliases only where needed.
3. Refactor the `smDash*` multiclasses to stop generating single-dash variants by default. Introduce a separate `smDashCompat*` variant (or similar) for the handful of options that genuinely need both forms for backward compatibility.
4. Use the existing `FF<>` / `JJ<>` helper classes (double-dash only) for new and eld-specific options, consistent with the comment already present in the file at lines 125–128.

---

## Affected Files

- `include/eld/Driver/GnuLinkerOptions.td`
- `include/eld/Driver/ARMLinkerOptions.td`
- `include/eld/Driver/HexagonLinkerOptions.td`

Nuance worth preserving in the issue: some single-dash forms (e.g. -shared, -static, -rpath, -Bsymbolic) are genuinely accepted by GNU ld for historical reasons, so they shouldn't be removed outright — just demoted to explicit compatibility aliases. The issue calls this out clearly
so the fix doesn't blindly break existing users.

Steven6798:
Update help messages to reflect canonical use of --word options. Single-dash forms are kept for compatibility but are now labeled as either deprecated (eld-specific) or compatibility aliases (GNU-compatible), guiding users to migrate without breaking existing workflows.

Fix: #1162

### DIFF
--- a/include/eld/Driver/GnuLinkerOptions.td
+++ b/include/eld/Driver/GnuLinkerOptions.td
@@ -6,6 +6,20 @@
 
 include "llvm/Option/OptParser.td"
 
+// The definitions below are copied from lld.
+
+// For options whose names are multiple letters, either one dash or
+// two can precede the option name except those that start with 'o'.
+class F<string name> : Flag<["--", "-"], name>;
+class J<string name> : Joined<["--", "-"], name>;
+
+// Convenience classes for long options which only accept two dashes. For lld
+// specific or newer long options, we prefer two dashes to avoid collision with
+// short options. For many others, we have to accept both forms to be compatible
+// with GNU ld.
+class FF<string name> : Flag<["--"], name>;
+class JJ<string name> : Joined<["--"], name>;
+
 //===----------------------------------------------------------------------===//
 /// Utility Functions
 //===----------------------------------------------------------------------===//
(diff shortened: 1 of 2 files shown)


Steven6798:
After these changes are merged, I’ll create a follow-up issue to migrate any deprecated flags used within the project.

include/eld/Driver/GnuLinkerOptions.td
```diff
@@ -353,66 +509,66 @@ def l : JoinedOrSeparate<["-"], "l">,
...
       HelpText<"This option negates most of the effects of the -N option."
                "Disable linking with shared libraries">,
       Group<grp_main>;
 def omagic_alias
     : Flag<["-"], "N">,
-      HelpText<"Set the text and data sections to be readable and writable."
-               " Also, do not page-align the data segment, and"
-               " disable linking against shared libraries.">,
+      HelpText<"Set the text and data sections to be readable and writable. "
+               "Also, do not page-align the data segment, and"
```
quic-areg: I think this needs the trailing space or the help text will read "…the data segment, *and**disable*** linking against shared libraries."

Steven6798: Thanks!

include/eld/Driver/GnuLinkerOptions.td
```diff
@@ -16,7 +30,174 @@ multiclass smDash<string opt1, string opt2, string help> {
   def opt1_eq : Joined<["-"], opt1 #"=">, Alias<!cast<Option>(opt1)>;
   // Compatibility aliases
   def opt2_dashdash : Separate<["--"], opt2>, Alias<!cast<Option>(opt1)>;
-  def opt2_dashdash_eq : Joined<["--"], opt2 #"=">, Alias<!cast<Option>(opt1)>;
+  def opt2_dashdash_eq : JJ<opt2 #"=">, Alias<!cast<Option>(opt1)>;
```
parth-07, line 33: For these new tablegen definitions, we may need to update the script docs/userguide/GenerateOptionsDocsFromTblGen.py that is used to automatically generate the command-line options documentation. Can you please make sure that the script works correctly with this change?

parth-07 (changes requested):

include/eld/Driver/GnuLinkerOptions.td
Steven6798: Of course. Thanks for the heads up.

Steven6798: It looks like that documentation was out-of-date even before my changes. I'll update it. Nice catch!

Steven6798: > It looks like that documentation was out-of-date even before my changes. I'll update it. Nice catch!

This is wrong. I assumed that all option aliases had to be printed but that is not the case. Everything works fine.

quic-shaksuma (approved):

Steven6798:
Is this ready to merge?

Steven6798:
There are issues with the description of the "output" and "fpic" flags. I'll fix that.

parth-07 (approved):

PR #1187, opened 2026-05-17
[ELD] Fix --wrap archive extraction for __real_<sym> references

### LINKED ISSUE #936: `--wrap=<sym>`: archive member not pulled when wrapper calls `__real_<sym>`
There is an issue with the `--wrap` flag where it does not pull the required symbol which is to be wrapped from an archive.

Reproductible example:

```bash
#!/usr/bin/env bash

mkdir -p build

cat > add.c <<'EOF'
int add(int a, int b) {
    return a + b;
}
EOF

cat > wrap.c <<'EOF'
#include <stdio.h>
extern int __real_add(int a, int b);
int __wrap_add(int a, int b) {
    int result = __real_add(a, b);
    printf("[WRAP add] real=%d, returns %d\n", result, result + result);
    return result;
}
EOF

cat > main.c <<'EOF'
#include <stdio.h>
extern int add(int a, int b);
int main() {
    int x = 40, y = 2;
    int sum = add(x, y);
    printf("add(%d, %d) = %d\n", x, y, sum);
    return 0;
}
EOF

clang -c add.c         -o build/add.o
clang -c wrap.c        -o build/wrap.o
clang -c main.c        -o build/main.o
llvm-ar rcs build/libfoo.a build/add.o

# set $ELD to test ELD behavior
clang -o build/demo_eld build/main.o build/wrap.o -fuse-ld=$ELD -Lbuild -lfoo -Wl,--wrap=add
```

Output:

```
Error: build/wrap.o(.text+0x15): undefined reference to `add'
Fatal: Linking had errors.
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

vijayvarghese:
## Problem
Fixes #936

When `--wrap=<sym>` is active and an object file references `__real_<sym>`,
the archive containing `<sym>` was not being extracted, producing:

    Error: wrap.o(.text+0x15): undefined reference to `add'

## Root Cause
`IRBuilder::addSymbol` renames `__real_<sym>` to `<sym>` in the namepool
before archive scanning begins. The existing check in `shouldIncludeSymbol`
only looked for `__real_<sym>` in the namepool — which no longer exists
under that name — so the archive member was never pulled.

## Fix
In `shouldIncludeSymbol`, added `|| info->isUndef()` as a second condition,
covering the case where the `__real_` reference was renamed before archive
scanning.

## Test
Added `test/Common/standalone/WrapSymbolArchivePull/` reproducing the exact
scenario from the issue: `main.o` and `wrap.o` as direct inputs, `add.o`
inside a static archive, linked with `--wrap=add`.

### DIFF
--- a/lib/Readers/ArchiveParser.cpp
+++ b/lib/Readers/ArchiveParser.cpp
@@ -494,10 +494,15 @@ ArchiveParser::shouldIncludeSymbol(const ArchiveFile &archive,
     return ArchiveFile::Symbol::Include;
 
   // If there is a wrap option specified for that symbol, we only pull from
-  // an archive, if the real symbol is still undefined.
+  // an archive, if the real symbol is still undefined. There are two ways this can manifest:
+  //   1. __real_<sym> exists directly in the name pool (uncommon).
+  //   2. __real_<sym> was renamed to <sym> by IRBuilder::addSymbol before
+  //      archive scanning began, so it never has its own name pool entry.
+  //      In that case <sym> itself is the renamed reference and info->isUndef()
+  //      captures it.
   std::string realSymbol = ("__real_" + SymName).str();
   ResolveInfo *RealSymbol = m_Module.getNamePool().findInfo(realSymbol);
-  if (RealSymbol && RealSymbol->isUndef())
+  if ((RealSymbol && RealSymbol->isUndef()) || info->isUndef())
     return ArchiveFile::Symbol::Include;
 
   return ArchiveFile::Symbol::Exclude;
--- a/lib/Readers/ArchiveParser.cpp
+++ b/lib/Readers/ArchiveParser.cpp
@@ -500,10 +500,19 @@ ArchiveParser::shouldIncludeSymbol(const ArchiveFile &archive,
   //      archive scanning began, so it never has its own name pool entry.
   //      In that case <sym> itself is the renamed reference and info->isUndef()
   //      captures it.
+  // Emit a trace message when pulling archive members for wrapped symbols to aid
+  // in debugging the wrap chain and archive resolution process.
+  bool TraceWrap = m_Module.getPrinter()->traceWrapSymbols();
   std::string realSymbol = ("__real_" + SymName).str();
   ResolveInfo *RealSymbol = m_Module.getNamePool().findInfo(realSymbol);
-  if ((RealSymbol && RealSymbol->isUndef()) || info->isUndef())
+  if ((RealSymbol && RealSymbol->isUndef()) || info->isUndef()){
+    if (TraceWrap){
+        m_Module.getConfig().raise(Diag::trace_wrap_archive_pull)
+        << archive.getInput()->decoratedPath()    // archive filename
+        << SymName;           // wrapped symbol name  
+    }
     return ArchiveFile::Symbol::Include;
+  }
 
   return ArchiveFile::Symbol::Exclude;
 }
--- a/lib/Readers/ArchiveParser.cpp
+++ b/lib/Readers/ArchiveParser.cpp
@@ -494,26 +494,34 @@ ArchiveParser::shouldIncludeSymbol(const ArchiveFile &archive,
     return ArchiveFile::Symbol::Include;
 
   // If there is a wrap option specified for that symbol, we only pull from
-  // an archive, if the real symbol is still undefined. There are two ways this can manifest:
-  //   1. __real_<sym> exists directly in the name pool (uncommon).
-  //   2. __real_<sym> was renamed to <sym> by IRBuilder::addSymbol before
-  //      archive scanning began, so it never has its own name pool entry.
-  //      In that case <sym> itself is the renamed reference and info->isUndef()
-  //      captures it.
-  // Emit a trace message when pulling archive members for wrapped symbols to aid
-  // in debugging the wrap chain and archive resolution process.
-  bool TraceWrap = m_Module.getPrinter()->traceWrapSymbols();
+  // an archive if the wrap chain actually requires the symbol.
+
+  // Restrict the fallback archive extraction path to non-LTO builds.
+  bool isElfPath = !(m_Module.getConfig().options().hasLTO() ||
+                     m_Module.needLTOToBeInvoked() || isPostLTOPhase);
+
   std::string realSymbol = ("__real_" + SymName).str();
   ResolveInfo *RealSymbol = m_Module.getNamePool().findInfo(realSymbol);
-  if ((RealSymbol && RealSymbol->isUndef()) || info->isUndef()){
-    if (TraceWrap){
-        m_Module.getConfig().raise(Diag::trace_wrap_archive_pull)
-        << archive.getInput()->decoratedPath()    // archive filename
-        << SymName;           // wrapped symbol name  
+  // Emit a trace message when pulling archive members for wrapped symbols to
+  // aid in debugging the wrap chain and archive resolution process.
+  bool TraceWrap = m_Module.getPrinter()->traceWrapSymbols();
+  if (RealSymbol && RealSymbol->isUndef()) {
+    if (TraceWrap) {
+      m_Module.getConfig().raise(Diag::trace_wrap_archive_pull)
+          << archive.getInput()->decoratedPath() // archive filename
+          << SymName;                            // wrapped symbol name
     }
     return ArchiveFile::Symbol::Include;
   }
-
+  // Non-LTO ELF path:
+  // Allow unresolved wrapped symbols to pull archive members.
+  if (isElfPath) {
+    if (TraceWrap) {
+      m_Module.getConfig().raise(Diag::trace_wrap_archive_pull)
+          << archive.getInput()->decoratedPath() << SymName;
+    }
+    return ArchiveFile::Symbol::Include;
+  }
   return ArchiveFile::Symbol::Exclude;
 }
 
--- a/test/Common/standalone/WrapSymbolArchivePull/WrapSymbolArchivePull.test
+++ b/test/Common/standalone/WrapSymbolArchivePull/WrapSymbolArchivePull.test
@@ -1,16 +1,8 @@
 
 #---WrapSymbolArchivePull.test------------------------ Executable ----------#
 #BEGIN_COMMENT
-# --wrap=<sym>: archive member not pulled when wrapper calls __real_<sym>
-#
-# When --wrap=add is used and wrap.o has an undefined reference to __real_add,
-# IRBuilder renames that reference to 'add' in the name pool.  Later, when the
-# archive is scanned, shouldIncludeSymbol() must still recognise the undefined
-# 'add' entry as a demand for the archive member that defines 'add'.
-#
-# Before the fix, shouldIncludeSymbol() looked only for __real_add in the name
-# pool (which no longer existed under that name after the rename) and therefore
-# returned Exclude, causing an "undefined reference to `add`" error.
+# Test that an archive member providing the real symbol is pulled when a
+# wrapper calls __real_<sym> (verifies archive extraction for wrapped symbols).
 #END_COMMENT
 #START_TEST
 
--- a/test/Common/standalone/WrapSymbolArchivePull/WrapSymbolArchivePull.test
+++ b/test/Common/standalone/WrapSymbolArchivePull/WrapSymbolArchivePull.test
@@ -1,7 +1,6 @@
 
 #---WrapSymbolArchivePull.test------------------------ Executable ----------#
 #BEGIN_COMMENT
-# Regression test for GitHub issue #936:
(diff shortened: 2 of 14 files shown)


lib/Readers/ArchiveParser.cpp
```diff
@@ -494,10 +494,15 @@ ArchiveParser::shouldIncludeSymbol(const ArchiveFile &archive,
...
+  // an archive, if the real symbol is still undefined. There are two ways this can manifest:
+  //   1. __real_<sym> exists directly in the name pool (uncommon).
+  //   2. __real_<sym> was renamed to <sym> by IRBuilder::addSymbol before
+  //      archive scanning began, so it never has its own name pool entry.
+  //      In that case <sym> itself is the renamed reference and info->isUndef()
+  //      captures it.
   std::string realSymbol = ("__real_" + SymName).str();
   ResolveInfo *RealSymbol = m_Module.getNamePool().findInfo(realSymbol);
-  if (RealSymbol && RealSymbol->isUndef())
+  if ((RealSymbol && RealSymbol->isUndef()) || info->isUndef())
```
quic-seaswara: can we update tracing for wrap symbols with --trace=wrap-symbols to show this information ?

quic-seaswara: Please update the docs too.

test/Common/standalone/WrapSymbolArchivePull/WrapSymbolArchivePull.test
```diff
@@ -0,0 +1,34 @@
...
+#
+# Before the fix, shouldIncludeSymbol() looked only for __real_add in the name
+# pool (which no longer existed under that name after the rename) and therefore
+# returned Exclude, causing an "undefined reference to `add`" error.
+#END_COMMENT
+#START_TEST
+
+# Compile inputs.
+RUN: %clang %clangopts -c %p/Inputs/1.c -o %t1.main.o
+RUN: %clang %clangopts -c %p/Inputs/2.c -o %t1.wrap.o
```
quic-seaswara, line 11: check that using LTO also makes sure the behavior is consistent.

test/Common/standalone/WrapSymbolArchivePull/WrapSymbolArchivePull.test
```diff
@@ -0,0 +1,34 @@
+
+#---WrapSymbolArchivePull.test------------------------ Executable ----------#
+#BEGIN_COMMENT
+# Regression test for GitHub issue #936:
```
quic-seaswara: Remove this line referencing github issue

lib/Readers/ArchiveParser.cpp
quic-seaswara: Can we make sure that __wrap_<symbol> is also handled accordingly ?

quic-seaswara: Can we also check if __real or __wrap if not defined and what the error is ?

quic-seaswara (changes requested):

lib/Readers/ArchiveParser.cpp
vijayvarghese: Added --trace=wrap-symbols tracing in ArchiveParser::shouldIncludeSymbol() to log archive member extraction triggered during wrapped symbol resolution. The trace now reports the wrapped symbol along with, from which archive it being pulled.

vijayvarghese: Added additional documentation for --wrap usage covering archive symbol extraction behavior, how \_\_real\_\<sym\> references can trigger archive member inclusion, and how to use --trace=wrap-symbols for debugging these cases.

test/Common/standalone/WrapSymbolArchivePull/WrapSymbolArchivePull.test
vijayvarghese: Added coverage for this in the latest update.

lib/Readers/ArchiveParser.cpp
vijayvarghese: Added coverage for this.
Verified \_\_wrap\_\<sym\> handling with an additional test where \_\_wrap\_\<sym\> is defined inside an archive member. In this case it is included through the existing archive inclusion flow (renameMap lookup path), and the test confirms correct resolution.

vijayvarghese: Added test coverage for both cases.
Validated linker diagnostics for:
missing \_\_wrap\_\<sym\> → unresolved \_\_wrap\_\<sym\>
missing \_\_real\_\<sym\> → No archive pull needed, linking finished with no errors.

Also by default behavior verified if \<sym\> is not defined and there is a call to \_\_real\_\<sym\>  will result in a Error: build/wrap.o: undefined reference to `\<sym\>'

test/Common/standalone/WrapSymbolArchivePull/WrapSymbolArchivePullLTO.test
```diff
@@ -0,0 +1,28 @@
+#---WrapSymbolArchivePullLTO.test---------- Executable ------------------#
+#BEGIN_COMMENT
+# Test that --wrap archive extraction works correctly under LTO.
+# When --wrap=add is active and wrap.o references __real_add,
+# IRBuilder renames __real_add to add in the name pool. Later,
+# when the archive is scanned, shouldIncludeSymbol() must still
```
quic-seaswara: remove comments not relevant for the test, test should contain only test specific comments.

test/Common/standalone/WrapSymbolArchivePull/WrapSymbolArchivePullLTO.test
```diff
@@ -0,0 +1,28 @@
...
+
+# Link with LTO: main.o + wrap.o are direct inputs;
+# add.o is only reachable via the archive.
+RUN: %link %linkopts -flto %t1.main.o %t1.wrap.o %t2.libfoo.a \
+RUN:   --wrap=add -o %t2.out -t 2>&1 | %filecheck %s
+
+# Verify the archive member was pulled in.
+#CHECK: {{.*}}libfoo.a{{.*}}add.o
+
+#UNSUPPORTED: windows, ndk-build, riscv32, riscv64
```
quic-seaswara: Why they are not supported ?

lib/Readers/ArchiveParser.cpp
```diff
@@ -500,10 +500,19 @@ ArchiveParser::shouldIncludeSymbol(const ArchiveFile &archive,
   //      archive scanning began, so it never has its own name pool entry.
   //      In that case <sym> itself is the renamed reference and info->isUndef()
   //      captures it.
+  // Emit a trace message when pulling archive members for wrapped symbols to aid
+  // in debugging the wrap chain and archive resolution process.
+  bool TraceWrap = m_Module.getPrinter()->traceWrapSymbols();
   std::string realSymbol = ("__real_" + SymName).str();
   ResolveInfo *RealSymbol = m_Module.getNamePool().findInfo(realSymbol);
-  if ((RealSymbol && RealSymbol->isUndef()) || info->isUndef())
+  if ((RealSymbol && RealSymbol->isUndef()) || info->isUndef()){
```
quic-seaswara: should this be __real_<symbol> or __wrap_<symbol> ?

quic-seaswara (changes requested):

lib/Readers/ArchiveParser.cpp
vijayvarghese: I investigated both \_\_real\_\<sym\> and \_\_wrap\_\<sym\> flows.

The original fallback using info->isUndef() was too broad. During LTO, wrapped symbols can remain unresolved due to wrap-reference tracking and rename-map processing even when there is no corresponding \_\_real\_\<sym\> reference. This caused unrelated archive members to be pulled.

After tracing the LTO flow, I kept the existing LTO archive extraction behavior intact and restricted the fallback archive extraction path to the non-LTO ELF flow. I also added test coverage for \_\_wrap\_\<sym\> definitions residing in archives to verify that case separately.

For the \_\_wrap\_\<sym\>  case, I verified that wrapper definitions residing in archives are still handled correctly and added dedicated test coverage for that scenario. \_\_wrap\_\<sym\>  definitions do not reach this wrap-specific archive extraction path because they are not matched as wrapped-symbol keys in the rename map and continue to follow the default archive inclusion behavior.

lib/Readers/ArchiveParser.cpp
```diff
@@ -494,12 +494,34 @@ ArchiveParser::shouldIncludeSymbol(const ArchiveFile &archive,
     return ArchiveFile::Symbol::Include;
 
   // If there is a wrap option specified for that symbol, we only pull from
-  // an archive, if the real symbol is still undefined.
+  // an archive if the wrap chain actually requires the symbol.
+
+  // Restrict the fallback archive extraction path to non-LTO builds.
+  bool isElfPath = !(m_Module.getConfig().options().hasLTO() ||
+                     m_Module.needLTOToBeInvoked() || isPostLTOPhase);
```
quic-seaswara, line 501: isPostLTOPhase is ELF, and this will turn isElfPath to false. Not sure what are you wanting to check.

Why are you restricting this to non LTO builds ?

quic-seaswara (changes requested):

lib/Readers/ArchiveParser.cpp
vijayvarghese: The reason for including `isPostLTOPhase` was that while tracing the regression I observed that `shouldIncludeSymbol()` is revisited during both the pre- and post-LTO archive scans. The broad fallback fixed the ELF issue, but it also tampered with the existing working LTO archive-pull logic and started pulling `<sym>` even when there was no corresponding `__real_<sym>` reference.

My intent was to keep the existing LTO archive extraction behavior unchanged and apply the fallback only to the flow where the original issue was observed. Since the post-LTO scan still operates on symbol state established during the LTO flow, I treated `isPostLTOPhase` as part of that path as well.

Regarding the check itself, the intent was not to classify ELF vs non-ELF object handling. The fallback is only safe when we are outside the LTO wrap-resolution flow. I included `hasLTO()`, `needLTOToBeInvoked()`, and `isPostLTOPhase` together to make it explicit that the fallback should not be applied during either the pre-LTO or post-LTO archive scans. Even if `isPostLTOPhase` ends up being redundant with the other conditions, I added it to make that intent clearer in the code.

That said, I agree the naming `isElfPath` is misleading. The check is really trying to distinguish the wrap-resolution flow where the fallback is safe to apply, rather than ELF vs non-ELF object handling.

parth-07: I don't fully understand why the fallback is only safe when we are outside the LTO wrap-resolution flow. Can you please expand upon this?

> I included hasLTO(), needLTOToBeInvoked(), and isPostLTOPhase together to make it explicit that the fallback should not be applied during either the pre-LTO or post-LTO archive scans.

Wouldn't this disable the fallback for both the pre and post LTO phase? Is that what we want?

It seems that we still do not behave correctly when the `__real_<sym>` undefined reference is in an LTO file. Reproducer:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
int __real_add(int u, int v);

int fn(int u, int v) {
  return __real_add(u, v);
}
EOF

cat >2.c <<\EOF
int add(int u, int v) {
  return u + v;
}
EOF

cat >main.c <<\EOF
int fn(int u, int v);

int main() {
  return fn(3, 5);
}
EOF

TARGET_OPT="-target x86_64-linux-gnu"
#TARGET_OPT=""
CC=clang

$CC ${TARGET_OPT} -o 1.o 1.c -c -flto
$CC ${TARGET_OPT} -o 2.o 2.c -c
$CC ${TARGET_OPT} -o main.o main.c -c

llvm-ar cr lib2.a 2.o

LDs=(ld.eld)
SFs=(eld)

for i in "${!SFs[@]}"; do
  ${LDs[$i]} -o a.${SFs[$i]}.out main.o 1.o lib2.a --wrap=add
done
```

test/Common/standalone/WrapSymbolArchivePull/WrapSymbolNoRealRef.test
```diff
@@ -0,0 +1,19 @@
...
+
+RUN: %clang %clangopts -c %p/Inputs/1.c -o %t1.main.o
+RUN: %clang %clangopts -c %p/Inputs/3.c -o %t1.add.o
+RUN: %clang %clangopts -c %p/Inputs/4.c -o %t1.wrap.o
+RUN: %ar cr %aropts %t2.libfoo.a %t1.add.o
+
+
+# Should link successfully — __real_add never referenced
+RUN: %link %linkopts %t1.main.o %t1.wrap.o %t2.libfoo.a \
+RUN:   --wrap=add -o %t2.out
```
parth-07, line 16: %filecheck directive is missing here.

parth-07 (changes requested):

PR #1190, opened 2026-05-18
Make last conflicting option take precedence

### LINKED ISSUE #1179: Mutually Exclusive Options Do Not Override Each Other
### **Description**

When passing mutually exclusive options, the **last occurrence does not override the previous one**, leading to inconsistent behavior.

For example:

```bash
$ ld.eld --no-warn-mismatch --warn-mismatch ...
# Warning is shown (expected)

$ ld.eld --warn-mismatch --no-warn-mismatch ...
# Warning is still shown (unexpected)
```

### **Expected Behavior**

Command-line option parsing should follow the standard convention where:

> **The last occurrence of a conflicting option takes precedence.**

So the second invocation should disable the warning:

```bash
$ ld.eld --warn-mismatch --no-warn-mismatch ...
# Warning should NOT be shown
```

### **Actual Behavior**

The `--no-warn-mismatch` option does not override a prior `--warn-mismatch`, even when it appears later in the argument list.

### **Impact**

*   Breaks expected CLI semantics
*   Can lead to confusing or misleading behavior for users
*   Makes it difficult to reliably control linker behavior via scripts or build systems

### **Notes**

This issue may affect other pairs of mutually exclusive options that follow a similar pattern (`--foo` / `--no-foo`).

Steven6798:
Update option handling so that the last occurrence of conflicting flags (e.g. --warn-mismatch/--no-warn-mismatch) takes precedence, matching standard command-line semantics and user expectations.

Fix: #1179

### DIFF
--- a/lib/LinkerWrapper/GnuLdDriver.cpp
+++ b/lib/LinkerWrapper/GnuLdDriver.cpp
@@ -202,13 +202,14 @@ void GnuLdDriver::printVersionInfo() const {
 // This function checks for such errors.
 template <class T>
 bool GnuLdDriver::checkOptions(llvm::opt::InputArgList &Args) const {
+  // TODO: Disabled since this will get replaced with a warning.
   // check --thread-count and if threads are disabled.
-  if (Args.getLastArg(T::thread_count)) {
-    if (!Config.options().threadsEnabled()) {
-      Config.raise(Diag::thread_count_with_no_threads);
-      return false;
-    }
-  }
+  // if (Args.getLastArg(T::thread_count)) {
+  //   if (!Config.options().threadsEnabled()) {
+  //     Config.raise(Diag::thread_count_with_no_threads);
+  //     return false;
+  //   }
+  // }
   return true;
 }
 
@@ -339,12 +340,9 @@ bool GnuLdDriver::processOptions(llvm::opt::InputArgList &Args) {
   if (llvm::opt::Arg *arg = Args.getLastArg(T::sysroot))
     Config.setSysRoot(arg->getValue());
 
-  // --fatal-warnings
-  Config.options().setFatalWarnings(Args.hasArg(T::fatal_warnings));
-
-  // --no-fatal-warnings
-  if (Args.hasArg(T::no_fatal_warnings))
-    Config.options().setFatalWarnings(false);
+  // --[no-]fatal-warnings
+  Config.options().setFatalWarnings(
+      Args.hasFlag(T::fatal_warnings, T::no_fatal_warnings, /*default=*/false));
 
   // --opt-record-file
   if (Args.hasArg(T::opt_record_file)) {
@@ -480,8 +478,9 @@ bool GnuLdDriver::processOptions(llvm::opt::InputArgList &Args) {
     }
   }
 
-  // --export-dynamic, -E
-  Config.options().setExportDynamic(Args.hasArg(T::export_dynamic));
+  // --[no-]export-dynamic, -E
+  Config.options().setExportDynamic(
+      Args.hasFlag(T::export_dynamic, T::no_export_dynamic, /*default=*/false));
 
   // --export-dynamic-symbol
   for (auto *Arg : Args.filtered(T::export_dynamic_symbol))
@@ -583,68 +582,57 @@ bool GnuLdDriver::processOptions(llvm::opt::InputArgList &Args) {
     }
   }
 
-  // --warn-shared-textrel
-  if (Args.hasArg(T::warn_shared_textrel))
-    Config.options().setWarnSharedTextrel(true);
+  // --[no-]warn-shared-textrel
+  Config.options().setWarnSharedTextrel(Args.hasFlag(
+      T::warn_shared_textrel, T::no_warn_shared_textrel, /*default=*/false));
 
   // --warn-common
   if (Args.hasArg(T::warn_common))
     Config.options().setWarnCommon();
 
-  // --no-warn-shared_textrel
-  if (Args.hasArg(T::no_warn_shared_textrel))
-    Config.options().setWarnSharedTextrel(false);
-
-  // --enable-newdtags, --disable-newdtags
-  if (Args.hasArg(T::enable_newdtags) && Args.hasArg(T::disable_newdtags)) {
-    errs() << "Cannot specify enable and disable  DTAGS at same time!\n";
-    return false;
-  }
-
-  // --enabld-new-dtags
-  if (Args.hasArg(T::enable_newdtags))
-    Config.options().setNewDTags(true);
-
-  // --enabld-disable-new-dtags
-  if (Args.hasArg(T::disable_newdtags))
-    Config.options().setNewDTags(false);
+  // --enable-new-dtags, --disable-new-dtags
+  bool newdtags =
+      Args.hasFlag(T::enable_newdtags, T::disable_newdtags, /*default=*/false);
+  Config.options().setNewDTags(newdtags);
 
   // --enable-linker-version / --disable-linker-version
-  if (Args.hasArg(T::enable_linker_version) &&
-      Args.hasArg(T::disable_linker_version)) {
-    errs() << "Cannot specify enable and disable LINKER_VERSION at same time!\n";
-    return false;
-  }
-
-  if (Args.hasArg(T::enable_linker_version)) {
-    Config.options().setLinkerVersionDirectiveEnabled(true);
+  bool linkerVersion = Args.hasFlag(
+      T::enable_linker_version, T::disable_linker_version, /*default=*/false);
+  Config.options().setLinkerVersionDirectiveEnabled(linkerVersion);
+  if (linkerVersion)
     Config.addCommandLine(Table->getOptionName(T::enable_linker_version), true);
-  }
-
-  if (Args.hasArg(T::disable_linker_version)) {
-    Config.options().setLinkerVersionDirectiveEnabled(false);
-    Config.addCommandLine(Table->getOptionName(T::disable_linker_version), true);
-  }
+  else
+    Config.addCommandLine(Table->getOptionName(T::disable_linker_version),
+                          true);
 
   bool recordCommandLine = Args.hasFlag(
       T::record_command_line, T::no_record_command_line, /*default=*/false);
   Config.options().setRecordCommandLine(recordCommandLine);
 
-  // --emit-relocs
-  if (Args.hasArg(T::emit_relocs)) {
-    Config.options().setEmitGNUCompatRelocs(true);
-    Config.options().setEmitRelocs(true);
-    Config.addCommandLine(Table->getOptionName(T::emit_relocs), true);
-  }
-
-  // --emit-relocs-llvm
-  if (Args.hasArg(T::emit_relocs_llvm))
-    Config.options().setEmitRelocs(true);
-
-  // --no-emit-relocs
-  if (Args.hasArg(T::no_emit_relocs)) {
-    Config.options().setEmitGNUCompatRelocs(false);
-    Config.options().setEmitRelocs(false);
+  // --[no-]emit-relocs, --emit-relocs-llvm
+  if (llvm::opt::Arg *arg = Args.getLastArg(T::emit_relocs, T::no_emit_relocs,
+                                            T::emit_relocs_llvm)) {
+    bool emitRelocs = false;
+    switch (arg->getOption().getID()) {
+    case T::emit_relocs_llvm: {
+      emitRelocs = true;
+      break;
+    }
+    case T::emit_relocs: {
+      emitRelocs = true;
+      Config.options().setEmitGNUCompatRelocs(true);
+      Config.addCommandLine(Table->getOptionName(T::emit_relocs), true);
+      break;
+    }
+    case T::no_emit_relocs: {
+      emitRelocs = false;
+      Config.options().setEmitGNUCompatRelocs(false);
+      break;
+    }
+    default:
+      break;
+    }
+    Config.options().setEmitRelocs(emitRelocs);
   }
 
   // --no-merge-strings
@@ -652,12 +640,13 @@ bool GnuLdDriver::processOptions(llvm::opt::InputArgList &Args) {
   Config.addCommandLine(Table->getOptionName(T::no_merge_strings),
                         Args.hasArg(T::no_merge_strings));
 
-  // --{no-}warn-mismatch
-  if (Args.getLastArg(T::no_warn_mismatch))
-    Config.options().setWarnMismatch(false);
-
-  if (Args.getLastArg(T::warn_mismatch))
-    Config.options().setWarnMismatch(true);
+  // --[no-]warn-mismatch
+  // This flag defaults to none so only set it if at least one of the two flags
+  // are present.
+  if (Args.hasArg(T::warn_mismatch) || Args.hasArg(T::no_warn_mismatch)) {
+    Config.options().setWarnMismatch(
+        Args.hasFlag(T::warn_mismatch, T::no_warn_mismatch, /*default=*/true));
+  }
 
   // --no-trampolines
   if (Args.hasArg(T::no_trampolines)) {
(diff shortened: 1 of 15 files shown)


lib/LinkerWrapper/GnuLdDriver.cpp
```diff
@@ -62,11 +62,12 @@ using namespace eld;
 static constexpr llvm::opt::OptTable::Info infoTable[] = {
 #define OPTION(PREFIXES_OFFSET, PREFIXED_NAME_OFFSET, ID, KIND, GROUP, ALIAS,  \
                ALIASARGS, FLAGS, VISIBILITY, PARAM, HELPTEXT,                  \
-               HELPTEXTSFORVARIANTS, METAVAR, VALUES, SUBCOMMANDIDS_OFFSET)                          \
-  LLVM_CONSTRUCT_OPT_INFO(                                                     \
-      PREFIXES_OFFSET, PREFIXED_NAME_OFFSET, GnuLdOptTable::ID, KIND,          \
-      GnuLdOptTable::GROUP, GnuLdOptTable::ALIAS, ALIASARGS, FLAGS,            \
-      VISIBILITY, PARAM, HELPTEXT, HELPTEXTSFORVARIANTS, METAVAR, VALUES, SUBCOMMANDIDS_OFFSET),
+               HELPTEXTSFORVARIANTS, METAVAR, VALUES, SUBCOMMANDIDS_OFFSET)    \
```
quic-areg: These changes seem unrelated

lib/LinkerWrapper/GnuLdDriver.cpp
```diff
@@ -328,12 +329,16 @@ bool GnuLdDriver::processOptions(llvm::opt::InputArgList &Args) {
   if (llvm::opt::Arg *arg = Args.getLastArg(T::sysroot))
     Config.setSysRoot(arg->getValue());
 
-  // --fatal-warnings
-  Config.options().setFatalWarnings(Args.hasArg(T::fatal_warnings));
-
-  // --no-fatal-warnings
-  if (Args.hasArg(T::no_fatal_warnings))
-    Config.options().setFatalWarnings(false);
+  // --[no-]fatal-warnings
```
quic-areg, line 343: You can simplify it like this:

```cpp
Config.options().setFatalWarnings(
    Args.hasFlag(T::fatal_warnings, T::no_fatal_warnings, /*default=*/false));

Config.options().setWarnMismatch(
    Args.hasFlag(T::warn_mismatch, T::no_warn_mismatch, /*default=*/true));
```
etc

quic-areg (commented):
It should apply to these options too: 

```
--gc-sections / --no-gc-sections	
--export-dynamic / --no-export-dynamic	
--relax / --no-relax (riscv)
```

Unrelated to this PR, these options are hard errors today when both forms are present; we might want to relax them in the future for consistency:
```
--enable-linker-version / --disable-linker-version	
--thread-count / --no-threads	
--enable-new-dtags / --disable-new-dtags	
```

Another tangent, Hexagon has no `--no-relax` option like riscv does, and the `--help` text for hexagon's `--relax` says "Enable Hexagon Relaxation (default behavior)", but it is not actually the default.

lib/LinkerWrapper/GnuLdDriver.cpp
Steven6798: Thanks. I'll fix my clang formatting options.

lib/LinkerWrapper/GnuLdDriver.cpp
Steven6798: Thanks!

test/Common/standalone/FlagOverwrite/FlagOverwrite.test
```diff
@@ -0,0 +1,74 @@
+#---FlagOverwrite.test--------------------- Executable --------------------#
```
quic-areg, line 1: I don't think this test needs to be exhaustive. Slimming it down to two or three cases should demonstrate that the ` hasFlag(positive, negative, default)` pattern works and make the test simpler and easier to read.

quic-areg (commented):
LGTM aside from the minor test comment

test/Common/standalone/FlagOverwrite/FlagOverwrite.test
```diff
@@ -0,0 +1,74 @@
...
+#    --[no-]demangle
+#    --[no-]check-sections
+#    --[no-]omagic
+#    --[no-]gc-sections
+#    --[no-]align-segments
+#    --[no-]emit-relocs
+#    --[no-]threads
+#    --[no-]export-dynamic
+# TODO: Add check for --[no-]warn-mismatch when support for --no-warn-mismatch
+# is added to ELD
```
quic-seaswara: Is this comment valid ?

test/Common/standalone/FlagOverwrite/FlagOverwrite.test
```diff
@@ -0,0 +1,74 @@
...
+RUN: %readelf -s %t1.out | %filecheck %s -check-prefix=NOEXPDYN
+
+TEXTREL: Warning: Add DT_TEXTREL in a shared object!
+NOTEXTREL-NOT: Warning: Add DT_TEXTREL in a shared object!
+FATAL: Fatal: Linking had errors.
+NOFATAL-NOT: Fatal: Linking had errors.
+DEMANGLE: Verbose: Adding symbol foo()
+NODEMANGLE: Verbose: Adding symbol _Z3foov
+CHECKSECT: section .comment file range overlaps with .text
+NOCHECKSECT-NOT: section .comment file range overlaps with .text
```
quic-seaswara, line 103: This looks like a bug ?

test/AArch64/standalone/TLS_IE_ALIGN/IEAlign.test
```diff
@@ -1,7 +1,7 @@
 #---IEAlign.test----------------------------------------------------------#
 # Check that TLS IE GOT entries account for TLS block alignment padding.
 RUN: %clang %clangopts -target aarch64 -c %p/Inputs/ie-align.c -o %t.o -fPIC -ftls-model=initial-exec
-RUN: %link %linkopts -march aarch64 %t.o -o %t.out --no-threads
+RUN: %link -march aarch64 %t.o -o %t.out --no-threads
```
quic-areg: I am guessing that `linkopts` was removed because it includes `--thread-count` and `--no-threads` which leads to an error? Can we use `--thread-count=1` instead of `--no-threads` and add back `linkopts`?

lib/LinkerWrapper/GnuLdDriver.cpp
```diff
@@ -1117,13 +1109,15 @@ bool GnuLdDriver::processOptions(llvm::opt::InputArgList &Args) {
...
-    Config.options().setDemangleStyle("demangle");
+  // --[no-]demangle
+  // This flag sets the same thing as --demangle-style so only set it if at
+  // least one of the two flags are present.
+  if (Args.getLastArg(T::demangle) || Args.getLastArg(T::no_demangle)) {
+    if (Args.hasFlag(T::demangle, T::no_demangle, /*default=*/true))
+      Config.options().setDemangleStyle("demangle");
+    else
+      Config.options().setDemangleStyle("none");
+  }
```
quic-seaswara, line 1146: Can we look into adding this as a -Wcommand-line option when there are conflicting options specified. 

If we plan to pursue this, then the old style is something that we want ?

test/Common/standalone/FlagOverwrite/FlagOverwrite.test
Steven6798: Sounds good. Thanks!

test/AArch64/standalone/TLS_IE_ALIGN/IEAlign.test
Steven6798: Of course. Thanks.

test/Common/standalone/FlagOverwrite/FlagOverwrite.test
Steven6798: Until the PR that add support for no-warn-mismatch is still open then yes. Adarsha mentioned that this test does not have to be exhaustive so I can remove the comment and no need to include that test case.

test/Common/standalone/FlagOverwrite/FlagOverwrite.test
Steven6798: You mean a bug in my test or that the check itself found a bug?

If it is a bug in eld then awesome. Since you understand why it is a bug then raise an issue. Great catch.

lib/LinkerWrapper/GnuLdDriver.cpp
Steven6798: Assuming that the -Wcommand-line option is a warning, I think we still need some consistency. The warning is already in my TODO list but making the last option take precedence is still a good idea. Otherwise, the order depends on which option is defined last in this function.

If you mean that the -Wcommand-line option is the one that controls if the last option takes precedence or not, then we would need both the old and new style.

I can update the PR if needed, let me know what approach you think is best.

test/AArch64/standalone/TLS_GD_ALIGN/GDAlign.test
```diff
@@ -1,7 +1,7 @@
 #---GDAlign.test----------------------------------------------------------#
 # Check that TLS GD GOT entries account for TLS block alignment padding.
 RUN: %clang %clangopts -target aarch64 -c %p/Inputs/gd-align.c -o %t.o -fPIC -ftls-model=global-dynamic
-RUN: %link %linkopts -march aarch64 %t.o -o %t.out --no-threads
+RUN: %link -march aarch64 %t.o -o %t.out --no-threads
```
parth-07: We need to keep `%linkopts` in the tests. Also, until now using both `--thread-count` and `--no-threads` was not an error. Why is it a problem now? Do you think it is a right behavior to make it an error?

Steven6798: --no-threads was being ignored if it was added after --threads. We now see the error because that is no longer the case.

parth-07: Do you think it is a right behavior to make it an error?

Steven6798: I did not make it an error, that logic was already there. I have not looked at why that check is there and the implications of it, but I can look into it and let you know.

parth-07: Yes, you did not make it an error, however, it is changing the already existing user-observed behavior that is seemingly unrelated to the PR description and the commit message.

Steven6798: Agreed. I'll look into the check. Thanks!

lib/LinkerWrapper/GnuLdDriver.cpp
quic-areg: I think the `-Wcommand-line` proposal may be a reasonable idea as a complement to last wins but it should not affect this PR.

test/AArch64/standalone/TLS_GD_ALIGN/GDAlign.test
quic-areg: It might be easier to revert the threads change out of this PR and land it as a follow-up. 

I think we could relax the error to be a warning: `--thread-count 4 ignored because threads are disabled`

There are two other pairs that lead to hard errors:
```
--enable-linker-version / --disable-linker-version	
--enable-new-dtags / --disable-new-dtags
```

Steven6798: Yes, it looks like the error is not necessary. I will still need to address the error in this PR otherwise the tests I modified will still fail; but with the relaxation of the error, I can revert back all the changes to these tests.

Still, now I would need to modify the Hexagon's Threads test since it checks for this error. Let me know what you think.

lib/LinkerWrapper/GnuLdDriver.cpp
Steven6798: I will address the warning in another PR

test/AArch64/standalone/TLS_GD_ALIGN/GDAlign.test
Steven6798: I added --thread-count=1 for now as Adarsha suggested. Any changes to ethe error thrown when --no-threads and thread-count are used at the same time will be address in another Issue/PR.

Steven6798:
Is this ready to merge?

Steven6798:
There is a special case that i plan to handle here. --[no-]align-segments, --script, -T, --[no-]omagic, and -N modify the alignment. Since --script, -T, --[no-]omagic, and -N require the alignment to be set to false, this will take precedence over --align-segments

Steven6798:
I added more missing cases. I also disabled some errors when passing conflicting options. These errors will become warning in the fix of issue #1222

lib/LinkerWrapper/GnuLdDriver.cpp
```diff
@@ -202,13 +202,14 @@ void GnuLdDriver::printVersionInfo() const {
...
-      Config.raise(Diag::thread_count_with_no_threads);
-      return false;
-    }
-  }
+  // if (Args.getLastArg(T::thread_count)) {
+  //   if (!Config.options().threadsEnabled()) {
+  //     Config.raise(Diag::thread_count_with_no_threads);
+  //     return false;
+  //   }
+  // }
```
quic-seaswara, line 212: why commented code ?

test/RISCV/standalone/FlagOverwrite/FlagOverwrite.test
```diff
@@ -0,0 +1,16 @@
...
+#START_TEST
+RUN: %clang %clangopts -c %p/Inputs/1.s -o %t1.1.o
+RUN: %link %linkopts --emit-relocs --no-relax --relax  -T %p/Inputs/script.t  %t1.1.o -o %t1.out
+RUN: %objdump --no-print-imm-hex -d %t1.out |  %filecheck %s -check-prefix=RELAX
+RUN: %link %linkopts --emit-relocs --relax --no-relax  -T %p/Inputs/script.t  %t1.1.o -o %t1.out
+RUN: %objdump --no-print-imm-hex -d %t1.out |  %filecheck %s -check-prefix=NORELAX
+
+RELAX-NOT: auipc	ra, 0
+NORELAX: auipc	ra, 0
+#END_TEST
```
quic-seaswara, line 16: can you update the test for all the options covered by this change ? one test per option group would be great.

quic-seaswara (changes requested):

lib/LinkerWrapper/GnuLdDriver.cpp
Steven6798: This error will become a warning in another PR. I commented the code out to prevent having to modify test that will trigger the error after the changes of this PR.

test/RISCV/standalone/FlagOverwrite/FlagOverwrite.test
Steven6798: I can make it exhaustive. Thanks.

Steven6798: Making it exhaustive will result in over 30 test cases. Is that ok?

quic-seaswara: Sure, if have the code to detect options are conflicting, supporting by tests is the right way to move forward.

Also make a table in the documentation to know that the options are conflicting.

Steven6798: Sounds good. I'll move the PR to draft until this is done.

Steven6798:
@quic-seaswara I updated the tests. They are mostly snippet of existing tests. After this PR gets merged the fix for issue #1222 will replace the snippets with the new warning, simplifying the test and eliminating duplication.

parth-07 (approved):

quic-seaswara (approved):

PR #1199, opened 2026-05-20
Omit empty sections in map file output

### LINKED ISSUE #1169: Remove empty sections from text map file
The current format of the text map file is below :

```
Linker Script and memory map
Build Statistics
# <file> <start time> <duration>
Linker scripts used (including INCLUDE command)
Linker Plugin Information
Linker Plugin Run Information

```

These sections are usually empty if  you just link a simple object file

`ld.eld a.o -Map x`

We should remove the empty sections in the map file if there is no content.

Steven6798:
Remove unused sections from the generated text map file when they contain no data. This change improves readability by only including sections with actual content.

Fix #1169

### DIFF
--- a/lib/LayoutMap/TextLayoutPrinter.cpp
+++ b/lib/LayoutMap/TextLayoutPrinter.cpp
@@ -358,6 +358,14 @@ void TextLayoutPrinter::printGlobalPluginInfo(Module &M, bool UseColor) {
   const LinkerScript::PluginVectorT &UniversalPlugins =
       M.getPluginManager().getUniversalPlugins();
 
+  bool hasPluginInfo =
+      !UniversalPlugins.empty() || !PluginListForSectionMatcher.empty() ||
+      !PluginListForSectionIterator.empty() ||
+      !PluginListForOutputSectionIterator.empty() ||
+      !PluginListForMemorySize.empty() || !PluginListForFileSize.empty();
+  if (!hasPluginInfo)
+    return;
+
   if (UseColor)
     outputStream().changeColor(llvm::raw_ostream::GREEN);
   outputStream() << "Linker Plugin Information"
@@ -415,6 +423,9 @@ void TextLayoutPrinter::printGlobalPluginInfo(Module &M, bool UseColor) {
     for (auto &P : PluginListForFileSize)
       PrintPlugin(*P);
   }
+
+  if (!M.getScript().getPluginRunList().size())
+    return;
   outputStream() << "Linker Plugin Run Information"
                  << "\n";
   int Index = 0;
--- a/test/Common/standalone/CMakeLists.txt
+++ b/test/Common/standalone/CMakeLists.txt
@@ -1,3 +1,4 @@
 add_subdirectory(EmitMapFileEvenInCrash)
+add_subdirectory(EmptyMapSections)
 add_subdirectory(PluginAPI)
 add_subdirectory(SymbolResolutionReport)
--- a/test/Common/standalone/CommandLine/BuildAndReproduce/BuildAndReproduce.test
+++ b/test/Common/standalone/CommandLine/BuildAndReproduce/BuildAndReproduce.test
@@ -20,7 +20,6 @@ RUN: %filecheck %s < %t2.out.map
 #CHECK: Linker scripts used (including INCLUDE command)
 #CHECK: {{.*}}script.t({{.*}})
 #CHECK: 	bar.t({{.*}})
-#CHECK: Linker Plugin Information
 #CHECK: foo.o*(.text*) #Rule 1, {{.*}}script.t {{.*}}
 #CHECK: bar.o*(.text*) #Rule 3, bar.t {{.*}}
 #END_TEST
--- a/test/Common/standalone/EmptyMapSections/CMakeLists.txt
+++ b/test/Common/standalone/EmptyMapSections/CMakeLists.txt
@@ -1,0 +1,21 @@
+set(SOURCES EmptyPlugin.cpp)
+
+if(NOT CYGWIN AND LLVM_ENABLE_PIC)
+  set(SHARED_LIB_SOURCES ${SOURCES})
+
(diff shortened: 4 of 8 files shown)


Steven6798:
I also updated the BuildAndReproduce test to no longer check for the empty "Linker Plugin Information" section

quic-areg (approved):

lib/LayoutMap/TextLayoutPrinter.cpp
```diff
@@ -357,6 +357,13 @@ void TextLayoutPrinter::printGlobalPluginInfo(Module &M, bool UseColor) {
       M.getScript().getPluginForType(plugin::Plugin::Type::ControlFileSize);
   const LinkerScript::PluginVectorT &UniversalPlugins =
       M.getPluginManager().getUniversalPlugins();
+  bool hasPluginInfo =
+      !UniversalPlugins.empty() || !PluginListForSectionMatcher.empty() ||
+      !PluginListForSectionIterator.empty() ||
+      !PluginListForOutputSectionIterator.empty() ||
+      !PluginListForMemorySize.empty() || !PluginListForFileSize.empty();
+  if (!hasPluginInfo)
+    return;
```
quic-seaswara, line 367: nit : Can we make this a private function ?

lib/LayoutMap/TextLayoutPrinter.cpp
```diff
@@ -1141,10 +1157,13 @@ void TextLayoutPrinter::printBuildStatistics(Module &CurModule, bool UseColor) {
 
 // Print the loaded Linker scripts and Memory Map files.
 void TextLayoutPrinter::printInputActions(bool UseColor) {
+  auto &InputActions = ThisLayoutInfo->getInputActions();
+  if (InputActions.empty())
+    return;
```
quic-seaswara: Do we have a test that checks this ?

lib/LayoutMap/TextLayoutPrinter.cpp
Steven6798: Of course.

lib/LayoutMap/TextLayoutPrinter.cpp
Steven6798: I forgot to add the tests. Thanks!

parth-07:
@quic-seaswara @quic-areg 
I feel that in the map-file seeing the sections actually empty and the stats as actually `0` is more assuring than not seeing the section/link-stat at all. Users, if they do not see the section/link-stat, may wonder is the stat `0` or did the linker not print the stat, or perhaps this stat is not available in this link version / link configuration. Please share your thoughts on it.

lib/LayoutMap/TextLayoutPrinter.cpp
quic-areg: I don't think it is valuable to extract this to a function. It is only ever used once and is more readable as-is imo.

lib/LayoutMap/TextLayoutPrinter.cpp
```diff
@@ -1114,8 +1121,15 @@ void TextLayoutPrinter::printLinkerInsertedTimingStats(Module &CurModule) {
...
   if (UseColor)
     outputStream().changeColor(llvm::raw_ostream::MAGENTA);
-  outputStream() << "\nBuild Statistics";
-  outputStream() << "\n# <file> <start time> <duration>\n";
+  bool InsertedBuildStats = false;
+  auto PrintBuildStatsHeader = [this, &InsertedBuildStats]() {
+    if (InsertedBuildStats)
+      return;
+    outputStream() << "\nBuild Statistics";
+    outputStream() << "\n# <file> <start time> <duration>\n";
```
quic-areg: In practice I don't think we ever show this information. It was a hexagon-specific feature (timing section) that I do not think is in use. If that is the case, I think it should be removed entirely. @quic-seaswara

quic-seaswara: Good suggestion, let us remove emitting Build Statistics entirely then. Please raise a seperate PR for this removal.

lib/LayoutMap/TextLayoutPrinter.cpp
Steven6798: For this case I don't see a way how this section will ever be empty. We always have at least 1 entry for the input file, and we can't run without input files. Any thoughts?

quic-seaswara: Yes, InputActions will never be empty I suppose.

Steven6798:
I removed the changes from printInputActions function since it is never empty. Also, from printBuildStatistics since it is not used and will be removed in another issue/PR.

lib/LayoutMap/TextLayoutPrinter.cpp
```diff
@@ -358,6 +358,14 @@ void TextLayoutPrinter::printGlobalPluginInfo(Module &M, bool UseColor) {
   const LinkerScript::PluginVectorT &UniversalPlugins =
       M.getPluginManager().getUniversalPlugins();
 
+  bool hasPluginInfo =
+      !UniversalPlugins.empty() || !PluginListForSectionMatcher.empty() ||
+      !PluginListForSectionIterator.empty() ||
+      !PluginListForOutputSectionIterator.empty() ||
+      !PluginListForMemorySize.empty() || !PluginListForFileSize.empty();
+  if (!hasPluginInfo)
+    return;
```
quic-seaswara, line 367: Can we simply use M.getLinkerScript().getPluginRunList().size() and see if it is non zero, then we print the plugin information ?

Steven6798: I'll try that and let you know. Thanks!

Steven6798: That's not enough. The EmitMapFileEvenInCrash test fails it is expecting plugin information that is now missing. It looks like we need to check for both plugin information and plugin run information, so I need to add your suggestion at the end of the function to have both checks. Nice catch.

PR #1203, opened 2026-05-21
Fix evaluation order of (before/in/after)-SECTIONS assignments

### LINKED ISSUE #430: Assignments after `SECTIONS` command should be evaluated after evaluating assignments in `SECTIONS` command

### LINKED ISSUE #1245: Map file incorrectly has all the assignments outside `SECTIONS` at the top in the `Output section and layout` heading
Map file incorrectly has all the assignments outside `SECTIONS` at the top in the `Output section and layout` heading of the map-file.

Reproducer:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
int foo() { return 1; }
EOF

cat >script.t <<\EOF
u = 0x13;
SECTIONS {
  text : {
    u = 0x15;
    *(.text)
  }
}
u = 0x17;
EOF

clang -o 1.o -target x86_64-linux-gnu 1.c -c
LDs=(ld.eld ld.lld ld.bfd)
SFs=(eld lld bfd)

for i in "${!SFs[@]}"; do
done

grep -E 'u\(.*\) =|text' 1.eld.map.txt
```

**Output:**

```
+ grep -E 'u\(.*\) =|text' 1.eld.map.txt
u(0x13) = 0x13; # u = 0x13; script.t
u(0x17) = 0x17; # u = 0x17; script.t
text	0x400000	0xb # Offset: 0x1000, LMA: 0x400000, Alignment: 0x10, Flags: SHF_ALLOC|SHF_EXECINSTR, Type: SHT_PROGBITS
	u(0x15) = 0x15; # u = 0x15; script.t
*(.text) #Rule 1, script.t [1:0]
```

`u(0x170 = 0x17;` assignment should be at the end instead of at the top.

parth-07:
Until now, eld used to incorrectly evaluate before SECTIONS and after SECTIONS assignments both before and after the in-SECTIONS assignments. This results in incorrect linker script symbol values and incorrect layout.

This patch fixes this linker script assignment evaluation order. Now, we evaluate the assignments after the SECTIONS command as part of the layout re-iteration loop in create(Script)ProgramHeaders.

One of the key design change in this patch is to store AfterSections assignments with the last output section. It is required to properly evaluate the assignments when the link contains multiple SECTIONS commands.

Resolves #430
Resolves #1245

### DIFF
--- a/include/eld/Object/OutputSectionEntry.h
+++ b/include/eld/Object/OutputSectionEntry.h
@@ -98,17 +98,21 @@ public:
 
   void append(RuleContainer *PInput) { Inputs.push_back(PInput); }
 
-  const SymbolAssignments &sectionsEndAssignments() const {
-    return SectionEndAssignments;
+  const SymbolAssignments &getPostOutputSectionAssignments() const {
+    return postOutputSectionAssignments;
   }
-  SymbolAssignments &sectionEndAssignments() { return SectionEndAssignments; }
+  SymbolAssignments &getPostOutputSectionAssignments() {
+    return postOutputSectionAssignments;
+  }
 
-  sym_iterator sectionendsymBegin() { return SectionEndAssignments.begin(); }
-  sym_iterator sectionendsymEnd() { return SectionEndAssignments.end(); }
+  void movePostOutputSectionAssignments(OutputSectionEntry *Out) {
+    postOutputSectionAssignments = Out->getPostOutputSectionAssignments();
+    Out->getPostOutputSectionAssignments().clear();
+  }
 
-  void moveSectionAssignments(OutputSectionEntry *Out) {
-    SectionEndAssignments = Out->sectionEndAssignments();
-    Out->sectionEndAssignments().clear();
+  template <typename Func> void forEachPostOutputSectionAssignment(Func F) {
+    for (auto *A : postOutputSectionAssignments)
+      F(A);
   }
 
   // A section may be part of multiple segments, this only returns the
@@ -221,7 +225,7 @@ private:
   size_t Order;
   bool IsDiscard;
   InputList Inputs;
-  SymbolAssignments SectionEndAssignments;
+  SymbolAssignments postOutputSectionAssignments;
   RuleContainer *FirstNonEmptyRule;
   RuleContainer *LastRule;
   std::vector<BranchIsland *> BranchIslands;
--- a/include/eld/Script/Assignment.h
+++ b/include/eld/Script/Assignment.h
@@ -35,10 +35,11 @@ class ELFSection;
 class Assignment : public ScriptCommand {
 public:
   enum Level {
-    BEFORE_SECTIONS,  // Assignments before SECTIONS command
-    AFTER_SECTIONS,   // Assignments after SECTIONS command
-    INPUT_SECTION,    // related to an input section
-    SECTIONS_END
+    BeforeSections,     // Assignments before any SECTIONS command
+    AfterInputSectDesc, // Assignments inside output section body (with input
+                        // rules)
+    AfterOutputSection, // Assignments between output sections inside SECTIONS
+    AfterSections       // Assignments after SECTIONS command
   };
 
   enum Type { DEFAULT, HIDDEN, PROVIDE, PROVIDE_HIDDEN, FILL, ASSERT, PRINT };
@@ -46,8 +47,6 @@ public:
 public:
   Assignment(Level AssignmentLevel, Type AssignmentType, std::string Symbol,
              Expression *ScriptExpression);
-
-
 
   Level level() const { return AssignmentLevel; }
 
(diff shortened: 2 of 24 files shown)


quic-seaswara (approved):

Steven6798 (approved):
Looks good.

quic-seaswara (commented):
How does the map file print the assignments, can you please add a test ?

include/eld/Object/OutputSectionEntry.h
```diff
@@ -98,17 +98,28 @@ class OutputSectionEntry {
...
-  SymbolAssignments &sectionEndAssignments() { return SectionEndAssignments; }
 
-  sym_iterator sectionendsymBegin() { return SectionEndAssignments.begin(); }
-  sym_iterator sectionendsymEnd() { return SectionEndAssignments.end(); }
+  sym_iterator postOutputSectionAssignmentsBegin() {
+    return postOutputSectionAssignments.begin();
+  }
+  sym_iterator postOutputSectionAssignmentsEnd() {
+    return postOutputSectionAssignments.end();
+  }
```
quic-seaswara: Can we use llvm:::make_range ?

include/eld/Script/Assignment.h
```diff
@@ -35,10 +35,11 @@ class ELFSection;
 class Assignment : public ScriptCommand {
 public:
   enum Level {
-    BEFORE_SECTIONS,  // Assignments before SECTIONS command
-    AFTER_SECTIONS,   // Assignments after SECTIONS command
-    INPUT_SECTION,    // related to an input section
-    SECTIONS_END
+    BeforeSections,     // Assignments before any SECTIONS command
+    AfterInputSectDesc, // Assignments inside output section body (with input
```
quic-seaswara, line 39: AfterRules ? or AfterRuleContainer ?

lib/LayoutMap/TextLayoutPrinter.cpp
```diff
@@ -1405,12 +1405,8 @@ void TextLayoutPrinter::printLayout(eld::Module &Module) {
...
     }
 
-    // Evaluate all assignments at the end of the output section.
-    for (OutputSectionEntry::sym_iterator It = (*Out)->sectionendsymBegin(),
-                                          Ie = (*Out)->sectionendsymEnd();
-         It != Ie; ++It) {
-      printAssignment(**It, Module, UseColor);
-    }
+    (*Out)->forEachPostOutputSectionAssignment(
+        [&](Assignment *A) { printAssignment(*A, Module, UseColor); });
```
quic-seaswara: const Assignment

parth-07:
> How does the map file print the assignments, can you please add a test ?

I have modified tests to also check the map-file contents.

quic-seaswara (changes requested):
general : can you add a test for assert ?

include/eld/Script/Assignment.h
parth-07: I have used the term InputSectDesc here to be consistent with the term used during the parsing. In parsing context, we predominantly use InputSectDesc over RuleContainer (For example, eld/Script/InputSectDesc, and ScriptParser::readInputSectionDescription). Assignment levels are assigned and primarily used during parsing, hence the use of the term AfterInputSectDesc.

lib/LayoutMap/TextLayoutPrinter.cpp
parth-07: Done.

include/eld/Object/OutputSectionEntry.h
parth-07: It turned out that these two functions were no longer getting used. Their usages were replaced with the newly added function `forEachPostOutputSectionAssignment`. I have removed these two functions.

parth-07:
> general : can you add a test for assert ?

I have added tests with ASSERTs.

parth-07:
@quic-seaswara can this patch be merged?

quic-seaswara:
We can definitely merge this, but the highest priority is to get the map file to show the evaluation order.

quic-seaswara (approved):

parth-07:
> We can definitely merge this, but the highest priority is to get the map file to show the evaluation order.

I have added the tests with map-file as well. The evaluation order is the same as the listed order in the linker script (and the map-file).

quic-seaswara:
Please update the docs about how assignments are evaluated and how the map file shows information.

PR #1205, opened 2026-05-21
Remove unnecessary and incorrect sorting of input sections

### LINKED ISSUE #1204: Incorrect sections order in the output image with LTO
When a linker script rule matches some input sections and there is no explicit ordering requirement specified, then the output image should contain the matched input sections in the same order as the input order. That is, the order of input files in the command-line and the order of sections within those input files.

For example, if a rule is `*(.text*)` and the input section order is 1.o(.text.foo), 2.o(.text.bar), and 3.o(.text.baz), then the corresponding output section of the rule must contain the input sections in the same order -- [1.o(.text.foo), 2.o(.text.bar), 3.o(.text.baz)]

However, when LTO is used, it is not always possible to maintain this order because the input file information for the bitcode files cannot be preserved. Hence, the ordering constraint is relaxed for the LTO-generated object file sections. But the linker should still maintain the ordering of sections from non-LTO object files. Currently, the input section ordering in the layout is incorrect in some cases when LTO is enabled.

Reproducer:

```bash
#!/usr/bin/env bash

cat >gen1.py <<\EOF
N = 300
with open("1.c", 'w') as f:
  f.write("int bar();")
  f.write("int baz();")
  for i in range(N):
    f.write(f"int foo_{i}() {{ return {i};}}\n")
  f.write("int main() {\n\treturn ")
  for i in range(N):
    f.write(f"foo_{i}()")
    if i != N-1:
      f.write(" + ");
  f.write(" + baz() + bar();\n}")

EOF

python3 gen1.py

cat >2.c <<\EOF
__attribute__((section(".ctors")))
int bar() { return 6; }
EOF

cat >3.c <<\EOF
int baz() { return 5; }
EOF

cat >script.t <<\EOF
SECTIONS {
  .text : { *(*.text*) }
}
EOF

clang -o 1.o -target aarch64-unknown-elf 1.c -c -ffunction-sections
clang -o 2.o -target aarch64-unknown-elf 2.c -c -flto -ffunction-sections
clang -o 3.o -target aarch64-unknown-elf 3.c -c -flto -ffunction-sections

LDs=(aarch64-link )
SFs=(eld ) 

for i in "${!SFs[@]}"; do
  ${LDs[$i]} -o a.${SFs[$i]}.out 2.o 1.o 3.o -flto-options=codegen="-function-sections" -T script.t -Map a.${SFs[$i]}.map.txt --save-temps
done

#start_line_no=$(grep -n '# Output Section' a.eld.map.txt | cut -f1 -d:)
#tail -n +${start_line_no} a.eld.map.txt | head -n 50
```

The .text output section content is:

```
.text.foo_62
.text.foo_63
.text.foo_64
....
....
.text.foo_298
.text.foo_299
.text.main
.text.foo_0
.text.foo_1
....
...
.text.foo_61
...
...
```

The order of .text.foo_ sections should have been correct because they are from non-LTO object file.

parth-07:
This commit removes the unnecessary and incorrect sorting of input sections before the merge input sections phase when the LTO is enabled.

Resolves #1204

### DIFF
--- a/lib/Object/ObjectLinker.cpp
+++ b/lib/Object/ObjectLinker.cpp
@@ -784,44 +784,6 @@ void ObjectLinker::markDiscardFileFormatSections() {
   }
 }
 
-bool ObjectLinker::mayBeSortSections(std::vector<Section *> &Sections) {
-  // If no linker scripts, we dont store the original input. Lets not sort.
-  if (!ThisModule->getScript().linkerScriptHasSectionsCommand())
-    return true;
-  if (ThisConfig.options().disableLTOLinkOrder())
-    return true;
-  // If we are doing partial link, lets not sort it.
-  bool IsPartialLink = (LinkerConfig::Object == ThisConfig.codeGenType());
-  if (IsPartialLink || LtoObjects.empty())
-    return true;
-  std::stable_sort(Sections.begin(), Sections.end(),
-                   [](Section *ASection, Section *BSection) {
-                     ELFSection *A = llvm::dyn_cast<ELFSection>(ASection);
-                     ELFSection *B = llvm::dyn_cast<ELFSection>(BSection);
-                     if (A == nullptr or B == nullptr)
-                       return false;
-                     // FIXME: Redundant checks. All files have original input.
-                     if (!A->originalInput())
-                       return false;
-                     if (!B->originalInput())
-                       return false;
-                     if ((A->name().starts_with(".ctors")) ||
-                         (B->name().starts_with(".ctors")))
-                       return false;
-                     if ((A->name().starts_with(".dtors")) ||
-                         (B->name().starts_with(".dtors")))
-                       return false;
-                     int64_t AOrdinal =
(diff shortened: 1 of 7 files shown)


quic-areg (approved):

lib/Object/ObjectLinker.cpp
```diff
@@ -784,44 +784,6 @@ void ObjectLinker::markDiscardFileFormatSections() {
   }
 }
 
-bool ObjectLinker::mayBeSortSections(std::vector<Section *> &Sections) {
-  // If no linker scripts, we dont store the original input. Lets not sort.
-  if (!ThisModule->getScript().linkerScriptHasSectionsCommand())
-    return true;
-  if (ThisConfig.options().disableLTOLinkOrder())
-    return true;
```
quic-seaswara, line 792: Can we remove the option too to disable lto link order ? We will need to eventually sort it for LTO with linker scripts support.

quic-seaswara (approved):

PR #1206, opened 2026-05-21
[test] fix fatlto tests on windows

quic-seaswara:
/dev/null filename is not yet supported on windows. change the file that the linker writes to.


quic-seaswara:
thanks @quic-shaksuma for catching this!

quic-shaksuma:
@quic-seaswara Can you please trigger a win build to verify fix

Issue #1110, opened 2026-04-28
eld creates invalid dynamic relocations with pie/TLS


quic-seaswara:
cat > tls.c << \!
static __thread int a = 10;
int foo() { return a; }
!

clang -c tls.c -fPIC -ftls-model=initial-exec 
ld.eld tls.o -pie -o libt.so 
llvm-readelf -r libt.so

quic-seaswara:
This bug exists for all targets that support pie. add this test to Common to make sure that the test works for all targets.

quic-areg:
This is not the case for X86, which has a different issue: https://github.com/qualcomm/eld/pull/1207

PR #1207, opened 2026-05-21
[X86] Emit R_X86_64_TPOFF64 for IE-mode TLS in shared objects.

quic-areg:
For IE TLS, x86_64 accesses a GOT entry containing the symbol's offset from the thread pointer. For executable outputs, the TP offset is fixed at link time, so the GOT entry can be resolved statically.

For shared objects, the offset is assigned by the dynamic loader. ELD was still writing the link-time static TLS offset into the GOT entry for local IE TLS references, leaving no dynamic relocation for the loader to apply.

Emit `R_X86_64_TPOFF64` for IE TLS GOT entries in shared objects, and mark the output with `DF_STATIC_TLS` when static TLS is required. Keep PIE output statically resolved.

### DIFF
--- a/lib/Target/X86/x86_64Relocator.cpp
+++ b/lib/Target/X86/x86_64Relocator.cpp
@@ -274,9 +274,17 @@ void x86_64Relocator::scanLocalReloc(InputFile &pInputFile, Relocation &pReloc,
     std::lock_guard<std::mutex> relocGuard(m_RelocMutex);
     if (rsym->reserved() & ReserveGOT)
       return;
-    // Create a TLS IE GOT entry.
     x86_64GOT *G = m_Target.createGOT(GOT::TLS_IE, Obj, rsym);
-    G->setValueType(GOT::TLSStaticSymbolValue);
+    // For executables, the symbol's offset from the thread pointer is fixed at
+    // link time. For shared objects, the dynamic loader must compute the offset
+    // at load time, so emit R_X86_64_TPOFF64.
+    if (config().isBuildingExecutable()) {
+      G->setValueType(GOT::TLSStaticSymbolValue);
+    } else {
+      helper_DynRel_init(Obj, &pReloc, rsym, G, 0x0,
+                         llvm::ELF::R_X86_64_TPOFF64, m_Target);
+      m_Target.setHasStaticTLS();
+    }
     rsym->setReserved(rsym->reserved() | ReserveGOT);
     return;
   }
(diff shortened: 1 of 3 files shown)


test/x86_64/linux/TLSIELocalSymbol/TLSIELocalSymbol.test
```diff
@@ -0,0 +1,36 @@
...
+  movl (%rax), %eax
+  retq
+  .size foo, .-foo
+
+  .section .tdata,"awT",@progbits
+  .p2align 2
+  .type a, @object
+  .size a, 4
+a:
+  .long 10
```
quic-seaswara: I would prefer having the code in 'C' and moving the test to Common.

lib/Target/X86/x86_64Relocator.cpp
```diff
@@ -274,9 +274,17 @@ void x86_64Relocator::scanLocalReloc(InputFile &pInputFile, Relocation &pReloc,
...
+    // For executables, the symbol's offset from the thread pointer is fixed at
+    // link time. For shared objects, the dynamic loader must compute the offset
+    // at load time, so emit R_X86_64_TPOFF64.
+    if (config().isBuildingExecutable()) {
+      G->setValueType(GOT::TLSStaticSymbolValue);
+    } else {
+      helper_DynRel_init(Obj, &pReloc, rsym, G, 0x0,
+                         llvm::ELF::R_X86_64_TPOFF64, m_Target);
+    }
+    m_Target.setHasStaticTLS();
```
quic-seaswara: Why are you setting StaticTLS ?

**Relevant code in ELFDynamic.cpp**

```
226   if (m_Backend.hasStaticTLS() &&
227       (LinkerConfig::DynObj == m_Config.codeGenType()))
228     dt_flags |= llvm::ELF::DF_STATIC_TLS;
```

quic-seaswara (changes requested):
It would be good to add a build and run test to make sure that the executable / shared object returns the value that we are looking for. 

I posted this here

https://github.com/qualcomm/eld/discussions/885#discussioncomment-16996361

test/x86_64/linux/TLSIELocalSymbol/TLSIELocalSymbol.test
quic-areg: I mentioned in #1110 that X86 has a different issue than the rest of the targets. I handle the remaining targets in #1208 which includes a Common test.

lib/Target/X86/x86_64Relocator.cpp
quic-areg: `DF_STATIC_TLS` is required for IE mode shared objects. Hexagon is the only target that sets this today (though it also sets it for executables where the flag is meaningless). The remaining targets I handled in #1208.

quic-seaswara (approved):

PR #1208, opened 2026-05-21
Resolve IE TLS GOT entries statically in PIE

### LINKED ISSUE #1110: eld creates invalid dynamic relocations with pie/TLS

quic-areg:
For non-preemptible symbols accessed through `R_*_GOTTPOFF / R_*_TLS_IE`, the IE-mode GOT entry was being initialized with a dynamic TPREL/TPOFF relocation in any non-static link, including -pie. PIE executables are not preempted, so the static TLS offset is known at link time and a dynamic relocation is both unnecessary and rejected by the loader.
    
Also set `DF_STATIC_TLS` on AArch64, ARM, and RISCV (Hexagon and X86 already do).

Fixes #1110

### DIFF
--- a/lib/Target/AArch64/AArch64Relocator.cpp
+++ b/lib/Target/AArch64/AArch64Relocator.cpp
@@ -307,8 +307,15 @@ void AArch64Relocator::scanLocalReloc(InputFile &pInput, Relocation &pReloc,
     if (config().isCodeStatic())
       return;
     AArch64GOT *G = m_Target.createGOT(GOT::TLS_IE, Obj, rsym);
+    if (config().isBuildingExecutable()) {
+      G->setValueType(GOT::TLSStaticSymbolValue);
+      if (rsym->reserved() == Relocator::None)
+        rsym->setReserved(rsym->reserved() | ReserveGOT);
+      return;
+    }
     helper_DynRel_init(Obj, &pReloc, rsym, G, 0x0,
                        llvm::ELF::R_AARCH64_TLS_TPREL64, m_Target);
+    m_Target.setHasStaticTLS();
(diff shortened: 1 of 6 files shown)


test/Common/standalone/TLS/TLSIELocalSymbol/TLSIELocalSymbol.test
```diff
@@ -0,0 +1,25 @@
...
+///   R_ARM_TLS_TPOFF32
+///   R_RISCV_TLS_TPREL64
+///   R_RISCV_TLS_TPREL32
+///   R_HEX_TPREL_32
+// SHARED: FLAGS
+// SHARED-SAME: STATIC_TLS
+// SHARED: R_{{.*}}TP{{(REL|OFF)_?[0-9]+}}
+// PIE: There are no relocations in this file.
+
+static __thread int a = 10;
```
parth-07: What do you think about also adding a non-local thread variable in the test as well to ensure `scanGlobalReloc` path is also tested.

quic-areg: Done

quic-seaswara (commented):
Can we also add a build/run test to make sure that the values printed are consistent and match expectations ?

parth-07 (approved):

quic-seaswara (approved):

quic-seaswara (approved):

PR #1212, opened 2026-05-22
[cmake] Normalize ELD_TARGETS_TO_BUILD target names to correct casing

### LINKED ISSUE #104: eld errors out at CMake step when LLVM_TARGETS_TO_BUILD contains `x86`.
eld errors out at CMake step with the below error when LLVM_TARGETS_TO_BUILD contains `x86` as one of the targets:

```
CMake Error at tools/eld/lib/Target/CMakeLists.txt:28 (add_subdirectory):
  add_subdirectory given source X86 which is not an existing directory.
```

The error happens because eld expects the target name to be `X86_64` instead of `X86`.

deepakshirkem:
When LLVM_TARGETS_TO_BUILD contains target names with incorrect casing (e.g. 'x86' instead of 'X86'), ELD inherits them and fails at CMake step because add_subdirectory expects the correct casing.

Normalize target names case-insensitively against ELD_ALL_TARGETS to ensure correct casing regardless of how the user specifies them.

### Output:
Screenshot from 2026-05-23 03-53-15

Fixes #104

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -150,6 +150,13 @@ if(NOT ELD_TARGETS_TO_BUILD)
       CACHE STRING "ELD Target(Targets) to build")
 endif()
 
+# Validate ELD_TARGETS_TO_BUILD against known targets
+foreach(t ${ELD_TARGETS_TO_BUILD})
+  if(NOT t IN_LIST ELD_ALL_TARGETS)
+    message(WARNING "Unknown ELD target '${t}'. Known targets are: ${ELD_ALL_TARGETS}")
+  endif()
+endforeach()
+
 # Set ELD_ENUM_TARGETS
 set(ELD_ENUM_TARGETS_STR "")
 set(ELD_ENUM_LINKERS_STR "")


CMakeLists.txt
```diff
@@ -150,6 +150,21 @@ if(NOT ELD_TARGETS_TO_BUILD)
...
+    string(TOLOWER ${t} t_lower)
+    string(TOLOWER ${known} known_lower)
+    if(t_lower STREQUAL known_lower)
+      list(APPEND ELD_NORMALIZED_TARGETS ${known})
+    endif()
+  endforeach()
+endforeach()
+if(ELD_NORMALIZED_TARGETS)
+  set(ELD_TARGETS_TO_BUILD ${ELD_NORMALIZED_TARGETS})
+endif()
```
quic-seaswara: If we keep the same name as LLVM this wont be needed right ?

deepakshirkem: You are right. LLVM also uses X86 as the target name. Instead of normalization, should we add a warning when an unrecognized target name is passed, to give users a clearer error message.

CMakeLists.txt
```diff
@@ -150,6 +150,14 @@ if(NOT ELD_TARGETS_TO_BUILD)
       CACHE STRING "ELD Target(Targets) to build")
 endif()
 
+# Validate ELD_TARGETS_TO_BUILD against known targets
+cmake_policy(SET CMP0057 NEW)
```
Steven6798: Is this needed? I think the cmake_minimum_required set this by default if it's a modern version. Other than that, everything looks good.

CMakeLists.txt
parth-07: @deepakshirkem Does LLVM CMake normalize the target names -- does it accept both 'arm' and 'ARM' as the target name in `LLVM_TARGETS_TO_BUILD`?

deepakshirkem: LLVM does not normalize target names. If a user passes an incorrect target name like x86, LLVM itself throws a FATAL_ERROR 

Screenshot from 2026-05-26 23-59-21

CMakeLists.txt
deepakshirkem: Thank You.  Updated the patch.

CMakeLists.txt
quic-seaswara: What does LLVM do if you pass a target that has a different case ?

quic-seaswara: If so, does LLVM cmake have a function ? I think this is overkill IMHO.

deepakshirkem: > What does LLVM do if you pass a target that has a different case ?

 LLVM throws a FATAL_ERROR.



> If so, does LLVM cmake have a function ? I think this is overkill IMHO.

LLVM has no such function. You are right, this is overkill. Should I close this PR as the fix is not needed?

quic-seaswara: Yes, no fix is needed.

PR #1216, opened 2026-05-26
Add LinkerWrapper::sortInputSectionsForSectionMerging

parth-07:
This commit adds the LinkerWrapper APIs
sortInputSectionsForSectionMerging and
getInputSectionsForSectionMerging. These APIs make it possible for plugins to change the order of the input sections, using a custom compare function, in a linker script rule and consequently in the output section.


include/eld/PluginAPI/LinkerWrapper.h
```diff
@@ -248,6 +248,47 @@ class DLL_A_EXPORT LinkerWrapper {
...
+  /// compare equivalent under \p Cmp retain their relative order from the
+  /// pre-sort list. \p Cmp must define a strict weak ordering.
+  ///
+  /// \param Cmp Comparator returning \c true when its first argument should
+  /// be ordered before its second.
+  ///
+  /// \note This function must only be used in the
+  /// \em ActBeforeSectionMerging link state.
+  eld::Expected<void>
+  sortInputSectionsForSectionMerging(InputSectionComparator Cmp);
```
quic-seaswara: Can we do this implicitly instead of explictly invoking this API ? 

sortInputSections instead of sortInputSectionsForSectionMerging ?

lib/LinkerWrapper/LinkerWrapper.cpp
```diff
@@ -212,6 +213,30 @@ eld::Expected<void> LinkerWrapper::finishAssignOutputSections() {
...
+      m_Module.getLinker()->getObjLinker()->getAllInputSections();
+  std::vector<plugin::Section> Sections;
+  Sections.reserve(AllInputSections.size());
+  for (eld::Section *S : AllInputSections)
+    Sections.emplace_back(S);
+  return Sections;
+}
+
+eld::Expected<void>
+LinkerWrapper::sortInputSectionsForSectionMerging(InputSectionComparator Cmp) {
```
quic-seaswara: Do  you want to sortInputSections with any comparator function, I think this is error prone.

We will need to record this action in the map file, if the order is changed by some plugin.

quic-seaswara (changes requested):

include/eld/PluginAPI/LinkerWrapper.h
parth-07: > Can we do this implicitly instead of explictly invoking this API ?

Do you mean adding a plugin hook that is called implicitly if a comparator function is registered by the plugin? If so, I do not think we should do this. I believe that it is best to keep plugin hooks count as low as possible for users to easily understand the plugin framework. I believe that plugin hooks should indicate either the position in the link pipeline (`ActBeforeRuleMatching` ...), or a component (`VisitSymbol`), and the `LinkerWrapper` APIs should be used for action/inspection. If we add too many hooks, then it will be difficult for users to understand when to use which hook and how the hooks are related to each other. The current use case is performing an action, and hence, I believe that a LinkerWrapper API is more suited for it.

> sortInputSections instead of sortInputSectionsForSectionMerging ?

I first went with the name `sortInputSections` but I feel that it does not give enough context, in particular, it does not tell whether we are sorting input sections within input files, or a global pool of all input sections. Also, it does not provide hint regarding restrictions that this API has.

lib/LinkerWrapper/LinkerWrapper.cpp
parth-07: > Do you want to sortInputSections with any comparator function, I think this is error prone.

I think we need to allow a custom comparator for the current use-case. 

> We will need to record this action in the map file, if the order is changed by some plugin.

Is it okay if we record this action in the plugin activity log file instead?

test/Common/Plugin/SortInputSectionsForMerging/SortInputSectionsForMerging.test
```diff
@@ -0,0 +1,23 @@
...
+#END_TEST
+
+CHECK:      Before sort:
+CHECK-DAG:    .text.aaa
+CHECK-DAG:    .text.bbb
+CHECK-DAG:    .text.ccc
+CHECK:      After sort:
+CHECK:        .text.ccc
+CHECK:        .text.bbb
+CHECK:        .text.aaa
```
quic-areg, line 26: Can we check that the sort survives section merging and is visible in the ELF or map?

test/Common/Plugin/SortInputSectionsForMerging/SortInputSectionsForMerging.test
```diff
@@ -0,0 +1,23 @@
+#---SortInputSectionsForMerging.test---------------------- Executable,LS --------------------#
+#BEGIN_COMMENT
+# Tests that LinkerWrapper::getInputSectionsForMerging exposes the
```
quic-areg: `getInputSectionsForSectionMerging` and `sortInputSectionsForSectionMerging`

include/eld/Object/ObjectLinker.h
```diff
@@ -331,6 +331,8 @@ class ObjectLinker {
     return AllInputSections;
   }
 
+  std::vector<Section *> &getAllInputSections() { return AllInputSections; }
```
quic-areg: can we change this to something like `sortAllInputSections(llvm::function_ref<bool(const Section*, const Section*)>)` to avoid exposing a mutable section vector?

lib/LinkerWrapper/LinkerWrapper.cpp
parth-07: I have recorded the information in the plugin activity log-file.

test/Common/Plugin/SortInputSectionsForMerging/SortInputSectionsForMerging.test
parth-07: I have updated the test to check the input section sorting through the map-file.

test/Common/Plugin/SortInputSectionsForMerging/SortInputSectionsForMerging.test
parth-07: Done. Thanks for catching this.

include/eld/Object/ObjectLinker.h
parth-07: Thanks for the suggestion. I have made this change.

quic-seaswara:
nit : spelling error in commit message,

fnuction -> function

parth-07:
> nit : spelling error in commit message,
> 
> fnuction -> function

Thanks for pointing it out. I have fixed it.

quic-areg (approved):

quic-seaswara (approved):

PR #1225, opened 2026-05-27
Add R_ARM_CALL runnable test

### LINKED ISSUE #1224: [ARM] Add static runnable test for relocations
[ARM] Add static runnable test for relocations

Steven6798:
Fix qualcomm#1224

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -129,6 +129,12 @@ jobs:
         working-directory: pr-${{ env.PR_NUMBER }}/obj
         run: ccache --evict-older-than 7200s --dir=$CCACHE_DIR
 
+      - name: Install dependencies
+        shell: bash
+        run:  |
+          sudo apt update
+          sudo apt install -y libsdl2-2.0-0
+
       - name: Run tests
         if: steps.file-check.outputs.skip_build == 'false'
         working-directory: pr-${{ env.PR_NUMBER }}/obj
--- a/test/ARM/RunTests/R_ARM_CALL/Inputs/hidden_callee.c
+++ b/test/ARM/RunTests/R_ARM_CALL/Inputs/hidden_callee.c
@@ -1,0 +1,22 @@
+// Hidden symbol: non-preemptible, direct relocation, no PLT entry.
+#define UART0_DR ((volatile unsigned int *)0x10009000)
+#define UART0_FR ((volatile unsigned int *)0x10009018)
+
+__attribute__((noinline, visibility("hidden"))) int hidden_callee() {
+  return 7;
+}
+
+int main(void) {
+  if (hidden_callee() != 7) {
+    return 1;
+  }
+  const char *s = "success\n";
+
+  while (*s) {
+    while (*UART0_FR & (1 << 5))
+      ;
+
+    *UART0_DR = *s++;
+  }
+  return 0;
+}
--- a/test/ARM/RunTests/R_ARM_CALL/R_ARM_CALL.test
+++ b/test/ARM/RunTests/R_ARM_CALL/R_ARM_CALL.test
@@ -1,0 +1,54 @@
+REQUIRES: run_test
+#---R_ARM_CALL.test--------------------- Executable------------------#
+#BEGIN_COMMENT
+# Test the ARM R_ARM_CALL relocation.
+# Covers: basic ARM-to-ARM call (static/PIE/dynamic), local and hidden
+# (non-preemptible) symbols, weak undefined symbol NOP conversion,
+# Thumb target BL->BLX rewrite, PLT call via shared library, and
+# trampoline creation for out-of-range branches.
+#END_COMMENT
+#START_TEST
+## --- Basic ARM-to-ARM call: local and global callee ---
+#RUN: %run_cc -nostdlib -c -o %t.startup.o %p/Inputs/startup.s
+#RUN: %run_cc -nostdlib -c -o %t.main.o %p/Inputs/main.c
+# Static
+#RUN: %link -T %p/Inputs/script.ld -static -o %t.static.out %t.startup.o %t.main.o
+#RUN: %not timeout .1 %run_sys -M vexpress-a9 -m 32M -kernel %t.static.out | %filecheck %s
+# PIE
+#RUN: %link -T %p/Inputs/script.ld -pie -o %t.pie.out %t.startup.o %t.main.o
+#RUN: %not timeout .1 %run_sys -M vexpress-a9 -m 32M -kernel %t.pie.out | %filecheck %s
+# Dynamic
+#RUN: %link -T %p/Inputs/script.ld -Wl,-dy -o %t.dyn.out %t.startup.o %t.main.o
+#RUN: %not timeout .1 %run_sys -M vexpress-a9 -m 32M -kernel %t.dyn.out | %filecheck %s
+
+## --- Weak undefined symbol: BL rewritten to NOP, execution continues ---
+#RUN: %run_cc -nostdlib -c -o %t.weak.o %p/Inputs/weak.c
+#RUN: %link -T %p/Inputs/script.ld -static -o %t.weak.out %t.startup.o %t.weak.o
+#RUN: %not timeout .1 %run_sys -M vexpress-a9 -m 32M -kernel %t.weak.out | %filecheck %s
+
+## --- Hidden (non-preemptible) symbol: direct call, no PLT ---
+#RUN: %run_cc -nostdlib -c -o %t.hidden.o %p/Inputs/hidden_callee.c
+#RUN: %link -T %p/Inputs/script.ld -static -o %t.hidden.out %t.startup.o %t.hidden.o
+#RUN: %not timeout .1 %run_sys -M vexpress-a9 -m 32M -kernel  %t.hidden.out | %filecheck %s
+
+## --- Thumb target: BL rewritten to BLX at link time ---
+#RUN: %run_cc -nostdlib -c -o %t.thumb.o %p/Inputs/thumb_callee.c
+#RUN: %link -T %p/Inputs/script.ld -static -o %t.thumb.out %t.startup.o %t.thumb.o
+#RUN: %not timeout .1 %run_sys -M vexpress-a9 -m 32M -kernel  %t.thumb.out | %filecheck %s
+
+## --- Preemptible symbol: call goes through PLT at runtime ---
+#RUN: %run_cc -fPIC -o %t.shared_callee.o -c %p/Inputs/shared_callee.c
+#RUN: %run_cc -fPIC -o %t.shared_caller.o -c %p/Inputs/shared_caller.c
+#RUN: %run_cc -shared -o %t.callee.so %t.shared_callee.o
+#RUN: %run_cc -o %t.shared.out %t.shared_caller.o %t.callee.so -Wl,-dy
+#RUN: %run %t.shared.out
+
+## --- Trampoline: out-of-range branch requires trampoline creation ---
+#RUN: %run_cc -nostdlib -c -o %t.distant_callee.o %p/Inputs/distant_callee.c
+#RUN: %run_cc -nostdlib -c -o %t.trampoline_caller.o %p/Inputs/trampoline_caller.c
+#RUN: %link -T %p/Inputs/script_trampoline.ld -static -o %t.trampoline.out %t.startup.o %t.trampoline_caller.o %t.distant_callee.o
+#RUN: %not timeout .1 %run_sys -M vexpress-a9 -m 64M -kernel %t.trampoline.out | %filecheck %s
+
+#CHECK: success
+
+#END_TEST
--- a/test/lit.cfg
+++ b/test/lit.cfg
@@ -214,10 +214,12 @@ dirname = 'dirname'
 datalayout = ''
 link_path = which(link)
 run=""
+run_sys=""
 run_cc = ''
 run_cxx = ''
 sysroot = ''
 emulator = ''
+emulator_system = ''
 
 # Control RISC-V Compression extension option, on by default
 clangnorvcopts = ''
@@ -540,9 +542,11 @@ if enable_run_tests:
       run_cxx  = resolve_env_tool(f'ELD_{arch_env}_CXX_COMPILER')
       sysroot  = os.environ.get(f'ELD_{arch_env}_SYSROOT', '')
       fallback_emulator = "qemu-" + arch_env.lower()
+      fallback_emulator_system = "qemu-system-" + arch_env.lower()
       emulator = resolve_env_tool(f'ELD_{arch_env}_EMULATOR') or which(fallback_emulator)
+      emulator_system = resolve_env_tool(f'ELD_{arch_env}_EMULATOR_SYSTEM') or which(fallback_emulator_system)
 
-  if run_cc and sysroot and emulator:
+  if run_cc and sysroot and emulator and emulator_system:
       config.available_features.add('run_test')
       run_cc += f" --ld-path={link_path} --sysroot={sysroot}"
       # This is required to ensure that libc.so from sysroot is loaded by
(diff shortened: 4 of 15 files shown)


Steven6798:
I'll appreciate some feedback on this, as this will serve as my template for adding more relocation tests. I'm sure these changes can be improved even more.

test/Common/RunTests/MyTest/MyTest.test
```diff
@@ -0,0 +1,29 @@
+REQUIRES: run_test
```
quic-areg: `MyTest` is a placeholder name; it should be changed to something descriptive - probably the name of the relocation in this case.

This is an ARM-specific test; I don't think it should belong in a directory called `Common`.

test/ARM/RunTests/CallReloc/CallReloc.test
```diff
@@ -0,0 +1,29 @@
...
+# Test the ARM R_ARM_CALL relocation.
+#END_COMMENT
+#START_TEST
+#RUN: %run_cc -o %t1.o -c %p/Inputs/1.s
+# Partial
+#RUN: %run_cc -o %t1.o %t1.o -r
+#RUN: llvm-readelf -r %t1.o | %filecheck %s -check-prefix RELOCS
+# Static
+#RUN: %run_cc -o %t1.out %t1.o -static
+#RUN: llvm-objdump -d --no-show-raw-insn %t1.out | FileCheck %s -check-prefix STATIC_RELOCS
```
quic-seaswara: %objdump

test/ARM/RunTests/CallReloc/Inputs/1.s
```diff
@@ -0,0 +1,54 @@
...
+                                        @ -- End function
+	.type	.L.str,%object                  @ @.str
+	.section	.rodata.str1.1,"aMS",%progbits,1
+.L.str:
+	.asciz	"Hello World"
+	.size	.L.str, 12
+
+	.ident	"clang version 23.0.0git (https://github.com/llvm/llvm-project.git 832285ab7dffc8381ae76182b9b1f0f4cbae23d0)"
+	.section	".note.GNU-stack","",%progbits
+	.addrsig
```
quic-seaswara: prefer 'C' code instead of assembly.

docs/DeveloperDocs/ARM/relocations.md
```diff
@@ -0,0 +1,15 @@
+# ARM Relocations
+
+## Relocation Table
+
+| Code | Name       | Type   | Class | Operation           |
+|------|------------|--------|-------|---------------------|
+| 28   | R_ARM_CALL | Static | Arm   | ((S + A) \| T) - P  |
```
quic-seaswara: add a link to the ARM ELF ABI. 

Can we have the table to also show the number of bits available and when the linker decides to insert a trampoline.

Also it may be useful to consider the differences between

* Static Linking
* PIE
* Shared objects

to make the table complete.

quic-seaswara (changes requested):

test/Common/RunTests/MyTest/MyTest.test
```diff
@@ -0,0 +1,29 @@
...
+#---MyTest.test--------------------- Executable------------------#
+#BEGIN_COMMENT
+# Test the ARM R_ARM_CALL relocation.
+#END_COMMENT
+#START_TEST
+#RUN: %run_cc -o %t1.o -c %p/Inputs/1.s
+# Partial
+#RUN: %run_cc -o %t1.o %t1.o -r
+#RUN: llvm-readelf -r %t1.o | %filecheck %s -check-prefix RELOCS
```
quic-areg: The `readelf` and `objdump` checks belong in traditional static lit tests rather than runtime tests. Runtime tests should only check runtime behavior and the two test types should not overlap, imo.

test/ARM/RunTests/CallReloc/Inputs/1.s
Steven6798: Sounds good. I'll use C code as much as I can and assembly only when needed.

test/ARM/RunTests/CallReloc/CallReloc.test
Steven6798: Thanks!

test/Common/RunTests/MyTest/MyTest.test
```diff
@@ -0,0 +1,29 @@
...
+#RUN: %run %t1.out | %filecheck %s
+# Dynamic
+#RUN: %run_cc -o %t1.out %t1.o -Wl,-dy
+#RUN: llvm-objdump -d --no-show-raw-insn %t1.out | FileCheck %s -check-prefix DYN_RELOCS
+#RUN: %run %t1.out | %filecheck %s
+
+#RELOCS: Relocation section '.rel.text'
+#RELOCS: {{.*}} R_ARM_CALL
+#STATIC_RELOCS: bl	0x8049158 <puts>
+#DYN_RELOCS: bl	0x3a4 <__cxa_finalize@plt+0xc>
```
quic-areg: These hard-coded address checks may be brittle (but I don't think they belong in runnable tests)

test/Common/RunTests/MyTest/MyTest.test
```diff
@@ -0,0 +1,29 @@
+REQUIRES: run_test
+#---MyTest.test--------------------- Executable------------------#
+#BEGIN_COMMENT
+# Test the ARM R_ARM_CALL relocation.
+#END_COMMENT
+#START_TEST
+#RUN: %run_cc -o %t1.o -c %p/Inputs/1.s
+# Partial
```
quic-areg: I don't see what the partial link step achieves here.

test/Common/RunTests/MyTest/Inputs/1.s
```diff
@@ -0,0 +1,54 @@
+	.syntax	unified
```
quic-areg: This input might be more readable as 'C'. If assembly is kept, it can be greatly simplified. Test inputs, whether C or asm, should be minimal. Eg these sections are not required:

```
.ident	"clang version 23.0.0git (https://github.com/llvm/llvm-project.git 832285ab7dffc8381ae76182b9b1f0f4cbae23d0)"	
.section	".note.GNU-stack","",%progbits	.addrsig
```

docs/DeveloperDocs/ARM/relocations.md
```diff
@@ -0,0 +1,15 @@
+# ARM Relocations
```
quic-areg: I do not see the point of the docs. We are duplicating the ARM ABI which is already well-documented [here](https://github.com/ARM-software/abi-aa/blob/main/aaelf32/aaelf32.rst). I don't think our documentation will be any better than the official docs, nor will it be maintained as well. 

This doc so far contains general ARM facts, not eld specifics. Official ARM sources are probably a better source for this information.

quic-areg (commented):
In general, I think we should prefer C inputs for runnable tests, get rid of the docs if they are going to just parrot the ABI, and separate the types of checks we make in static vs runtime tests.

test/Common/RunTests/MyTest/MyTest.test
Steven6798: Yes, that was a placeholder for my initial testing. I forgot to update that. It's fixed. Thanks!

test/Common/RunTests/MyTest/MyTest.test
Steven6798: Got it. Thanks!

test/Common/RunTests/MyTest/MyTest.test
Steven6798: They were indeed brittle but will be removed as you said. Thanks!

test/Common/RunTests/MyTest/Inputs/1.s
Steven6798: Thanks. I'll use C as much as I can and only use assembly when needed. In those corner cases I'll clean the assembly file as much as possible.

docs/DeveloperDocs/ARM/relocations.md
Steven6798: I agree. @quic-seaswara thoughts?

quic-seaswara: I am thinking that to use a single point of reference to show what relocations are implemented by eld, there are a lot of relocations, that we have not seen. This is useful for code review and any changes that we need to make to support bug fixes. 

We should definitely look to add the eld specifics to the relocation, number of bits, does it create a PLT or no PLT for shared objects and PLT/no PLT for PIE's etc.

quic-seaswara:
Rename the test to R_ARM_CALL.test

Steven6798:
> Rename the test to R_ARM_CALL.test

Sounds good. Thanks!

docs/DeveloperDocs/ARM/relocations.md
Steven6798: That make sense. I'll only add ELD specific information and remove any generic information documented in the ARM ABI.

test/Common/RunTests/MyTest/MyTest.test
Steven6798: Just testing that the relocation was still there after partial link. But as you said, these test don't belong in runnable tests. Thanks!

quic-seaswara: taking the output of partial links and building an executable / shared object and running it belongs, if that is what you were looking to do.

Steven6798: Got it. Then I'll keep the partial link and remove the additional checks so we can still test partial linking. Thanks!

quic-seaswara: you need the partial linked object file to be used in a shared library / executable link and run it. This will ensure that eld did not break some assumption in partial link.

docs/DeveloperDocs/ARM/relocations.md
Steven6798: So far, I don't see anything eld specific for R_ARM_CALL. It looks like it is fully supported. Thoughts?

Steven6798:
I addressed the issues. Let me know if I missed something or if I can improve it even more.

quic-areg (commented):
LGTM mostly.

For future PRs, if there is no existing static test for the relocation it might be a good idea to add one in addition to the runtime test. `R_ARM_CALL` is already well tested, so that does not apply here, but it might for future relocations.

The doc looks better than before imo, but I still think it adds little value. Checking the list of supported relocations for a reviewer or developer was already a pretty trivial task (they are all defined in *RelocationFunctions.h) , and I worry that the doc risks going stale.

test/ARM/RunTests/R_ARM_CALL/R_ARM_CALL.test
```diff
@@ -0,0 +1,21 @@
+REQUIRES: run_test
+#---R_ARM_CALL.test--------------------- Executable------------------#
+#BEGIN_COMMENT
+# Test the ARM R_ARM_CALL relocation.
+#END_COMMENT
+#START_TEST
+#RUN: %run_cc -o %t1.o -c %p/Inputs/1.c
```
quic-seaswara: can we establish that this generates a R_ARM_CALL relocation ?

quic-seaswara: compile with -fPIC and -fpic options

test/ARM/RunTests/R_ARM_CALL/R_ARM_CALL.test
```diff
@@ -0,0 +1,21 @@
...
+# Partial
+#RUN: %run_cc -o %t1.o %t1.o -r
+# Static
+#RUN: %run_cc -o %t1.out %t1.o -static
+#RUN: %run %t1.out | %filecheck %s
```
quic-seaswara: This brings in all the libraries etc along with testing for a reloc, is this designed as intended ? 

The sysroot might bring in tons of relocations such as R_ARM_CALL already.

docs/DeveloperDocs/ARM/relocations.md
```diff
@@ -0,0 +1,9 @@
...
+
+Information about ARM relocations can be found here: https://github.com/ARM-software/abi-aa/blob/main/aaelf32/aaelf32.rst
+
+## Status
+
+| Name       | Status          |
+|------------|-----------------|
+| R_ARM_CALL | Fully supported |
```
quic-seaswara: We need eld specifics here. 

Total bits available in the relocation, when does the linker create trampolines for branch range overflows, does it create a PLT for shared library links, how does it work for local symbols or non premptible symbols.

quic-seaswara (changes requested):

test/ARM/RunTests/R_ARM_CALL/R_ARM_CALL.test
Steven6798: I can add the llvm-readelf again to check for the relocation like I was doing before. Is that enough?

Steven6798: Sounds good

test/ARM/RunTests/R_ARM_CALL/R_ARM_CALL.test
Steven6798: In theory we would like to isolate this test as much as possible. The c code uses libraries, so we need to include them and also test static linking. Is there a better way to do this without including so much code?

quic-seaswara: https://dev.to/ripan030/my-first-bare-metal-program-from-reset-to-hello-4cn8 is an example

Steven6798:
I'm making progress. I'll leave the documentation for another issue to focus on 1 thing at a time. I improved the testing coverage of this relocation. I also started using Bare-Metal runs as much as possible to isolate the tests.

Steven6798:
I also updated the lit.cfg to include qemu-system-target to use Bare-Metal runs

.github/workflows/ci.yml
```diff
@@ -129,6 +129,12 @@ jobs:
         working-directory: pr-${{ env.PR_NUMBER }}/obj
         run: ccache --evict-older-than 7200s --dir=$CCACHE_DIR
 
+      - name: Install dependencies
+        shell: bash
+        run:  |
+          sudo apt update
+          sudo apt install -y libsdl2-2.0-0
```
parth-07, line 136: Why do we need this?

test/lit.cfg
```diff
@@ -214,10 +214,12 @@ dirname = 'dirname'
 datalayout = ''
 link_path = which(link)
 run=""
+run_sys=""
 run_cc = ''
 run_cxx = ''
 sysroot = ''
 emulator = ''
+emulator_system = ''
```
parth-07, line 222: Can you please explain why do we need system emulator? For which cases linux emulator is not enough?

test/ARM/RunTests/R_ARM_CALL/Inputs/hidden_callee.c
```diff
@@ -0,0 +1,22 @@
+// Hidden symbol: non-preemptible, direct relocation, no PLT entry.
+#define UART0_DR ((volatile unsigned int *)0x10009000)
```
parth-07, line 2: Can we use printf and use non-system qemu emulator?

.github/workflows/ci.yml
Steven6798: I'm trying to use the qemu system emulator. the machine that runs the CI workflow is missing a library. Shakti suggested adding it to the workflow but clearly that did not work. If you have any suggestions about how to address this let me know.

test/lit.cfg
Steven6798: Shankar suggested using the system emulator to isolate the testing as much as possible. For example, when static linking a lot gets included in the binary, with this approach, only a minimal amount of code is included. What do you think?

test/ARM/RunTests/R_ARM_CALL/Inputs/hidden_callee.c
Steven6798: We are discussing using qemu system in another comment so I'll close this one.

test/lit.cfg
parth-07: @quic-seaswara Hi Shankar, I am concerned that writing baremetal code and using system emulator requires more boilerplate and makes testing more difficult without providing any apparent benefit. 

> to isolate the testing as much as possible

We are only removing the libc dependency -- all other dependency are still there, and we are unlikely to encounter bugs/corner-cases in libc when doing extended testing of different relocations computations because we won't be using any advanced feature of libc.

> For example, when static linking a lot gets included in the binary, with this approach, only a minimal amount of code is included. 

I agree with this, but I am not sure if less size is worth the benefit here. We can always delete the test artifacts after the testing.

Steven6798:
The way the ARM testing is going to be done changed, I will close this PR and open a new one.

PR #1226, opened 2026-05-28
Simplify emit relocs support

quic-seaswara:
Modularize and add helper functions to support emit relocs.


quic-seaswara:
Relanding the patch after the fix to https://github.com/qualcomm/eld/pull/1192

PR #1241, opened 2026-06-03
ELFReaderBase::Create now use input file format

### LINKED ISSUE #1231: ELFReader::Create should not use target configuration
The `ELFReader::Create` function selects the reader implementation based on the target configuration rather than the actual properties of the input file. This leads to incorrect reader instantiation when the input file and configuration do not match.

## Why this is wrong

The reader must reflect the actual format of the input file, not the target configuration.

If there is a mismatch between:

- input file properties (e.g., ELF64 BE)
- target configuration (e.g., ELF32 LE)

the current implementation will instantiate a reader based on the configuration rather than the file, resulting in:

- incorrect parsing behavior
- potential runtime failures or undefined behavior

## Expected Behavior

`ELFReader::Create` should:

- Inspect the `InputFile`
- Determine its actual format (endianness, bit width, etc.)
- Instantiate the appropriate `ELFReader` specialization that matches the input file

The target configuration should not influence reader selection.

Steven6798:
ELF files are now read based on their real format (bit width and endianness) instead of the configured target.

Previously, mismatches between the input file and target configuration could cause files to be interpreted incorrectly, leading to parsing errors or unexpected failures.

With this change, ELF files are consistently handled according to their true format, improving reliability and preventing incorrect reads when configurations do not match the input.

Fix qualcomm#1231

### DIFF
--- a/include/eld/Readers/ELFReaderBase.h
+++ b/include/eld/Readers/ELFReaderBase.h
@@ -93,11 +93,9 @@ public:
 
   /// Creates and returns an instance of a reader class.
   ///
-  /// The reader class depends on the input type and the target
-  /// configuration. For example, if an input is a dyanmic input library
-  /// and the target configuration is little-endian 32-bits architecture,
-  /// then the reader class would be
-  /// DynamicELFReader<llvm::object::ELF32LE>.
+  /// The reader class depends on the input type. For example, if an input is a
+  /// dynamic input library and a little-endian 32-bits architecture, then the
+  /// reader class would be DynamicELFReader<llvm::object::ELF32LE>.
   static eld::Expected<std::unique_ptr<ELFReaderBase>>
   Create(Module &module, InputFile &inputFile);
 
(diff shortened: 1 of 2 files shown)


quic-seaswara (approved):

parth-07:
I am not sure what are we solving here.

> Previously, mismatches between the input file and target configuration could cause files to be interpreted incorrectly, leading to parsing errors or unexpected failures.

Wouldn't it be better to error out early if there is a mismatch between input file and the target configuration?

Steven6798:
> I am not sure what are we solving here.
> 
> > Previously, mismatches between the input file and target configuration could cause files to be interpreted incorrectly, leading to parsing errors or unexpected failures.
> 
> Wouldn't it be better to error out early if there is a mismatch between input file and the target configuration?

We can add a check here for --[no]-warn-mismatch if needed but still this function must create the reader based on the file, not the configuration.

quic-areg (approved):

quic-seaswara (approved):

PR #1247, opened 2026-06-04
[linkerscript] Add column numbers for linker script diagnostics

quic-seaswara:
Also update tests that linker-script diagnostic column expectations


test/Common/standalone/linkerscript/UnclosedQuoteTest/UnclosedQuote.test
```diff
@@ -7,6 +7,6 @@
 RUN: %clang %clangopts -c %p/Inputs/1.c -o %t1.1.o
 RUN: %not %link %linkopts %t1.1.o -T %p/Inputs/script.t -o %t2.out 2>&1 | %filecheck %s
 
-CHECK: Error: {{.*}}script.t:3: unclosed quote
-CHECK-COUNT-1: unclosed quote
```
quic-areg, line 11: Why did this change?

quic-seaswara: the script parser errors have been centralized to this function `setError` now. Previously the linker would continue even after producing this error, now will stop.

lib/ScriptParser/ScriptLexer.cpp
```diff
@@ -67,12 +67,24 @@ StringRef ScriptLexer::getLine() const {
...
 }
 
-std::string ScriptLexer::getCurrentLocation() const {
-  std::string Filename = std::string(getCurrentMB().getBufferIdentifier());
-  return (Filename + ":" + Twine(PrevTokLine)).str();
+std::string ScriptLexer::getCurrentLocation(
+    std::optional<llvm::StringRef> columnTok) const {
+  size_t lineNumber = PrevTokLine;
+  size_t columnNumber = 0;
+  size_t displayColumn = 1;
```
quic-areg: should this default to 1? It could be misleading when the error is actually somewhere else and we use a placeholder of 1.

test/Common/standalone/linkerscript/UnclosedQuoteTest/UnclosedQuote.test
quic-areg: It might be worth mentioning in the commit msg/pr description that this cascading diagnostics issue was fixed.

lib/ScriptParser/ScriptLexer.cpp
quic-seaswara: so far found that the error is shown correctly and the caret points to the column. Initializing to 1 does make sense to me.

lib/ScriptParser/ScriptLexer.cpp
```diff
@@ -67,12 +67,24 @@ StringRef ScriptLexer::getLine() const {
...
+  {
```
quic-areg: does this block scope do anything?

lib/ScriptParser/ScriptLexer.cpp
```diff
@@ -92,39 +104,40 @@ bool ScriptLexer::diagnose() const {
   return ThisConfig.getDiagEngine()->diagnose();
 }
 
+std::string ScriptLexer::formatDiagnosticMessage(
+    const Twine &Msg, std::optional<llvm::StringRef> columnTok) const {
+  std::string message = (getCurrentLocation(columnTok) + ": " + Msg).str();
+  {
```
quic-areg: same here. The braces look unnecessary

quic-seaswara: just scoped the variables around it, so that we know what variables we don't need after the scope.

lib/ScriptParser/ScriptLexer.cpp
quic-seaswara: just scoped the variables around it, so that we know what variables we don't need after the scope.

lib/ScriptParser/ScriptLexer.cpp
quic-seaswara: we can remove it too since this is not followed in the rest of the code.

lib/ScriptParser/ScriptLexer.cpp
quic-areg: The column isn't always known, and in cases when it's not known the placeholder value of 1 is likely to be wrong:
```
$ cat script.t
                   /* unterminated
SECTIONS { .text : { *(.text) } }

$ ld.eld -T script.t ...
Error: script.t:1:1: unclosed comment in a linker script
```
The /* is at column 20, but the diagnostic says column 1. My suggestion was to omit the column number when we don't know it and point the caret to the start of the line.

quic-seaswara: Good catch!. Thank you for the test!

test/Common/standalone/linkerscript/UnclosedQuoteTest/UnclosedQuote.test
quic-seaswara: Done

lib/ScriptParser/ScriptLexer.cpp
quic-seaswara: Done and updated diagnostics.

lib/ScriptParser/ScriptLexer.cpp
quic-seaswara: Done

lib/ScriptParser/ScriptLexer.cpp
quic-seaswara: Done

Steven6798 (approved):
Nice addition to diagnostics.

PR #1248, opened 2026-06-04
[workflow] Use latest stable clang-format

quic-seaswara:


lenary (commented):
This seems supremely complicated.

Why can't we just have a variable that says "latest stable is 22", and then bump it manually after a release?

lenary:
Oh it's already merged. Fair enough

quic-seaswara:
> This seems supremely complicated.
> 
> Why can't we just have a variable that says "latest stable is 22", and then bump it manually after a release?

When LLVM moves to the next release, it just moves on.

PR #1253, opened 2026-06-08
[ARM/Baremetal] Remove --compact and -z compactdyn options

quic-seaswara:
--compact (ARM-specific) produced smaller ELF images by incrementing file offsets by section size rather than virtual-address difference, requiring the loader to satisfy a strict LMA/offset congruence invariant.  -z compactdyn suppressed DT_PLTGOT and DT_DEBUG from the dynamic section.

Both options add meaningful complexity to the layout and dynamic-section code paths, are supported only in a narrow set of environments, and are no longer required.


.github/workflows/clang-format-pr.yml
```diff
@@ -17,7 +17,8 @@ jobs:
 
       - name: Install clang-format
         run: |
-          set -euo pipefail
+          set -uo pipefail
```
quic-areg: unrelated

quic-seaswara: mainly for debugging this failure, related to clang-format.  will make this as a seperate PR.

quic-seaswara:
part of #960

PR #1258, opened 2026-06-09
[RISCV] Error out when relax-tbljal is mixed with shared/pie

svs-quic:
The `RISCV Zcmt` extension is not compatible with shared libraries or when position independent code is enabled.


quic-seaswara (approved):

lenary (approved):
LGTM

Issue #1098, opened 2026-04-24
Add support for --no-warn-mismatch


quic-seaswara:
The --no-warn-mismatch option is intended to suppress warnings emitted by the linker when input files are incompatible or mismatched, while otherwise preserving normal link semantics.
In particular, this option disables diagnostics that warn about:

- Object files built for different architectures, ABIs, endianness, or ELF classes

- Other forms of input incompatibility that would normally produce a warning during the link step

The option does not make incompatible inputs linkable and does not change error handling.

Expected behavior

Without --no-warn-mismatch:
The linker emits a warning indicating mismatched or incompatible input files.

With --no-warn-mismatch:
The same warning is suppressed, but the link result (success or failure) is otherwise unchanged.

Steven6798:
I'm working on this.

Steven6798:
I noticed that this warning currently supports three options: None, WarnMismatch, and NoWarnMismatch. As part of addressing this issue, I’m considering removing the None option and converting this into a simple Boolean setting to streamline and simplify the configuration.

@parth-07 @quic-areg I would appreciate some feedback.

quic-areg:
> I noticed that this warning currently supports three options: None, WarnMismatch, and NoWarnMismatch. As part of addressing this issue, I’m considering removing the None option and converting this into a simple Boolean setting to streamline and simplify the configuration.
> 
> [@parth-07](https://github.com/parth-07) [@quic-areg](https://github.com/quic-areg) I would appreciate some feedback.

Currently, different backends have different `WarnMismatch` defaults, so the enum is necessary: `None` lets the target backend choose the default. I think when this is fully implemented all the backends should have the same default policy and the option collapsed to a boolean.

Steven6798:
> Currently, different backends have different `WarnMismatch` defaults, so the enum is necessary: `None` lets the target backend choose the default. I think when this is fully implemented all the backends should have the same default policy and the option collapsed to a boolean.

Thanks, that makes sense. What are your thoughts on the default behavior? I tend to prefer emitting as much information as possible, especially for warnings and errors, since it can help catch issues earlier. Curious to hear your perspective.

quic-seaswara:
default behavior should be None. Any warnings would be promoted to fatal errors with --fatal-warnings. So we should be careful when setting defaults.

Steven6798:
> default behavior should be None. Any warnings would be promoted to fatal errors with --fatal-warnings. So we should be careful when setting defaults.

[...]

PR #1260, opened 2026-06-09
Add support for --no-warn-mismatch for ABI information

### LINKED ISSUE #1215: Add support for --no-warn-mismatch for ABIs
Check for compatibility between object files using ABI information.

Steven6798:
Allow users to suppress warnings about incompatible or mismatched input files using the --no-warn-mismatch option. This change disables mismatch diagnostics while preserving normal link behavior and outcomes.

Fix: qualcomm#1215

### DIFF
--- a/lib/Target/RISCV/RISCVInfo.cpp
+++ b/lib/Target/RISCV/RISCVInfo.cpp
@@ -45,6 +45,9 @@ bool RISCVInfo::isABIFlagSet(uint64_t inputFlag, uint32_t ABIFlag) const {
 }
 
 bool RISCVInfo::isCompatible(uint64_t pFlag, const std::string &pFile) const {
+  if (m_Config.options().noWarnMismatch())
+    return true;
+
   assert(m_OutputFlag);
   if (isABIFlagSet(pFlag, llvm::ELF::EF_RISCV_FLOAT_ABI_SOFT) ^
       isABIFlagSet(*m_OutputFlag, llvm::ELF::EF_RISCV_FLOAT_ABI_SOFT)) {
--- a/test/ARM/standalone/AttributesWarn/AttributesWarn.test
+++ b/test/ARM/standalone/AttributesWarn/AttributesWarn.test
@@ -7,5 +7,7 @@
 #RUN: %clang %clangopts -c -fshort-enums %p/Inputs/foo.c -o %t1.o
 #RUN: %clang %clangopts -c -fno-short-enums %p/Inputs/bar.c -o %t2.o
 #RUN: %link --warn-mismatch -o %t1.out %linkopts %t1.o %t2.o 2>&1 | %filecheck %s --check-prefix="ENUM"
+#RUN: %link --no-warn-mismatch -o %t1.out %linkopts %t1.o %t2.o 2>&1 | %filecheck %s --check-prefix="ENUM_NOERR" --allow-empty
 #
 #ENUM:  Warning: the size of enumerated data item
+#ENUM_NOERR-NOT:  Warning: the size of enumerated data item
--- a/test/ARM/standalone/PCSAttrs/PCSAttr.test
+++ b/test/ARM/standalone/PCSAttrs/PCSAttr.test
@@ -8,12 +8,14 @@
 RUN: %clang %clangopts -fno-pie -march=armv7 -fropi -c %p/Inputs/1.c -o %t1.o
 RUN: %clang %clangopts -fno-pie -march=armv7 -frwpi -c %p/Inputs/2.c -o %t2.o
 RUN: %not %link --warn-mismatch %linkopts -Map %t1.map -march armv7 %t1.o %t2.o -o %t2.out 2>&1 | %filecheck %s
+RUN: %link --no-warn-mismatch %linkopts -Map %t1.map -march armv7 %t1.o %t2.o -o %t2.out 2>&1 | %filecheck %s -check-prefix=NOERR --allow-empty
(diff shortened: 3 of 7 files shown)


Steven6798:
There is an issue with the riscv-attribute tests. These tests fail when no-warn-mismatch is used even though gnu ld links successfully. I narrow it down to the "RISCVAttributeFragment::mergeArch" function. I will raise an issue bout this so the team can address it.

lib/Target/RISCV/RISCVInfo.cpp
```diff
@@ -46,32 +46,37 @@ bool RISCVInfo::isABIFlagSet(uint64_t inputFlag, uint32_t ABIFlag) const {
 
 bool RISCVInfo::isCompatible(uint64_t pFlag, const std::string &pFile) const {
   assert(m_OutputFlag);
-  if (isABIFlagSet(pFlag, llvm::ELF::EF_RISCV_FLOAT_ABI_SOFT) ^
-      isABIFlagSet(*m_OutputFlag, llvm::ELF::EF_RISCV_FLOAT_ABI_SOFT)) {
+  if (!m_Config.options().noWarnMismatch() &&
```
quic-areg: can we have this check once and exit at the top of the function? 
```c++
if (m_Config.options().noWarnMismatch())
  return true;
```

Steven6798: Of course. Thanks!

quic-seaswara (commented):
The RISC-V ABI specifies that these errors below are detected and raised.

Do we want to override this by --no-warn-mismatch ?

https://riscv-non-isa.github.io/riscv-elf-psabi-doc/#_file_header

Steven6798:
> The RISC-V ABI specifies that these errors below are detected and raised.
> 
> Do we want to override this by --no-warn-mismatch ?
> 
> https://riscv-non-isa.github.io/riscv-elf-psabi-doc/#_file_header

Thanks for letting me know. I'll move the PR to draft and think about this.

lenary:
Still reading up on what the intentions are here, which are still a bit unclear to me.

IIRC, we can only ever warn for attribute mismatches in `.riscv.attributes`, because it may just be that the linker doesn't know about some extensions and whether they are compatible or not.

As for RISC-V e_flags, these are cases where it is clearer what is compatible, but I think the intention is to be able to ignore this based on what @quic-areg was writing in #1098 (linking soft float to hard float when the users know there is no fp use).

I guess this says "don't even emit the warning", which is most useful when warnings are promoted to errors.

So this seems fine, though I don't really condone its use, I do see how we need it for compatibility.

Steven6798:
> So this seems fine, though I don't really condone its use, I do see how we need it for compatibility.

Thanks for the feedback.

quic-areg:
Current changes LGTM.

GNU ld's `--no-warn-mismatch` has a larger scope than this PR covers (ARM CPU arch profile, RISCV attribute parse failures, attribute mismatches, etc). We should either match the same scope or specify where we intentionally diverge.

quic-seaswara:
The description of the --no-warn-mismatch feature in the GNU documentation is quite vague, which makes it difficult to fully understand the expected behavior and implications.

While I am fine with this patch as-is, I am concerned that we may be suppressing diagnostics that could be important for users.

I am generally in favor of preserving diagnostics. Instead of completely suppressing these warnings, could we consider downgrading them to notes when --no-warn-mismatch is specified?

This approach would:

1. Still provide visibility into potential mismatches
2. Help users remain aware of issues without being overly disruptive

Steven6798:
> The description of the --no-warn-mismatch feature in the GNU documentation is quite vague, which makes it difficult to fully understand the expected behavior and implications.

This flag is one of those things that do not follow the norm. Even the name mentions warning even though it deals with errors and changes linking behavior. It is more of a special case.

> I am generally in favor of preserving diagnostics. Instead of completely suppressing these warnings, could we consider downgrading them to notes when --no-warn-mismatch is specified?

We could add it to --verbose. That way it is still visible but with a way to hide it. What do you think?

quic-seaswara:
> > The description of the --no-warn-mismatch feature in the GNU documentation is quite vague, which makes it difficult to fully understand the expected behavior and implications.
> 
> This flag is one of those things that do not follow the norm. Even the name mentions warning even though it deals with errors and changes linking behavior. It is more of a special case.
> 
> > I am generally in favor of preserving diagnostics. Instead of completely suppressing these warnings, could we consider downgrading them to notes when --no-warn-mismatch is specified?
> 
> We could add it to --verbose. That way it is still visible but with a way to hide it. What do you think?

verbose currently is too verbose. I would prefer adding as a note.

Steven6798:
> verbose currently is too verbose. I would prefer adding as a note.

Sounds good.

quic-areg:
Would that require two copies of every diagnostic? one an error/warning and the other a note?

quic-seaswara:
> Would that require two copies of every diagnostic? one an error/warning and the other a note?

Good point. May be we can have a function to override or downgrade the severity from a warning to a note by explicitly requesting, and reuse the same diag. 

For now, we can duplicate the diag that fall under the no warn mismatch category to a note. 

In any case both of the points above can be treated as a follow up.

quic-seaswara (approved):

Steven6798:
> In any case both of the points above can be treated as a follow up.

I'll raise an issue. Thanks for the feedback.

quic-areg (approved):

quic-seaswara (approved):

PR #1262, opened 2026-06-10
docs: add ELD code coverage guide

bsaharsh:
Added a step by step code coverage guide for ELD using llvm-cov. It covers environment setup, build configuration and the full llvm cov pipeline in 4 steps. Updated CMakeLists.txt to use ELD_COVERAGE cmake option to default to llvm cov. Currently the percentages are as below
Line:     82.73%
Function: 78.07%
Region:   79.37%
Branch:   76.39%

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -217,9 +217,9 @@ include_directories(
   ${CMAKE_CURRENT_SOURCE_DIR}/include
   ${CMAKE_CURRENT_SOURCE_DIR}/include/eld/PluginAPI/)
 
-option(ELD_COVERAGE "Build ELD with coverage" OFF)
+option(ELD_COVERAGE "Build ELD with LLVM source-based coverage instrumentation" OFF)
 if(ELD_COVERAGE)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} --coverage")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fprofile-instr-generate -fcoverage-mapping")
 endif()
 
 option(ELD_SANITIZE "Build ELD with Sanitizer" OFF)
--- a/docs/DeveloperDocs/EldCodeCoverage.md
+++ b/docs/DeveloperDocs/EldCodeCoverage.md
@@ -1,0 +1,113 @@
+# ELD Code Coverage Guide
+
+## LLVM Native Coverage (llvm-cov source-based)
+
+This pipeline uses Clang's source-based coverage instrumentation, which provides finer-grained (expression-level) coverage data.
+
+> **Toolchain requirement**: This pipeline requires the **LLVM/Clang toolchain** (`clang`/`clang++`). It is not compatible with GCC or the GNU toolchain.
+
+The pipeline relies on two artifact types:
+
+- **`.profraw`** — Generated at **runtime** by each executed binary. Contains raw profile counters per function region.
+- **`.profdata`** — Merged from one or more `.profraw` files using `llvm-profdata merge`. This is the indexed form that `llvm-cov` consumes.
+
+## Prerequisites
+
+Ensure `llvm-cov` and `llvm-profdata` are available:
+
+```bash
+llvm-cov --version
+llvm-profdata --version
+```
+Export the paths of clang and clang++
+
+```bash
+export CC=clang
+export CXX=clang++
+```
+
+---
+
+## Code Coverage Build
+
+Note:
+- We use the CMake option `ELD_COVERAGE=ON` which enables the coverage flags (`-fprofile-instr-generate -fcoverage-mapping`).
+- `CMAKE_BUILD_TYPE=Release` can also be used for source-based coverage — Clang's instrumentation is inserted before the optimizer runs and is immutable, so optimizations do not affect report accuracy. See [Clang source-based coverage docs](https://clang.llvm.org/docs/SourceBasedCodeCoverage.html).
+
+```bash
+cmake -G Ninja \
+  -DCMAKE_BUILD_TYPE=Debug \
+  -DLLVM_ENABLE_ASSERTIONS:BOOL=ON \
+  -DLLVM_ENABLE_PROJECTS='llvm;clang' \
+  -DLLVM_EXTERNAL_PROJECTS=eld \
+  -DLLVM_EXTERNAL_ELD_SOURCE_DIR=${PWD}/llvm-project/eld \
+  -DLLVM_TARGETS_TO_BUILD='ARM;AArch64;RISCV;Hexagon;X86' \
+  -DCMAKE_CXX_FLAGS='-stdlib=libc++' \
+  -DLLVM_USE_LINKER=lld \
+  -DELD_ENABLE_SYMBOL_VERSIONING=ON \
+  -DELD_ENABLE_RUN_TESTS=ON \
+  -DELD_COVERAGE=ON \
+  -B ./obj \
+  -S ./llvm-project/llvm
+```
+
+Build:
+
+```bash
+cmake --build obj
+```
+
+---
+
+## Generating Code Coverage Report
+
+### Step 1 — Run Tests with Profile Output
+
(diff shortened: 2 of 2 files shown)


docs/DeveloperDocs/EldCodeCoverage.md
```diff
@@ -0,0 +1,194 @@
...
+- **`.gcno` (Graph Notes)** — Generated at **compile time**. Contains the control flow graph of the program used to map coverage data back to source lines.
+- **`.gcda` (Graph Data)** — Generated at **runtime** when instrumented binary is executed. Contains the actual hit counts for lines, functions, and branches.
+
+---
+
+## Prerequisites
+
+Before starting, ensure lcov is available. If not, install it:
+- Install on Debian/Ubuntu: `sudo apt-get install lcov`
+- genhtml comes bundled with the lcov package.
```
quic-seaswara: Can we use llvm-cov instead of relying on lcov package ?

quic-seaswara (changes requested):

docs/DeveloperDocs/EldCodeCoverage.md
bsaharsh: If we want to drop the lcov package then we may proceed in two ways:
1. Using the Gcov-Compatible Mode (llvm-cov gcov): According to the llvm-cov documentation, this subcommand specifically takes a SOURCEFILE as an argument, making it a file-level tool. It lacks a project-wide command (like lcov --directory) to automatically scan and aggregate coverage across the entire build. To obtain project-wide metrics, a custom script would be required to aggregate the individual file results.
2. Using the Native LLVM Pipeline: The documentation suggests a native pipeline that avoids the gcov format entirely. This approach requires building the project with the -fprofile-instr-generate and -fcoverage-mapping flags. The LLVM_PROFILE_FILE environment variable is used at runtime to produce raw profile data, which is then merged into a .profdata file using llvm-profdata. But this pipeline is not GNU Gcov compatible.

quic-seaswara: Prefer Native LLVM.

bsaharsh: I have updated the document and CMakeLists.txt, now it uses the LLVM native pipeline.

CMakeLists.txt
```diff
@@ -222,6 +222,14 @@ if(ELD_COVERAGE)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} --coverage")
 endif()
 
+option(ELD_LLVM_COV "Build ELD with LLVM source-based coverage instrumentation" OFF)
+if(ELD_LLVM_COV)
+  if(ELD_COVERAGE)
```
quic-seaswara: Can we just have ELD_COVERAGE and default to llvm cov.

quic-seaswara (changes requested):

CMakeLists.txt
bsaharsh: Yes, I have removed the --coverage entirely and just used the below
option(ELD_COVERAGE "Build ELD with LLVM source-based coverage instrumentation" OFF)
if(ELD_COVERAGE)
  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fprofile-instr-generate -fcoverage-mapping")
endif()

I have run the pipeline, the results are the same. Are there any other changes which you would recommend?

bsaharsh:
@quic-seaswara Please Have a look I have updated the CMakeLists.txt.

docs/DeveloperDocs/EldCodeCoverage.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Code Coverage Build
+
+Note:
+- We use the CMake option `ELD_COVERAGE=ON` which enables the coverage flags (`-fprofile-instr-generate -fcoverage-mapping`).
+
+```bash
+cmake -G Ninja \
+  -DCMAKE_BUILD_TYPE=Debug \
+  -DCMAKE_INSTALL_PREFIX=../inst \
```
quic-seaswara: Can we remove the install ?

docs/DeveloperDocs/EldCodeCoverage.md
```diff
@@ -0,0 +1,127 @@
...
+  -DLLVM_ENABLE_ASSERTIONS:BOOL=ON \
+  -DLLVM_ENABLE_PROJECTS='llvm;clang' \
+  -DLLVM_EXTERNAL_PROJECTS=eld \
+  -DLLVM_EXTERNAL_ELD_SOURCE_DIR=${PWD}/llvm-project/eld \
+  -DLLVM_TARGETS_TO_BUILD='ARM;AArch64;RISCV;Hexagon;X86' \
+  -DCMAKE_CXX_FLAGS='-stdlib=libc++' \
```
quic-seaswara, line 45: enable symbol versioning, and also enable run tests

docs/DeveloperDocs/EldCodeCoverage.md
```diff
@@ -0,0 +1,127 @@
...
+  --object obj/lib/libLW.so \
+  --ignore-filename-regex='llvm-project/(llvm|clang)/|third-party/|/obj/tools/[^e]|/obj/tools/e[^l]|/obj/tools/el[^d]|/obj/(include|lib|bin)/|llvm-project/eld/test/|llvm-project/eld/Plugins/HelloWorldPlugin/')
+
+TOTAL_LINE=$(echo "$COV_SUMMARY" | grep '^TOTAL')
+echo "$TOTAL_LINE" #To get the final line which has the results
+
+REGION_COV=$(echo "$TOTAL_LINE" | awk '{print $4}')
+FUNC_COV=$(echo   "$TOTAL_LINE" | awk '{print $7}')
+LINE_COV=$(echo   "$TOTAL_LINE" | awk '{print $10}')
+BRANCH_COV=$(echo "$TOTAL_LINE" | awk '{print $13}')
```
quic-seaswara: Why this ?

quic-seaswara (changes requested):

docs/DeveloperDocs/EldCodeCoverage.md
bsaharsh: Yes, we can remove the install, updated.

docs/DeveloperDocs/EldCodeCoverage.md
bsaharsh: Updated.

docs/DeveloperDocs/EldCodeCoverage.md
bsaharsh: llvm-cov report command writes only to stdout — there is no -o flag to save output. So due to this I have saved the report using the COV_SUMMARY , which contains the final summary line so piped it to extract with the awk commands. We can use the export command which emits json output, we make a json file for it. What do you suggest?

bsaharsh:
@quic-seaswara Please have a look, I have updated the Markdown File.

parth-07 (approved):
LGTM.

docs/DeveloperDocs/EldCodeCoverage.md
quic-seaswara: `-format=html -output-dir=coverage_report`

quic-seaswara (changes requested):

docs/DeveloperDocs/EldCodeCoverage.md
bsaharsh: Hi @quic-seaswara , we already generate the html file with the show command which generates the line by line coverage , it has source annotation, it has per file html generated under the coverage sub directory and index.html is the summary of each file. The report command gives out the summary table, which is almost the same as the generated summary (index.html) in the coverage-report directory, mainly I have used it to get a quick look on numbers without opening up the browser to check. We can remove all the commands from TOTAL_LINE , 116 and just store it in COV_SUMMARY for the documentation.

quic-seaswara: Let us use the coverage report as the source of truth. No need to get the numbers directly. We dont have a use case for it.

quic-seaswara (approved):

PR #1263, opened 2026-06-10
Make symbol table order deterministic

### LINKED ISSUE #1261: fix non-deterministic order of entries in .dynsym

quic-areg:
Multithreaded relocation scanning made the order in which symbols were appended to the module symbol vector dependent on thread timing. Both .dynsym and .symtab inherited that order: existing sorts had address ties everywhere, and stable_sort fell back to the unpredictable insertion order.

Sort each table deterministically by (input ordinal, input .symtab index).

Fixes #1261.


parth-07 (approved):

lib/Target/GNULDBackend.cpp
```diff
@@ -800,6 +800,9 @@ void GNULDBackend::sizeDynNamePools() {
   if (!SetSymbolsToBeExported())
     return;
 
+  // Sort before partitioning so dynsym indices are deterministic across runs.
+  m_Module.sortSymbols();
```
parth-07: We call sortSymbols also in GNULDBackend::emitRegNamePools. Should we now remove it from here?

include/eld/Config/LinkerConfig.h
```diff
@@ -60,6 +60,9 @@ class LinkerConfig {
     Unset             ///< Undetermine code position mode
   };
 
+  // FIXME: ARM/RISCV/X86 disable ScanRelocations and ApplyRelocations by
+  // default, likely to aviod non-determinism. With dynsym order now fixed,
```
quic-seaswara: nit : avoid

test/Common/standalone/DeterministicDynsym/DeterministicDynsym.test
```diff
@@ -0,0 +1,47 @@
+## Repeated links of the same inputs into a shared library must produce
+## identical output, even with multithreaded scanning.
+
+# RUN: rm -rf %t && split-file %s %t
+# RUN: %clang %clangopts -c -fPIC %t/a.c -o %t/a.o
```
quic-seaswara: compile with -ffunction-sections and -fdata-sections

test/Common/standalone/DeterministicDynsym/DeterministicDynsym.test
```diff
@@ -0,0 +1,47 @@
+## Repeated links of the same inputs into a shared library must produce
+## identical output, even with multithreaded scanning.
```
quic-seaswara, line 2: Check this on windows too by running the workflow on this PR.

test/Common/standalone/DeterministicDynsym/DeterministicDynsym.test
```diff
@@ -0,0 +1,47 @@
...
+# RUN: %link %linkopts -shared --threads --enable-threads=all %t/a.o %t/b.o %t/c.o -o %t.out6.so
+# RUN: %link %linkopts -shared --threads --enable-threads=all %t/a.o %t/b.o %t/c.o -o %t.out7.so
+# RUN: %link %linkopts -shared --threads --enable-threads=all %t/a.o %t/b.o %t/c.o -o %t.out8.so
+# RUN: diff %t.out1.so %t.out2.so
+# RUN: diff %t.out1.so %t.out3.so
+# RUN: diff %t.out1.so %t.out4.so
+# RUN: diff %t.out1.so %t.out5.so
+# RUN: diff %t.out1.so %t.out6.so
+# RUN: diff %t.out1.so %t.out7.so
+# RUN: diff %t.out1.so %t.out8.so
```
quic-seaswara: Can we use `readelf -a`  or the linker map file to compare instead of doing a binary diff output ? If something fails it may be easy to figure out the error.

lib/Core/Module.cpp
```diff
@@ -382,28 +382,38 @@ bool Module::sortCommonSymbols() {
...
-                       }));
+  auto Cmp = [](ResolveInfo *A, ResolveInfo *B) -> bool {
+    // Section symbols always appear first.
+    if (A->type() == ResolveInfo::Section &&
+        (B->type() != ResolveInfo::Section))
+      return true;
+    if (A->type() != ResolveInfo::Section &&
+        (B->type() == ResolveInfo::Section))
+      return false;
+    if (A->isLocal() && !B->isLocal())
```
quic-seaswara: This is not specific to this patch, but since you are changing this function. We need to handle absolute FILE symbols. Can be a seperate bug.

quic-seaswara (changes requested):

lib/Target/GNULDBackend.cpp
quic-areg: I think we need both: this sort runs before layout to produce a deterministic dynsym partition and the existing sort in `emitRegNamePools` runs after layout when final addresses exist. Removing the second one would change `.symtab` order.

parth-07: I see. Thanks for the explanation. I am still concerned about a few things. Now, the dynsym order is deterministic but still arbitrary for the user. We have the below code in the `Module::sortSymbols`:

```cpp
    auto ValA = A->outSymbol()->value();
    auto ValB = B->outSymbol()->value();
    if (ValA != ValB)
      return ValA < ValB;
    // Break ties deterministically by name, then input ordinal, so the final
    // order does not depend on insertion order which can vary across runs due
    // to multithreaded relocation scanning.
    int NameCmp = A->getName().compare(B->getName());
    if (NameCmp != 0)
      return NameCmp < 0;
    assert(A->resolvedOrigin() && B->resolvedOrigin());
    return A->resolvedOrigin()->getInput()->getInputOrdinal() <
           B->resolvedOrigin()->getInput()->getInputOrdinal();
```
Symbol value comparison seems misleading here because at this stage the symbols contain the original values from the input files. Hence, the "sort by address" intent is not actually being realized here. If we have the following symbols: lib1.so[foo] (value: 0x10), lib1.so[bar] (value: 0x20), lib2.so[baz] (0x10), then the output order of dynsym table will be: [`lib1.so[foo]`, `lib2.so[baz]`, `lib1.so[bar]`]. The order is deterministic but it is still difficult to reason without seeing the input values of each individual shared library symbols. Later, if the value of `lib1.so[foo]` changes, then the `.dynsym` order will also change.

Additionally, I believe that it would be better to only sort dynamic symbols here (`GNULDBackend::DynamicSymbols` vector) as that is both significantly more efficient and makes the intent clear from the code itself.

lib/Core/Module.cpp
```diff
@@ -378,28 +378,32 @@ bool Module::sortCommonSymbols() {
...
+    if (A->type() == ResolveInfo::Section &&
+        (B->type() != ResolveInfo::Section))
+      return true;
+    if (A->type() != ResolveInfo::Section &&
+        (B->type() == ResolveInfo::Section))
+      return false;
+    // ELF requires all locals before all globals.
+    if (A->isLocal() != B->isLocal())
+      return A->isLocal();
+    // Deterministic order: by (input ordinal, input .symtab index).
```
quic-seaswara, line 392: How are undef's sorted ?

lib/Target/GNULDBackend.cpp
quic-areg: Fully agree. Addressed this in the latest revision. Only `DynamicSymbols` is sorted for `.dynsym` rather than the whole module symbol vector, and the value/name comparison is gone from the `.dynsym` key.

lib/Core/Module.cpp
quic-areg: They're not treated as a separate class anymore, so they sort like any other symbol. I don't think the undef vs defined split has any ELF significance for `.symtab`. The splits that matter are section symbols first and locals before globals, and both of those are preserved. My goal is a predictable, deterministic order, and treating undefs separately makes the output order unpredictable. This also matches lld's approach.

quic-seaswara (approved):

include/eld/Config/LinkerConfig.h
```diff
@@ -60,6 +60,9 @@ class LinkerConfig {
     Unset             ///< Undetermine code position mode
   };
 
+  // FIXME: ARM/RISCV/X86 disable ScanRelocations and ApplyRelocations by
```
parth-07:
```suggestion
  // FIXME: ARM/RISCV/X86 disable multithreading in ScanRelocations and ApplyRelocations by
```

test/AArch64/standalone/KeepCommonSymbols/KeepCommonSymbols.test
```diff
@@ -8,9 +8,9 @@ RUN: %clang %clangopts -fcommon -c %p/Inputs/1.c -o %t1.1.o
 RUN: %link %linkopts -o %t1.1.out %t1.1.o --gc-sections -e main -T %p/Inputs/1.script
 RUN: %readelf -s %t1.1.out | %filecheck %s
 #END_TEST
-#CHECK: b_common
-#CHECK: a_common
-#CHECK: f_common
+#CHECK-DAG: b_common
```
parth-07: Why do we need `-DAG` when the symbol table order is deterministic?

Steven6798 (approved):

test/AArch64/standalone/KeepCommonSymbols/KeepCommonSymbols.test
quic-areg: Done

quic-areg:
windows: https://github.com/qualcomm/eld/actions/runs/28829117365/job/85498895127?pr=1263

parth-07 (approved):
LGTM.

PR #1268, opened 2026-06-15
[RISCV] Remove outdated `c.lui rd, 0` to `c.li rd, 0` rewriting. NFCI

jonathonpenix:
This code should be dead as `R_RISCV_RVC_LUI` is an internal relocation to help with relaxation and the only place we emit this relocation already checks that the immediate is non-zero.

Cleaning this up is also helpful as a follow up patch will introduce a new internal relocation for `c.li` (for handling GOT load relaxations). The existing code is incorrect for that case, and cleaning this up will let us handle these two consistently (and more consistently with other instruction types).

### DIFF
--- a/lib/Target/RISCV/RISCVHelper.h
+++ b/lib/Target/RISCV/RISCVHelper.h
@@ -276,8 +276,6 @@ template <typename T> T encodeCJ(T Result) {
 }
 
 template <typename T> T encodeCI(T Result) {
-  // `c.lui rd, 0` is illegal, it will be converted  to `c.li rd, 0` when
-  // applying
   uint16_t Imm17 = extractBits(Result, 17, 17) << 12;
   uint16_t Imm16_12 = extractBits(Result, 16, 12) << 2;
   Result = Imm17 | Imm16_12;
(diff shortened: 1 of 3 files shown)


svs-quic (approved):
LGTM!

quic-seaswara (changes requested):
Prefer adding a build/run test to verify that the fix is good.

quic-seaswara (commented):
Can we also start adding a developer docs for relaxations which will be extremely helpful for RISC-V as a follow up ?

jonathonpenix:
>Prefer adding a build/run test to verify that the fix is good.

Not entirely sure if it is what you had in mind, but I added explicit checks that we aren't creating c.lui with imm 0 along (AFAIK) the only path we can possibly generate `c.lui`.

It's not really a check that the fix is good as much as checking that the existing code already guards against this sort of thing.

I'm not really sure what else I can check though--the deleted code should be dead and I don't think a run test adds anything meaningful/anything not already covered by the build test. Experimentally, deleting the `hi_imm != 0` check breaks building things in weird ways already when building just empty `main` tests.)

jonathonpenix:
Sorry, I take this patch back.

I don't think this code is actually dead since we don't actually know where anything is going to be at the time we are relaxing.

So this code can help paper over things like this:
```
cat > 1.s << '!'
.option relax

.data
.global var
var:
.word 10

.section foo_sec, "ax", @progbits
.global foo
foo:
  lui a0, %hi(var)
  lw a0, %lo(var)(a0)
  lui a0, %hi(var)
  lw a0, %lo(var)(a0)
  ret
!

cat > script.t << '!'
SECTIONS {
  foo_text 0x7F0 : { *(foo_sec) }
  .data : { *(.data) }
}
!

clang --target=riscv32-unknown-elf 1.s -c
ld.eld 1.o -T script.t --no-relax-zero
```

Just going to close and redo this elsewhere.

PR #1269, opened 2026-06-15
[RISCV] Implement encoding for CI-type instructions in terms of its `Shift`. NFCI

jonathonpenix:
A follow-up patch will add a new internal relocation to handle `c.li`.
Currently, `encodeCI` looks for the immediate in the higher bits of `Result`
(bits 17 and 16:12) which makes sense for `c.lui` handling (and when `c.lui` is
rewritten to `c.li`) but not really for the new relocation for `c.li` (where we
want to look at bits 5 and 4:0).

I don't think this would work as-is for some of the other CI-type instructions
if they ever need to be handled so it isn't entirely generic. But, this should
be sufficient for supporting both `c.li` and `c.lui` relocations. I also think
this is marginally simpler/more straightforward than creating a separate
`EncTy_*` or checking the instruction type, etc. so I went this route for now.

Note `isValidRVCLUIType` is now in a (more) broken state as I didn't update the
call to `encodeCI`. As far as I can tell, this code is already dead/broken (it
ultimately calls `extractRVCImmediate` which itself has a typo (`exractBits`)
so has never been compile-able as far as I can tell. Not sure if we want to fix
this up or just remove it, so I'm leaving it be for the purposes of this patch.

### DIFF
--- a/lib/Target/RISCV/RISCVHelper.h
+++ b/lib/Target/RISCV/RISCVHelper.h
@@ -275,9 +275,9 @@ template <typename T> T encodeCJ(T Result) {
   return Result;
 }
 
-template <typename T> T encodeCI(T Result) {
-  uint16_t Imm17 = extractBits(Result, 17, 17) << 12;
-  uint16_t Imm16_12 = extractBits(Result, 16, 12) << 2;
+template <typename T> T encodeCI(T Result, uint32_t Shift) {
+  uint16_t Imm17 = extractBits(Result, 5 + Shift, 5 + Shift) << 12;
+  uint16_t Imm16_12 = extractBits(Result, 4 + Shift, Shift) << 2;
   Result = Imm17 | Imm16_12;
   return Result;
 }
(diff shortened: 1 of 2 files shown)


lenary (approved):
LGTM.

Instruction types are a bit annoying in the C extension, potentially less relevant, but this seems the best approach.

quic-seaswara (approved):

jonathonpenix:
Sorry for the churn/force-pushing here. Put it on top of #1384 now instead--this should be basically the same as it should still work for the c.lui -> c.li rewriting.

lenary (approved):
LGTM

jonathonpenix:
Final rebasing before merging to unstack this stuff.

@quic-seaswara are you fine with this change?

I think you looked at it before it was put on #1384. Which, shouldn't really change anything, but I don't think eld has things set to dismiss reviews after more changes were pushed.

Just want to confirm before going ahead

PR #1270, opened 2026-06-15
[RISCV] Implement GOT load relaxation

jonathonpenix:
The relaxation is defined in the RISC-V psABI at [1].

Broadly, it allows relaxing a GOT-relative sequence to either a PC-relative
sequence or absolute addi/c.li.

There's a few limitations or items left for future work:
- The psABI allows the linker to omit the GOT slot for a symbol if all
  references to the symbol's GOT slot is removed. This is not implemented.
- Similar to other relaxations in eld, relaxations for symbols of value
  0 are generally avoided.
- Weak undef symbols cannot be relaxed currently under `-static -pie` as they
  are considered preemptible still. It's not clear to me if this intended
  (and, we do have `-static -pie` users so it might be worth cleaning this
  up if possible)--this is left to resolve elsewhere.

[1] https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/ffde5a7f62a1bb12a4d775afa45342ac24138849/riscv-elf.adoc?plain=1#L1774

Assisted-by: claude


jonathonpenix:
There's an XPASS on this currently that I still need to look into.

But, in the meantime any initial feedback would be greatly appreciated!

test/Common/RunTests/HelloWorld/HelloWorld.test
```diff
@@ -9,6 +9,8 @@ UNSUPPORTED: x86
 #END_COMMENT
 #START_TEST
 RUN: %run_cc -o %t1.o %p/Inputs/HelloWorld.c -c -fno-pic
+RUN: %run_cc -o %t.norelax.out %t1.o -static -Wl,--no-relax-got
```
jonathonpenix, line 12: Adding this as the test otherwise unexpectedly passes with this patch (it's xfail'ed for RISC-V currently). AFAICT the PR here is working as intended, it just happens to paper over an unrelated bug. See https://github.com/qualcomm/eld/issues/1342

So I think it is worth covering both relax and no-relax cases rather than just un-XFAIL'ing the test or something.

jonathonpenix:
Sorry for the churn + force pushing here--wanted to put this on top of #1384 instead of #1268 and needed to update some things after #1289 so just rebased + squashed everything down again.

Hopefully things should be settled down. Please take a look!

jonathonpenix:
FWIW the eld internal HF runs look clean.

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -1406,6 +1407,142 @@ bool RISCVLDBackend::isGOTReloc(const Relocation &reloc) const {
...
+  if (!region)
+    return false;
+
+  const Relocation *BaseReloc = Reloc.type() == llvm::ELF::R_RISCV_GOT_HI20
+                                    ? &Reloc
+                                    : getBaseReloc(Reloc);
+  if (!BaseReloc)
+    return false;
+
+  // Until recently, the calculation for R_RISCV_GOT_HI20 was `G + GOT + A - P`
```
quic-seaswara, line 1427: I thought addends are always 0.

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -1406,6 +1407,142 @@ bool RISCVLDBackend::isGOTReloc(const Relocation &reloc) const {
...
+  // The psABI only includes "it’s bound at link time to be within the object"
+  // as a condition for the relative case. But, absolute symbols are seemingly
+  // also preemptible, so ignoring this check in the absolute case could mean a
+  // change in behavior if the relaxation is performed. Treat this as a general
+  // requirement for the relaxation.
+  if (isSymbolPreemptible(*SymInfo))
+    return false;
+
+  if (SymInfo->isIFunc())
+    return false;
```
quic-seaswara, line 1445: I dont think you need this check for isIFunc as isSymbolPreemptible would  return. Did you find that isSymbolPreemptible is returning false for IFUNC symbols ?

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -1406,6 +1407,142 @@ bool RISCVLDBackend::isGOTReloc(const Relocation &reloc) const {
...
+
+  Relocator::DWord S = getSymbolValuePLT(*BaseReloc);
+  uint64_t Offset = Reloc.targetRef()->offset();
+  StringRef SymName = SymInfo->name();
+  const char *RelaxName = "RISCV_GOT_LOAD";
```
quic-seaswara: StringRef

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -1406,6 +1407,142 @@ bool RISCVLDBackend::isGOTReloc(const Relocation &reloc) const {
...
+  bool GOTRelaxEnabled = config().options().getRISCVRelax() &&
+                         config().options().getRISCVRelaxGOT();
+  if (SymInfo->isAbsolute() || SymInfo->isWeakUndef()) {
+    // Comments elsewhere mention eld uses zero as an indicator of an unknown
+    // symbol value, so conservatively follow suit and skip this relaxation for
+    // symbols of value zero.
+    bool SymbolValueMayBeUnknown = S == 0 && !SymInfo->isWeakUndef();
```
quic-seaswara, line 1457: You would not need this anymore. of using 0 as unknown symbol value, since we fixed the Layout convergence issue.

@parth-07 can you pls comment here ?

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -1406,6 +1407,142 @@ bool RISCVLDBackend::isGOTReloc(const Relocation &reloc) const {
...
+
+    return false;
+  }
+
+  Relocator::Address P = BaseReloc->place(m_Module);
+  // Again avoid relaxing symbols with value zero in case they indicate a symbol
+  // with an unknown value. Make an exception for RV32 as a
+  // PCREL_HI20/PCREL_LO12_I pair can reach the entire address space.
+  bool SymKnownInRange = config().targets().is32Bits() ||
+                         (S != 0 && llvm::isInt<32>((S - P) + 0x800));
```
quic-seaswara, line 1523: Wouldn't this be always true in riscv 32bit ?

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -1635,10 +1773,27 @@ void RISCVLDBackend::mayBeRelax(int relaxation_pass, bool &pFinished) {
...
             doRelaxationPC(relocation, GP);
             break;
+          case llvm::ELF::R_RISCV_PCREL_LO12_I: {
+            const Relocation *HIReloc = getBaseReloc(*relocation);
+            if (!HIReloc)
+              break;
+
+            if (HIReloc->type() == llvm::ELF::R_RISCV_GOT_HI20 ||
+                relocWasGOTLoadRelaxed(HIReloc)) {
+              if (hasRelax(*HIReloc) && allGOTLOsRelaxable(*HIReloc, rs))
```
quic-seaswara: Is there just one HI GOT reloc for all the low GOT relocs ? Is this why  you are having this function ?

quic-seaswara (changes requested):

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -1635,10 +1773,27 @@ void RISCVLDBackend::mayBeRelax(int relaxation_pass, bool &pFinished) {
...
           case llvm::ELF::R_RISCV_PCREL_LO12_S:
             doRelaxationPC(relocation, GP);
             break;
+          case llvm::ELF::R_RISCV_PCREL_LO12_I: {
+            const Relocation *HIReloc = getBaseReloc(*relocation);
+            if (!HIReloc)
+              break;
+
+            if (HIReloc->type() == llvm::ELF::R_RISCV_GOT_HI20 ||
+                relocWasGOTLoadRelaxed(HIReloc)) {
```
quic-akaryaki, line 1790: Does this check work when LO12 is encountered before GOT_HI20?

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -1635,10 +1773,27 @@ void RISCVLDBackend::mayBeRelax(int relaxation_pass, bool &pFinished) {
...
-          case llvm::ELF::R_RISCV_PCREL_LO12_I:
           case llvm::ELF::R_RISCV_PCREL_LO12_S:
             doRelaxationPC(relocation, GP);
             break;
+          case llvm::ELF::R_RISCV_PCREL_LO12_I: {
+            const Relocation *HIReloc = getBaseReloc(*relocation);
+            if (!HIReloc)
+              break;
+
+            if (HIReloc->type() == llvm::ELF::R_RISCV_GOT_HI20 ||
```
quic-akaryaki, line 1789: Should it be `&&` ?

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -1635,10 +1773,27 @@ void RISCVLDBackend::mayBeRelax(int relaxation_pass, bool &pFinished) {
...
+                relocWasGOTLoadRelaxed(HIReloc)) {
+              if (hasRelax(*HIReloc) && allGOTLOsRelaxable(*HIReloc, rs))
+                doRelaxationGOT(*relocation);
```
quic-akaryaki, line 1792: doRelaxationGOT() returns a value but it's unused. Should you try PC relaxation if the GOT one did not work?

quic-akaryaki (commented):
Sorry, just high-level comments, I did not read deeply.

lib/Target/RISCV/RISCVLDBackend.cpp
jonathonpenix: This was just recently (in the last week) changed to be the case: https://github.com/riscv-non-isa/riscv-elf-psabi-doc/commit/eb794e8ccfabfbb370db2070565b588252c4f8b7

Currently eld accepts + seems to honor the addend.

I was thinking about pushing a change for eld to ignore this now that the ABI says it must be zero, but I don't want to keep stacking this PR on top of more things.

So I'm hoping that can be separate/a follow-up.

lib/Target/RISCV/RISCVLDBackend.cpp
jonathonpenix: Yeah, `isSymbolPreemptible` will return false for a lot of cases involving ifuncs currently--the ifunc check is almost last so any of the checks above it will "preempt" that part. (And, ifunc + pie, hidden ifunc, etc. are valid IIUC. I think even it just being a static executable is enough to get it to return false for ifuncs?)

So for now I think explicitly checking for ifuncs is the safest route.

If there should be changes to `isSymbolPreemptible`, I hope that can be done separately.

And last thought: the current ifunc check in GOT_LOAD_RELAX_REL.test does fail if the `if(isIFUNC())` check here is deleted, so we have some coverage on this (though it isn't exhaustive).

lib/Target/RISCV/RISCVLDBackend.cpp
jonathonpenix: Done

lib/Target/RISCV/RISCVLDBackend.cpp
jonathonpenix: Yep--the first `config().targets().is32Bits()` is intended to make this always true for rv32 (as we can reach the entire address space with pcrel lo/hi pair).

The `|| (S != 0 && llvm::isInt<32>((S - P) + 0x800));` part is intended to be rv64-specific

lib/Target/RISCV/RISCVLDBackend.cpp
jonathonpenix: My understanding is that:
- A pcrel lo must be matched to one GOT hi
- A GOT HI can have an arbitrary number of pcrel los referring to it

So before we do any relaxation we have to make sure the specific GOT HI we're looking at and all LOs that point at it are valid candidates for this relaxation.

lib/Target/RISCV/RISCVLDBackend.cpp
jonathonpenix: Originally I was just following the other relaxations that all have this and I couldn't write a (simple) case that required this. So I'm happy to remove this if it isn't needed anymore on that front.

That said, I _think_ the `S == 0` is still needed for some syms like `__rela_iplt_start`/`__rela_iplt_end` since IIUC they are initially defined with value 0 and aren't given their final value (via finalizeSymbols()) until after relaxation (via `layout()`) happens.

(For an example of this happening, see the comment at the top of GOT_LOAD_RELAX_RUN.test--I think that is what blocks the relaxation currently.)

`__rela_iplt_start`/`_end` should fall under the "pcrel" case below, but I _think_ other standard syms (ex: those in `initStandardSymbols()`) would a) have a similar issue of having a value of 0x0 until _after_ relaxation and b) can fulfil the absolute requirement here. I need to check that though to confirm.

Will wait to hear from @parth-07 before making any changes on this.

lib/Target/RISCV/RISCVLDBackend.cpp
jonathonpenix: `||` is correct I think.

This relaxation either:
* Deletes the insn associated with the GOT HI (so we rewrite it to a `R_RISCV_NONE` reloc)
* Changes the GOT-relative reloc to a PCREL-reloc

So if the HIReloc was relaxed, it shouldn't have a `R_RISCV_GOT_HI20` type anymore. If the HIReloc wasn't relaxed, we want to check the type == `R_RISCV_GOT_HI20` and `relocWasGOTLoadRelaxed` should return false.

lib/Target/RISCV/RISCVLDBackend.cpp
jonathonpenix: I think it should--if we are processing the LO reloc first, the HI Reloc should still have its type as `R_RISCV_GOT_HI20`

(then in the typical case where we process HI -> LO, the type would have been rewritten to NONE/PCREL, hence the `relocWasGOTLoadRelaxed` check)

It isn't the most elaborate test, but this is what `bar` in nonsequential.S is trying to check

lib/Target/RISCV/RISCVLDBackend.cpp
jonathonpenix: > doRelaxationGOT() returns a value but it's unused

Bit of a side note, but yeah--this part is a bit odd. I kind of just followed what the other relaxations were doing (returning true/false depending on what happened, even if they weren't used). 

> Should you try PC relaxation if the GOT one did not work

For now I don't think so, the GOT relaxation is GOTHI/PCREL_LO relocations and I think the PC relaxation is expecting PCREL_HI/PC_REL ([it bails if it sees a GOT HI relocation](https://github.com/qualcomm/eld/blob/48c07b400c6713c4ad9ba2357eb65ee0ecc6f24b/lib/Target/RISCV/RISCVLDBackend.cpp#L1446)).

So for now I think they're separate "categories".

If there's ever a relaxation that applies to both GOTHI and PCREL HI I guess we might have to revisit this.

jonathonpenix: >Should you try PC relaxation if the GOT one did not work?

One other thought on this, I'm not entirely sure if I'm allowed to "chain" relaxations. Ex: we can rewrite the GOT-relative access to be a PCREL one--I'm not sure if we are allowed to (or would want to) then reprocess and potentially do the PC relaxations on these rewritten relocations.

The impression I got was "no" so I avoided this sort of thing for now.

lenary: Yeah, you're not supposed to "chain" relaxations, but maybe in this case nothing bad would happen. I really need to make a start on the docs that Shankar asked for...

lib/Target/RISCV/RISCVLDBackend.cpp
quic-akaryaki: Thanks!

lib/Target/RISCV/RISCVLDBackend.cpp
quic-akaryaki: I didn't notice you rewrite the type, thanks.

lib/Target/RISCV/RISCVLDBackend.cpp
quic-akaryaki: > So for now I think they're separate "categories".

You are right. It's been a long time I looked at this code. This would be more obvious if we checked the type of HI relocation upfront.

quic-akaryaki (approved):
Thanks!

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -2540,6 +2696,19 @@ RISCVLDBackend::postProcessing(llvm::FileOutputBuffer &pOutput) {
   return {};
 }
 
+bool RISCVLDBackend::allGOTLOsRelaxable(const Relocation &HIReloc,
+                                        const ELFSection *S) const {
```
quic-areg: looks like `S` is unused

quic-areg (approved):

lib/Target/RISCV/RISCVLDBackend.cpp
jonathonpenix: Oops, yep! Missed that when rebasing, thanks for catching this!

lib/Target/RISCV/RISCVLDBackend.cpp
jonathonpenix: >This would be more obvious if we checked the type of HI relocation upfront.

Please let me know if you have something specific in mind--I'm not entirely sure what you mean by "checked the type of HI relocation upfront".

quic-akaryaki: I mean, in the prior code, we tried PC relaxation even the HI relocation was of the GOT type just to exit from the middle of the doRelaxationPC(). 
Now you have clarified this.
But, doRelaxationPC() still have the check for the GOT relocation, should it assert allowed types?

lib/Target/RISCV/RISCVLDBackend.cpp
lenary: Yes please can it be a follow-up.

Compilers should only produce addend=0 code, but hand-assembly might do something funky and wrong. bfd ld gives an error when the addend is non-zero, which might be the sort of thing ELD wants because I know you love a good diagnostic @quic-seaswara.

lib/Target/RISCV/RISCVLDBackend.cpp
lenary: iirc there's a draft PR from Alexey to make it easier to work out why a symbol has value 0 (#276), and suppress relaxations when the zero is spurious. Until that lands, we should keep the "don't do this when the symbol is zero" check IMO.

lib/Target/RISCV/RISCVLDBackend.cpp
lenary: I am not sure the second part wraps correctly on rv32, but the entire condition as written seems correct to me.

lenary (approved):
LGTM

apazos (approved):

quic-seaswara (approved):

PR #1272, opened 2026-06-15
[test] Port tests to work with the internal LIT shell

svs-quic:
With LLVM recently making the internal shell default for the LIT tests some of the ELD tests were failing due to multiple incompatibilities with how the internal shell executes tests.

The tests relied on shell features that are not supported by lit's internal shell like subshell grouping, for-loops, inline VAR=... cmd assignment, command substitution, and cd - behavior.  Update the affected tests to use internal-shell-safe RUN lines.

Assisted by: GPT-5.3


quic-seaswara (approved):

svs-quic:
Hopefully these changes work fine on windows as well.

PR #1274, opened 2026-06-15
[lit] Remove EldShellTestFormat and use lit.formats.ShTest directly

quic-seaswara:
EldShellTestFormat duplicated lit's built-in shell test handling. Replace it with a plain lit.formats.ShTest, controlled by LIT_USE_INTERNAL_SHELL (defaulting to the internal shell). get_test_format now accepts a pre-built base_format, removing the execute_external parameter from both it and ThinArchiveTestFormat.

Add run_and_print_exit.py which runs a binary and prints its exit code, and wire it up as the %run_and_print_exit substitution in lit.cfg.

Assisted by Claude.


quic-seaswara:
Related discussion : https://discourse.llvm.org/t/rfc-enabling-the-lit-internal-shell-by-default/80179

test/eld_test_formats.py
```diff
@@ -1,120 +1,9 @@
-import os
```
svs-quic, line 1: Can the contents of this file be moved in to `lit.cfg`? It looks like there is only one use of `get_test_format`?

svs-quic (approved):
LGTM!

test/eld_test_formats.py
quic-seaswara: Sure. I will do this as a follow up.

Steven6798 (approved):

PR #1275, opened 2026-06-15
[ARM] Remove RegionTable support

quic-seaswara:
The RegionTable was an ARM-specific feature that emitted a __region_table__ section describing BSS region start/end addresses for static executables. It was never emitted for PIE, shared libraries, or partial links, and the feature is no longer needed.

Remove RegionTableFragment (header, implementation, CMakeLists entry), the RegionTable enum values from TargetFragment::Kind and Module::InternalInputType, the corresponding Module::createInternalSection switch case, the include from LayoutInfo.h, and all three ARM backend methods (finishAssignOutputSections, updateTargetSections, handleBSS) along with their associated member variables. Also remove the three lit test directories that covered this feature.


parth-07 (commented):
Nitpick: We have this below comment in GNULDBackend.h:620 from which region-table part needs to be removed.:

```
// The target can decide on how we want to handle BSS sections.
  // If the linker is emitting a region table, its possible to mix BSS
  // DATA in one segment.
```

quic-seaswara:
> Nitpick: We have this below comment in GNULDBackend.h:620 from which region-table part needs to be removed.:
> 
> ```
> // The target can decide on how we want to handle BSS sections.
>   // If the linker is emitting a region table, its possible to mix BSS
>   // DATA in one segment.
> ```

Will remove.

Steven6798 (approved):

PR #1281, opened 2026-06-16
[nightly] Upload ELD-only binaries to GitHub Releases biannually

quic-seaswara:
- After the nightly build, install only the ld.eld component into a separate prefix using `cmake --install --component ld.eld`.
- Collect shared library dependencies for ld.eld into the install tree using a new reusable script BundleSharedLibDeps.sh, which replaces the duplicated inline logic in the full-toolchain bundling step.
- Pack the result as eld-nightly.tar.xz and upload it to a rolling 'nightly' pre-release on GitHub Releases on Jan 1 and Jul 1.
- manual workflow_dispatch triggers an upload unconditionally.

Assisted by Claude.

### DIFF
--- a/.github/workflows/nightly-test.yml
+++ b/.github/workflows/nightly-test.yml
@@ -12,12 +12,12 @@ permissions:
 jobs:
   build-and-runtest-nightly:
     if: github.repository == 'qualcomm/eld'
-    runs-on: ubuntu-latest
-
+    runs-on: ubuntu-22.04
     env:
       BASE_BRANCH_NAME: ${{ github.event.pull_request.base.ref }}
       ELD_REPO: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name || github.repository }}
       ELD_REF: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.ref || github.ref_name }}
+      SCRIPTS_DIR: ${{ github.workspace }}/llvm-project/llvm/tools/eld/.github/workflows/scripts
 
     steps:
       - name: Set up Clang 22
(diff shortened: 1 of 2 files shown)


quic-seaswara:
https://github.com/qualcomm/eld/actions/runs/27653189608

Steven6798:
> https://github.com/qualcomm/eld/actions/runs/27653189608

The workflow failed. Could you take a look at that?

quic-seaswara:
> > https://github.com/qualcomm/eld/actions/runs/27653189608
> 
> The workflow failed. Could you take a look at that?

Yes need to rerun the workflow :)

quic-seaswara:
> > > https://github.com/qualcomm/eld/actions/runs/27653189608
> > 
> > 
> > The workflow failed. Could you take a look at that?
> 
> Yes need to rerun the workflow :)

https://github.com/qualcomm/eld/actions/runs/27790731971

The way I have configured this action is to whenever the action is triggered manually, it will generate a nightly release otherwise bi-annually.

Nightly release available of eld now : https://github.com/qualcomm/eld/releases/download/nightly/eld-nightly.tar.xz

Steven6798 (approved):

quic-areg:
The binary does not run on my machine: 
```
bin/ld.eld: symbol lookup error: /tmp/install-eld-only/bin/../lib/libc.so.6: undefined symbol: __tunable_is_initialized, version             
GLIBC_PRIVATE       
  ```

I think the default extracted directory name `install-eld-only` is an awkward name for a user who downloads a tarball. It matches the workflow's internal directory name.

**Comment by quic-seaswara** [comment]
> The binary does not run on my machine:
> 
> ```
> bin/ld.eld: symbol lookup error: /tmp/install-eld-only/bin/../lib/libc.so.6: undefined symbol: __tunable_is_initialized, version             
> GLIBC_PRIVATE       
> ```
> 
> I think the default extracted directory name `install-eld-only` is an awkward name for a user who downloads a tarball. It matches the workflow's internal directory name.

You need to run on ubuntu 24.04.

```
$ /tmp/install-eld-only/bin/ld.eld --version
eld 23.0 (GNU Compatible linker)
Supported Targets: arm aarch64 riscv32 riscv64 hexagon x86_64 

$ cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=24.04
DISTRIB_CODENAME=noble
DISTRIB_DESCRIPTION="Ubuntu 24.04.3 LTS"
```


What do you suggest ? eld-nightly  ?

**Comment by quic-areg** [comment]
> You need to run on ubuntu 24.04.

Right, and I think that could be a problem for users wanting to try eld. 

Can we:
1. not bundle glibc components as part of `BundleSharedLibDeps.sh` and let the host's glibc resolve them.
2. run the workflow on an older host like `ubuntu-22.04` rather than `ubuntu-latest` to support a wider range of hosts? 

This is exactly what LLVM's release workflow does:
```
# llvm/llvm-project .github/workflows/release-binaries.yml
runs-on: ubuntu-22.04
# "We use ubuntu-22.04 rather than the latest version to make the built
#  binaries more portable (eg functional against older glibc)."
```
>  eld-nightly 

sounds good.

quic-seaswara:
> > You need to run on ubuntu 24.04.
> 
> Right, and I think that could be a problem for users wanting to try eld.
> 
> Can we:
> 
> 1. not bundle glibc components as part of `BundleSharedLibDeps.sh` and let the host's glibc resolve them.

Sure. 

> 2. run the workflow on an older host like `ubuntu-22.04` rather than `ubuntu-latest` to support a wider range of hosts?

Sure.

PR #1283, opened 2026-06-17
Add eld::plugin::InputFile::isLTOGeneratedObject API

parth-07:
This commit adds a new API,
eld::plugin::InputFile::isLTOGeneratedObject, in the plugin framework.


lib/LinkerWrapper/PluginADT.cpp
```diff
@@ -1296,6 +1296,15 @@ bool plugin::InputFile::isBitcode() const {
   return m_InputFile->isBitcode();
 }
 
+bool plugin::InputFile::isLTOGeneratedObject() const {
+  if (!m_InputFile)
+    return false;
```
quic-seaswara, line 1301: probably it should return Expected<bool> and false should be an error.

test/Common/Plugin/IsLTOGeneratedObject/IsLTOGeneratedObject.test
```diff
@@ -0,0 +1,16 @@
+#---IsLTOGeneratedObject.test----------------------- Executable,LS ----------------#
+#BEGIN_COMMENT
+# Tests that isLTOGeneratedObject() returns true for ELF objects generated
+# by LTO and false for regular ELF object files.
+#END_COMMENT
+#START_TEST
+UNSUPPORTED: x86
```
quic-seaswara: Why unsupported x86 ?

lib/LinkerWrapper/PluginADT.cpp
```diff
@@ -1296,6 +1296,15 @@ bool plugin::InputFile::isBitcode() const {
...
+  eld::ELFObjectFile *ObjFile = llvm::dyn_cast<eld::ELFObjectFile>(m_InputFile);
```
quic-seaswara, line 1302: What about ThinLTO ?

quic-seaswara (changes requested):
Add more tests for

- ThinLTO
- FatLTO
- Embedded LTO

parth-07:
> Add more tests for

> ThinLTO
> FatLTO
> Embedded LTO

Hi Shankar, this API simply exposes an internal data-structure value that should already be correct for these cases. This patch is not adding/modifying any existing functionality. I do not think we should exhaustively test a plugin framework API that is simply used for inspecting a linker state. We never exhaustively test the plugin framework inspection APIs. Please let me know your thoughts on this.

lib/LinkerWrapper/PluginADT.cpp
parth-07: We never return Expected from the `eld::plugin::` inspection APIs. They all follow the pattern if the object is empty, then the API will return false/0 value. I believe that we should not change a well-established usage pattern now. If we want to change this pattern, then I believe we should change it for all the APIs. Please let me know your thoughts on this.

lib/LinkerWrapper/PluginADT.cpp
parth-07: The same API `ELFObjectFile::setLTOObject` is treated the same way for all the LTO cases. Is there any particular reason due to which you think ThinLTO should be handled differently here?

test/Common/Plugin/IsLTOGeneratedObject/IsLTOGeneratedObject.test
parth-07: I have removed it now. It was due to a copy-paste mistake. Thanks for catching it.

lib/LinkerWrapper/PluginADT.cpp
quic-seaswara: Sure.

lib/LinkerWrapper/PluginADT.cpp
quic-seaswara: Please include ThinLTO in the testcase.

quic-seaswara:
> > Add more tests for
> 
> > ThinLTO
> > FatLTO
> > Embedded LTO
> 
> Hi Shankar, this API simply exposes an internal data-structure value that should already be correct for these cases. This patch is not adding/modifying any existing functionality. I do not think we should exhaustively test a plugin framework API that is simply used for inspecting a linker state. We never exhaustively test the plugin framework inspection APIs. Please let me know your thoughts on this.

I believe we can expand the testing to include thin lto and embedded lto and see that the input file returned after LTO shows the desired property.

parth-07:
> I believe we can expand the testing to include thin lto and embedded lto and see that the input file returned after LTO shows the desired property.

I have extended the test with thin-lto and fat-lto.

lib/LinkerWrapper/PluginADT.cpp
parth-07: I have extended the testcase with ThinLTO.

include/eld/PluginAPI/PluginADT.h
```diff
@@ -1166,6 +1166,10 @@ struct DLL_A_EXPORT InputFile {
   /// Returns true if the input is LLVM bitcode; Otherwise returns false.
   bool isBitcode() const;
 
+  /// Returns true if the input file is an ELF object file generated by LTO;
```
quic-seaswara, line 1169: nit : remove ; at the end of the line ?

parth-07: `;` is separating the two points: 1) Returns true if the input file is an ELF object file generated by LTO, 2)  Otherwise returns false.

quic-seaswara (approved):

Steven6798 (approved):

PR #1166, opened 2026-05-12
[object] Fix reproducer crash when input file has no contents

deepakshirkem:
###  Problem
ELD crashes with a segfault when a reproduce tarball contains a linker script that references a file not found in the sysroot.

### Testing
Added test `SysrootScriptCrash` that verifies ELD errors out  gracefully instead of segfaulting when a linker script references an unreadable file.

### Screenshot
Screenshot from 2026-05-12 18-52-19

Fixes #818

cc @quic-seaswara , @parth-07 , @quic-areg

Issue #818, opened 2026-02-12
Reproducer crashes when the link contains a script that is found in the sysroot


parth-07:
The reproducer crashes when the link contains a script that is found in the sysroot.

Reproducer:

```bash
#!/usr/bin/env bash

mkdir -p A/lib64
mkdir -p A/script

cat >1.c <<\EOF
int foo() { return 1; }
EOF

cat >script.t <<\EOF
GROUP(/lib64/lib1.so)
EOF

cat >main.c <<\EOF
int main() { return 0; }
EOF

clang -o lib1.so -target x86_64-unknown-elf 1.c -shared -fPIC
clang -o main.o -target x86_64-unknown-elf main.c -c 
mv lib1.so A/lib64/lib1.so
mv script.t A/script.t

ld.eld --sysroot=A -T A/script.t main.o --reproduce a.tar
rm -rf unpacked && mkdir unpacked
tar -xvf a.tar -C unpacked --strip-components=1
cd unpacked
bash -x response.txt
```

PR #1284, opened 2026-06-17
Revert "[object] Fix reproducer crash when input file has no contents"

deepakshirkem:
Reverts qualcomm/eld#1166


quic-seaswara:
Please sign the commit.

deepakshirkem:
Hi @parth-07, The `MissingMappingFileInput` test reproduces the scenario described in issue #818:

1. Create a sysroot containing `lib1.so` and a linker script that references `GROUP(/lib64/lib1.so)`.
2. Link with `--reproduce` to generate a tarball. This succeeds because `lib1.so` exists in the sysroot.
3. Extract the tarball.
4. Remove `lib1.so` from the extracted tarball to simulate the missing file case.
5. Remove `lib1.so` from the sysroot as well.
6. Replay the link using the `mapping.ini` file generated by `--reproduce`.

Without the fix:
 `resolvePathMappingFile()` calls `createMemoryArea()`, which returns `nullptr` for the missing file.
 `MemArea` is set to `nullptr`, but the function still returns `true`.
 The null `MemArea` eventually reaches `readAndProcessInput()`, resulting in a segfault.

With the fix:
`resolvePathMappingFile()` returns `false` when `createMemoryArea()` returns `nullptr`.
ELD reports a proper fatal error instead of crashing.

The test verifies this by checking for the error message: `cannot read file /lib64/lib1.so`  instead of a segfault.

lib/Input/Input.cpp
```diff
@@ -70,7 +70,7 @@ bool Input::resolvePathMappingFile(const LinkerConfig &PConfig) {
   if (!InputMem)
     InputMem = createMemoryArea(FileName, PConfig.getDiagEngine());
   if (!InputMem)
-    return false; // File does not exist;
+    return false;
```
parth-07: @deepakshirkem It does not seem like this PR is changing any functionality at all.

test/Common/standalone/MissingMappingFileInput/Inputs/mapping_missing.ini
```diff
@@ -1,2 +1,2 @@
 [Hash]
-missing.t=DOESNOTEXIST
+script.t=DOESNOTEXIST
```
parth-07: It has the same issue as the last PR. The linker sees `%t.dir/scripts/script.t` and the mapping file has entry for `script.t`. Hence, mapping-file entry is never used.

parth-07:
Hi @deepakshirkem 

The revert patch should just revert the patch and do nothing else. Please add the actual functionality/fix in a follow-up patch _after_ the revert patch.

> Without the fix:
resolvePathMappingFile() calls createMemoryArea(), which returns nullptr for the missing file.
MemArea is set to nullptr, but the function still returns true.
The null MemArea eventually reaches readAndProcessInput(), resulting in a segfault.

> With the fix:
resolvePathMappingFile() returns false when createMemoryArea() returns nullptr.
ELD reports a proper fatal error instead of crashing.

Yes, this seems correct, however, in your last patch the fix was right but the test was not testing the fix.

deepakshirkem:
Hi @parth-07, 
>The revert patch should just revert the patch and do nothing else. Please add the actual functionality/fix in a follow-up patch after the revert patch.

I have reverted the changes to keep this PR as a pure revert  of commit bbc88918, with no additional changes.

quic-seaswara (approved):

quic-seaswara (changes requested):
again the revert is not reverting the change completely. I may be pressed to do the revert myself.

deepakshirkem:
@quic-seaswara, Closing this PR as the revert is incomplete Input.cpp was not fully reverted. I will create a clean revert PR. Sorry for the confusion!

PR #1285, opened 2026-06-17
[RISCV] Relax QC.E.J/JAL to CM.JT/JALT when possible

svs-quic:
This can save either 2 or 4 bytes depending on whether the QC.E.J/JAL can be relaxed to a JAL.

### DIFF
--- a/lib/Target/RISCV/RISCVLDBackend.cpp
+++ b/lib/Target/RISCV/RISCVLDBackend.cpp
@@ -519,7 +519,7 @@ bool RISCVLDBackend::doRelaxationJal(Relocation *reloc) {
 
 bool RISCVLDBackend::doRelaxationQCCall(Relocation *reloc) {
   // This function performs the relaxation to replace: QC.E.JAL or QC.E.J with
-  // one of JAL, C.J, or C.JAL.
+  // one of CM.JT/CM.JALT, JAL, C.J, or C.JAL.
 
   Fragment *frag = reloc->targetRef()->frag();
   RegionFragmentEx *region = llvm::dyn_cast<RegionFragmentEx>(frag);
@@ -536,19 +536,29 @@ bool RISCVLDBackend::doRelaxationQCCall(Relocation *reloc) {
   Relocator::DWord P = reloc->place(m_Module);
   Relocator::DWord X = S + A - P;
 
+  bool doRelax = config().options().getRISCVRelax();
   bool DoCompressed = config().options().getRISCVRelaxToC();
   bool canRelaxXqci =
       config().targets().is32Bits() && config().options().getRISCVRelaxXqci();
-  bool canRelax =
-      config().options().getRISCVRelax() && canRelaxXqci && llvm::isInt<21>(X);
-  bool canCompress = DoCompressed && llvm::isInt<12>(X);
+  bool canRelax = doRelax && canRelaxXqci;
+  bool canCompress = canRelax && DoCompressed && llvm::isInt<12>(X);
+  bool canRelaxTbljal = canRelax && DoCompressed &&
+                        config().options().getRISCVRelaxTbljal() &&
+                        llvm::isInt<32>(X);
+  bool canRelaxJal = canRelax && llvm::isInt<21>(X);
+  auto reportMissedQCCall = [&]() {
+    reportMissedRelaxation("RISCV_QC_E_CALL", *region, offset,
+                           (canCompress || canRelaxTbljal) ? 4 : 2,
+                           reloc->symInfo()->name());
+  };
 
   if (!canRelax) {
-    reportMissedRelaxation("RISCV_QC_E_CALL", *region, offset,
-                           canCompress ? 4 : 2, reloc->symInfo()->name());
+    reportMissedQCCall();
     return false;
   }
 
+  // Prefer direct C.J/C.JAL over table-jump since both end up 2 bytes and
+  // direct compression avoids the JVT entry overhead.
   if (canCompress) {
     uint32_t compressed = isTailCall ? 0xa001 : 0x2001;
     const char *msg = isTailCall ? "RISCV_QC_E_J_C" : "RISCV_QC_E_JAL_C";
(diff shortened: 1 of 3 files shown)


lenary (approved):
LGTM

quic-areg (approved):

lib/Target/RISCV/RISCVTableJump.cpp
```diff
@@ -86,6 +94,11 @@ void RISCVTableJumpFragment::scanTableJumpEntries(ELFSection &Sec) {
...
       R->targetRef()->memcpy(&Insn, sizeof(Insn), 0);
       Rd = (Insn >> 7) & 0x1f;
+    } else if (R->type() == eld::ELF::riscv::internal::R_RISCV_QC_E_CALL_PLT) {
+      R->targetRef()->memcpy(&Insn, sizeof(Insn), 0);
+      // QC.E.J has func2=0b00 and QC.E.JAL has func2=0b01. Use this for
+      // selecting Rd.
+      Rd = (Insn >> 15) & 0x3;
     } else {
       // CALL/CALL_PLT: read the following JALR to get rd.
       R->targetRef()->memcpy(&Insn, sizeof(Insn), 4);
```
quic-seaswara, line 104: Why are we explicitly calling memcpy ? The relocation word already gets populated when reading the relocation.

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -563,19 +573,40 @@ bool RISCVLDBackend::doRelaxationQCCall(Relocation *reloc) {
...
     reloc->setTargetData(compressed);
-    relaxDeleteBytes(msg, *region, offset + 2, 4,
-                     reloc->symInfo()->name());
+    relaxDeleteBytes(msg, *region, offset + 2, 4, reloc->symInfo()->name());
     return true;
   }
 
+  if (canRelaxTbljal && TableJumpFragment && TableJumpFragment->size() &&
+      m_pRISCVTableJumpSection && !m_pRISCVTableJumpSection->isIgnore() &&
+      !m_pRISCVTableJumpSection->isDiscard()) {
```
quic-seaswara, line 587: Can we add a test to check if the RISCVTableJumpSection is ignored or discarded ?

quic-seaswara (changes requested):

lib/Target/RISCV/RISCVTableJump.cpp
svs-quic: I wasn't aware of that and used it to get the Rd from the JALR of a AUIPC and JALR pair for the CALL/CALL_PLT case. Made changes to remove the memcpy calls now.

lib/Target/RISCV/RISCVLDBackend.cpp
svs-quic: Added a test case.

lib/Target/RISCV/RISCVTableJump.cpp
lenary: I think the two prior calls to memcpy are indeed useless - this one has a non-zero offset though, and I don't like inspecting the registers used by a different instruction to work out what the return address register is. Please can we put CALL/CALL_PLT memcpy back so we're looking at the `JALR` not the `AUIPC`?

svs-quic: I've put it back now.

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -536,19 +536,29 @@ bool RISCVLDBackend::doRelaxationQCCall(Relocation *reloc) {
...
   bool canRelaxXqci =
       config().targets().is32Bits() && config().options().getRISCVRelaxXqci();
-  bool canRelax =
-      config().options().getRISCVRelax() && canRelaxXqci && llvm::isInt<21>(X);
-  bool canCompress = DoCompressed && llvm::isInt<12>(X);
+  bool canRelax = doRelax && canRelaxXqci;
+  bool canCompress = canRelax && DoCompressed && llvm::isInt<12>(X);
+  bool canRelaxTbljal = canRelax && DoCompressed &&
+                        config().options().getRISCVRelaxTbljal() &&
+                        llvm::isInt<32>(X);
```
parth-07: Table jump is an indirect jump to an absolute 32-bit address, right? 

is `llvm::isInt<32>(X)` the right condition to use here considering `X` is a PC-relative value (`S + A - P`)?

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -536,19 +536,29 @@ bool RISCVLDBackend::doRelaxationQCCall(Relocation *reloc) {
...
 
+  bool doRelax = config().options().getRISCVRelax();
   bool DoCompressed = config().options().getRISCVRelaxToC();
   bool canRelaxXqci =
       config().targets().is32Bits() && config().options().getRISCVRelaxXqci();
-  bool canRelax =
-      config().options().getRISCVRelax() && canRelaxXqci && llvm::isInt<21>(X);
-  bool canCompress = DoCompressed && llvm::isInt<12>(X);
+  bool canRelax = doRelax && canRelaxXqci;
+  bool canCompress = canRelax && DoCompressed && llvm::isInt<12>(X);
```
parth-07, line 544: Previously, `canCompress` was independent of `canRelax`. Why did we add `canRelax` constraint to `canCompress`?

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -536,19 +536,29 @@ bool RISCVLDBackend::doRelaxationQCCall(Relocation *reloc) {
...
+  bool canRelaxTbljal = canRelax && DoCompressed &&
```
parth-07, line 545: We are missing the `DoCompressed` condition of enabling the TblJal relaxation in the constraints added in this patch at RISCVTableJump.cpp:80:

```cpp
if (R->type() == ...R_RISCV_QC_E_CALL_PLT &&
      (!getRISCVRelax() || !is32Bits() || !getRISCVRelaxXqci()))
    continue;
```

Can the difference in constraint here and the RISCVTableJump.cpp cause any issues?

lib/Target/RISCV/RISCVLDBackend.cpp
svs-quic: if we no-relax then we cannot relax to the compressed instructions. This is how it is for the other relaxations:

bool canRelaxToGP =
      config().options().getRISCVRelax() &&
      config().options().getRISCVGPRelax()

bool canCompressLUI = config().options().getRISCVRelax() &&
                          config().options().getRISCVRelaxToC()

lib/Target/RISCV/RISCVLDBackend.cpp
svs-quic: Yes thanks, I had missed this in the initia PR.

lib/Target/RISCV/RISCVLDBackend.cpp
svs-quic: Thanks for pointing this out as well. I have changed the check to just S+A. Also added check to verify that it is uint 32 given that that is what the size of each jump table entry is.

parth-07 (approved):
LGTM.

svs-quic:
@quic-seaswara could you please approve this patch if you have no further comments?

quic-seaswara (approved):

Issue #1278, opened 2026-06-16
TLSDESC relaxations can incorrectly fire if they're next to a instruction marked `RELAX` that will be deleted


jonathonpenix:
**Reproducer**
```
cat > a.s << EOF
.text
foo:
.option relax
  lui a7, %hi(var)
.option norelax
.Ltlsdesc_hi1:
  auipc a4, %tlsdesc_hi(b)
  lw  a5, %tlsdesc_load_lo(.Ltlsdesc_hi1)(a4)
  addi  a0, a4, %tlsdesc_add_lo(.Ltlsdesc_hi1)
  jalr  t0, 0(a5), %tlsdesc_call(.Ltlsdesc_hi1)
  add   a0, a0, tp
.option relax
  lw a7, %lo(var)(a7)

.section .data
.globl var
var:
.word 5

.section .tbss
.tbss
.globl b
b:
.zero 4
EOF

cat > script.t << EOF
SECTIONS {
  .text 0x10000 : { *(.text) }
  .data : {
    __global_pointer$ = .;
    *(.data)
  }
  .tbss : { *(.tbss) }
}
EOF

llvm-mc --triple riscv32 -filetype=obj -mattr=+c,+relax a.s -o a.o
ld.eld a.o -T script.t -o a.out --emit-relocs
llvm-objdump -d a.out
```
The relaxation seems to fire (at least partially):
```
00010000 <foo>:
   10000: 00000013      nop
   10004: 00000513      li      a0, 0x0
   10008: 9512          add     a0, a0, tp
   1000a: 0001a883      lw      a7, 0x0(gp)
```

lld's output looks correct to me:
```
ld.lld a.o -T script.t -o a.out --emit-relocs --verbose --relax-gp
llvm-objdump -d a.out

# 00010000 <foo>:
#    10000: 00000013      nop
#    10004: 00000013      nop
#    10008: 00000013      nop
#    1000c: 00000513      li      a0, 0x0
#    10010: 9512          add     a0, a0, tp
#    10012: 0001a883      lw      a7, 0x0(gp)
```
**Issue**
[...]

PR #1289, opened 2026-06-17
Fix issues associated with incorrect application of R_RISCV_RELAX

### LINKED ISSUE #1276: eld reorders `PCREL_LO12_*`/`RELAX` relocs when the LO is a forward reference
**Reproducer**
The underlying issue can be reproduced using eld's tests [here](https://github.com/qualcomm/eld/tree/57d9d0031041c33e7c1ddf2dcefae3d54c538d1a/test/RISCV/standalone/Relocs/PCREL_HI_LO_FWD_REF) though this input doesn't necessarily cause problems (see 'Issue/Impact' below).

A modified/more contrived case to get the linker to actually mislink something as a result:
```
cat > red.s << EOF
        .section .data.num
num:
        .word   1
        .section .data.val
val:
        .word   2

        .section .text.test
        .globl  test
        .type   test,@function
test:
.Lpcrel_hi0:
        auipc   a0, %pcrel_hi(num)
        lw      s0, %pcrel_lo(.Lpcrel_hi0)(a0)
        j       .LBB0_2
.LBB0_1:
        .option norelax
        lui     a3, %hi(num)
        .option relax
        sw      zero, %pcrel_lo(.Lpcrel_hi5)(a0)
        .option norelax
        lw      a3, %lo(num)(a3)
        .option relax
.LBB0_2:
        call    foo@plt
.Lpcrel_hi5:
        auipc   a0, %pcrel_hi(val)
        lw      a1, %pcrel_lo(.Lpcrel_hi5)(a0)
        bnez    a1, .LBB0_1
        j       .LBB0_2

        .section .text.foo
        .globl  foo
        .type   foo,@function
foo:
        nop
EOF

cat > red_script.t << EOF
SECTIONS {
   .text (0x1233000) : {
    *(.text.test)
   }
   .anothertext (0x1233560) : {
    *(.text.foo)
   }
   .data (0x1234560) : {
     __global_pointer$ = .;
     *(.data.*)
   }
}
EOF

clang --target=riscv32-unknown-elf -c red.s
ld.eld --emit-relocs -T red_script.t red.o
```

If you run `llvm-objdump -d a.out`, note that the `lui` has been deleted and the LO (correctly) not GP relaxed (but now the HI has been deleted):
```
01233006 <.LBB0_1>:
 1233006: 54052a23      sw      zero, 0x554(a0)
 123300a: 5606a683      lw      a3, 0x560(a3)
```

If you run `llvm-objdump -r red.o a.out`, note that the relative ordering of `R_RISCV_PCREL_LO12_S` and the associated `R_RISCV_RELAX` has flipped:

red.o:
```
0000000a R_RISCV_HI20             num
0000000e R_RISCV_PCREL_LO12_S     .Lpcrel_hi5
0000000e R_RISCV_RELAX            *ABS*
00000012 R_RISCV_LO12_I           num
```

a.out:
```
01233006 R_RISCV_NONE             num
01233006 R_RISCV_RELAX            *ABS*
01233006 R_RISCV_PCREL_LO12_S     val
0123300a R_RISCV_LO12_I           num
```

**Issue**
I think the re-ordering happens as a result of how [pending relocations](https://github.com/qualcomm/eld/blob/57d9d0031041c33e7c1ddf2dcefae3d54c538d1a/lib/Target/RISCV/RISCVLDBackend.cpp#L1448) are handled and we ultimately do a [stable_sort](https://github.com/qualcomm/eld/blob/57d9d0031041c33e7c1ddf2dcefae3d54c538d1a/lib/Target/RISCV/RISCVLDBackend.cpp#L1593) to put the relocations back in offset order (though, as noted a slightly different 'absolute' order).

This then becomes a problem in [mayBeRelax](https://github.com/qualcomm/eld/blob/57d9d0031041c33e7c1ddf2dcefae3d54c538d1a/lib/Target/RISCV/RISCVLDBackend.cpp#L1192-L1199) where we check whether something is relaxable based on whether the **following** relocation is `R_RISCV_RELAX`. So you can end up in places where the HI part of some relocation pair looks relaxable, but the lo part is never considered (and like above, `mayBeRelax` just associates the RELAX with the wrong relocation)

Note this is also somewhat insidious too as ex: checking for the `R_RISCV_RELAX` at the particular offset (via ex: `findRelocation`) *succeeds* since it is at the right offset, just not in the order `mayBeRelax` expects.

**Impact**
Right now I think this is relatively low impact--we disable GP relaxation for pcrel cases like [this](https://github.com/qualcomm/eld/blob/57d9d0031041c33e7c1ddf2dcefae3d54c538d1a/lib/Target/RISCV/RISCVLDBackend.cpp#L1468) but I hit this when trying to implement GOT load relaxations. Presumably whatever other/future relaxations that involve `PCREL_LO12_*` and some HI relocation will have this issue too.

But again, you can write contrived cases to make the linker do bad things with this.

quic-akaryaki:
Two related changes:
- fix issues when R_RISCV_RELAX applied to wrong relocations (#1276 and #1278)
- remove "pending relocations" logic to fix quadratic time.

Fixes #1276 and #1278.

### DIFF
--- a/lib/Target/RISCV/RISCVLDBackend.cpp
+++ b/lib/Target/RISCV/RISCVLDBackend.cpp
@@ -1530,10 +1530,18 @@ void RISCVLDBackend::mayBeRelax(int relaxation_pass, bool &pFinished) {
         continue;
       auto relocList = rs->getLink()->getRelocations();
       for (llvm::SmallVectorImpl<Relocation *>::iterator it = relocList.begin();
-           it != relocList.end(); ++it) {
+           config().getDiagEngine()->diagnose() && it != relocList.end();
+           ++it) {
         auto relocation = *it;
-        // Check if the next relocation is a RELAX relocation.
         Relocation::Type type = relocation->type();
+
+        // Processing of R_RISCV_ALIGN is unconditional.
+        if (relaxation_pass == RELAXATION_ALIGN) {
+          if (type == llvm::ELF::R_RISCV_ALIGN && doRelaxationAlign(relocation))
+            pFinished = false;
+          continue;
+        }
+
         llvm::SmallVectorImpl<Relocation *>::iterator it2 = it + 1;
         Relocation *nextRelax = nullptr;
         if (it2 != relocList.end()) {
@@ -1542,38 +1550,15 @@ void RISCVLDBackend::mayBeRelax(int relaxation_pass, bool &pFinished) {
             nextRelax = nullptr;
         }
 
-        // try to relax
-        switch (type) {
-        case llvm::ELF::R_RISCV_CALL:
-        case llvm::ELF::R_RISCV_CALL_PLT: {
-          if (nextRelax && relaxation_pass == RELAXATION_CALL)
-            doRelaxationCall(relocation);
-          break;
-        }
-        case llvm::ELF::R_RISCV_JAL: {
-          if (nextRelax && relaxation_pass == RELAXATION_CALL)
-            doRelaxationJal(relocation);
-          break;
-        }
-        case llvm::ELF::R_RISCV_PCREL_HI20:
-        case llvm::ELF::R_RISCV_PCREL_LO12_I:
-        case llvm::ELF::R_RISCV_PCREL_LO12_S: {
-          if (nextRelax && relaxation_pass == RELAXATION_PC)
-            doRelaxationPC(relocation, GP);
-          break;
-        }
-        case llvm::ELF::R_RISCV_LO12_S:
-        case llvm::ELF::R_RISCV_LO12_I:
-        case llvm::ELF::R_RISCV_HI20: {
-          if (nextRelax && relaxation_pass == RELAXATION_LUI)
-            doRelaxationLui(relocation, GP);
-          break;
-        }
-        case llvm::ELF::R_RISCV_TLSDESC_HI20:
-        case llvm::ELF::R_RISCV_TLSDESC_LOAD_LO12:
-        case llvm::ELF::R_RISCV_TLSDESC_ADD_LO12:
-        case llvm::ELF::R_RISCV_TLSDESC_CALL:
-          if (relaxation_pass == RELAXATION_TLSDESC) {
+        if (relaxation_pass == RELAXATION_TLSDESC) {
+          // doRelaxationTLSDESC is used for both TLSDESC optimizations and
+          // relaxations, therefore this function should be called regardless
+          // of whether relaxations are enabled.
+          switch (type) {
+          case llvm::ELF::R_RISCV_TLSDESC_HI20:
+          case llvm::ELF::R_RISCV_TLSDESC_LOAD_LO12:
+          case llvm::ELF::R_RISCV_TLSDESC_ADD_LO12:
+          case llvm::ELF::R_RISCV_TLSDESC_CALL:
             // In the TLSDESC relaxation sequence, only the instruction with
             // R_RISCV_TLSDESC_HI20 can be marked with R_RISCV_RELAX to indicate
             // that the whole sequence is relaxable. So the other three
@@ -1583,25 +1568,48 @@ void RISCVLDBackend::mayBeRelax(int relaxation_pass, bool &pFinished) {
               if (const Relocation *HIReloc = getBaseReloc(*relocation))
                 nextRelax = rs->getLink()->findRelocation(
                     HIReloc->targetRef()->offset(), llvm::ELF::R_RISCV_RELAX);
-            // Note that doRelaxationTLSDESC is used for both optimizations and
-            // relaxations, therefore this function should be called regardless
-            // of whether relaxations are enabled.
             doRelaxationTLSDESC(*relocation, nextRelax);
+            break;
+          }
+          continue;
+        }
+
+        // try to relax
+        if (!nextRelax)
+          continue;
+
+        switch (relaxation_pass) {
+        case RELAXATION_CALL:
+          switch (type) {
+          case llvm::ELF::R_RISCV_CALL:
+          case llvm::ELF::R_RISCV_CALL_PLT:
+            doRelaxationCall(relocation);
+            break;
+          case llvm::ELF::R_RISCV_JAL:
+            doRelaxationJal(relocation);
+            break;
+          case ELF::riscv::internal::R_RISCV_QC_E_CALL_PLT:
+            doRelaxationQCCall(relocation);
+            break;
           }
           break;
-        case llvm::ELF::R_RISCV_ALIGN: {
-          if (relaxation_pass == RELAXATION_ALIGN)
-            if (doRelaxationAlign(relocation))
-              pFinished = false;
+        case RELAXATION_PC:
+          switch (type) {
+          case llvm::ELF::R_RISCV_PCREL_HI20:
+          case llvm::ELF::R_RISCV_PCREL_LO12_I:
+          case llvm::ELF::R_RISCV_PCREL_LO12_S:
+            doRelaxationPC(relocation, GP);
+            break;
+          }
           break;
-        }
-        case ELF::riscv::internal::R_RISCV_QC_E_CALL_PLT: {
-          if (nextRelax && relaxation_pass == RELAXATION_CALL)
-            doRelaxationQCCall(relocation);
-          break;
-        }
-        case ELF::riscv::internal::R_RISCV_QC_E_32: {
-          if (nextRelax && relaxation_pass == RELAXATION_LUI) {
+        case RELAXATION_LUI:
+          switch (type) {
+          case llvm::ELF::R_RISCV_LO12_S:
+          case llvm::ELF::R_RISCV_LO12_I:
+          case llvm::ELF::R_RISCV_HI20:
+            doRelaxationLui(relocation, GP);
+            break;
+          case ELF::riscv::internal::R_RISCV_QC_E_32: {
             uint64_t access_offset = relocation->targetRef()->offset() + 6;
             bool relaxed = false;
             if (Relocation *acc32 = rs->getLink()->findRelocation(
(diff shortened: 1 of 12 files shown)


lenary:
This is one approach but there might be a simpler one. How about:

Before we call relaxDeleteBytes, we can go through and change the `R_RISCV_RELAX` that caused us to perform the relaxation (i.e. at the current offset) into an `R_RISCV_NONE`. This then ensures that when we delete bytes and move the following relocations a little earlier, that we don't have a leftover `R_RISCV_RELAX` on the "wrong" offset.

This might be simpler than per-target relocation flags? Alternatively we could have an internal relocation that meant "there was a R_RISCV_RELAX here, but it relates to something deleted", which might make reconstructing reasonable output for emit-relocs easier.

quic-akaryaki:
> This is one approach but there might be a simpler one. How about:
> 
> Before we call relaxDeleteBytes, we can go through and change the `R_RISCV_RELAX` that caused us to perform the relaxation (i.e. at the current offset) into an `R_RISCV_NONE`. This then ensures that when we delete bytes and move the following relocations a little earlier, that we don't have a leftover `R_RISCV_RELAX` on the "wrong" offset.
> 
> This might be simpler than per-target relocation flags? Alternatively we could have an internal relocation that meant "there was a R_RISCV_RELAX here, but it relates to something deleted", which might make reconstructing reasonable output for emit-relocs easier.

There are two approaches, not one? The first one is simple, but it will remove R_RISCV_RELAX from -emit-relocs output. It is debatable whether they should be there, but correctly they are, and lld keeps them too.

The second one (internal temporary relocation) does not seem simper than a flag. It is really easier to check a bit than "what is the next relocation?" every time so the code that deals with R_RISCV_RELAX can be in one location. Adding internal relocations is also not simple.

However, I think a longer-term direction would be to not load R_RISC_RELAX internally at all to reduce memory footprint, and later reconstruct them in -emit-relocs if needed. I looked into this, but we would need a target-specific hook in emitRelocation(). I didn't feel like I want to go that far.

lib/Target/RISCV/RISCVLDBackend.h
```diff
@@ -31,6 +31,12 @@ class RISCVRelaxationStats;
 /// RISCVLDBackend - linker backend of RISCV target of GNU ELF format
 ///
 class RISCVLDBackend : public GNULDBackend {
+
+  // Relocation backend flags.
+  enum {
```
lenary: Can you expose a Typedef from `Relocation.h` so that this can be given an underlying type to match the type of backend flags?

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -1385,6 +1390,11 @@ bool RISCVLDBackend::handleRelocation(ELFSection *pSection,
   // real symbols but not this one. We need to map it to null, otherwise
   // --emit relocs will not find symbol in index map.
   case llvm::ELF::R_RISCV_RELAX: {
+    auto Relocs = pSection->getRelocations(); // llvm::iterator_range
+    if (!Relocs.empty()) {
+      Relocation &LastR = **(Relocs.end() - 1);
+      LastR.setBackendFlags(LastR.backendFlags() | REL_FLAG_RELAXABLE);
```
lenary: This doesn't work if the R_RISCV_RELAX comes before the relocation, which I think can happen in #1276

include/eld/Readers/Relocation.h
```diff
@@ -35,7 +35,7 @@ class Relocation {
   typedef uint64_t DWord;   // FIXME: use SizeTrait<T>::Word instead
   typedef int64_t SWord;    // FIXME: use SizeTrait<T>::SWord instead
   typedef uint32_t Size;
-  typedef uint32_t Type;
+  typedef uint16_t Type;
```
quic-seaswara: `#define ELF64_R_TYPE(i)                 ((i) & 0xffffffff)`

We need this to be 32bit :( 

Sorry for my bad suggestion

include/eld/Readers/Relocation.h
```diff
@@ -93,6 +93,9 @@ class Relocation {
 
   size_t getOffset() const;
 
+  /// backendFlags - backend-specific flags
+  uint16_t backendFlags() const { return m_BackendFlags; }
```
quic-seaswara: getBackendFlags

lib/Target/RISCV/RISCVLDBackend.h
quic-akaryaki: I don't think this will add any value because the caller will pass an int (or unsigned int) anyway (due to C promotion rules). If we are paranoic, I can a check for valid bits and use unsigned as the parameter type.

quic-seaswara (commented):
It looks like we need to deal with creating a unordered map of Relocation * to bool.

lib/Target/RISCV/RISCVLDBackend.cpp
quic-akaryaki: I haven't seen #1276 ... I guess this can be fixed in #1276?

quic-akaryaki: Yes, this was the critical flaw in my first version. I assign them now after sorting.

include/eld/Readers/Relocation.h
quic-akaryaki: Backend flags are gone.

lib/Target/RISCV/RISCVLDBackend.h
```diff
@@ -332,7 +337,10 @@ class RISCVLDBackend : public GNULDBackend {
   /// A map to keep track of the relocation that defines the base address for
   /// relative relocations. This is a concept in RISC-V and applies to
   /// relocations consisting of a HI20 and LO12 pairs.
-  llvm::DenseMap<const Relocation *, const Relocation *> m_BaseRelocs;
```
quic-akaryaki, line 335: I had to make the pointed relocation pointer non-const because R_RISCV_QC_E_32 relaxation updates the access relocation and m_BaseRelocs are used to store pointers to access relocation for it.

quic-akaryaki:
I the latest push, I reverted the behavior change to apply R_RISCV_RELAX to all relocations at one offset. Now and as before, R_RISCV_RELAX is only applied to the previous relocation.

lib/Target/RISCV/RISCVLDBackend.h
```diff
@@ -56,6 +57,10 @@ class RISCVLDBackend : public GNULDBackend {
 
   bool initStubFactory() override;
 
+  bool HasRelax(const Relocation &R) const {
```
quic-seaswara: nit : hasRelax ?

lib/Target/RISCV/RISCVLDBackend.h
```diff
@@ -332,7 +333,10 @@ class RISCVLDBackend : public GNULDBackend {
   /// A map to keep track of the relocation that defines the base address for
   /// relative relocations. This is a concept in RISC-V and applies to
   /// relocations consisting of a HI20 and LO12 pairs.
-  llvm::DenseMap<const Relocation *, const Relocation *> m_BaseRelocs;
+  llvm::DenseMap<const Relocation *, Relocation *> m_BaseRelocs;
+
+  /// Identify relocations that have an associated R_RISCV_RELAX.
+  llvm::DenseSet<const Relocation *> m_RelocsWithRelax;
```
quic-seaswara, line 339: can this be llvm:DenseSet<Relocation *> ?

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -1565,6 +1534,14 @@ void RISCVLDBackend::mayBeRelax(int relaxation_pass, bool &pFinished) {
         auto relocation = *it;
         // Check if the next relocation is a RELAX relocation.
         Relocation::Type type = relocation->type();
+
+        // Processing of R_RISCV_ALIGN is unconditional.
+        if (relaxation_pass == RELAXATION_ALIGN) {
+          if (type == llvm::ELF::R_RISCV_ALIGN && doRelaxationAlign(relocation))
+            pFinished = false;
+          continue;
```
quic-seaswara, line 1542: we might want to return if there is failure 

if (!diagnose()) return false;

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -1614,25 +1568,48 @@ void RISCVLDBackend::mayBeRelax(int relaxation_pass, bool &pFinished) {
...
             doRelaxationTLSDESC(*relocation, nextRelax);
+            break;
           }
-          break;
-        case llvm::ELF::R_RISCV_ALIGN: {
-          if (relaxation_pass == RELAXATION_ALIGN)
-            if (doRelaxationAlign(relocation))
-              pFinished = false;
-          break;
+          continue;
```
quic-seaswara, line 1570: Same thing, probably if there is an error return from this function

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -1565,6 +1534,14 @@ void RISCVLDBackend::mayBeRelax(int relaxation_pass, bool &pFinished) {
         auto relocation = *it;
         // Check if the next relocation is a RELAX relocation.
```
quic-seaswara: Remove this comment ?

quic-seaswara (commented):
some initial comments, will review in detail shortly.

quic-akaryaki:
@lenary ping pls.

lib/Target/RISCV/RISCVLDBackend.h
quic-akaryaki: I would like keep the const because there is no need to update it through this set.

quic-seaswara (approved):

lenary (approved):
LGTM, thanks for fixing this, how it works is now quite a bit clearer to me with your offline explanations.

PR #1290, opened 2026-06-18
Add primitive support for versioned symbols in regular object files

parth-07:
This patch adds support for properly resolving versioned symbols in
regular object files.

The patch extends the canonical/non-canonical alias-pair model that is
used for shared-library versioned symbols to the relocatable object versioned
symbols.

Detailed developer documentation for symbol resolution of versioned
symbols with examples is added in docs/DeveloperDocs/SymbolVersioning.md
as part of this patch.

This patch does not add support for local versioned symbols. The local
versioned symbol support will be added as a follow-up patch.

### DIFF
--- a/docs/DeveloperDocs/SymbolVersioning.md
+++ b/docs/DeveloperDocs/SymbolVersioning.md
@@ -1,0 +1,425 @@
+# Symbol versioning
+
+All symbol versioning related functionality is guarded by the `ELD_ENABLE_SYMBOL_VERSIONING`
+build macro.
+
+## Symbol resolution of versioned symbols
+
+A versioned symbol is of two types:
+
+- **default** versioned symbol (`bar@@V1`): a default versioned symbol
+  satisfies undefined references of both the versioned and
+  unversioned references of the symbol. `bar@@V1` definition can satisfy
+  undefined references of both `bar@V1` and plain `bar`.
+- **non-default** versioned symbol (`bar@V1`): a non-default versioned
+  symbol only satisfies undefined references of the versioned references
+  of the symbol. `bar@V1` definition can only satisfy undefined references
+  of `bar@V1`.
+
+eld supports symbol resolution of versioned symbols by inserting two
+symbols for default-versioned symbols into the already-existing symbol resolution
+machinery and adding a symbol normalization phase that runs after symbol resolution
+to combine the two symbol nodes for the default-versioned symbols into one, if required.
+
+Whenever eld sees a non-default-versioned symbol defintion `bar@V1`, then eld
+inserts `bar@V1` to the symbol resolution machinery, and it automatically
+gets used to resolve `bar@V1` undefined references.
+
+Here is the interesting part, whenever eld sees a default-versioned symbol
+definition `bar@@V1`, then eld inserts two symbols to the symbol resolution
+machinery: a canonical symbol `bar@V1` and a non-canonical symbol `bar`.
+With this, `bar@@V1` definition resolves undefined references to both
+`bar` and `bar@V1`, as it should.
+
+One fundamental rule of versioned-symbol resolution is that, for a default
+versioned symbol, both the canonical symbol (`foo@V1`) and the non-canonical
+(unversioned symbol `foo`) must resolve to the same definition. Allowing them
+to resolve to different symbols is incorrect because, the unversioned
+symbol is simply an alias for the versioned symbol (`foo@V1`), and both are
+originally the same symbol (`foo@@V1`).
+
+### Why is symbol normalization required?
+
+As we saw above, for a default-versioned symbol definition `bar@@V1`,
+eld creates two symbols `bar@V1` and `bar`. Some undefined symbol references
+may have resolved to `bar@V1` definition and some may have resolved to `bar`
+definition. This is a problem because in reality there is only one symbol
+`bar@@V1`. If we keep two symbol nodes per default-versioned symbol definition,
+then we will need to also emit both these symbols into the symbol table and
+create duplicate GOT/PLT slots for these symbols.
+
+Symbol normalization resolves this duplicate symbol issue by rewriting
+all the references to non-canonical symbols (`bar`) into the canonical symbols
+(`bar@V1`). A key thing to note here is that we only require the non-canonical
+aliases during symbol resolution phase.
+
+LLD does not always normalize the symbols and thus creates duplicate GOT/PLT slots for the same
+symbol when a default versioned symbol is accessed as both plain unversioned reference
+and versioned reference.
+
+Reproducer:
+
+```bash
+#!/usr/bin/env bash
+
+cat >1.c <<\EOF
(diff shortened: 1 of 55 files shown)


docs/DeveloperDocs/SymbolVersioning.md
```diff
@@ -0,0 +1,424 @@
...
+
+The order matters in these examples, (A) `bar@@V1`, (B) `bar@V2`, means
+that the linker first sees `bar@@V1` and then sees `bar@V2`.
+
+The symbols originates from regular object files unstated otherwise.
+
+1) (A) `bar@@V1`, (B) `bar@V1`
+
+mutliple definition error.
+
```
quic-seaswara, line 143: nit : spelling error

docs/DeveloperDocs/SymbolVersioning.md
```diff
@@ -0,0 +1,424 @@
...
+and versioned reference.
+
+Reproducer:
+
+```bash
+#!/usr/bin/env bash
+
+cat >1.c <<\EOF
+__asm__(".symver foo1,foo@@V1");
+int foo1() {
```
quic-seaswara, line 67: I am confused by this normalization, on how we will handle weak and strong symbols.

Though the behavior is right, I am unable to trace how is it being handled.

```
cat > lib.c << \!
__asm__(".symver foo, foo@@V1");
__attribute__((weak)) int foo() { return 1; }
int bar() { return foo(); }
!

cat > strong.c << \!
int foo() { return 99; }
!

cat > main.c << \!
int foo();
int _start() {
  return foo();
}
!

# weak reference to non-default version foo@V1 (single @)
cat > weakref.s << \!
        .section .text.bar
        .global bar
        .weak foo@V1
        .symver foo, foo@V1
        jmp foo
!

clang -c lib.c -fPIC
clang -c strong.c -fPIC
clang -c main.c
clang -c weakref.s

$ELD weakref.o lib.o strong.o main.o
```

lib/SymbolResolver/IRBuilder.cpp
```diff
@@ -238,76 +238,153 @@ LDSymbol *IRBuilder::addSymbolFromObject(
...
+  };
+
+#ifdef ELD_ENABLE_SYMBOL_VERSIONING
+  ParsedVersionedName P = parseVersionedName(SymbolName);
+
+  if (P.IsMalformed) {
+    ThisConfig.raise(Diag::error_malformed_versioned_symbol_name)
+        << Input.getInput()->decoratedPath() << SymbolName;
+    return nullptr;
   }
```
quic-seaswara: This should be handled in readSymbols.

test/Common/standalone/SymbolVersioning/VersionedSymbolsInRelocatableObjects/ValidResolution.test
```diff
@@ -0,0 +1,166 @@
...
+ExecDefault-DAG:           bar@@V1
+
+## Case 11: executable whose static .symtab must contain bar@V1 and no @@.
+RUN: echo '__asm__(".symver bar_v1, bar@V1"); extern int bar_v1(void); int main(void){return bar_v1();}' > %t.main_nondefault.c
+RUN: %clang %clangopts -o %t.main_nondefault.o %t.main_nondefault.c -c
+RUN: %link %linkopts -o %t.exec_nondefault.exe %t.strong_nondefault.o %t.main_nondefault.o
+RUN: %readelf -sW %t.exec_nondefault.exe | %filecheck %s --check-prefix=ExecNonDefault
+ExecNonDefault-DAG:        bar@V1
+ExecNonDefault-NOT:        bar@@V1
+#END_TEST
```
quic-seaswara, line 129: One test have 11 cases means this needs to be split.

quic-seaswara (changes requested):
This particular change deserves the google unit tests for symbol versioning to capture symbol versioning. Please add that in this commit. 

This commit is changing a lot of things simultaneously, introducing canonical symbols, and symbol resolution, also handles all the symbol resolution cases in one single commit. 

Smaller commits make it easy to review as well. 

I prefer this commit be split as it touches a lot of critical code. 

I am extremely sorry for the late review.

docs/DeveloperDocs/SymbolVersioning.md
parth-07: Fixed!

docs/DeveloperDocs/SymbolVersioning.md
parth-07: We handle the above example as follows:

Linker creates and resolves the symbols in this order:

1) UND weakref.o(foo@V1)
    - `UND foo@V1` is inserted in the name pool.

2) WEAK lib.o(foo@@V1)
    - `WEAK foo@V1` and `WEAK foo` is inserted in the name pool. `lib.o(foo)` is an alias of `lib.o(foo@V1)`.
    - `foo@V1` resolveInfo now points to `lib.o(foo@V1)`
    
3) DEF strong.o(foo)
    - `foo` is inserted into the name pool.
    - `foo` resolveInfo now points to `strong.o(foo)`

4) UND main.o(foo)
    - `UND foo` is inserted into the name pool.

In the symbol normalization phase, the linker sees that the alias `foo` of `foo@V1` have been resolved to
different symbols and tries to resolve them by using the already-existing symbol resolution machinery (StaticResolver). `lib.o(foo)` is resolved to `strong.o(foo)` and `lib.o(foo@V1)` is resolved to `lib.o(foo@V1)`.
`lib.o(foo)` wins the resolution here because the strong definition wins in `strong.o(foo)` vs `lib.o(foo@V1)`.
The normalization phase then collapses lib.o(foo) and lib.o(foo@V1) into one symbol that points to the
resolution winner, which is `strong.o(foo)` in this case.

lib/SymbolResolver/IRBuilder.cpp
parth-07: Yes, I have moved this check in readSymbols.

test/Common/standalone/SymbolVersioning/VersionedSymbolsInRelocatableObjects/ValidResolution.test
parth-07: I have split the test into 4 tests now.

parth-07:
> This particular change deserves the google unit tests for symbol versioning to capture symbol versioning. Please add that in this commit.

I have added google unit tests for symbol versioning.

> This commit is changing a lot of things simultaneously, introducing canonical symbols, and symbol resolution, also handles all the symbol resolution cases in one single commit.
>
> Smaller commits make it easy to review as well.

> I prefer this commit be split as it touches a lot of critical code.

Most of the changes are due to the tests. Out of the 1811 lines added, 1300+ lines are from the tests. It will be difficult to break the canonical symbols change and symbol resolution in separate commits because they are very closely dependent upon each other.

quic-seaswara (approved):

PR #1305, opened 2026-06-19
[Input] Fix null dereference in resolvePathMappingFile

### LINKED ISSUE #1335: ELD crashes when input file is missing from reproduce tarball or mapping file contains incorrect mapping
ELD crashes with a segfault instead of reporting a clean fatal error  when a file referenced in a linker script is missing from the reproduce 
tarball or the mapping.ini contains an incorrect mapping pointing to a  non-existent file.

**Reproducer:**
```bash
mkdir -p A/lib64 A/scripts

echo 'int foo() { return 1; }' > 1.c
echo 'int main() { return 0; }' > main.c
echo 'GROUP(/lib64/lib1.so)' > A/scripts/script.t

clang -o lib1.so 1.c -shared -fPIC
clang -o main.o main.c -c
mv lib1.so A/lib64/lib1.so

ld.eld --sysroot=A -T A/scripts/script.t main.o --reproduce a.tar -o a.out
rm -rf unpacked && mkdir unpacked
tar -xvf a.tar -C unpacked --strip-components=1
cd unpacked
rm -rf SharedLibrary
bash -x response.txt
```

**Expected:** `Fatal: cannot read file /lib64/lib1.so`
**Actual:** SIGSEGV

Related to #818

deepakshirkem:
ELD crashes with a segfault instead of reporting a clean fatal error when a file referenced in a linker script is missing from the reproduce tarball or mapping.ini contains an incorrect mapping.

The root cause is `Input::resolvePathMappingFile()` calls `createMemoryArea()` which returns `nullptr` when the file does not exist, but still calls `setMemArea(nullptr)` and returns `true`, propagating a null `MemArea` to `readAndProcessInput()` where it is dereferenced by `ScriptLexer`.

Fixes #1335

### DIFF
--- a/lib/Input/Input.cpp
+++ b/lib/Input/Input.cpp
@@ -68,6 +68,8 @@ bool Input::resolvePathMappingFile(const LinkerConfig &PConfig) {
       Input::getMemoryAreaForPath(FileName, PConfig.getDiagEngine());
   if (!InputMem)
     InputMem = createMemoryArea(FileName, PConfig.getDiagEngine());
+  if (!InputMem)
+    return false; // File does not exist;
   setMemArea(InputMem);
   // All queries to return the name of the Input return FileName for the main
   // driver.
--- a/test/Common/standalone/MissingMappingFileInput/MissingMappingFileInput.test
+++ b/test/Common/standalone/MissingMappingFileInput/MissingMappingFileInput.test
@@ -1,0 +1,26 @@
+#UNSUPPORTED: windows, reproduce_fail
+#---MissingMappingFileInput.test----------- Executable -----------------#
+#BEGIN_COMMENT
+# Verify that ELD does not crash when a reproduce tarball is replayed
+# and an input file referenced in a linker script is missing from the
+# sysroot. ELD should report a fatal error instead of segfaulting.
+#END_COMMENT
+#START_TEST
+RUN: rm -rf %t.dir && mkdir -p %t.dir/lib64 %t.dir/scripts
+RUN: %clang %clangopts -c %p/Inputs/main.c -o %t.main.o
+RUN: %clang %clangopts -c %p/Inputs/lib1.c -o %t.lib1.o -fPIC
+RUN: %link %linkopts -shared %t.lib1.o -o %t.dir/lib64/lib1.so
+RUN: %cp %p/Inputs/script.t %t.dir/scripts/script.t
+RUN: %link %linkopts %t.main.o \
+RUN:   --sysroot=%t.dir \
+RUN:   -T %t.dir/scripts/script.t \
+RUN:   --reproduce %t.tar \
+RUN:   --dump-response-file %t.response \
+RUN:   -o %t.out
+RUN: mkdir -p %t.reproduce
+RUN: %tar %gnutaropts -xvf %t.tar -C %t.reproduce --strip-components=1
+RUN: rm -rf %t.reproduce/SharedLibrary
+RUN: rm -rf %t.dir/lib64/lib1.so
+RUN: cd %t.reproduce && %not %link %linkopts @%t.response 2>&1 | %filecheck %s
+#END_TEST
+CHECK: cannot read file /lib64/lib1.so
--- a/test/Common/standalone/MissingMappingFileInput/Inputs/lib1.c
+++ b/test/Common/standalone/MissingMappingFileInput/Inputs/lib1.c
@@ -1,0 +1,1 @@
+int foo() { return 1; }
--- a/test/Common/standalone/MissingMappingFileInput/Inputs/main.c
+++ b/test/Common/standalone/MissingMappingFileInput/Inputs/main.c
@@ -1,0 +1,1 @@
+int main() { return 0; }
--- a/test/Common/standalone/MissingMappingFileInput/Inputs/script.t
+++ b/test/Common/standalone/MissingMappingFileInput/Inputs/script.t
@@ -1,0 +1,1 @@
+GROUP(/lib64/lib1.so)


parth-07 (commented):
I don't understand why is the PR modifying an already exiting test and how is it fixing anything by not modifying the source-code at all.

deepakshirkem:
Hi @parth-07, The PR has been update.  It now includes the fix in Input.cpp and  a correct test. Please review when you get a chance!

deepakshirkem:
@parth-07  ping.

test/Common/standalone/MissingMappingFileInput/MissingMappingFileInput.test
```diff
@@ -0,0 +1,26 @@
...
+RUN:   --sysroot=%t.dir \
+RUN:   -T %t.dir/scripts/script.t \
+RUN:   --reproduce %t.tar \
+RUN:   --dump-response-file %t.response \
+RUN:   -o %t.out
+RUN: mkdir -p %t.reproduce
+RUN: %tar %gnutaropts -xvf %t.tar -C %t.reproduce --strip-components=1
+RUN: rm -rf %t.reproduce/SharedLibrary
+RUN: rm -rf %t.dir/lib64/lib1.so
+RUN: cd %t.reproduce && %not %link --no-threads @%t.response 2>&1 | %filecheck %s
```
parth-07: Why do we need --no-threads here?

test/Common/standalone/MissingMappingFileInput/MissingMappingFileInput.test
```diff
@@ -0,0 +1,26 @@
+#UNSUPPORTED: windows, reproduce_fail
```
parth-07, line 1: Why is this test unsupported for windows and reproduce_fail?

test/Common/standalone/MissingMappingFileInput/MissingMappingFileInput.test
parth-07: The fix for https://github.com/qualcomm/eld/issues/818 is not avoiding the crash and reporting an error message. It should instead be making reproducer work correctly for the case when the shared library should be found under the sysroot.

lib/Input/Input.cpp
```diff
@@ -69,6 +69,8 @@ bool Input::resolvePathMappingFile(const LinkerConfig &PConfig) {
       Input::getMemoryAreaForPath(FileName, PConfig.getDiagEngine());
   if (!InputMem)
     InputMem = createMemoryArea(FileName, PConfig.getDiagEngine());
+  if (!InputMem)
+    return false; // File does not exist;
```
parth-07, line 72: This fix is correct for the case when the reproduce tarball is tampered, but it is a wrong fix for https://github.com/qualcomm/eld/issues/818. In the issue, there is no tampering of the reproduce tarball. The core issue is not that the linker is crashing, but that the linker is incorrectly handling the sysroot during reproduce tarball. 

Can you create a separate issue for the link crashing when the files has been removed from the reproduce tarball/setup (or when the mapping.ini file contains incorrect mapping) and then link this PR with that issue?

parth-07 (changes requested):

test/Common/standalone/MissingMappingFileInput/MissingMappingFileInput.test
deepakshirkem: >Why do we need --no-threads here?

was used to match  the original link command in the reproduce tarball's response.txt will update the test to use %link with %linkopts instead,

test/Common/standalone/MissingMappingFileInput/MissingMappingFileInput.test
deepakshirkem: After checking existing reproduce tests (Reproduce.test, ArchiveFile.test, Namespec.test), all use the same `#UNSUPPORTED: windows, reproduce_fail` pattern since our test uses `--reproduce` flag and Unix-specific commands, so this is correct.

test/Common/standalone/MissingMappingFileInput/MissingMappingFileInput.test
deepakshirkem: Yeah this makes sense to me. Thank-You

deepakshirkem:
@parth-07 Ping++ ::))

test/Common/standalone/MissingMappingFileInput/MissingMappingFileInput.test
parth-07: Can you please remove this UNSUPPORTD line? I believe it is not required. You will need to update the `tar` usage in the test as was done in this PR: https://github.com/qualcomm/eld/pull/1709

test/Common/standalone/MissingMappingFileInput/MissingMappingFileInput.test
```diff
@@ -0,0 +1,26 @@
...
+#---MissingMappingFileInput.test----------- Executable -----------------#
+#BEGIN_COMMENT
+# Verify that ELD does not crash when a reproduce tarball is replayed
+# and an input file referenced in a linker script is missing from the
+# sysroot. ELD should report a fatal error instead of segfaulting.
+#END_COMMENT
+#START_TEST
+RUN: rm -rf %t.dir && mkdir -p %t.dir/lib64 %t.dir/scripts
```
parth-07:
```suggestion
RUN: %rm %t.dir && mkdir -p %t.dir/lib64 %t.dir/scripts
```

test/Common/standalone/MissingMappingFileInput/MissingMappingFileInput.test
```diff
@@ -0,0 +1,26 @@
...
+RUN: %link %linkopts %t.main.o \
+RUN:   --sysroot=%t.dir \
+RUN:   -T %t.dir/scripts/script.t \
+RUN:   --reproduce %t.tar \
+RUN:   --dump-response-file %t.response \
+RUN:   -o %t.out
+RUN: mkdir -p %t.reproduce
+RUN: %tar %gnutaropts -xvf %t.tar -C %t.reproduce --strip-components=1
+RUN: rm -rf %t.reproduce/SharedLibrary
+RUN: rm -rf %t.dir/lib64/lib1.so
```
parth-07:
```suggestion
RUN: %rm %t.reproduce/SharedLibrary
RUN: %rm %t.dir/lib64/lib1.so
```

parth-07 (commented):
Please shorten the commit title to fit 72 characters, and update the PR description. PR description is incorrect. This PR has nothing to do with sysroot files.

test/Common/standalone/MissingMappingFileInput/MissingMappingFileInput.test
deepakshirkem: Done.

test/Common/standalone/MissingMappingFileInput/MissingMappingFileInput.test
deepakshirkem: Done.

deepakshirkem:
@parth-07 Fixed the suggested changes. Please do one more cycle of review when you get a chance?

test/Common/standalone/MissingMappingFileInput/MissingMappingFileInput.test
```diff
@@ -0,0 +1,25 @@
...
+#BEGIN_COMMENT
+# Verify that ELD does not crash when a reproduce tarball is replayed
+# and an input file referenced in a linker script is missing.
+# ELD should report a fatal error instead of segfaulting.
+#END_COMMENT
+#START_TEST
+RUN: %rm %t.dir && mkdir -p %t.dir/lib64 %t.dir/scripts
+RUN: %clang %clangopts -c %p/Inputs/main.c -o %t.main.o
+RUN: %clang %clangopts -c %p/Inputs/lib1.c -o %t.lib1.o -fPIC
+RUN: %link %linkopts -shared %t.lib1.o -o %t.dir/lib64/lib1.so
```
parth-07: @deepakshirkem Does the test case for the current fix needs to be this complicated? In this test, we are creating a sysroot that contains the linker script, and linker is implicitly finding the shared libraries in the sysroot because the linker script is in the sysroot. For the issue being fixed, do we need to have a sysroot, or a linker script? If no, then can you please make this test simpler?

parth-07 (changes requested):
Please shorten the commit title to fit 72 characters, and update the PR description. PR description is incorrect. This PR has nothing to do with sysroot files.

test/Common/standalone/MissingMappingFileInput/MissingMappingFileInput.test
deepakshirkem: @parth-07, you are right! The test was unnecessarily complicated. I have simplified it to directly test the fix without any sysroot or reproduce tarball setup The simplified test still fails without the fix and passes with it.

parth-07 (commented):
lib1.c file added in the test is not being used and thus should be removed.

deepakshirkem:
>lib1.c file added in the test is not being used and thus should be removed.

Done. Thank You

parth-07 (approved):

deepakshirkem:
@quic-seaswara @quic-areg ping.

deepakshirkem:
[Shankar Easwaran (@quic-seaswara)](https://github.com/quic-seaswara) [quic-areg](https://github.com/quic-areg) ping :))

PR #1307, opened 2026-06-19
[Diagnostics] Emit mismatch diagnostics as notes with --no-warn-mismatch

### LINKED ISSUE #1264: Downgrade severity of errors affected by --no-warn-mismatch
### Issue Description

Diagnostics affected by `--no-warn-mismatch` (such as `incompatible_architecture_versions`) are currently suppressed entirely.

This behavior is confusing for users who want visibility into mismatch information without treating it as a warning or error.

#### Desired Behavior
- Downgrade these diagnostics from errors or warnings to **notes** when they are affected by `--no-warn-mismatch`.
- **Always emit** the diagnostic, even when `--no-warn-mismatch` is present, instead of omitting it.

This ensures users consistently see relevant mismatch information while clearly signaling that it is informational rather than actionable.

Steven6798:
Diagnostics affected by --no-warn-mismatch
(e.g., incompatible_architecture_versions) are no longer suppressed entirely.

Instead, downgrade them to notes and always emit them, ensuring users have visibility into mismatch information without treating it as actionable warnings or errors.

Fix: #1264

### DIFF
--- a/lib/Diagnostics/DiagnosticEngine.cpp
+++ b/lib/Diagnostics/DiagnosticEngine.cpp
@@ -230,6 +230,25 @@ void DiagnosticEngine::ignoreLLVMError(llvm::Error E) {
     ASSERT(!IgnoreErr, "ignoreErr must always be false");
 }
 
+bool DiagnosticEngine::isWarnMismatch(DiagIDType Id) {
+  if (Id == Diag::warn_mismatch_enum_size ||
+      Id == Diag::incompatible_architecture ||
+      Id == Diag::incompatible_input_architecture)
+    return true;
+
+  return false;
+}
+
+bool DiagnosticEngine::isErrorMismatch(DiagIDType Id) {
+  if (Id == Diag::err_mismatch_r9_use ||
+      Id == Diag::incompatible_architecture_versions ||
+      Id == Diag::attribute_parsing_error ||
+      Id == Diag::err_unrecognized_input_file || Id == Diag::invalid_elf_class)
+    return true;
+
+  return false;
+}
+
 // Initializes the default diagnostic IDs.
 // Diagnostic IDs are formed of 2 components: diagnostic severity and base ID.
 // Both diagnostic severity and the base diagnostic ID can be extract out of the
--- a/lib/Readers/ELFReader.cpp
+++ b/lib/Readers/ELFReader.cpp
@@ -424,22 +424,35 @@ template <class ELFT>
 eld::Expected<bool> ELFReader<ELFT>::isCompatible() const {
   const InputFile &inputFile = this->m_InputFile;
   LinkerConfig &config = this->m_Module.getConfig();
+  bool WarnMismatch = config.options().warnMismatch();
 
-  if (config.options().warnMismatch() && !checkMachine())
-    return std::make_unique<plugin::DiagnosticEntry>(plugin::DiagnosticEntry(
+  if (!checkMachine()) {
+    auto DE = std::make_unique<DiagnosticEntry>(
         Diag::err_unrecognized_input_file,
-        {inputFile.getInput()->getResolvedPath().native(),
-         config.targets().getArch()}));
+        std::vector<std::string>{
+            inputFile.getInput()->getResolvedPath().native(),
+            config.targets().getArch()});
+    if (!WarnMismatch)
+      config.raiseDiagEntry(std::move(DE));
+    else
+      return std::move(DE);
+  }
 
   eld::Expected<bool> expCheckFlags = checkFlags();
   ELDEXP_RETURN_DIAGENTRY_IF_ERROR(expCheckFlags);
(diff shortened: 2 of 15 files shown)


quic-seaswara (approved):

test/Common/standalone/WrongArchObject/WrongArchInput.test
```diff
@@ -13,6 +13,6 @@ RUN: %not %link %emulation -Bdynamic --warn-mismatch %t.2.o -L%p/Inputs %t.lib.s
 RUN: %link %emulation -Bdynamic --no-warn-mismatch %t.2.o -L%p/Inputs %t.lib.so 2>&1 | %filecheck %s -check-prefix=NOERR --allow-empty
 
 CHECK: Error: cannot recognize the format of file {{[` -\(\)_A-Za-z0-9.\\/:]+}}.3.o'. object format or given target machine ({{[` -\(\)_A-Za-z0-9.\\/:]+}}) is incompatible.
-NOERR-NOT: Error: cannot recognize the format of file
+NOERR: Note: cannot recognize the format of file {{.*}}'. object format or given target machine ({{.*}}) is incompatible.
```
parth-07: Shouldn't the link error out if we cannot read the input file?

Steven6798: We are following GNU LD for the behavior of this flag.

lib/Target/RISCV/RISCVInfo.cpp
```diff
@@ -45,39 +45,38 @@ bool RISCVInfo::isABIFlagSet(uint64_t inputFlag, uint32_t ABIFlag) const {
...
-    return true;
+  bool NoWarnMismatch = m_Config.options().noWarnMismatch();
 
   assert(m_OutputFlag);
   if (isABIFlagSet(pFlag, llvm::ELF::EF_RISCV_FLOAT_ABI_SOFT) ^
       isABIFlagSet(*m_OutputFlag, llvm::ELF::EF_RISCV_FLOAT_ABI_SOFT)) {
     m_Config.raise(Diag::incompatible_architecture_versions)
         << flagString(pFlag) << pFile << flagString(*m_OutputFlag);
-    return false;
+    return NoWarnMismatch ? true : false;
```
quic-areg: can we `return true;` early if `NoWarnMismatch` is set and leave the rest of the checks alone?

Steven6798: You mean returning at the beginning of the function? (line 48)

lib/Diagnostics/DiagnosticEngine.cpp
```diff
@@ -154,6 +154,8 @@ DiagnosticEngine::Severity DiagnosticEngine::getSeverity(DiagIDType Id) {
     ADD_CASE(Warning);
     ADD_CASE(Trace);
     ADD_CASE(InternalError);
+    ADD_CASE(WarnMismatchWarning);
```
parth-07: Why are these part of diagnostic severity?

parth-07 (changes requested):

lib/Diagnostics/DiagnosticEngine.cpp
Steven6798: Shankar suggested this new diagnostic to abstract away the changes of severity from the targets and handle everything in the Diagnostic code.

lib/Readers/ELFReader.cpp
```diff
@@ -424,22 +424,37 @@ template <class ELFT>
...
+  if (!checkClass()) {
+    if (NoWarnMismatch) {
+      config.raise(Diag::invalid_elf_class)
+          << inputFile.getInput()->decoratedPath()
+          << config.targets().getArch();
+    } else {
+      return std::make_unique<plugin::DiagnosticEntry>(plugin::DiagnosticEntry(
+          Diag::invalid_elf_class,
+          {inputFile.getInput()->decoratedPath(), config.targets().getArch()}));
+    }
```
parth-07: What do you think about reducing the diagnostic duplication here using the pattern below?

```suggestion
    auto DE = std::make_unique<DiagnosticEntry>(Diag::invalid_elf_class, 
      std::vector<std::string>(inputFile.getInput()->decoratedPath(), 
        config.targets().getArch());
    if (NoWarnMismatch) {
      config.raiseDiagEntry(std::move(DE));
    } else {
      return std::move(DE);
    }
```

Steven6798: Looks great!

quic-shaksuma (approved):

lib/Target/RISCV/RISCVInfo.cpp
Steven6798: Resolved after speaking with Adarsha

parth-07 (approved):
LGTM.

PR #1308, opened 2026-06-19
Unify default behavior for --warn-mismatch flags

### LINKED ISSUE #1293: Default to --warn-mismatch when --[no-]warn-mismatch flags are not present
Currently, each architecture initializes differently the --[no-]warn-mismatch flags if they are not present in the command line. This must be removed and the default behavior be set to --warn-mismatch.

The follows issues #1214 and #1215

Steven6798:
Remove architecture-specific initialization of --[no-]warn-mismatch when not explicitly provided. Set the default behavior consistently to --warn-mismatch across all architectures.

Fix: qualcomm#1293

### DIFF
--- a/include/eld/Config/GeneralOptions.h
+++ b/include/eld/Config/GeneralOptions.h
@@ -58,8 +58,6 @@ public:
     StripAllSymbols
   };
 
-  enum class WarnMismatchMode { None, WarnMismatch, NoWarnMismatch };
-
   enum OrphanMode { Place, Warn, Error, Invalid };
 
   enum ErrorStyleType { gnu, llvm };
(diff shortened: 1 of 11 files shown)


parth-07 (approved):

include/eld/Config/GeneralOptions.h
```diff
@@ -481,25 +479,9 @@ class GeneralOptions {
...
-  }
-
-  bool hasOptionWarnNoWarnMismatch() const {
-    return (WarnMismatch != WarnMismatchMode::None);
-  }
-
-  bool noWarnMismatch() const {
-    return (WarnMismatch == WarnMismatchMode::NoWarnMismatch);
-  }
+  void setWarnMismatch(bool PEnable) { WarnMismatch = PEnable; }
```
quic-areg: can we change the argument name to `Enable`

lib/Target/AArch64/AArch64Info.h
```diff
@@ -63,11 +63,7 @@ class AArch64Info : public TargetInfo {
 
   std::string flagString(uint64_t flag) const override;
 
-  void initializeAttributes(InputBuilder &pBuilder) override {
-    // Warn on mismatch.
-    if (!m_Config.options().hasOptionWarnNoWarnMismatch())
-      m_Config.options().setWarnMismatch(false);
-  }
+  void initializeAttributes(InputBuilder &pBuilder) override {}
```
quic-areg: do we need to override the function at all anymore?

quic-areg (commented):
Can you add a test? IIRC before this patch we would allow 32-bit and 64-bit objects to be merged without a warning on RISC-V and ARM, and that should no longer be the case.

LGTM otherwise

include/eld/Config/GeneralOptions.h
Steven6798: Of course

lib/Target/AArch64/AArch64Info.h
Steven6798: No. I can remove all the empty overrides created by this fix. I kept it just in case it is needed in the future. What do you think?

quic-shaksuma (approved):

quic-areg (commented):
Test missing?

Steven6798:
> Test missing?

Sorry I completely missed your first comment. I'll update the patch

Steven6798:
> Can you add a test? IIRC before this patch we would allow 32-bit and 64-bit objects to be merged without a warning on RISC-V and ARM, and that should no longer be the case.

I updated the WrongArchInput test to include test cases without the --wan-mismatch flag. This should be enough to test the default behavior. Thanks for catching this.

PR #1314, opened 2026-06-19
[TBSS] Fix spurious PT_LOAD created for TBSS-only sections

### LINKED ISSUE #353: empty segment created when the image has only tbss

quic-seaswara:
TBSS (.tbss) sections occupy no file space and do not advance the dot counter; their VMA is shared with the section that *follows* them.  The program-header loop treated TBSS like any other allocatable section when deciding whether to open a new PT_LOAD, producing empty segments whenever the link contained only TBSS sections or TBSS appeared between two real sections.

After placing a TBSS section into a segment, the loop updated prev, prev_flag, prevOut, and prev_mem_region.  Because TBSS shares VMA with the section that follows it and does not advance the dot counter, advancing those pointers made the next real section compare against a non existent boundary.

The now-redundant isPrevTBSS variable is removed.

Resolves #353

### DIFF
--- a/lib/Target/CreateProgramHeaders.hpp
+++ b/lib/Target/CreateProgramHeaders.hpp
@@ -247,8 +247,6 @@ bool GNULDBackend::createProgramHdrs() {
       continue;
     }
 
-    bool isPrevTBSS = prev && prev->isTBSS();
-
     bool isPrevRelRO =
         config().options().hasRelro() && prev && isRelROSection(prev);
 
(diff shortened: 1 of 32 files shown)


quic-seaswara:
No changes from previous commit, there was an issue with RELRO handling (rebasing)

PR #1318, opened 2026-06-23
[GNULDBackend] Replace std::move with std::copy in sizeDynNamePools

### LINKED ISSUE #1301: `std::move(...)` in `GNULDBackend::sizeDynNamePools()` is not really moving
```cpp
std::move(PartitionBegin, RVect.end(), std::back_inserter(DynamicSymbols));
```

The above `std::move` call in `GNULDBackend::sizeDynNamePools()` is not moving anything. It is simply calling the move constructor on each pointer stored in `RVect` (`std::vector<ResolveInfo *>`) and storing the move-result in `DynamicSymbols`. However, there is no difference in moving and copying a pointer. We should change the `std::move` here to `std::copy` to make the code easier to understand and less misleading.

bsaharsh:
std::move on a range of raw pointers (ResolveInfo *) is identical to std::copy — raw pointers are trivially copyable and moving them does not null the source. Using std::move here is misleading.
Resolves #1301 .

### DIFF
--- a/lib/Target/GNULDBackend.cpp
+++ b/lib/Target/GNULDBackend.cpp
@@ -816,9 +816,8 @@ void GNULDBackend::sizeDynNamePools() {
 
     DynamicSymbols.push_back(LDSymbol::null()->resolveInfo());
 
-    // Move all the DynamicSymbols.
-    // FIXME: Is it really moving?
-    std::move(PartitionBegin, RVect.end(), std::back_inserter(DynamicSymbols));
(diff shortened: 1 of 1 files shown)


parth-07:
@bsaharsh Can you please make the clang-format builder happy, and add the 'Resolves #<issue-number>' in both the commit message and the PR description?

parth-07 (approved):

quic-seaswara (approved):

PR #1319, opened 2026-06-23
Fix PT_TLS p_filesz corruption when segment contains .tbss

quic-k:


lib/Target/GNULDBackend.cpp
```diff
@@ -2423,7 +2423,7 @@ bool GNULDBackend::setupSegment(ELFSegment *E) {
 
     if (cur->addr() < lower)
       lower = cur->addr();
-    if (cur->offset() < lower_offset)
+    if (!isCurBSS && cur->offset() < lower_offset)
```
quic-seaswara: .tbss is SHT_NOBITS and its `offset` is inherited from the previous output section. I would prefer that we just update the higher and lower address and skip offset updates for `TBSS` sections.

```
f (isTBSS && E->type() == llvm::ELF::PT_TLS) {
  // Only update memory bounds; skip file-offset tracking below.
  if (cur->addr() < lower)
    lower = cur->addr();
  if ((cur->addr() + cur->size()) > upper)
    upper = cur->addr() + cur->size();
  continue;
}

```
Since this code is super critical, can we accompany this by a run test that will run for all the targets.

quic-seaswara (changes requested):

quic-k:
closed in favour of https://github.com/qualcomm/eld/pull/1326

PR #1320, opened 2026-06-23
[RISCV] Implement the Xqccmt Qualcomm Vendor Extension

svs-quic:
This patch adds support for relaxing RISC-V calls and jumps to Xqccmt table-jump instructions. Xqccmt uses the same table-jump encoding as Zcmt, but uses the vendor mnemonics qc.cm.jt and qc.cm.jalt, and also supports a  qc.cm.jalt form that links through t0.

We extend the --relax-tbljal flag from a boolean option into a mode selector. The default --relax-tbljal  continues to enable the Zcmt relaxation and can also be explicitly enable through --relax-tbljal=zcmt , while --relax-tbljal=xqccmt selects Xqccmt relaxation. 

  Update the table-jump implementation to:
    - track whether the current table-jump mode is Zcmt or Xqccmt
    - select qc.cm.jt for tail calls and qc.cm.jalt for link-register calls
    - support the Xqccmt t0-link form by storing bit 0 in the JVT entry
    - keep ra and t0 call-table entries distinct even when they target the
    same symbol
    - preserve existing Zcmt behavior for the default --relax-tbljal mode

The Xqccmt spec can be found at https://github.com/quic/riscv-unified-db/releases/download/Xqccmt-0.1.0/Xqccmt-0.1.0.pdf

Assisted by gpt-5.5

### DIFF
--- a/lib/LinkerWrapper/RISCVLinkDriver.cpp
+++ b/lib/LinkerWrapper/RISCVLinkDriver.cpp
@@ -155,8 +155,8 @@ RISCVLinkDriver::parseOptions(ArrayRef<const char *> Args,
       TbljalMode =
           llvm::StringSwitch<GeneralOptions::RISCVRelaxTbljalMode>(
               Arg->getValue())
-              .Case("zcmt", GeneralOptions::RISCVRelaxTbljalMode::Zcmt)
-              .Case("xqccmt", GeneralOptions::RISCVRelaxTbljalMode::Xqccmt)
+              .CaseLower("zcmt", GeneralOptions::RISCVRelaxTbljalMode::Zcmt)
+              .CaseLower("xqccmt", GeneralOptions::RISCVRelaxTbljalMode::Xqccmt)
               .Default(GeneralOptions::RISCVRelaxTbljalMode::None);
       if (TbljalMode == GeneralOptions::RISCVRelaxTbljalMode::None) {
         Config.raise(Diag::invalid_value_for_option)
--- a/lib/LinkerWrapper/RISCVLinkDriver.cpp
+++ b/lib/LinkerWrapper/RISCVLinkDriver.cpp
@@ -140,19 +140,41 @@ RISCVLinkDriver::parseOptions(ArrayRef<const char *> Args,
   if (ArgList.hasArg(OPT_RISCVLinkOptTable::no_relax_tlsdesc))
     Config.options().setRISCVRelaxTLSDESC(false);
 
-  // --relax-tbljal, --no-relax-tbljal (default)
-  bool EnableTbljal =
-      ArgList.hasFlag(OPT_RISCVLinkOptTable::relax_tbljal,
-                      OPT_RISCVLinkOptTable::no_relax_tbljal, false);
-  if (EnableTbljal && (ArgList.hasArg(OPT_RISCVLinkOptTable::shared) ||
-                       ArgList.hasFlag(OPT_RISCVLinkOptTable::pie,
-                                       OPT_RISCVLinkOptTable::no_pie, false))) {
+  // --relax-tbljal[=zcmt|xqccmt], --no-relax-tbljal (default)
+  GeneralOptions::RISCVRelaxTbljalMode TbljalMode =
+      GeneralOptions::RISCVRelaxTbljalMode::None;
+  if (llvm::opt::Arg *Arg =
+          ArgList.getLastArg(OPT_RISCVLinkOptTable::relax_tbljal,
+                             OPT_RISCVLinkOptTable::relax_tbljal_eq,
+                             OPT_RISCVLinkOptTable::no_relax_tbljal)) {
+    if (Arg->getOption().matches(OPT_RISCVLinkOptTable::no_relax_tbljal)) {
+      TbljalMode = GeneralOptions::RISCVRelaxTbljalMode::None;
+    } else if (Arg->getOption().matches(OPT_RISCVLinkOptTable::relax_tbljal)) {
+      TbljalMode = GeneralOptions::RISCVRelaxTbljalMode::Zcmt;
+    } else {
+      TbljalMode =
+          llvm::StringSwitch<GeneralOptions::RISCVRelaxTbljalMode>(
+              Arg->getValue())
+              .Case("zcmt", GeneralOptions::RISCVRelaxTbljalMode::Zcmt)
+              .Case("xqccmt", GeneralOptions::RISCVRelaxTbljalMode::Xqccmt)
+              .Default(GeneralOptions::RISCVRelaxTbljalMode::None);
+      if (TbljalMode == GeneralOptions::RISCVRelaxTbljalMode::None) {
+        Config.raise(Diag::invalid_value_for_option)
+            << Arg->getOption().getPrefixedName() << Arg->getValue();
+        return LINK_FAIL;
+      }
+    }
+  }
+  if (TbljalMode != GeneralOptions::RISCVRelaxTbljalMode::None &&
+      (ArgList.hasArg(OPT_RISCVLinkOptTable::shared) ||
+       ArgList.hasFlag(OPT_RISCVLinkOptTable::pie,
+                       OPT_RISCVLinkOptTable::no_pie, false))) {
     Config.raise(Diag::not_supported)
-        << "Zcmt table jump relaxation"
+        << "Zcmt/Xqccmt table jump relaxation"
         << "shared libraries or position independent code";
     return LINK_FAIL;
   }
-  Config.options().setRISCVRelaxTbljal(EnableTbljal);
+  Config.options().setRISCVRelaxTbljal(TbljalMode);
 
   // --enable-bss-mixing
   if (ArgList.hasArg(OPT_RISCVLinkOptTable::enable_bss_mixing))
--- a/lib/Target/RISCV/RISCVRelocationHelper.h
+++ b/lib/Target/RISCV/RISCVRelocationHelper.h
@@ -90,6 +90,19 @@ enum Reg {
   return inst;
 }
 
+[[maybe_unused]] unsigned getQCEJumpRd(uint64_t insn) {
+  switch ((insn >> 15) & 0x3) {
+  case 0:
+    return X_ZERO;
+  case 1:
+    return X_RA;
+  case 2:
+    return X_T0;
+  default:
+    return 32;
+  }
+}
+
 // Extract bits v[begin:end], where range is inclusive, and begin must be < 63.
 [[maybe_unused]] uint32_t extractBits(uint64_t v, uint32_t begin,
                                       uint32_t end) {
--- a/lib/Target/RISCV/RISCVRelocationHelper.h
+++ b/lib/Target/RISCV/RISCVRelocationHelper.h
@@ -96,8 +96,6 @@ enum Reg {
     return X_ZERO;
   case 1:
     return X_RA;
-  case 2:
-    return X_T0;
   default:
     return 32;
   }
--- a/lib/Target/RISCV/RISCVRelocationHelper.h
+++ b/lib/Target/RISCV/RISCVRelocationHelper.h
@@ -91,14 +91,11 @@ enum Reg {
 }
 
 [[maybe_unused]] unsigned getQCEJumpRd(uint64_t insn) {
-  switch ((insn >> 15) & 0x3) {
-  case 0:
-    return X_ZERO;
-  case 1:
-    return X_RA;
-  default:
-    return 32;
-  }
+  return (insn >> 15) & 0x3;
+}
+
+[[maybe_unused]] bool isValidQCEJumpRd(unsigned Rd) {
+  return Rd == X_ZERO || Rd == X_RA;
 }
 
 // Extract bits v[begin:end], where range is inclusive, and begin must be < 63.
--- a/lib/Target/RISCV/RISCVTableJump.cpp
+++ b/lib/Target/RISCV/RISCVTableJump.cpp
@@ -4,16 +4,23 @@
 // SPDX-License-Identifier: BSD-3-Clause
 //===----------------------------------------------------------------------===//
 //
-// This file implements RISC-V Zcmt table-jump handling for linker relaxation:
-// - scans call/jal relocations and selects profitable cm.jt/cm.jalt entries
+// This file implements RISC-V table-jump handling for linker relaxation:
+// - scans call/jal relocations and selects profitable table entries
 // - materializes and emits the .riscv.jvt table
 // - provides map-dump support for selected table-jump entries
+//
+// Zcmt and Xqccmt use the same 16-bit instruction encoding. Zcmt calls the
+// instructions cm.jt/cm.jalt. Xqccmt calls them qc.cm.jt/qc.cm.jalt.
+// Xqccmt also lets qc.cm.jalt write the return address to t0 (x5). It does
+// that by storing bit 0 as metadata in the JVT entry. A clear bit means ra
+// (x1). A set bit means t0 (x5). The real jump target always has bit 0 clear.
 //
 // ABI reference:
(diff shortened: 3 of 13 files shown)


lib/LinkerWrapper/RISCVLinkDriver.cpp
```diff
@@ -140,19 +140,41 @@ RISCVLinkDriver::parseOptions(ArrayRef<const char *> Args,
...
+                             OPT_RISCVLinkOptTable::no_relax_tbljal)) {
+    if (Arg->getOption().matches(OPT_RISCVLinkOptTable::no_relax_tbljal)) {
+      TbljalMode = GeneralOptions::RISCVRelaxTbljalMode::None;
+    } else if (Arg->getOption().matches(OPT_RISCVLinkOptTable::relax_tbljal)) {
+      TbljalMode = GeneralOptions::RISCVRelaxTbljalMode::Zcmt;
+    } else {
+      TbljalMode =
+          llvm::StringSwitch<GeneralOptions::RISCVRelaxTbljalMode>(
+              Arg->getValue())
+              .Case("zcmt", GeneralOptions::RISCVRelaxTbljalMode::Zcmt)
```
lenary: Can we do `.CaseLower` here, so that the string case doesn't matter?

lib/Target/RISCV/RISCVRelocationHelper.h
```diff
@@ -90,6 +90,19 @@ enum Reg {
...
 }
 
+[[maybe_unused]] unsigned getQCEJumpRd(uint64_t insn) {
+  switch ((insn >> 15) & 0x3) {
+  case 0:
+    return X_ZERO;
+  case 1:
+    return X_RA;
+  case 2:
+    return X_T0;
```
lenary: What's going on here? Neither of the two `qc.e.j*` instructions can encode `t0` as a destination according to the v0.13.0 pdf I have.

lenary: I think maybe something has invented a `qc.e.jalt` instruction that doesn't exist, but I'm not sure what's happened here.

lib/Target/RISCV/RISCVTableJump.cpp
```diff
@@ -176,14 +200,83 @@ static void selectEntries(
...
+      return A.Entry.Saved > B.Entry.Saved;
+    const uint64_t AVA = getTableJumpTargetVA(Backend, A.Sym);
+    const uint64_t BVA = getTableJumpTargetVA(Backend, B.Sym);
+    if (AVA != BVA)
+      return AVA < BVA;
+    if (A.Sym->name() != B.Sym->name())
+      return A.Sym->name() < B.Sym->name();
+    // Keep ra before t0 when every other key is the same. The order is only a
+    // tie-breaker, but stable output makes tests and maps easier to read.
+    return !A.UseT0 && B.UseT0;
```
lenary: Can we replace this with a comparison using `std::tie`? I agree stable is more important, and yes it's worth prioritising `ra` over `t0`, if the saving and name is the same.

lib/Target/RISCV/RISCVTableJump.cpp
```diff
@@ -213,9 +314,11 @@ int RISCVTableJumpFragment::getCMJTEntryIndex(const ResolveInfo *Sym) const {
   return It != CMJTCandidates.end() ? It->second.Index : -1;
 }
 
-int RISCVTableJumpFragment::getCMJALTEntryIndex(const ResolveInfo *Sym) const {
-  auto It = CMJALTCandidates.find(Sym);
-  return It != CMJALTCandidates.end()
+int RISCVTableJumpFragment::getCMJALTEntryIndex(const ResolveInfo *Sym,
+                                                unsigned Rd) const {
+  const auto &Candidates = Rd == 5 ? QCCMJALTTCandidates : CMJALTCandidates;
```
lenary, line 312: I think what this means is that `QCCMJALTTCandidates` and `CMJALTCandidates` are exclusive by this point?

I.e. if there's a symbol -> index mapping in one of them, then the other won't contain any symbol mapping to the same index, right?

Presumably when this is called we've already made choices about laying out the table?

lib/Target/RISCV/RISCVTableJump.cpp
```diff
@@ -269,6 +375,10 @@ writeEntries(RISCVLDBackend &Backend, uint8_t *Buf,
 
   for (auto &KV : Entries) {
     uint64_t VA = getTableJumpTargetVA(Backend, KV.first);
+    // Xqccmt qc.cm.jalt uses bit 0 as link-register metadata. The hardware
+    // clears that bit before jumping, so this does not change the target.
+    if (SetEntryBit0)
+      VA |= 1;
```
lenary: I don't understand this change, especially as I think that `writeEntries` is now only used for the `cm.jt` entries?

lib/Target/RISCV/RISCVTableJump.cpp
```diff
@@ -277,12 +387,39 @@ writeEntries(RISCVLDBackend &Backend, uint8_t *Buf,
...
+    else
+      llvm::support::endian::write64le(EntryBuf, VA);
+  };
+
+  // Call entries from both maps share one index space. Write every entry to
+  // its exact index so ra and t0 entries never overwrite each other.
+  for (auto &KV : CMJALTCandidates)
+    WriteOne(KV.first, KV.second, /*SetEntryBit0=*/false);
+  for (auto &KV : QCCMJALTTCandidates)
+    WriteOne(KV.first, KV.second, /*SetEntryBit0=*/true);
```
lenary, line 405: Do we always lay out the `t0` entries after all the `ra` entries? I didn't get that impression but maybe I wasn't reading closely enough.

lib/Target/RISCV/RISCVTableJump.h
```diff
@@ -21,8 +21,9 @@ struct RISCVTableJumpEntry {
   int Index = -1;
 };
 
-// Implements the Zcmt table jump section (.riscv.jvt) used by table jump
-// relaxation (cm.jt/cm.jalt).
+// Implements the RISC-V table jump section (.riscv.jvt) used by table jump
+// relaxation. Zcmt uses cm.jt/cm.jalt. Xqccmt uses the same instruction
+// encoding, but the assembler/disassembler names are qc.cm.jt/qc.cm.jalt.
```
lenary: I think this comment should document the table and its contents, rather than the instructions per-se.

So we should maybe say that the table contains jump addresses, and if we're doing qccmt, then the lowest bit of those addresses might be set but the address being jumped to will not have it set.

lenary (commented):
Coming together. I think AI has hallucinated an instruction that doesn't exist.

lib/LinkerWrapper/RISCVLinkDriver.cpp
svs-quic: I have made the change, thanks.

lib/Target/RISCV/RISCVRelocationHelper.h
svs-quic: Yeah sorry I thought I had removed this before I pushed the patch. Removed it now.

lib/Target/RISCV/RISCVTableJump.cpp
svs-quic: I've tried using std::tie now.

lib/Target/RISCV/RISCVTableJump.cpp
svs-quic: Yes, the QCCMJALTTCandidates is only for those that have x5 as the RA. CMJALTCandidates has those that have the x1 as the RA. A symbol -> index mapping will go into only one of them.

lib/Target/RISCV/RISCVTableJump.cpp
svs-quic: Yeah I had initially tried to do it in the same function but then split it into separate ones. Removed it.

lib/Target/RISCV/RISCVTableJump.cpp
svs-quic: That is not what is happening here. The index is stored in KV.second so it gets written to the correct index.

lib/Target/RISCV/RISCVTableJump.h
svs-quic: Thanks I've changed the comment.

svs-quic (commented):
Thanks for your comments, Sam. I think I have addressed them.

lib/Target/RISCV/RISCVTableJump.cpp
lenary: Sorry, yeah, I see that now.

test/RISCV/FromLLD/riscv-xqccmt-tbljal-call.s
```diff
@@ -4,6 +4,7 @@
 #
 # RUN: %llvm-mc -filetype=obj -mattr=+relax,+experimental-xqccmt %s -o %t.o
 # RUN: %link %linkopts %t.o --relax-tbljal=xqccmt --defsym foo=0x800000 --defsym bar=0x900000 -o %t
+# RUN: %link %linkopts %t.o --relax-tbljal=XQCCMT --defsym foo=0x800000 --defsym bar=0x900000 -o %t.upper
```
lenary: We probably want to make some assertions against `%t.upper`? the same ones as `%t`? Rather than just "this didn't exit with an error"

svs-quic: Yes have done that now.

lenary (approved):
LGTM

lib/Target/RISCV/RISCVTableJump.cpp
```diff
@@ -175,14 +199,80 @@ static void selectEntries(
...
+    const uint64_t BVA = getTableJumpTargetVA(Backend, B.Sym);
+    const int ASaved = -A.Entry.Saved;
+    const int BSaved = -B.Entry.Saved;
+    const llvm::StringRef AName = A.Sym->name();
+    const llvm::StringRef BName = B.Sym->name();
+    return std::tie(ASaved, AVA, AName, A.UseT0) <
+           std::tie(BSaved, BVA, BName, B.UseT0);
+  });
+
+  if (Entries.size() > MaxSize)
```
quic-akaryaki, line 240: Instead of adding elements, sorting them and then removing, you can check for MaxSize and acceptable Saved values when inserting.

lib/Target/RISCV/RISCVTableJump.cpp
```diff
@@ -175,14 +199,80 @@ static void selectEntries(
...
+    Entries.resize(MaxSize);
+
+  const int WordSize = Backend.config().targets().is32Bits() ? 4 : 8;
+  while (!Entries.empty() && Entries.back().Entry.Saved < WordSize)
+    Entries.pop_back();
+
+  for (size_t I = 0; I < Entries.size(); ++I) {
+    if (Entries[I].UseT0)
+      QCCMJALTTCandidates[Entries[I].Sym].Index = static_cast<int>(I);
```
quic-akaryaki: You should not need a static_cast here.

lib/Target/RISCV/RISCVTableJump.cpp
```diff
@@ -175,14 +199,80 @@ static void selectEntries(
...
+    else
+      CMJALTCandidates[Entries[I].Sym].Index = static_cast<int>(I);
+  }
+
+  llvm::SmallVector<const ResolveInfo *, 0> ToErase;
```
quic-akaryaki: You can use std::erase_if + std::erase or llvm::erase_if, not clear why you need a temporary array.

lib/Target/RISCV/RISCVTableJump.cpp
```diff
@@ -175,14 +199,80 @@ static void selectEntries(
...
+  for (const auto &KV : CMJALTCandidates)
+    if (KV.second.Index < 0)
+      ToErase.push_back(KV.first);
+  for (const ResolveInfo *Sym : ToErase)
+    CMJALTCandidates.erase(Sym);
+
+  ToErase.clear();
```
quic-akaryaki: Same here

lib/Target/RISCV/RISCVTableJump.cpp
```diff
@@ -175,14 +199,80 @@ static void selectEntries(
...
+  selectCallEntries(Backend, CMJALTCandidates, QCCMJALTTCandidates,
+                    MaxCMJALTEntrySize);
 
   const int WordSize = Backend.config().targets().is32Bits() ? 4 : 8;
+  const unsigned CallEntryCount =
+      getCallTableCandidateCount(CMJALTCandidates, QCCMJALTTCandidates);
 
-  int SavedBoth = -static_cast<int>(
-      (StartCMJALTEntryIdx + CMJALTCandidates.size()) * WordSize);
+  int SavedBoth =
```
quic-akaryaki, line 268: Why do you use negative numbers? Looks like they are always negative. Wouldn't it be easier to use unsigned or uint64_t? Although it's existing code.

lib/Target/RISCV/RISCVRelocationHelper.h
```diff
@@ -90,6 +90,17 @@ enum Reg {
...
 }
 
+[[maybe_unused]] unsigned getQCEJumpRd(uint64_t insn) {
+  switch ((insn >> 15) & 0x3) {
+  case 0:
+    return X_ZERO;
+  case 1:
+    return X_RA;
+  default:
+    return 32;
```
quic-akaryaki: Returning 32 as a register number is weird and potentially error prone. Isn't X_ZERO == 0 and X_RA == 1? Why do you need this function at all if the valid values are the same? It seems you need to boolean check for acceptable register numbers instead.

lib/Target/RISCV/RISCVTableJump.cpp
quic-akaryaki: But I know it's to squash a warning :(

lib/Target/RISCV/RISCVRelocationHelper.h
svs-quic: Thanks, I've removed the cases and kept the (insn >> 15) & 0x3 alone in the function given that it is used in a couple of places. I've also added a function to check for valid register numbers.

lib/Target/RISCV/RISCVTableJump.cpp
svs-quic: I've made the changes to use WordSize and simplified the candidate selection.

lib/Target/RISCV/RISCVTableJump.cpp
svs-quic: We no longer need the ToErase vector.

lib/Target/RISCV/RISCVTableJump.cpp
svs-quic: We no longer need the ToErase vector.

lib/Target/RISCV/RISCVTableJump.cpp
svs-quic: The number represents the profitability from emitting the jump table and can be negative.

SavedBoth: cost of emitting a table that includes jump entries plus the padded call-entry region.
SavedCMJTOnly: cost of emitting only the cm.jt jump entries.

For each of the candidates in CMJTCandidates, CMJALTCandidates and QCCMJALTTCandidates we add the bytes saved to this value to get the final profitability. A negative value means the table would increase the code size and the candidates need to be discarded.

```
 for (auto &KV : CMJALTCandidates)
     SavedBoth += KV.second.Saved;
   for (auto &KV : QCCMJALTTCandidates)
     SavedBoth += KV.second.Saved;
```

I guess I can start with both of them as zero and keep appending the savings and finally subtract the cost of emitting the table as well but I would like to leave it as is in this PR.

svs-quic (commented):
Thank you for your comments, Alexey. I have addressed them in the latest commit.

quic-akaryaki (approved):

PR #1326, opened 2026-06-23
[TLS] Fix PT_TLS file size and memory size

quic-seaswara:
Exclude .tbss from file-offset (lower_offset / upper_offset) tracking for PT_TLS segments. .tbss is SHT_NOBITS and occupies no file space. its sh_offset is inherited from the preceding non-NOBITS section. range. Including it corrupts the filesz computation for PT_TLS. Memory-size bounds (lower / upper) are still updated so PT_TLS p_memsz accounts for the per-thread TBSS reservation.


quic-seaswara:
@quic-k I added a run test along with this to make sure we have the right fix. 

https://github.com/qualcomm/eld/pull/1319 can be abandoned.

PR #1330, opened 2026-06-24
[AArch64] Add overflow check for R_AARCH64_ADR_PREL_PG_HI21

quic-seaswara:
The ADRP instruction encodes (PG(S+A) - PG(P)) >> 12 in a 21-bit signed immediate field. Add a signed 21-bit range check for R_AARCH64_ADR_PREL_PG_HI21; the _NC (no-check) variant intentionally skips the check.

Add a lit test that verifies a link within range succeeds and a link that overflows the 21-bit field is caught with an overflow diagnostic.


efriedma-quic (approved):
LGTM

PR #1331, opened 2026-06-24
[RISCV] JVT Symbol Fixes

lenary:
This PR includes two fixes to the Zcmt symbols:
- It now updates the size of the `__jvt_base$` symbol to accurately reflect the size of the table it creates.
- It adds a `$d` data mapping symbol so that when the jump table is put in an executable section (as is required), it is disassembled as data, not code.


lenary:
Marking as draft as the tests need more work.

svs-quic (approved):
LGTM! Thanks for doing this, Sam.

quic-seaswara (approved):

PR #1334, opened 2026-06-25
[AArch64] Fix regression in abs relocation

nezetic:
Fix a regression in aarch64 abs relocation.

The pauth path changed the original non-pauth logic, and this was never intended.

### DIFF
--- a/lib/Target/AArch64/AArch64Relocator.cpp
+++ b/lib/Target/AArch64/AArch64Relocator.cpp
@@ -699,7 +699,6 @@ Relocator::Result abs(Relocation &pReloc, AArch64Relocator &pParent) {
     // Create GOT slot for rsym with the fragment reference of P
   } else
     S = pParent.getSymValue(&pReloc);
-  Relocator::Address targetVal = S + A;
 
   bool isAuthAbs = pReloc.type() == llvm::ELF::R_AARCH64_AUTH_ABS64;
   ELFSection *target_sect = pReloc.targetRef()->getOutputELFSection();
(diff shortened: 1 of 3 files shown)


efriedma-quic (commented):
The revised logic makes sense to me.

I think I'd reorganize the code a bit to keep the ptrauth relocations closer to the non-ptrauth counterparts.  There are some weird inconsistencies, like `rsym && rsym->isWeak() && rsym->isUndef()` vs. `rsym && rsym->isWeakUndef() && (pParent.config().codeGenType() == LinkerConfig::Exec)`.

I'd like to see testcases for absolute relocations against PLT symbols, and against weak undefined symbols, with and without ptrauth,

quic-seaswara (commented):
Can we accompany this with a build/run test ? 

We have support for running tests.

quic-seaswara:
@nezetic  can you please add a test ?

nezetic:
> The revised logic makes sense to me.
> 
> I think I'd reorganize the code a bit to keep the ptrauth relocations closer to the non-ptrauth counterparts. There are some weird inconsistencies, like `rsym && rsym->isWeak() && rsym->isUndef()` vs. `rsym && rsym->isWeakUndef() && (pParent.config().codeGenType() == LinkerConfig::Exec)`.
> 
> I'd like to see testcases for absolute relocations against PLT symbols, and against weak undefined symbols, with and without ptrauth,

I changed the code to better match the original logic.

Regarding the tests, I already wrote a few for the weak undefined symbols (see [PAuthRelocWeak.s](https://github.com/qualcomm/eld/tree/main/test/AArch64/standalone/Relocations/PAuthRelocWeak.s)).

And about the PLT, the current PAuth implementation don't support PLT GOT signing.

nezetic:
> Cédric Tessier (@nezetic) can you please add a test ?

I added a test that should cover the regression.

efriedma-quic (approved):
LGTM

quic-seaswara (approved):

PR #1348, opened 2026-06-26
[Zephyr] Add opt-in PR workflow 

quic-areg:
Factor the Zephyr build+twister body into zephyr-run.yml, which zephyr-nightly.yml now calls. Adds zephyr-pr.yml, a manual workflow_dispatch that downloads the PR's own ld.eld artifact from a successful ci.yml run and runs tests/kernel on qemu_cortex_a53.

ci.yml packages the built ld.eld + libLW.so.4 + libc++ so zephyr-pr.yml can consume it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -164,6 +164,34 @@ jobs:
         working-directory: pr-${{ env.PR_NUMBER }}/obj
         run: ninja check-eld-summary
 
+      # Package the PR-built ld.eld (+ libs) for zephyr-pr.yml. `inst/` layout
+      # matches FetchNightlyToolset (~18MB compressed).
+      - name: Package PR ELD toolset
+        if: steps.file-check.outputs.skip_build == 'false'
+        working-directory: pr-${{ env.PR_NUMBER }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          mkdir -p inst/bin inst/lib
+
+          cp obj/bin/ld.eld          inst/bin/
+          cp -P obj/lib/libLW.so.4   inst/lib/
+          ln -sf libLW.so.4          inst/lib/libLW.so
+
+          # Bundle libc++/libc++abi dependencies.
+          bash llvm-project/llvm/tools/eld/.github/workflows/scripts/BundleSharedLibDeps.sh \
+            inst ld.eld
+
+          XZ_OPT=-T0 tar -Jcf install-pr-toolset.tar.xz inst
+
+      - name: Upload PR ELD toolset
+        if: steps.file-check.outputs.skip_build == 'false'
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a #v7.0.1
+        with:
+          name: pr-eld-toolset
+          retention-days: 1
+          path: pr-${{ env.PR_NUMBER }}/install-pr-toolset.tar.xz
+
       - name: Clean up
         if: always()
         run: |
--- a/.github/workflows/zephyr-nightly.yml
+++ b/.github/workflows/zephyr-nightly.yml
@@ -9,214 +9,25 @@ on:
 
 permissions:
   contents: write
+  actions: read
 jobs:
-  zephyr-build-and-test:
+  zephyr:
     if: github.repository == 'qualcomm/eld' || github.event_name == 'workflow_dispatch'
-    runs-on: ubuntu-latest
-    container:
-      image: ghcr.io/zephyrproject-rtos/ci-base:main
-      options: --user 0
     strategy:
       fail-fast: false
       matrix:
         target:
-          - board: qemu_riscv32
-            arch: riscv32
-            extra_args: CONFIG_PICOLIBC_USE_TOOLCHAIN=y
-          - board: qemu_riscv32_xip
-            arch: riscv32
-            extra_args: CONFIG_PICOLIBC_USE_TOOLCHAIN=y
-          - board: qemu_riscv64
-            arch: riscv64
-            extra_args: CONFIG_PICOLIBC_USE_TOOLCHAIN=y
-          - board: qemu_cortex_m3
-            arch: arm
-            # cpullvm ARM picolibc is missing a _nopic version, so its
-            # objects have GOT relocations that fail when .got is discarded.
-            extra_args: CONFIG_PICOLIBC_USE_MODULE=y
-          - board: qemu_cortex_a53
-            arch: aarch64
-            extra_args: CONFIG_PICOLIBC_USE_TOOLCHAIN=y
-    env:
-      CPULLVM_VERSION: "22.1.0"
-      CPULLVM_DIR: /opt/cpullvm
-      CCACHE_DIR: /root/.ccache
-      CCACHE_BASEDIR: ${{ github.workspace }}
-      CCACHE_NOHASHDIR: "1"
-      CCACHE_MAXSIZE: "2G"
-      CCACHE_COMPRESS: "1"
-      CCACHE_COMPILERCHECK: "content"
-
-    steps:
-      - name: Checkout eld
-        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
-        with:
-          path: llvm-project/llvm/tools/eld
-
-      - name: Prepare ccache dir
-        run: |
-          mkdir -p "$CCACHE_DIR"
-          ccache --version | head -1
-
-      - name: Restore ccache
-        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 #v6.1.0
-        with:
-          path: ${{ env.CCACHE_DIR }}
-          key: ccache-${{ matrix.target.arch }}-cpullvm${{ env.CPULLVM_VERSION }}-${{ github.run_id }}
-          restore-keys: |
-            ccache-${{ matrix.target.arch }}-cpullvm${{ env.CPULLVM_VERSION }}-
-
-      - name: ccache stats (before build)
-        run: ccache -s -v || ccache -s
-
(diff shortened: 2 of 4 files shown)


quic-areg:
Run: https://github.com/qualcomm/eld/actions/runs/28616045854/job/84862934625

.github/workflows/zephyr-run.yml
```diff
@@ -0,0 +1,256 @@
...
+        if: inputs.eld_source == 'nightly'
+        uses: ./llvm-project/llvm/tools/eld/.github/workflows/FetchNightlyToolset
+
+      - name: Download PR ELD artifact
+        if: inputs.eld_source == 'artifact'
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c #v8.0.1
+        with:
+          name: pr-eld-toolset
+          run-id: ${{ inputs.artifact_run_id }}
+          github-token: ${{ github.token }}
```
quic-seaswara, line 105: Remove the artifact after download to conserve space ? 

https://github.com/marketplace/actions/delete-artifact

.github/workflows/zephyr-pr.yml
```diff
@@ -0,0 +1,107 @@
...
+      contents: write # zephyr-run.yml requires write (unused outside the nightly schedule path)
+      actions: read
+    uses: ./.github/workflows/zephyr-run.yml
+    with:
+      board: qemu_cortex_a53
+      arch: aarch64
+      extra_args: CONFIG_PICOLIBC_USE_TOOLCHAIN=y
+      test_scope: tests/kernel
+      eld_source: artifact
+      artifact_run_id: ${{ needs.resolve.outputs.run_id }}
```
quic-seaswara, line 106: can we run all the architectures supported by eld ?

.github/workflows/ci.yml
```diff
@@ -164,6 +164,41 @@ jobs:
...
+            case "$(basename "${dep}")" in
+              libc++.so.1*|libc++abi.so.1*)
+                cp -P "${dep}" inst/lib/
+                cp -n "$(readlink -f "${dep}")" inst/lib/ || true
+                ;;
+            esac
+          done
+
+          ls -la inst/bin inst/lib
+          XZ_OPT=-T0 tar -Jcf install-pr-toolset.tar.xz inst
```
quic-seaswara, line 185: You can use the script BundleSharedLibDeps.sh to bundle dependencies that you need.

.github/workflows/ci.yml
```diff
@@ -164,6 +164,41 @@ jobs:
...
+
+      - name: Upload PR ELD toolset
+        if: steps.file-check.outputs.skip_build == 'false'
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a #v7.0.1
+        with:
+          name: pr-eld-toolset
+          path: pr-${{ env.PR_NUMBER }}/install-pr-toolset.tar.xz
```
quic-seaswara, line 193: Have you considered running this on the self hosted machine, that will allow the artifacts to be passed between  the workflows without uploading to github ?

.github/workflows/zephyr-run.yml
```diff
@@ -0,0 +1,256 @@
...
+
+      - name: Upload build artifacts
+        if: always()
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a #v7.0.1
+        with:
+          name: zephyr-build-artifacts-${{ inputs.board }}
+          retention-days: 2
+          path: |
+            zephyr/build/zephyr/zephyr.elf
+            zephyr/build/zephyr/zephyr.map
```
quic-seaswara, line 238: Do we use these artifacts ?

quic-seaswara (changes requested):
Very cool. 

Have few comments.

.github/workflows/zephyr-pr.yml
quic-areg: yes i am going to add them in a follow-up

.github/workflows/zephyr-run.yml
quic-areg: retention days is set to 1

.github/workflows/zephyr-run.yml
quic-areg: I have never used them myself. I can remove it separately.

.github/workflows/ci.yml
quic-areg: I think even on a self-hosted machine the artifact should be passed through github. Not every CI build will have an associated zephyr build, so I think the self-hosted runner would accumulate too many unused artifacts.

.github/workflows/zephyr-run.yml
```diff
@@ -0,0 +1,256 @@
...
+          record-mode: "record"
+          status: "fail"
+          run-id: ${{github.run_id}}
+          arch-name: ${{ inputs.arch }}
+          branch-name: "main"
+
+      # Both branches normalize the payload to ./inst so the splice is identical.
+      - name: Fetch nightly ELD
+        if: inputs.eld_source == 'nightly'
+        uses: ./llvm-project/llvm/tools/eld/.github/workflows/FetchNightlyToolset
```
jonathonpenix, line 97: This isn't at all an important/blocking comment, but since I've been tinkering with similar stuff in CPULLVM:

How similar are your nightly and PR installs/tarballs?

We're doing similar stuff in CPULLVM right now and essentially what I landed on was giving our Zephyr workflow run id and package name input parameters and we'll have an action that'll fetch a toolchain based on those two things (which, that action is basically a clone of/riff on eld's FetchNightlyToolset).

Our packaging is basically identical between our nightlies and PRs, so I think that ends up being a pretty generic "interface" so the Zephyr workflow doesn't have to care too much about where the toolchain it'll use is coming from (PR vs nightly, different branches with different package naming schemes, different hosts).

Maybe something similar would work here? Might be more trouble than its worth (and I might be misunderstanding something) but it might let you hoist these PR vs Nightly `if`s into their respective workflows, and let this workflow assume a common format.

.github/workflows/zephyr-nightly.yml
```diff
@@ -9,214 +9,25 @@ on:
...
-          record-mode: "update"
-          status: "pass"
-          run-id: ${{github.run_id}}
-          arch-name: ${{ matrix.target.arch }}
-          branch-name: "main"
+          - { board: qemu_riscv32,     arch: riscv32,  extra_args: CONFIG_PICOLIBC_USE_TOOLCHAIN=y }
+          - { board: qemu_riscv32_xip, arch: riscv32,  extra_args: CONFIG_PICOLIBC_USE_TOOLCHAIN=y }
+          - { board: qemu_riscv64,     arch: riscv64,  extra_args: CONFIG_PICOLIBC_USE_TOOLCHAIN=y }
+          # cpullvm ARM picolibc is missing a _nopic version, so its objects
+          # have GOT relocations that fail when .got is discarded.
```
jonathonpenix, line 24: Two misc. comments not at all related to this PR but that I'm just now thinking about seeing this and don't remember if it was mentioned elsewhere:
* We did add a variant for this, but it isn't in the 22.x release (only mainline right now). Not sure how you all feel about fetching nightly toolchains instead but that'd be one option. Or, it should be there in the 23.x release once that is out.
* We [recently turned on qemu_cortex_a9 testing](https://github.com/qualcomm/cpullvm-toolchain/pull/411) too since we have some users doing 32bit armv7a stuff. A compatible multilib is only available in CPULLVM mainline right now though. There are a fair number of failures (including in kernel tests IIRC), but haven't looked into them at all (no idea if they are compiler vs lib vs linker issues).

jonathonpenix (approved):
A few scattered thoughts, but none are at all blocking/etc.

Overall, looks good to me!

.github/workflows/ci.yml
quic-areg: done

quic-shaksuma (approved):

Steven6798 (approved):

.github/workflows/zephyr-nightly.yml
quic-seaswara: I think this would be a follow up to add qemu cortex a9. We will need to get to a stable point before adding to the workflow. 

I think @quic-areg  is planning to use the Zephyr LLVM SDK instead of CPULLVM as we move forward.

@quic-areg ?

.github/workflows/zephyr-run.yml
quic-areg: Thanks! I think we could benefit from doing something like this in a follow-up. I will give it a try.

quic-seaswara: We currently have nightly that contains the entire toolchain and the PR that contains only the eld tarball.

We can package the eld tarball as a seperate artifact too from the nightly eld workflow.

I see this will simplify the workflow too. 

@quic-areg ?

.github/workflows/zephyr-nightly.yml
quic-areg: Fetching a nightly toolchain creates a moving target for both the compiler and linker which is something I wanted to avoid. Downloading CI artifacts across repos requiring setting up permissions/ access tokens etc that I also wanted to avoid. Right now we build picolibc instead of using the picolibc from the toolchain for ARM targets.

I will look at cortex a9 in a follow-up.

I tried out the Zephyr SDK for building but it has some limitations and uses an older LLVM version.

quic-seaswara (approved):

.github/workflows/zephyr-nightly.yml
quic-seaswara: will recommend a ticket to track this.

.github/workflows/zephyr-run.yml
quic-seaswara: will recommend a ticket to track this.

PR #1357, opened 2026-06-29
[Windows] Add support for dev/null output files on windows

### LINKED ISSUE #1259: Add support for dev/null on windows

Steven6798:
Fix: qualcomm#1259

### DIFF
--- a/lib/LinkerWrapper/GnuLdDriver.cpp
+++ b/lib/LinkerWrapper/GnuLdDriver.cpp
@@ -2150,7 +2150,7 @@ bool GnuLdDriver::processLTOOptions(llvm::lto::Config &Conf,
 
 std::string GnuLdDriver::getOutputFileName() const {
   std::string FileName = Config.options().outputFileName();
-  if (FileName != "/dev/null")
+  if (FileName != "/dev/null" && FileName != "NUL")
     FileName = eld::sys::fs::Path(FileName).filename().native();
 
   return FileName;
--- a/lib/Core/Linker.cpp
+++ b/lib/Core/Linker.cpp
@@ -765,7 +765,7 @@ bool Linker::emit() {
   if (layoutInfo)
     layoutInfo->recordOutputFileSize(OutputFileSize);
 
-  {
+  if (Path != "/dev/null" && Path != "NUL") {
     std::error_code Ec;
     int OutputFlag = 0;
     if (Perm & 0x755)
--- a/test/Common/standalone/devnull/devnull.test
+++ b/test/Common/standalone/devnull/devnull.test
@@ -1,3 +1,2 @@
-#UNSUPPORTED: windows
 RUN: %clang %clangopts -c %p/Inputs/1.c -o %t1.o
 RUN: %link %linkopts %t1.o -o /dev/null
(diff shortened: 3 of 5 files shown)


Steven6798:
There are 2 ways to fix this problem.

- Approach 1: The current solution. This forces output files names on windows to /dev/null, following linux's convention.
- Approach 2: We can use NUL (Windows /dev/null equivalent) for windows files and fix any test that checks just for /dev/null to also check for NUL.

Any feedback is welcome.

quic-seaswara (commented):
Add a /dev/null support specific test.

What does the map file show ? 

Can we emit a diagnostic saying that "writing to output file skipped ... " 

Please create a follow up for 

- Input file names also can be /dev/null.

lib/LinkerWrapper/GnuLdDriver.cpp
```diff
@@ -2150,6 +2150,11 @@ bool GnuLdDriver::processLTOOptions(llvm::lto::Config &Conf,
 
 std::string GnuLdDriver::getOutputFileName() const {
   std::string FileName = Config.options().outputFileName();
+  // On Windows, lit's internal shell replaces /dev/null with NUL (the Windows
```
quic-areg: I don't think we should have lit-specific stuff in the linker or reference unix-specific paths on windows.

Can we solve this on the lit side instead? How many tests use `dev/null`? Can they be rewritten to omit the path or be made more permissive to accept both forms?

Steven6798: Absolutely. That is the other approach that I mentioned. I can update the tests to accept either /test/null or NUL.

Steven6798:
> Add a /dev/null support specific test.

Sounds good. I can update the existing devnull test to run for windows too.

> What does the map file show?

The map file show nothing related to the /dev/null or NUL. Not sure what I should be looking for.

> Can we emit a diagnostic saying that "writing to output file skipped ... "
> 
> Please create a follow up for
> 
> * Input file names also can be /dev/null.

I'll create tickets for these 2 issues.

lib/LinkerWrapper/GnuLdDriver.cpp
quic-seaswara: > I don't think we should have lit-specific stuff in the linker or reference unix-specific paths on windows.
> 
> Can we solve this on the lit side instead? How many tests use `dev/null`? Can they be rewritten to omit the path or be made more permissive to accept both forms?

lld does it in the linker, we should continue to have it in the linker.

quic-areg: > lld does it in the linker

I don't think this is true. lld does the opposite mapping (`/dev/null` -> `NUL`) and only for input files, not `-o` output, and lld's tests don't rely on any of it as far as I can tell. 

The the only failure in our windows CI is ` RISCV/FromLLD/riscv-attributes.s` (with others marked unsupported). lit's internal shell already remaps `/dev/null` to a temp file and the linker writes there fine. 

expect: `Fatal: Linking had errors (/dev/null)`
actual:  `Fatal: Linking had errors (C:\Users\RUNNER~1\AppData\Local\Temp\lit-tmp-.../tmp...)`

the test can simply be fixed by changing the check to work on all platforms: 
` # INVALID_ARCH1_NOERR: Fatal: Linking had errors`

This seems like a better fix than emitting a file path on windows that does not actually exist.

quic-seaswara: `openAuxiliaryFile` does it for all the files lld writes to. 

I am fine with adding NUL support in the output filename as well. 

You can fix the test case as a separate fix. 

@Steven6798  please continue with your patch.

quic-seaswara (approved):

quic-shaksuma (approved):

PR #1379, opened 2026-07-01
[AArch64] Fix TLSDESC_ADD_LO12 encoding

### LINKED ISSUE #1378: [AArch64] TLSDESC_ADD_LO12 incorrectly encoded in shared libraries / dynamic executables
R_AARCH64_TLSDESC_ADD_LO12_NC applies to an add x0, x0, #imm instruction. 

The linker was instead encoding it using the helper for LDR/STR.

Test 

cat > tls.c << \!
__thread int tls_var;
!

```
clang -target aarch64 -fPIC -c tls.c -o tls.o
ld.eld  -shared tls.o -o tls.so
llvm-objdump -d tls.so

```

The ldr (R_AARCH64_TLSDESC_LD64_LO12_NC) and add (R_AARCH64_TLSDESC_ADD_LO12_NC) must point at the same GOT slot.

quic-seaswara:
R_AARCH64_TLSDESC_ADD_LO12_NC applies to an ADD instruction whose immediate field takes an unshifted 12-bit page offset.


efriedma-quic (approved):
LGTM

parth-07 (approved):

Issue #1370, opened 2026-06-30
[Linker] Improve behavior when output files are not emitted


Steven6798:
Skip file emission logic when output files are not generated to simplify code

PR #1381, opened 2026-07-01
Fix to support NUL output file

Steven6798:


parth-07 (approved):

quic-shaksuma (approved):

lib/Core/Linker.cpp
```diff
@@ -814,7 +814,8 @@ bool Linker::emit() {
   }
 
   LinkerProgress->incrementAndDisplayProgress();
-  if ((Path != "/dev/null") && (ThisConfig->options().verifyLink())) {
+  if ((Path != "/dev/null" && Path != "NUL") &&
```
quic-seaswara, line 817: Follow up : can we have a helper function  in Support isNullDevice(Path) and return true ?

Steven6798: Yes in issue #1370 I will make some improvements and I can include this.

PR #1382, opened 2026-07-01
Revert "Make version script pattern matching GNU-compatible"

parth-07:
Reverts qualcomm/eld#1151. This functionality, although correct, breaks an already existing test test/Common/standalone/VersionScript/VersionScript.test that only runs when symbol versioning is disabled. This test should be fixed as part of the change.


Steven6798 (approved):

pgodeq (approved):
This is required for unblock builders.

PR #1383, opened 2026-07-01
[Security] Fix more GitHub actions mutable action tags

### LINKED ISSUE #1374: [Security] Fix GitHub actions mutable action tags
Action must use the reference to a full 40-character commit SHA instead of the tag.

Steven6798:
Fix: #1374

### DIFF
--- a/.github/workflows/FetchNightlyToolset/action.yaml
+++ b/.github/workflows/FetchNightlyToolset/action.yaml
@@ -5,7 +5,7 @@ runs:
   steps:
     - name: Get latest run id of nightly-test.yml workflow
       id: latest
-      uses: actions/github-script@v8
+      uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 #v9.0.0
       with:
         github-token: ${{ github.token }}
         script: |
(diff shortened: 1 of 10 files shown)


.github/workflows/FetchNightlyToolset/action.yaml
```diff
@@ -5,7 +5,7 @@ runs:
   steps:
     - name: Get latest run id of nightly-test.yml workflow
       id: latest
-      uses: actions/github-script@v8
+      uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 #v9.0.0
```
parth-07, line 8: How often do we need to update these versions?

parth-07 (approved):

quic-shaksuma (approved):
It would be nice to have the source of the sha used documented as a comment in the yml files.

.github/workflows/FetchNightlyToolset/action.yaml
Steven6798: Not much. The core fix of this PR is to stop using mutable tags (v8) to fix Semgrep warnings. This is what LLVM does too.

Steven6798:
> It would be nice to have the source of the sha used documented as a comment in the yml files.

I'll add this as a follow up. Thanks!

PR #1384, opened 2026-07-01
[RISCV] Add tests for and minor rework of c.lui to c.li rewriting. NFCI

jonathonpenix:
A follow-up patch will introduce a new internal relocation for `c.li` for handling GOT load relaxations. The existing code is incorrect for that case, so limit this rewriting to only when we see a `c.lui`.

While here, try to align how `EncTy_CI` is handled with other types. It seems a bit confusing for this to be handled differently than all the other types.

Also, originally I tried deleting this thinking that this code was dead. Turns out it isn't, but it doesn't seem like there's any tests covering it currently. So try and add tests. These are intended more to document the current state of things rather than what necessarily ought to happen.

### DIFF
--- a/lib/Target/RISCV/RISCVHelper.h
+++ b/lib/Target/RISCV/RISCVHelper.h
@@ -276,8 +276,6 @@ template <typename T> T encodeCJ(T Result) {
 }
 
 template <typename T> T encodeCI(T Result) {
-  // `c.lui rd, 0` is illegal, it will be converted  to `c.li rd, 0` when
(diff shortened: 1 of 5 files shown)


jonathonpenix:
FWIW, I'm omitting a run test as I think triggering this requires mapping something into addresses 0-2047 which I don't think qemu usermode/linux are pleased with.

quic-seaswara (approved):

lenary (approved):
LGTM

PR #1391, opened 2026-07-02
[Map] Add line-aware linker script context in map output

quic-seaswara:
The map file and diagnostic output now include the source line number from the linker script alongside each rule, assignment, INCLUDE directive, MEMORY, and output section description. This makes it easier to trace every map-file entry back to the exact location in the linker script that produced it.

Users can use editor support to jump to the linker script to either fix or change the linker script to account for any error.


test/AArch64/standalone/linkerscript/LinkerScriptExpressions/LinkerScriptExpressions.test
```diff
@@ -8,7 +8,7 @@ RUN: %not %link %linkopts -march aarch64 %t1.o  -o %t.out --section-start=.my_st
 
 #CHECK-DAG: {{[0-9]+}}: {{[0-9]+}}   0 NOTYPE  GLOBAL DEFAULT  ABS _myvar2
 #CHECK-DAG: {{[0-9]+}}: {{[0-9]+}}   0 NOTYPE  GLOBAL DEFAULT  ABS _myvar3
-#ERR: Fatal: {{.*}}errorexpn.t: Error: undefined symbol '.my_start' referenced in expression
+#ERR: Fatal: {{.*}}errorexpn.t:{{.*}}: Error: undefined symbol '.my_start' referenced in expression
```
Steven6798, line 11: Maybe match numbers instead of everything?

Steven6798 (approved):
Looks good.

PR #1393, opened 2026-07-02
[AArch64] Fix TLSDESC relaxation for non-preemptible symbols

quic-seaswara:
When building a dynamic executable (-dy), TLSDESC relocations for non-preemptible TLS symbols must be relaxed.  Previously the relaxation only triggered for fully static binaries (isCodeStatic()), leaving dynamic executables with descriptor calls that would fault at runtime because no resolver is registered for the local symbol.

### DIFF
--- a/lib/Target/AArch64/AArch64Relocator.cpp
+++ b/lib/Target/AArch64/AArch64Relocator.cpp
@@ -331,7 +331,7 @@ void AArch64Relocator::scanLocalReloc(InputFile &pInput, Relocation &pReloc,
     if (rsym->reserved() & ReserveGOT)
       return;
 
-    if (config().isCodeStatic()) {
+    if (config().isCodeStatic() || config().isBuildingExecutable()) {
       AArch64GOT *G = m_Target.createGOT(GOT::TLS_IE, Obj, rsym);
       rsym->setReserved(rsym->reserved() | ReserveGOT);
       G->setValueType(GOT::TLSStaticSymbolValue);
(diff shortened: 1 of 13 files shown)


lib/Target/AArch64/AArch64Relocator.cpp
```diff
@@ -1251,10 +1264,10 @@ Relocator::Result tls_tlsdesc_lo(Relocation &pReloc,
                                  AArch64Relocator &pParent) {
   Relocator::DWord A = pReloc.addend();
 
-  if (!(pReloc.symInfo()->reserved() & Relocator::ReserveGOT)) {
+  if (pParent.isTLSDescStatic(pReloc)) {
     Relocator::DWord X =
         pParent.getSymValue(&pReloc) + AArch64LDBackend::getStaticTCBSize();
-    // Convert to movk, save to x0
+    // Convert to movk with x0 as destination
```
efriedma-quic, line 1270: Why do we want to write to x0, specifically?

quic-seaswara: I have not changed that part of the code. It looks like the old implementation also was using x0 as the destination. Preserving the behavior.

quic-seaswara: https://[raw.githubusercontent.com/ARM-software/abi-aa/main/design-documents/tlsdesc-resolvers.rst](https://raw.githubusercontent.com/ARM-software/abi-aa/main/design-documents/tlsdesc-resolvers.rst)

It appears that x0 contains the TP relative offset and looks like that is what was done to preserve the semantics.

Steven6798 (approved):

lib/Target/AArch64/AArch64Relocator.cpp
efriedma-quic: It feels weird, but okay, I guess.

efriedma-quic (approved):
LGTM

PR #1395, opened 2026-07-03
refactor: establish HasHighSectionCount once during file parsing

### LINKED ISSUE #1303: Move `setHasHighSectionCount()` call outside `ELFReader<ELFT>::createSymbol`
Currently, while creating symbols in the API `ELFReader<ELFT>::createSymbol`, we have this check:

```cpp
if (rawSym.st_shndx == llvm::ELF::SHN_XINDEX) {
    EFileBase->setHasHighSectionCount();
    is_ordinary = false;
  }
```

This check should not be per-symbol. We can instead use the value of `e_shnum` entry in the ELF header of the input, or check the presence of section of type `SHT_SYMTAB_SHNDX` to determine if an input file has high section count. This can change this check from per-symbol to per-input-file.

yael-shr:
### Description

This PR addresses the design issue where the file-level property
`HasHighSectionCount` was being updated from the per-symbol path in
`createSymbol()`.

### Changes
- Move the `setHasHighSectionCount()` call from `createSymbol()` to
  `setSectionInInputFile()` when `SHT_SYMTAB_SHNDX` is encountered.
- Keep `createSymbol()` responsible only for determining the local
  `is_ordinary` state.

This change moves the determination of a file-level property to the
ELF parsing stage, improving separation of responsibilities and
keeping file state management out of the per-symbol path.

Fixes #1303

### DIFF
--- a/lib/Readers/ELFReader.cpp
+++ b/lib/Readers/ELFReader.cpp
@@ -121,6 +121,7 @@ void ELFReader<ELFT>::setSectionInInputFile(ELFSection *S,
     break;
   case llvm::ELF::SHT_SYMTAB_SHNDX:
     EFileBase->setExtendedSymbolTable(S);
+    EFileBase->setHasHighSectionCount();
     break;
   case llvm::ELF::SHT_STRTAB:
     EFileBase->setStringTable(S);
@@ -239,12 +240,8 @@ ELFReader<ELFT>::createSymbol(llvm::StringRef stringTable, Elf_Sym rawSym,
   ELFFileBase *EFileBase = llvm::cast<ELFFileBase>(&m_InputFile);
 
   bool is_ordinary = true;
-  // FIXME: Instead of checking this when creating each symbol (which can be
-  // relatively expensive as there can be too many symbols), can we instead set
-  // this using 'e_shnum'? Or perhaps, by seeing if there is any entry in
-  // section of type SHT_SYMTAB_SHNDX?
+
   if (rawSym.st_shndx == llvm::ELF::SHN_XINDEX) {
-    EFileBase->setHasHighSectionCount();
     is_ordinary = false;
   }
 


parth-07 (commented):
Hi @yael-shr,

Can you please add sign-off information in the commit? (it can be done by running `git commit -s --amend --no-edit`)

parth-07 (approved):
LGTM.

parth-07:
@yael-shr Can you please make the clang-format workflow happy?

yael-shr:
Done, I fixed the clang-format issue and updated the PR. Thanks!

parth-07:
@quic-seaswara @quic-areg Can you please review this PR?

quic-seaswara (approved):

lib/Readers/ELFReader.cpp
```diff
@@ -121,6 +121,7 @@ void ELFReader<ELFT>::setSectionInInputFile(ELFSection *S,
     break;
   case llvm::ELF::SHT_SYMTAB_SHNDX:
     EFileBase->setExtendedSymbolTable(S);
+    EFileBase->setHasHighSectionCount();
```
parth-07, line 124: @yael-shr Instead of explicitly setting it here, can we just modify `ObjectFile::hasHighSectionCount()` function to return true if `ELFFileBase::getExtendedSymbolTable()` returns non-null?

parth-07 (changes requested):

lib/Readers/ELFReader.cpp
yael-shr: Thanks for the suggestion! I understand the idea of removing the duplicated `HasHighSectionCount` state and deriving it from the presence of `ExtendedSymbolTable`.

I noticed that `getExtendedSymbolTable()` is currently part of `ELFFileBase`, while `hasHighSectionCount()` is defined in `ObjectFile`. Could you please clarify the preferred approach for connecting these two? Should `hasHighSectionCount()` be moved/overridden in `ELFFileBase`, or is there another design you had in mind?

Thanks!

yael-shr:
Hi @parth-07, just a gentle follow-up on this PR when you have a moment.

I'm still waiting for your guidance on the requested design change before I proceed.

Thanks!

Issue #1362, opened 2026-06-30
[X86] GOTPCRELX relaxation falls back to GOT when displacement exceeds 32-bit range


bsaharsh:
When implementing GOTPCRELX relaxation, the linker must handle the case where the direct displacement S + A - P does not fit in a signed 32-bit field — for example when a linker script places the symbol more than 2 GB from the instruction. In this case relaxation is not valid and the linker must keep GOT indirection.

Reproduce :
```bash
cat >1.c <<\EOF
int u=5;
int* foo(){
    return &u;
}
EOF

int u = 5;
cat >script.t <<\EOF
SECTIONS {
    .text 0x0 : { *(.text*) }
    /DISCARD/ : { *(.eh_frame) *(.eh_frame_hdr) *(.note*) *(.comment) }
    .data 0x100000000 : { *(.data*) }
}
EOF
LDs=(ld.eld ld.lld ld.bfd)
SFs=(eld lld bfd)

clang -fdata-sections -ffunction-sections -target x86_64-linux-gnu 1.c -c -fPIC -o 1.o

for i in "${!SFs[@]}"; do
  ${LDs[$i]} -o a.${SFs[$i]}.out 1.o -T script.t
done
```
lld handles this correctly via a post-layout relaxOnce() pass that checks the computed displacement and reverts overflowing relaxations to GOT indirection, then re-runs layout. mold handles it via a conservative approach: always creates a GOT slot in scan, checks the range at apply time and falls back to the GOT path if overflow.

bfd does not handle this gracefully — it errors out:
1.c:(.text.foo+0x7): failed to convert GOTPCREL relocation against 'u'; relink with --no-relax
ld.bfd: final link failed

PR #1396, opened 2026-07-03
[x86] Support R_X86_64_GOTPCRELX relaxation

### LINKED ISSUE #1361: [X86] Add Support for GOTPCRELX relaxation for x86-64
ELD does not implement any GOTPCRELX relaxation for the x86-64 backend. The compiler emits R_X86_64_GOTPCRELX (type 41) and R_X86_64_REX_GOTPCRELX (type 42) as an explicit opt-in signal that the linker may rewrite the GOT-indirect instruction into a direct RIP-relative form
when the target symbol is non-preemptible. ELD currently treats both types identically to the old non-relaxable R_X86_64_GOTPCREL (type 9), always allocating a GOT slot and using GOT indirection even when the symbol is provably non-preemptible. This results in unnecessary GOT
slots, unnecessary dynamic relocations, and one extra memory load per access at runtime compared to lld and bfd.

This meta issue tracks all sub-tasks needed to implement correct GOTPCRELX relaxation, including the basic transform, the range overflow fallback, and the GOT+PLT redundancy fix.

bsaharsh:
Implement relaxation for R_X86_64_GOTPCRELX (type 41) relocations.

For non-preemptible symbols, GOT-indirect references may be rewritten to direct RIP-relative accesses when:

- The symbol is non-preemptible (hidden, protected, bound by -Bsymbolic, or resolved in an executable)
- The symbol is not STT_GNU_IFUNC
- --relax is enabled
- The resulting S + A - P displacement fits in a signed 32-bit range

If the relaxed displacement cannot be represented in 32 bits, the link fails with an error.

Introduces the x86-64 relaxation framework modeled on the RISCV pattern.

This change only handles R_X86_64_GOTPCRELX (41). Support for R_X86_64_REX_GOTPCRELX (42) will be added separately.

Tests:
- GOTPCRELXRelax_Basic.test
- GOTPCRELXRelax_Preemptible.test
- GOTPCRELXRelax_Type9.test
- GOTPCRELXRelax_Overflow.test
- GOTPCRELXRelax_AddendNonM4.test
Fixes https://github.com/qualcomm/eld/issues/1361

### DIFF
--- a/include/eld/Config/GeneralOptions.h
+++ b/include/eld/Config/GeneralOptions.h
@@ -924,6 +924,10 @@ public:
 
   bool getRISCVRelax() const { return BRiscvRelax; }
 
+  void setRelax(bool Value) { ShouldRelax = Value; }
+
+  bool getRelax() const { return ShouldRelax; }
+
   void setRISCVZeroRelax(bool Relax) { RiscvZeroRelax = Relax; }
 
   bool getRISCVZeroRelax() const { return RiscvZeroRelax; }
@@ -1323,6 +1327,7 @@ private:
   bool DisableGuardForWeakUndefs = false; // hexagon specific option to
                                           // disable guard functionality.
   bool BRiscvRelax = true;                // enable riscv relaxation
+  bool ShouldRelax = false; // x86-64 GOTPCRELX relaxation (opt-in via --relax)
   bool RiscvZeroRelax = true;             // Zero-page relaxation
   bool RiscvGPRelax = true;               // GP relaxation
   bool BRiscvRelaxToC = true; // enable riscv relax to compressed code
--- a/include/eld/Diagnostics/DiagBackends.inc
+++ b/include/eld/Diagnostics/DiagBackends.inc
@@ -90,6 +90,9 @@ DIAG(expected_group_section, DiagnosticEngine::InternalError,
      "Expected %0 to be a 'group' section.")
 DIAG(warn_ignore_reloc_addend, DiagnosticEngine::Warning,
      "%2: ignoring addend for relocation `%1' referring symbol %0")
+DIAG(error_x86_gotpcrelx_relax_overflow, DiagnosticEngine::Error,
+     "%0: relaxing relocation %1 out of range: %2 is not in [%3, %4]; "
+     "references '%5'; relink with --no-relax")
 DIAG(error_build_id_not_hex, DiagnosticEngine::Error,
     "build id specified is not hexadecimal : %0")
 DIAG(error_unknown_build_id, DiagnosticEngine::Error,
--- a/include/eld/Diagnostics/DiagVerbose.inc
+++ b/include/eld/Diagnostics/DiagVerbose.inc
@@ -34,6 +34,8 @@ DIAG(not_relaxed, DiagnosticEngine::Verbose,
 DIAG(relax_to_compress, DiagnosticEngine::Verbose,
      "%0 : relaxing instruction 0x%1 to compressed instruction 0x%2 for symbol "
      "%3 in section %4+0x%5 file %6")
+DIAG(trace_relax_gotpcrelx, DiagnosticEngine::Trace,
+     "%0: relaxed GOTPCRELX (%1) for symbol '%2' to PC-relative access")
 DIAG(verbose_ehframe_remove_fde, DiagnosticEngine::Verbose, "EhFrame %0")
 DIAG(verbose_ehframe_read_fde, DiagnosticEngine::Verbose, "EhFrame %0")
 DIAG(verbose_ehframe_read_cie, DiagnosticEngine::Verbose, "EhFrame %0")
--- a/include/eld/Driver/x86_64LinkerOptions.td
+++ b/include/eld/Driver/x86_64LinkerOptions.td
@@ -5,3 +5,13 @@
 //===----------------------------------------------------------------------===//
 
 include "GnuLinkerOptions.td"
+
+def grp_x86_64_linker : OptionGroup<"opts">, HelpText<"X86-64 Linker Options">;
+
+def relax : Flag<["--"], "relax">,
+            HelpText<"Enable GOTPCRELX relaxation (GOT-indirect to PC-relative)">,
+            Group<grp_x86_64_linker>;
+
+def no_relax : Flag<["--"], "no-relax">,
+               HelpText<"Disable GOTPCRELX relaxation">,
+               Group<grp_x86_64_linker>;
--- a/lib/Target/X86/x86_64LDBackend.cpp
+++ b/lib/Target/X86/x86_64LDBackend.cpp
@@ -7,6 +7,7 @@
 //===----------------------------------------------------------------------===//
 #include "x86_64LDBackend.h"
 #include "eld/Config/LinkerConfig.h"
+#include "eld/Fragment/RegionFragment.h"
 #include "eld/Fragment/Stub.h"
 #include "eld/Input/ELFObjectFile.h"
 #include "eld/Object/ObjectBuilder.h"
@@ -22,7 +23,10 @@
 #include "x86_64StandaloneInfo.h"
 #include "llvm/BinaryFormat/ELF.h"
 #include "llvm/Support/Casting.h"
+#include "llvm/Support/Endian.h"
+#include "llvm/Support/MathExtras.h"
 #include "llvm/TargetParser/Host.h"
+#include <limits>
 
 using namespace eld;
 using namespace llvm;
@@ -96,6 +100,190 @@ void x86_64LDBackend::initTargetSymbols() {
 bool x86_64LDBackend::initBRIslandFactory() { return true; }
 
 bool x86_64LDBackend::initStubFactory() { return true; }
+
+x86_64LDBackend::GOTPCRELXOpcode
+x86_64LDBackend::getGOTPCRELXOpcode(uint8_t op, uint8_t modrm) const {
+  if (op == 0x8b)
+    return GOTPCRELXOpcode::MOV;
+  if (op == 0xff && modrm == 0x15)
+    return GOTPCRELXOpcode::CALL;
+  if (op == 0xff && modrm == 0x25)
+    return GOTPCRELXOpcode::JMP;
+  return GOTPCRELXOpcode::Unknown;
+}
+
+bool x86_64LDBackend::isGOTPCRELXRelaxable(const Relocation *reloc) const {
+  // Partial links have no final layout, so relaxation is not possible; keep
+  // the GOT slot and let the final link decide.
+  if (config().isLinkPartial())
+    return false;
+  if (reloc->type() != llvm::ELF::R_X86_64_GOTPCRELX)
+    return false;
+  // GNU as may emit GOTPCRELX with addend != -4. Such an instruction does not
+  // load the full GOT entry (e.g. movl x@GOTPCREL+4(%rip) loads the high 32
+  // bits), so it cannot be relaxed. Matches lld's adjustGotPcExpr.
+  if (static_cast<int64_t>(reloc->addend()) != -4)
+    return false;
+  const ResolveInfo *rsym = reloc->symInfo();
+  // IFUNC symbols must retain their GOT slot: the IRELATIVE resolver runs at
+  // load time and deposits the selected implementation's address there.
+  // Relaxing to a PC-relative access would bypass that resolution. Note a
+  // hidden/local IFUNC is non-preemptible, so the isIFunc() check is not
+  // redundant with isSymbolPreemptible().
+  if (isSymbolPreemptible(*rsym) || rsym->isIFunc())
+    return false;
+  auto *RF = llvm::dyn_cast<RegionFragment>(reloc->targetRef()->frag());
+  if (!RF)
+    return false;
+  uint32_t offset = reloc->targetRef()->offset();
+  // The opcode and ModR/M byte sit at offset-2 and offset-1 relative to the
+  // displacement field. A corrupt or hand-crafted object could place the
+  // relocation at offset 0 or 1, making those reads out-of-bounds.
+  if (offset < 2)
+    return false;
+  // The two bytes preceding the 4-byte displacement hold the opcode and,
+  // for CALL/JMP, the ModR/M byte. Only MOV/CALL/JMP are relaxable.
+  const uint8_t *loc =
+      reinterpret_cast<const uint8_t *>(RF->getRegion().data()) + offset;
+  uint8_t op = loc[-2];
+  uint8_t modrm = loc[-1];
+  return getGOTPCRELXOpcode(op, modrm) != GOTPCRELXOpcode::Unknown;
+}
+
+bool x86_64LDBackend::shouldIgnoreRelocSync(Relocation *reloc) const {
+  return isGOTPCRELXRelaxCandidate(reloc);
+}
+
+eld::Expected<void>
+x86_64LDBackend::postProcessing(llvm::FileOutputBuffer &pOutput) {
+  ELDEXP_RETURN_DIAGENTRY_IF_ERROR(GNULDBackend::postProcessing(pOutput));
+
+  // Relaxation rewrites bytes in the laid-out output image; it is not
+  // applicable to partial links, which have no final layout.
+  if (config().options().getRelax() && !config().isLinkPartial())
+    ELDEXP_RETURN_DIAGENTRY_IF_ERROR(doRelax(pOutput));
+
+  return {};
+}
+
+eld::Expected<void> x86_64LDBackend::doRelax(llvm::FileOutputBuffer &pOutput) {
+  uint8_t *buf = pOutput.getBufferStart();
+  // The scan phase already identified every relaxation candidate (skipping
+  // debug relocations and internal files, which never carry a relaxable
+  // relocation). Iterate that cached set instead of re-walking all
+  // relocations, and dispatch on the relocation type. Future relaxable types
+  // (e.g. R_X86_64_REX_GOTPCRELX, TLS relaxations) add a case here.
+  for (Relocation *reloc : m_GOTPCRELXRelaxCandidates) {
+    switch (reloc->type()) {
+    case llvm::ELF::R_X86_64_GOTPCRELX:
+      ELDEXP_RETURN_DIAGENTRY_IF_ERROR(relaxGOTPCRELXReloc(reloc, buf));
+      break;
+    default:
+      // Only relaxable types are recorded as candidates; skip anything else.
+      break;
+    }
+  }
+  return {};
+}
+
+eld::Expected<void> x86_64LDBackend::relaxGOTPCRELXReloc(Relocation *reloc,
+                                                         uint8_t *buf) {
+  // Only RegionFragment relocations are recorded as relaxation candidates
+  // (isGOTPCRELXRelaxable requires a RegionFragment), so this cast must
+  // succeed.
+  auto *RF = llvm::dyn_cast<RegionFragment>(reloc->targetRef()->frag());
+  assert(RF && "GOTPCRELX relax candidate must reference a RegionFragment");
+
+  uint32_t offset = reloc->targetRef()->offset();
+  // isGOTPCRELXRelaxable checks offset >= 2 before recording a candidate,
+  // so this must hold for every relocation that reaches here.
+  assert(offset >= 2 && "GOTPCRELX relax candidate must have offset >= 2");
+  // Opcode/ModR/M bytes live just before the 4-byte displacement field.
+  const uint8_t *loc =
+      reinterpret_cast<const uint8_t *>(RF->getRegion().data()) + offset;
+
+  ResolveInfo *rsym = reloc->symInfo();
+  Relocator::Address S = reloc->symValue(m_Module);
+  Relocator::DWord A = reloc->addend();
+  Relocator::DWord P = reloc->place(m_Module);
+  // Relaxed form is a direct PC-relative access: displacement = S + A - P.
+  int64_t val = static_cast<int64_t>(S) + static_cast<int64_t>(A) -
+                static_cast<int64_t>(P);
+
+  if (!llvm::isInt<32>(val)) {
+    // Unlike lld (which relayouts and can fall back to the GOT slot), eld
+    // decides relaxability at scan time and has already dropped the GOT slot,
+    // so an out-of-range displacement here is a hard error. Report it with the
+    // source location and abort the link (matching GNU ld's --no-relax hint).
+    std::string location;
+    ELFSection *overflowSect = RF->getOwningSection();
+    assert(overflowSect &&
+           "RegionFragment in relax candidate must have an owning section");
+    location = overflowSect->getLocation(offset, config().options());
+    return std::make_unique<plugin::DiagnosticEntry>(plugin::DiagnosticEntry(
+        Diag::error_x86_gotpcrelx_relax_overflow,
+        {std::move(location), "R_X86_64_GOTPCRELX", llvm::itostr(val),
+         llvm::itostr(std::numeric_limits<int32_t>::min()),
+         llvm::itostr(std::numeric_limits<int32_t>::max()),
+         std::string(rsym->name())}));
+  }
+
+  FragmentRef::Offset off = reloc->targetRef()->getOutputOffset(m_Module);
+  if (off == (FragmentRef::Offset)-1)
+    return {};
+  size_t out_off = reloc->targetRef()->getOutputELFSection()->offset() + off;
+
+  uint8_t op = loc[-2];
+  uint8_t modrm = loc[-1];
+  const char *kind = nullptr;
+  switch (getGOTPCRELXOpcode(op, modrm)) {
+  case GOTPCRELXOpcode::MOV:
+    // mov foo@GOTPCREL(%rip), reg  ->  lea foo(%rip), reg
+    // The GOT-load opcode 0x8b becomes the address-load opcode 0x8d; the
+    // ModR/M and displacement encoding are otherwise identical.
+    buf[out_off - 2] = 0x8d;
+    llvm::support::endian::write32le(buf + out_off, static_cast<uint32_t>(val));
+    kind = "mov->lea";
+    break;
+  case GOTPCRELXOpcode::CALL:
+    // call *foo@GOTPCREL(%rip)  ->  addr32 call foo
+    // The 2-byte "ff 15" indirect call is 6 bytes; the direct "e8 rel32" is
+    // 5 bytes, so a 0x67 (addr32) prefix pads it back to 6 bytes.
+    buf[out_off - 2] = 0x67;
+    buf[out_off - 1] = 0xe8;
+    llvm::support::endian::write32le(buf + out_off, static_cast<uint32_t>(val));
+    kind = "call->direct";
+    break;
+  case GOTPCRELXOpcode::JMP:
+    // jmp *foo@GOTPCREL(%rip)  ->  jmp foo; nop
+    // "ff 25" indirect jmp (6 bytes) becomes "e9 rel32" direct jmp (5 bytes)
+    // followed by a 0x90 nop for padding. The displacement shifts one byte
+    // earlier (no ModR/M), so it is written at out_off-1 with val+1 to account
+    // for the one-byte-shorter instruction prefix.
+    buf[out_off - 2] = 0xe9;
+    llvm::support::endian::write32le(buf + out_off - 1,
+                                     static_cast<uint32_t>(val + 1));
+    buf[out_off + 3] = 0x90;
+    kind = "jmp->direct";
+    break;
+  case GOTPCRELXOpcode::Unknown:
+    // isGOTPCRELXRelaxable only records MOV/CALL/JMP candidates, so this is
+    // not expected; leave the instruction untouched if it is ever reached.
+    break;
+  }
+
+  if (m_Module.getPrinter()->traceRelax()) {
+    std::string location;
+    ELFSection *traceSect = RF->getOwningSection();
+    assert(traceSect &&
+           "RegionFragment in relax candidate must have an owning section");
+    location = traceSect->getLocation(offset, config().options());
+    config().raise(Diag::trace_relax_gotpcrelx)
+        << location << kind << rsym->name();
+  }
+
+  return {};
+}
 
 /// finalizeSymbol - finalize the symbol value
 bool x86_64LDBackend::finalizeTargetSymbols() {
--- a/lib/Target/X86/x86_64LDBackend.h
+++ b/lib/Target/X86/x86_64LDBackend.h
@@ -14,6 +14,7 @@
 #include "eld/Target/GNULDBackend.h"
 #include "x86_64PLT.h"
 #include "llvm/BinaryFormat/ELF.h"
+#include <unordered_set>
 
 namespace eld {
 
@@ -47,6 +48,10 @@ public:
   bool initBRIslandFactory() override;
 
   bool initStubFactory() override;
+
+  bool shouldIgnoreRelocSync(Relocation *Reloc) const override;
+
+  eld::Expected<void> postProcessing(llvm::FileOutputBuffer &pOutput) override;
 
   /// getTargetSectionOrder - compute the layout order of target section
   unsigned int getTargetSectionOrder(const ELFSection &pSectHdr) const override;
@@ -98,6 +103,27 @@ public:
 
   void sortRelocation(ELFSection &pSection) override;
 
+  /// Returns true if the relocation is eligible for GOTPCRELX relaxation.
+  /// Checks: type R_X86_64_GOTPCRELX, addend == -4, non-preemptible,
+  /// not IFUNC, and opcode is MOV/CALL/JMP (read directly from the input
+  /// fragment). Relaxation is disabled for partial links.
+  bool isGOTPCRELXRelaxable(const Relocation *reloc) const;
+
+  /// Records a relocation that the scan phase has decided to relax, so that
+  /// postProcessing can iterate only these candidates instead of re-walking
+  /// every relocation. Thread-safe: called from the parallel scan under the
+  /// relocator mutex.
+  void recordGOTPCRELXRelaxCandidate(Relocation *reloc) {
+    m_GOTPCRELXRelaxCandidates.insert(reloc);
+  }
+
+  /// Returns true if this relocation was recorded as a relaxation candidate
+  /// during the scan phase. O(1) lookup used by shouldIgnoreRelocSync and
+  /// the apply-path guard to avoid re-deriving relaxability.
+  bool isGOTPCRELXRelaxCandidate(Relocation *reloc) const {
+    return m_GOTPCRELXRelaxCandidates.count(reloc) != 0;
+  }
+
   DynRelocType getDynRelocType(const Relocation *X) const override {
     if (X->type() == llvm::ELF::R_X86_64_GLOB_DAT)
       return DynRelocType::GLOB_DAT;
@@ -144,6 +170,18 @@ private:
 
   uint64_t maxBranchOffset() override { return 0; }
 
+  enum class GOTPCRELXOpcode { MOV, CALL, JMP, Unknown };
+  GOTPCRELXOpcode getGOTPCRELXOpcode(uint8_t op, uint8_t modrm) const;
+
+  /// Rewrites the GOT-indirect instruction referenced by a single relaxation
+  /// candidate into its PC-relative form, patching the output buffer.
+  /// Returns an error (and does not patch) if the displacement overflows a
+  /// signed 32-bit field.
+  eld::Expected<void> relaxGOTPCRELXReloc(Relocation *reloc, uint8_t *buf);
+
+  /// Iterates the cached relaxation candidates and rewrites each one.
+  eld::Expected<void> doRelax(llvm::FileOutputBuffer &pOutput);
+
 private:
   Relocator *m_pRelocator;
 
@@ -155,6 +193,11 @@ private:
   llvm::DenseMap<ResolveInfo *, x86_64GOT *> m_GOTMap;
   llvm::DenseMap<ResolveInfo *, x86_64GOT *> m_GOTPLTMap;
   llvm::DenseMap<ResolveInfo *, x86_64PLT *> m_PLTMap;
+  /// Relocations selected for GOTPCRELX relaxation during the scan phase.
+  /// Populated under the relocator mutex; consumed single-threaded in
+  /// postProcessing. unordered_set for O(1) membership checks in
+  /// shouldIgnoreRelocSync and the apply-path guard.
+  std::unordered_set<Relocation *> m_GOTPCRELXRelaxCandidates;
 };
 } // namespace eld
 
--- a/lib/Target/X86/x86_64Relocator.cpp
+++ b/lib/Target/X86/x86_64Relocator.cpp
@@ -292,6 +292,25 @@ void x86_64Relocator::scanLocalReloc(InputFile &pInputFile, Relocation &pReloc,
     rsym->setReserved(rsym->reserved() | ReserveGOT);
     return;
   }
+  case llvm::ELF::R_X86_64_GOTPCRELX: {
+    std::lock_guard<std::mutex> relocGuard(m_RelocMutex);
+    // Relaxability is decided per relocation (it inspects this reference's own
+    // opcode bytes), so it must be checked before the per-symbol ReserveGOT
+    // short-circuit: a non-relaxable reference to the same symbol may have
+    // reserved a GOT slot first. With --relax, local symbols are always
+    // non-preemptible; skip the GOT slot for the relaxable reference and record
+    // it so postProcessing can patch it without re-walking.
+    if (config().options().getRelax() &&
+        m_Target.isGOTPCRELXRelaxable(&pReloc)) {
+      m_Target.recordGOTPCRELXRelaxCandidate(&pReloc);
+      return;
+    }
+    if (rsym->reserved() & ReserveGOT)
+      return;
+    CreateGOT(Obj, pReloc, !config().isCodeStatic(), m_Target);
+    rsym->setReserved(rsym->reserved() | ReserveGOT);
+    return;
+  }
   case llvm::ELF::R_X86_64_TLSLD: {
     std::lock_guard<std::mutex> relocGuard(m_RelocMutex);
     getTLSModuleID(pReloc.symInfo(), config().isCodeStatic());
@@ -402,9 +421,34 @@ void x86_64Relocator::scanGlobalReloc(InputFile &pInputFile, Relocation &pReloc,
     rsym->setReserved(rsym->reserved() | ReservePLT);
     return;
   }
-
-  case llvm::ELF::R_X86_64_GOTPCREL:
-  case llvm::ELF::R_X86_64_GOTPCRELX:
+  case llvm::ELF::R_X86_64_GOTPCRELX: {
+    std::lock_guard<std::mutex> relocGuard(m_RelocMutex);
+    // Relaxability is decided per relocation (it inspects this reference's own
+    // opcode bytes), so it must be checked before the per-symbol ReserveGOT
+    // short-circuit: a non-relaxable reference to the same symbol may have
+    // reserved a GOT slot first. With --relax, non-preemptible non-IFUNC
+    // symbols with a relaxable opcode are handled by postProcessing: no GOT
+    // slot is needed and an out-of-range displacement is a link error. An
+    // addend != -4 or an unknown opcode keeps the GOT slot.
+    if (config().options().getRelax() &&
+        m_Target.isGOTPCRELXRelaxable(&pReloc)) {
+      m_Target.recordGOTPCRELXRelaxCandidate(&pReloc);
+      return;
+    }
+    if (rsym->reserved() & ReserveGOT)
+      return;
+    CreateGOT(Obj, pReloc, !config().isCodeStatic(), m_Target);
+    rsym->setReserved(rsym->reserved() | ReserveGOT);
+    return;
+  }
+  case llvm::ELF::R_X86_64_GOTPCREL: {
+    std::lock_guard<std::mutex> relocGuard(m_RelocMutex);
+    if (rsym->reserved() & ReserveGOT)
+      return;
+    CreateGOT(Obj, pReloc, !config().isCodeStatic(), m_Target);
+    rsym->setReserved(rsym->reserved() | ReserveGOT);
+    return;
+  }
   case llvm::ELF::R_X86_64_REX_GOTPCRELX: {
     std::lock_guard<std::mutex> relocGuard(m_RelocMutex);
     if (rsym->reserved() & ReserveGOT)
@@ -688,6 +732,11 @@ Relocator::Result eld::relocGOTRelative(Relocation &pReloc,
   DiagnosticEngine *DiagEngine = pParent.config().getDiagEngine();
   ResolveInfo *symInfo = pReloc.symInfo();
   const GeneralOptions &options = pParent.config().options();
+
+  // For relaxable GOTPCRELX relocations, postProcessing handles the opcode
+  // patch and displacement. Skip apply here to avoid a null GOT entry lookup.
+  if (pParent.getTarget().isGOTPCRELXRelaxCandidate(&pReloc))
+    return Relocator::OK;
 
   Relocator::DWord A = pReloc.addend();
   Relocator::DWord P = pReloc.place(pParent.module());
--- a/test/x86_64/linux/GOTPCRELXRelaxation/GOTPCRELXRelax_BadInput.test
+++ b/test/x86_64/linux/GOTPCRELXRelaxation/GOTPCRELXRelax_BadInput.test
@@ -1,0 +1,30 @@
+#--GOTPCRELXRelax_BadInput.test----------Executable--------#
+# A GOTPCRELX relocation placed at offset 0 (via .reloc) means the opcode
+# bytes at loc[-2]/loc[-1] would be before the fragment buffer. The linker
+# must not crash; it must skip relaxation and keep the GOT slot instead.
+#START_TEST
+RUN: %clang %clangopts -c %p/Inputs/badinput.s -o %t.o
+RUN: %link %linkopts --relax -shared -o %t.so %t.o
+
+# Before: input object has a nop with a GOTPCRELX relocation at offset 0.
+RUN: %objdump -dr %t.o | %filecheck %s --check-prefix=BEFORE
+
+# After: output must not contain a leal (8d) — the relocation was not relaxed.
+RUN: %objdump -d %t.so | %filecheck %s --check-prefix=AFTER
+
+# A GOT slot must still exist (kept because the relocation was not relaxable).
+RUN: %readelf -r %t.so | %filecheck %s --check-prefix=RELOC
+#END_TEST
+
+# Input: nop at offset 0 with the GOTPCRELX reloc attached to it.
+BEFORE: <h>:
+BEFORE-NEXT: {{[ \t]+}}0: 90{{.*}}nop
+BEFORE: R_X86_64_GOTPCRELX
+
+# Output: not relaxed to leal (no opcode 8d).
+AFTER: <h>:
+AFTER-NOT: 8d {{.*}} leal
+
+# GOT slot kept with a RELATIVE dynamic relocation (foo is hidden/non-preemptible).
+RELOC: .rela.dyn
+RELOC: R_X86_64_RELATIVE
--- a/test/x86_64/linux/GOTPCRELXRelaxation/GOTPCRELXRelax_Overflow.test
+++ b/test/x86_64/linux/GOTPCRELXRelaxation/GOTPCRELXRelax_Overflow.test
@@ -1,0 +1,9 @@
+#--GOTPCRELXRelax_Overflow.test----------Shared Library--------#
+# Out-of-range: displacement overflows 32 bits, linker must error.
+#START_TEST
+RUN: %clang %clangopts -c %p/Inputs/overflow.s -o %t.o
+RUN: %not %link %linkopts --relax -T %p/Inputs/script_overflow.t -o %t.out %t.o 2>&1 \
+RUN:   | %filecheck %s
+#END_TEST
+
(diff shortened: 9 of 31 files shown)


include/eld/Driver/x86_64LinkerOptions.td
```diff
@@ -4,4 +4,10 @@
 // SPDX-License-Identifier: BSD-3-Clause
 //===----------------------------------------------------------------------===//
 
-include "GnuLinkerOptions.td"
\ No newline at end of file
+include "GnuLinkerOptions.td"
+
+def grp_x86_64_linker : OptionGroup<"opts">, HelpText<"X86-64 Linker Options">;
+
+def relax : Flag<["--"], "relax">,
```
parth-07, line 11: We should add `--no-relax` as well.

lib/Target/X86/x86_64Relocator.cpp
```diff
@@ -288,6 +288,15 @@ void x86_64Relocator::scanLocalReloc(InputFile &pInputFile, Relocation &pReloc,
     rsym->setReserved(rsym->reserved() | ReserveGOT);
     return;
   }
+  case llvm::ELF::R_X86_64_GOTPCRELX:
+  case llvm::ELF::R_X86_64_REX_GOTPCRELX: {
+    std::lock_guard<std::mutex> relocGuard(m_RelocMutex);
+    if (rsym->reserved() & ReserveGOT)
+      return;
+    CreateGOT(Obj, pReloc, !config().isCodeStatic(), m_Target);
```
parth-07, line 310: Do we always create GOT slot even when GOT-load can be relaxed, and thus may not be used?

test/x86_64/linux/GOTPCRELXRelaxation/GOTPCRELXRelax_Overflow.test
```diff
@@ -0,0 +1,10 @@
+#--GOTPCRELXRelax_Overflow.test----------Shared Library--------#
+# Out-of-range: displacement overflows 32 bits, linker must error.
+#START_TEST
+RUN: %clang %clangopts -c %p/Inputs/overflow.s -o %t.o
+RUN: not %link %linkopts --relax -T %p/Inputs/script_overflow.t -o %t.out %t.o 2>&1 \
```
parth-07:
```suggestion
RUN: %not %link %linkopts --relax -T %p/Inputs/script_overflow.t -o %t.out %t.o 2>&1 \
```

test/x86_64/linux/GOTPCRELXRelaxation/GOTPCRELXRelax_Overflow.test
```diff
@@ -0,0 +1,10 @@
...
+RUN:   | %filecheck %s
+#END_TEST
+
+CHECK: GOTPCRELX relaxation
```
parth-07: Can you add the full error message here?

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -99,6 +102,110 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
 
 bool x86_64LDBackend::initStubFactory() { return true; }
 
+bool x86_64LDBackend::readSection(InputFile &pInput, ELFSection *S) {
+  eld::LayoutInfo *layoutInfo = m_Module.getLayoutInfo();
+  if (S->isCode()) {
+    const char *Buf = pInput.getCopyForWrite(S->offset(), S->size());
+    eld::RegionFragmentEx *F =
+        make<RegionFragmentEx>(Buf, S->size(), S, S->getAddrAlign());
```
quic-seaswara: If you're only replacing instructions, RegionFragmentEx is likely overkill for this use case.

RegionFragmentEx is designed to replace code sequences and reduce code size by removing or rewriting code based on the final layout. If the replacement preserves the original code size, then RegionFragmentEx is not the right abstraction to use.

Another drawback is the additional memory overhead. RegionFragmentEx requires copying and modifying sections, which increases memory usage and complexity unnecessarily for a simple instruction replacement.

Instead, you could consider overriding shouldIgnoreRelocSync() and implementing the required logic in a postProcessing() hook to handle this relocation.

Do you think that approach would work?

quic-seaswara (changes requested):

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: Yes, I think It should work, as most of the relaxations doesn't change the size of the instruction, jmp and call are relaxed from 6 byte to 5 byte but they are padded with nop of 1 byte as per the ABI. 

Will update the approach. Thanks

include/eld/Driver/x86_64LinkerOptions.td
bsaharsh: Updated

lib/Target/X86/x86_64Relocator.cpp
bsaharsh: Updated

test/x86_64/linux/GOTPCRELXRelaxation/GOTPCRELXRelax_Overflow.test
bsaharsh: Updated

test/x86_64/linux/GOTPCRELXRelaxation/GOTPCRELXRelax_Overflow.test
bsaharsh: Added it.

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: I have updated the approach, Please Have a look

include/eld/Diagnostics/DiagBackends.inc
```diff
@@ -90,6 +90,9 @@ DIAG(expected_group_section, DiagnosticEngine::InternalError,
      "Expected %0 to be a 'group' section.")
 DIAG(warn_ignore_pcrel_lo_addend, DiagnosticEngine::Warning,
      "%2: ignoring addend for relocation `%1' referring symbol %0")
+DIAG(x86_gotpcrelx_relax_overflow, DiagnosticEngine::Error,
```
parth-07: Please prefix the diagnostic name with the diagnostic severity.

lib/Target/X86/x86_64Relocator.cpp
```diff
@@ -401,8 +413,28 @@ void x86_64Relocator::scanGlobalReloc(InputFile &pInputFile, Relocation &pReloc,
...
+  }
+  case llvm::ELF::R_X86_64_GOTPCRELX: {
+    std::lock_guard<std::mutex> relocGuard(m_RelocMutex);
+    if (rsym->reserved() & ReserveGOT)
+      return;
+    // With --relax, non-preemptible non-IFUNC symbols are handled by
+    // postProcessing. No GOT slot needed; overflow causes a link error.
+    if (config().options().getRelax() &&
+        !m_Target.isSymbolPreemptible(*rsym) &&
+        !rsym->isIFunc())
```
parth-07: Do we relax all instructions that can make use of `GOTPCRELX` relocation? or do we only relax subset of instructions that use `GOTPCRELX` relocation? It seems that we only relax `mov (0x8b)`, `call (0xff /2, modrm 0x15)`, and `jmp (0xff /4, modrm 0x25)` instructions. If there is any other instruction that uses `GOTPCRELX` and we do not relax that instruction, then we should continue creating GOT slot for that. Can you please verify this?

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -99,6 +102,97 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
 
 bool x86_64LDBackend::initStubFactory() { return true; }
 
+bool x86_64LDBackend::shouldIgnoreRelocSync(Relocation *reloc) const {
+  if (reloc->type() != llvm::ELF::R_X86_64_GOTPCRELX)
+    return false;
+  ResolveInfo *rsym = reloc->symInfo();
+  return config().options().getRelax() &&
```
parth-07: We are duplicating the relax-eligibility condition for a symbol at multiple places. Can you please refactor this into a single function?

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -99,6 +102,97 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
...
+  ELDEXP_RETURN_DIAGENTRY_IF_ERROR(GNULDBackend::postProcessing(pOutput));
+
+  if (config().options().getRelax())
+    ELDEXP_RETURN_DIAGENTRY_IF_ERROR(relaxGOTPCRELX(pOutput));
+
+  return {};
+}
+
+eld::Expected<void>
+x86_64LDBackend::relaxGOTPCRELX(llvm::FileOutputBuffer &pOutput) {
```
parth-07: lld has this below comment:

> // Only R_X86_64_[REX_]|[CODE_4_]GOTPCRELX can be relaxed. GNU as may emit
  // GOTPCRELX with addend != -4. Such an instruction does not load the full GOT
  // entry, so we cannot relax the relocation. E.g. movl x@GOTPCREL+4(%rip),
  // %rax (addend=0) loads the high 32 bits of the GOT entry.

Are we properly addressing this?

parth-07 (changes requested):

include/eld/Diagnostics/DiagBackends.inc
bsaharsh: Added the error prefix.

lib/Target/X86/x86_64Relocator.cpp
bsaharsh: I have refactored the multiple checks, into a single function now it checks for the addend and the opcode too, and if it doesn't match then it uses the GOT slot for it.

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: Refactored into a single function.

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: Have added the missing addend check too.

BhoumikPatidar:
Please add tests for call and jmp rewrite paths as well
And test for -pie mode as well.
And assert that the GOT slot is actually absent after relaxation.

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -99,6 +102,111 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
...
+          m_Module.setFailure(true);
+          continue;
+        }
+
+        FragmentRef::Offset off = reloc->targetRef()->getOutputOffset(m_Module);
+        if (off == (FragmentRef::Offset)-1)
+          continue;
+        size_t out_off =
+            reloc->targetRef()->getOutputELFSection()->offset() + off;
+
```
BhoumikPatidar: Can we have a single helper for the opcode decode and classify logic rather than repeating in `isGOTPCRELXRelaxable` and `relaxGOTPCRELX`?

test/x86_64/linux/GOTPCRELXRelaxation/GOTPCRELXRelax_Overflow.test
```diff
@@ -0,0 +1,9 @@
+#--GOTPCRELXRelax_Overflow.test----------Shared Library--------#
+# Out-of-range: displacement overflows 32 bits, linker must error.
+#START_TEST
+RUN: %clang %clangopts -c %p/Inputs/overflow.s -o %t.o
+RUN: %not %link %linkopts --relax -T %p/Inputs/script_overflow.t -o %t.out %t.o 2>&1 \
+RUN:   | %filecheck %s
+#END_TEST
+
+CHECK: GOTPCRELX relaxation: symbol 'sym' at address [0x{{[0-9a-fA-F]+}}] is unreachable with a 32-bit displacement [0x{{[0-9a-fA-F]+}} overflows]
```
parth-07: Can you improve this diagnostic by adding filename information? Please check the format of other relocation overflow diagnostics in eld. Additionally, please also add `use --no-relax` in the diagnostic as well.

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -99,6 +102,111 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
 
 bool x86_64LDBackend::initStubFactory() { return true; }
 
+bool x86_64LDBackend::isGOTPCRELXRelaxable(const Relocation *reloc) const {
```
parth-07, line 115: We should not relax when the link is partial.

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -99,6 +102,127 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
...
+bool x86_64LDBackend::shouldIgnoreRelocSync(Relocation *reloc) const {
+  return config().options().getRelax() && isGOTPCRELXRelaxable(reloc);
+}
+
+eld::Expected<void>
+x86_64LDBackend::postProcessing(llvm::FileOutputBuffer &pOutput) {
+  ELDEXP_RETURN_DIAGENTRY_IF_ERROR(GNULDBackend::postProcessing(pOutput));
+
+  if (config().options().getRelax())
+    ELDEXP_RETURN_DIAGENTRY_IF_ERROR(relaxGOTPCRELX(pOutput));
```
quic-seaswara: This function can be called doRelax instead of relaxGOTPCRELX.

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -99,6 +102,127 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
...
+    if (!ObjFile)
+      continue;
+    for (auto &rs : ObjFile->getRelocationSections()) {
+      if (rs->isIgnore() || rs->isDiscard())
+        continue;
+      for (Relocation *reloc : rs->getLink()->getRelocations()) {
+        if (reloc->type() != llvm::ELF::R_X86_64_GOTPCRELX)
+          continue;
+
+        if (!isGOTPCRELXRelaxable(reloc))
```
quic-seaswara: The body of the loop can be made to a function, create a switch case for what relocations would be candidates for relaxation. 

Follow up : Need map file annotations to know what relaxation was done, targets support map file annotations.  See RISC-V for example.

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -99,6 +102,127 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
...
+
+eld::Expected<void>
+x86_64LDBackend::relaxGOTPCRELX(llvm::FileOutputBuffer &pOutput) {
+  uint8_t *buf = pOutput.getBufferStart();
+
+  for (auto &input : m_Module.getObjectList()) {
+    ELFObjectFile *ObjFile = llvm::dyn_cast<ELFObjectFile>(input);
+    if (!ObjFile)
+      continue;
+    for (auto &rs : ObjFile->getRelocationSections()) {
```
quic-seaswara: I dont think you want to go through debug relocations here. 

Going through debug relocations will have an impact on link time.

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -99,6 +102,127 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
...
+  return {};
+}
+
+eld::Expected<void>
+x86_64LDBackend::relaxGOTPCRELX(llvm::FileOutputBuffer &pOutput) {
+  uint8_t *buf = pOutput.getBufferStart();
+
+  for (auto &input : m_Module.getObjectList()) {
+    ELFObjectFile *ObjFile = llvm::dyn_cast<ELFObjectFile>(input);
+    if (!ObjFile)
```
quic-seaswara: What about internal files, can we skip ?

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -99,6 +102,127 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
...
+                                           static_cast<uint32_t>(val));
+          break;
+        case GOTPCRELXOpcode::JMP:
+          buf[out_off - 2] = 0xe9;
+          llvm::support::endian::write32le(buf + out_off - 1,
+                                           static_cast<uint32_t>(val + 1));
+          buf[out_off + 3] = 0x90;
+          break;
+        case GOTPCRELXOpcode::Unknown:
+          break;
```
quic-seaswara: Add support for --trace=relax.

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -99,6 +102,127 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
...
+  return classifyGOTPCRELXOpcode(op, modrm) != GOTPCRELXOpcode::Unknown;
+}
+
+bool x86_64LDBackend::shouldIgnoreRelocSync(Relocation *reloc) const {
+  return config().options().getRelax() && isGOTPCRELXRelaxable(reloc);
+}
+
+eld::Expected<void>
+x86_64LDBackend::postProcessing(llvm::FileOutputBuffer &pOutput) {
+  ELDEXP_RETURN_DIAGENTRY_IF_ERROR(GNULDBackend::postProcessing(pOutput));
```
quic-seaswara, line 159: This should be skipped for partial linking.

Is this relaxation applicable for all kinds of linking modes ? 

- Shared libraries
- PIE 
- Dynamic linking 
- Static linking

If so, please update.

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: I have refactored into a single function.

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -99,6 +102,127 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
...
+  if (isSymbolPreemptible(*rsym) || rsym->isIFunc())
+    return false;
+  auto *RF = llvm::dyn_cast<RegionFragment>(reloc->targetRef()->frag());
+  if (!RF)
+    return false;
+  const uint8_t *loc =
+      reinterpret_cast<const uint8_t *>(RF->getRegion().data()) +
+      reloc->targetRef()->offset();
+  uint8_t op = loc[-2];
+  uint8_t modrm = loc[-1];
```
quic-seaswara, line 149: Add comments about what is being done here.

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -99,6 +102,127 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
...
+}
+
+bool x86_64LDBackend::isGOTPCRELXRelaxable(const Relocation *reloc) const {
+  if (reloc->type() != llvm::ELF::R_X86_64_GOTPCRELX)
+    return false;
+  if (static_cast<int64_t>(reloc->addend()) != -4)
+    return false;
+  const ResolveInfo *rsym = reloc->symInfo();
+  if (isSymbolPreemptible(*rsym) || rsym->isIFunc())
+    return false;
```
quic-seaswara, line 134: IFunc symbols are preemptible so  no need to check that.

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -99,6 +102,127 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
...
+    return GOTPCRELXOpcode::CALL;
+  if (op == 0xff && modrm == 0x25)
+    return GOTPCRELXOpcode::JMP;
+  return GOTPCRELXOpcode::Unknown;
+}
+
+bool x86_64LDBackend::isGOTPCRELXRelaxable(const Relocation *reloc) const {
+  if (reloc->type() != llvm::ELF::R_X86_64_GOTPCRELX)
+    return false;
+  if (static_cast<int64_t>(reloc->addend()) != -4)
```
quic-seaswara, line 125: Why this check ? lld does not seem to have this check.

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -99,6 +102,127 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
...
+  if (op == 0xff && modrm == 0x15)
+    return GOTPCRELXOpcode::CALL;
+  if (op == 0xff && modrm == 0x25)
+    return GOTPCRELXOpcode::JMP;
+  return GOTPCRELXOpcode::Unknown;
+}
+
+bool x86_64LDBackend::isGOTPCRELXRelaxable(const Relocation *reloc) const {
+  if (reloc->type() != llvm::ELF::R_X86_64_GOTPCRELX)
+    return false;
```
quic-seaswara, line 121: This condition is excessively tested. Can we reduce the checks for this reloc type ?

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -99,6 +102,127 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
 
 bool x86_64LDBackend::initStubFactory() { return true; }
 
+x86_64LDBackend::GOTPCRELXOpcode
+x86_64LDBackend::classifyGOTPCRELXOpcode(uint8_t op, uint8_t modrm) {
```
quic-seaswara: const function

quic-seaswara: Can this function be changed to getGOTPCRELXOpcode, can this be a lambda function ?

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -99,6 +102,127 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
...
+        Relocator::Address S = reloc->symValue(m_Module);
+        Relocator::DWord A = reloc->addend();
+        Relocator::DWord P = reloc->place(m_Module);
+        int64_t val = static_cast<int64_t>(S) + static_cast<int64_t>(A) -
+                      static_cast<int64_t>(P);
+
+        if (!llvm::isInt<32>(val)) {
+          config().raise(Diag::error_x86_gotpcrelx_relax_overflow)
+              << rsym->name() << llvm::utohexstr(S)
+              << llvm::utohexstr(static_cast<uint64_t>(val));
```
quic-seaswara: Do we have a test case ? 

Why do we return an error, can we continue as this is not relaxable ?

lib/Target/X86/x86_64LDBackend.cpp
quic-seaswara: GOTPCRELXOpcode::Unknown

Why even have this case ?

quic-seaswara (changes requested):
The commit message seems wrong.

Introduces the x86-64 relaxation framework modeled on RISC-V. Isn't the model here different from RISC-V ?

include/eld/Config/GeneralOptions.h
```diff
@@ -1323,6 +1327,7 @@ class GeneralOptions {
   bool DisableGuardForWeakUndefs = false; // hexagon specific option to
                                           // disable guard functionality.
   bool BRiscvRelax = true;                // enable riscv relaxation
+  bool BRelax = false; // x86-64 GOTPCRELX relaxation (opt-in via --relax)
```
quic-areg: Other linkers have it on by default; do we plan to do that later? 

nit: can we remove the `B` prefix? `ShouldRelax` or `RelaxationEnabled` or something is probably better

include/eld/Diagnostics/DiagBackends.inc
```diff
@@ -90,6 +90,9 @@ DIAG(expected_group_section, DiagnosticEngine::InternalError,
      "Expected %0 to be a 'group' section.")
 DIAG(warn_ignore_pcrel_lo_addend, DiagnosticEngine::Warning,
      "%2: ignoring addend for relocation `%1' referring symbol %0")
+DIAG(error_x86_gotpcrelx_relax_overflow, DiagnosticEngine::Error,
+     "GOTPCRELX relaxation: symbol '%0' at address [0x%1] is unreachable "
```
quic-areg: I think this error message does not match the overflow diagnostics of other targets

I think we have a central overflow path used by other targets that could also be used by X86:
```cpp
// lib/Readers/Relocation.cpp:217
void Relocation::issueSignedOverflow(Relocator &R, int64_t Value, int64_t Min, int64_t Max) const {
```

lib/Target/X86/x86_64LDBackend.cpp
quic-areg: A test with two link stages, partial and final, would be nice if we don't already have one for x86. Maybe it could go in `Common` too since we have seen patches break partial linking in the past. I think we have many `-r` tests but none that relink the partially linked output.

lib/Target/X86/x86_64LDBackend.cpp
quic-areg: It does; we can't relax `GOTPCRELX` with an addend other than `-4`:
```cpp
RelExpr X86_64::adjustGotPcExpr(RelType type, int64_t addend,
                                const uint8_t *loc) const {
  // Only R_X86_64_[REX_]|[CODE_4_]GOTPCRELX can be relaxed. GNU as may emit
  // GOTPCRELX with addend != -4. Such an instruction does not load the full GOT
  // entry, so we cannot relax the relocation. E.g. movl x@GOTPCREL+4(%rip),
  // %rax (addend=0) loads the high 32 bits of the GOT entry.
  if (!ctx.arg.relax || addend != -4 ||
```
that's what `GOTPCRELXRelax_AddendNonM4.test` checks for

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -99,6 +102,127 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
...
+          m_Module.setFailure(true);
+          continue;
```
quic-areg: Why do we `continue` instead of returning a proper `Expected` Error? I think this results in emitting a corrupt output file on overflow.

lib/Target/X86/x86_64LDBackend.cpp
quic-areg: lld and GNU ld never fail on a `GOTPCRELX` overflow; they simply don't relax that reference and fall back to the GOT slot. I think we should do that too and add a test.

lib/Target/X86/x86_64LDBackend.cpp
quic-areg: ifunc symbols are preemptible but I think they shouldn't be relaxed, so the check is necessary. If they were relaxed we would skip the required irelative/plt resolution at runtime. We should add a test for this.

lib/Target/X86/x86_64LDBackend.cpp
quic-areg: I don't think we can relax non mov/call/jmp instructions.

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -99,6 +102,127 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
...
+    ELDEXP_RETURN_DIAGENTRY_IF_ERROR(relaxGOTPCRELX(pOutput));
+
+  return {};
+}
+
+eld::Expected<void>
+x86_64LDBackend::relaxGOTPCRELX(llvm::FileOutputBuffer &pOutput) {
+  uint8_t *buf = pOutput.getBufferStart();
+
+  for (auto &input : m_Module.getObjectList()) {
```
quic-seaswara: I would also prefer you to cache all the relocations that linker  visited that are candidates for relaxation and just iterate over that, instead of iterating through all the relocations again.

This is for considering performance.

quic-seaswara (changes requested):

lib/Target/X86/x86_64LDBackend.cpp
quic-seaswara: isSymbolPreemptible already returns true for IFunc :)

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: Currently only RISCV and x86 only support the GOTPCRELX relaxation so other backends wont support it, so will add the tests in x86.

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: I believe for the local or hidden IFUNC the isSymbolPreemptible would return false as it checks for the non Default visibility(if (pSym.other() != ResolveInfo::Default return false;) before going to the IFUNC and so we need to keep the IFUNC check.

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: Actually lld has a framework some kind of a loop which runs atmost around 30 times to check after the layout has happened and check the symbols address overflows or not and creates a got slot and does a relayout so it doesnt fail but bfd is failing and I have opened the issue too, with a bash test script https://github.com/qualcomm/eld/issues/1362 .  But in x86 we dont have any framework currently so we error out it.

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: Yes currently mold only relaxes call/jmp and lld also supports the mov , and I also found a ABI which mentions the mov, call and jmp.
[abi.pdf](https://github.com/user-attachments/files/30045044/abi.pdf)
There are test and binop which are supported for REX_GOTPCRELX case as we need the register information too.

test/x86_64/linux/GOTPCRELXRelaxation/GOTPCRELXRelax_Overflow.test
bsaharsh: Currently I have changed the Diagnostics, here is an example
Error: 1.o:(.text+0x2): relaxing relocation R_X86_64_GOTPCRELX out of range: 4294967290 is not in [-2147483648, 2147483647]; references 'sym'; relink with --no-relax
Fatal: Linking had errors (a.eld.out)
Does this work?

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: I have added the guard for the partial link.

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: I have changed the naming.

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: I have made the suggested changes and made a issue for the follow up.

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: Have used a vector to cache.

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: Added a vector cache which will handle this.

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: Added

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: I have updated it.

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: Added

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: Yes updated

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: using as a static helper as it gets called twice.

bsaharsh: I have changed the function but not using as a lambda function, as we call it twice, once at scan and once at patching loop.

include/eld/Config/GeneralOptions.h
bsaharsh: Parth has suggested to disable by default , and enabled once after having a stable release. renamed the boolean

include/eld/Diagnostics/DiagBackends.inc
bsaharsh: I have updated.

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: Thanks, I have updated.

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: Thanks, I have updated.

bsaharsh:
Hi @partaror @quic-areg @quic-seaswara . Thanks for the comments, I have updated and used the suggested changes , Please Have a look when you get time.

include/eld/Diagnostics/DiagVerbose.inc
```diff
@@ -34,6 +34,8 @@ DIAG(not_relaxed, DiagnosticEngine::Verbose,
 DIAG(relax_to_compress, DiagnosticEngine::Verbose,
      "%0 : relaxing instruction 0x%1 to compressed instruction 0x%2 for symbol "
      "%3 in section %4+0x%5 file %6")
+DIAG(relax_gotpcrelx, DiagnosticEngine::Trace,
```
parth-07: Please prefix the diagnostic with diagnostic severity?

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -98,6 +103,175 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
...
+    return false;
+  auto *RF = llvm::dyn_cast<RegionFragment>(reloc->targetRef()->frag());
+  if (!RF)
+    return false;
+  // The two bytes preceding the 4-byte displacement hold the opcode and,
+  // for CALL/JMP, the ModR/M byte. Only MOV/CALL/JMP are relaxable.
+  const uint8_t *loc =
+      reinterpret_cast<const uint8_t *>(RF->getRegion().data()) +
+      reloc->targetRef()->offset();
+  uint8_t op = loc[-2];
```
parth-07, line 148: Ideally, `loc[-2]` and `loc[-1]` should always be valid, however, I worry that in a corrupt/bad object file it may not be the case. Can you please add an early return if offset is less than 2 here?

Perhaps a bad assembly test can be added to make sure that we are not crashing here.

```asm
.text
.start:
.reloc ., R_X86_64_GOTPCRELX, foo
```

lib/Target/X86/x86_64Relocator.cpp
```diff
@@ -288,6 +288,25 @@ void x86_64Relocator::scanLocalReloc(InputFile &pInputFile, Relocation &pReloc,
     rsym->setReserved(rsym->reserved() | ReserveGOT);
     return;
   }
+  case llvm::ELF::R_X86_64_GOTPCRELX: {
```
parth-07, line 295: Can you please open an issue regarding GOTPCREL and REX_GOTPCREL relocations are unsupported for local symbols?

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -98,6 +103,175 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
...
+  uint8_t modrm = loc[-1];
+  return getGOTPCRELXOpcode(op, modrm) != GOTPCRELXOpcode::Unknown;
+}
+
+bool x86_64LDBackend::shouldIgnoreRelocSync(Relocation *reloc) const {
+  return config().options().getRelax() && isGOTPCRELXRelaxable(reloc);
```
parth-07: Instead of calling `isGOTPCRELXRelaxable` again, can we instead check whether the relocation is stored in `m_GOTPCRELXCandidates`?

lib/Target/X86/x86_64Relocator.cpp
```diff
@@ -685,6 +732,12 @@ Relocator::Result eld::relocGOTRelative(Relocation &pReloc,
   ResolveInfo *symInfo = pReloc.symInfo();
   const GeneralOptions &options = pParent.config().options();
 
+  // For GOTPCRELX relocations, postProcessing handles the opcode patch and
+  // displacement. Skip apply here to avoid a null GOT entry lookup.
+  if (pReloc.type() == llvm::ELF::R_X86_64_GOTPCRELX && options.getRelax() &&
+      pParent.getTarget().isGOTPCRELXRelaxable(&pReloc))
```
parth-07: Can we use m_GOTPCRELXCandidates here as well? We will likely need to change it from vector to something like an unordered_set.

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -98,6 +103,175 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
...
+    buf[out_off + 3] = 0x90;
+    kind = "jmp->direct";
+    break;
+  case GOTPCRELXOpcode::Unknown:
+    // isGOTPCRELXRelaxable only records MOV/CALL/JMP candidates, so this is
+    // not expected; leave the instruction untouched if it is ever reached.
+    break;
+  }
+
+  if (config().options().trace() & DiagnosticPrinter::TraceRelax) {
```
parth-07: Can we use `traceRelax()` here?

include/eld/Diagnostics/DiagVerbose.inc
bsaharsh: Have added the trace_ prefix

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: Added the guard and test case

lib/Target/X86/x86_64Relocator.cpp
bsaharsh: Have opened the issue

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: added, now using the function which checks from the unordered set

lib/Target/X86/x86_64Relocator.cpp
bsaharsh: Added the unordered set instead of vector

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: yes have changed to use traceRelax

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -98,6 +103,184 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
...
+    // isGOTPCRELXRelaxable only records MOV/CALL/JMP candidates, so this is
+    // not expected; leave the instruction untouched if it is ever reached.
+    break;
+  }
+
+  if (m_Module.getPrinter()->traceRelax()) {
+    std::string location;
+    if (ELFSection *sect = RF->getOwningSection())
+      location = sect->getLocation(offset, config().options());
+    config().raise(Diag::relax_gotpcrelx) << location << kind << rsym->name();
```
parth-07: Now this should be `trace_relax_gotpcrelx`.

lib/Target/X86/x86_64Relocator.cpp
```diff
@@ -288,6 +288,25 @@ void x86_64Relocator::scanLocalReloc(InputFile &pInputFile, Relocation &pReloc,
...
+    std::lock_guard<std::mutex> relocGuard(m_RelocMutex);
+    // Relaxability is decided per relocation (it inspects this reference's own
+    // opcode bytes), so it must be checked before the per-symbol ReserveGOT
+    // short-circuit: a non-relaxable reference to the same symbol may have
+    // reserved a GOT slot first. With --relax, local symbols are always
+    // non-preemptible; skip the GOT slot for the relaxable reference and record
+    // it so postProcessing can patch it without re-walking.
+    if (config().options().getRelax() &&
+        m_Target.isGOTPCRELXRelaxable(&pReloc)) {
+      m_Target.recordGOTPCRELXCandidate(&pReloc);
```
parth-07: Can we please rename the function to `recordGOTPCRELXRelaxCandidate` and the corresponding function to `isGOTPCRELXRelaxCandidate`? `Relax` here is important because we may have GOTPCRELX relocations which cannot be relaxed and the current name does not clearly indicate that.

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -98,6 +103,184 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
...
+      // Only relaxable types are recorded as candidates; skip anything else.
+      break;
+    }
+  }
+  return {};
+}
+
+eld::Expected<void> x86_64LDBackend::relaxGOTPCRELXReloc(Relocation *reloc,
+                                                         uint8_t *buf) {
+  auto *RF = llvm::dyn_cast<RegionFragment>(reloc->targetRef()->frag());
```
parth-07, line 194: We only call `relaxGOTPCRELXReloc` for GOTPCRELXRelax candidates, and only relocations with targetRef() of type `RegionFragment` can be candidates, right? Do we need this if-condition? Do you think it would be better as an ASSERT?

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -98,6 +103,184 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
...
+    break;
+  case GOTPCRELXOpcode::Unknown:
+    // isGOTPCRELXRelaxable only records MOV/CALL/JMP candidates, so this is
+    // not expected; leave the instruction untouched if it is ever reached.
+    break;
+  }
+
+  if (m_Module.getPrinter()->traceRelax()) {
+    std::string location;
+    if (ELFSection *sect = RF->getOwningSection())
```
parth-07: Is there a case possible where we do not have an owning section for a region fragment where we are doing relaxation? If no, can you please add an assert that owning section must be present instead of the if-condition.

lib/Target/X86/x86_64LDBackend.h
```diff
@@ -148,6 +176,18 @@ class x86_64LDBackend : public GNULDBackend {
 
   uint64_t maxBranchOffset() override { return 0; }
 
+  enum class GOTPCRELXOpcode { MOV, CALL, JMP, Unknown };
+  static GOTPCRELXOpcode getGOTPCRELXOpcode(uint8_t op, uint8_t modrm);
```
parth-07: Please make it a private member function instead of a static function.

lib/Target/X86/x86_64Relocator.cpp
```diff
@@ -399,8 +420,34 @@ void x86_64Relocator::scanGlobalReloc(InputFile &pInputFile, Relocation &pReloc,
...
+        m_Target.isGOTPCRELXRelaxable(&pReloc)) {
+      m_Target.recordGOTPCRELXRelaxCandidate(&pReloc);
+      return;
+    }
+    if (rsym->reserved() & ReserveGOT)
+      return;
+    CreateGOT(Obj, pReloc, !config().isCodeStatic(), m_Target);
+    rsym->setReserved(rsym->reserved() | ReserveGOT);
+    return;
+  }
```
parth-07:
```suggestion
    case llvm::ELF::R_X86_64_GOTPCRELX:
    // Relaxability is decided per relocation (it inspects this reference's own
    // opcode bytes), so it must be checked before the per-symbol ReserveGOT
    // short-circuit: a non-relaxable reference to the same symbol may have
    // reserved a GOT slot first. With --relax, non-preemptible non-IFUNC
    // symbols with a relaxable opcode are handled by postProcessing: no GOT
    // slot is needed and an out-of-range displacement is a link error. An
    // addend != -4 or an unknown opcode keeps the GOT slot.
    if (config().options().getRelax() &&
        m_Target.isGOTPCRELXRelaxable(&pReloc)) {
      m_Target.recordGOTPCRELXRelaxCandidate(&pReloc);
      return;
    }
  case llvm::ELF::R_X86_64_GOTPCREL: {
    std::lock_guard<std::mutex> relocGuard(m_RelocMutex);
    if (rsym->reserved() & ReserveGOT)
      return;
    CreateGOT(Obj, pReloc, !config().isCodeStatic(), m_Target);
    rsym->setReserved(rsym->reserved() | ReserveGOT);
    return;
  }
```

lib/Target/X86/x86_64Relocator.cpp
```diff
@@ -685,6 +732,11 @@ Relocator::Result eld::relocGOTRelative(Relocation &pReloc,
   ResolveInfo *symInfo = pReloc.symInfo();
   const GeneralOptions &options = pParent.config().options();
 
+  // For GOTPCRELX relocations, postProcessing handles the opcode patch and
```
parth-07:
```suggestion
  // For relaxable GOTPCRELX relocations, postProcessing handles the opcode patch and
```

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -98,6 +103,184 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
...
+  // for CALL/JMP, the ModR/M byte. Only MOV/CALL/JMP are relaxable.
+  const uint8_t *loc =
+      reinterpret_cast<const uint8_t *>(RF->getRegion().data()) + offset;
+  uint8_t op = loc[-2];
+  uint8_t modrm = loc[-1];
+  return getGOTPCRELXOpcode(op, modrm) != GOTPCRELXOpcode::Unknown;
+}
+
+bool x86_64LDBackend::shouldIgnoreRelocSync(Relocation *reloc) const {
+  return config().options().getRelax() && isGOTPCRELXRelaxCandidate(reloc);
```
parth-07: Do we need a check for `config().options().getRelax()` here? Don't we only insert members into `GOTPCRELXRelaxCandidates` if relaxation is enabled?

parth-07 (changes requested):

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: Updated to have ASSERT instead of if-condition

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: Yes, it would be better to have ASSERTS, updated.

lib/Target/X86/x86_64LDBackend.h
bsaharsh: updated

lib/Target/X86/x86_64Relocator.cpp
bsaharsh: reordered them.

lib/Target/X86/x86_64Relocator.cpp
bsaharsh: added the relaxable word

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: yes the check is indeed redundant, have updated, Thanks

quic-areg (approved):
LGTM

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -98,6 +103,190 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
...
+}
+
+eld::Expected<void> x86_64LDBackend::doRelax(llvm::FileOutputBuffer &pOutput) {
+  uint8_t *buf = pOutput.getBufferStart();
+  // The scan phase already identified every relaxation candidate (skipping
+  // debug relocations and internal files, which never carry a relaxable
+  // relocation). Iterate that cached set instead of re-walking all
+  // relocations, and dispatch on the relocation type. Future relaxable types
+  // (e.g. R_X86_64_REX_GOTPCRELX, TLS relaxations) add a case here.
+  for (Relocation *reloc : m_GOTPCRELXRelaxCandidates) {
```
parth-07, line 176: A follow-up: we can process the relaxations parallelly. Can you please create an issue for this?

parth-07 (approved):

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -98,6 +103,190 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
...
+  // (isGOTPCRELXRelaxable requires a RegionFragment), so this cast must
+  // succeed.
+  auto *RF = llvm::dyn_cast<RegionFragment>(reloc->targetRef()->frag());
+  assert(RF && "GOTPCRELX relax candidate must reference a RegionFragment");
+
+  uint32_t offset = reloc->targetRef()->offset();
+  // Guard against a corrupt object where the relocation sits at offset < 2,
+  // which would make the loc[-2]/loc[-1] opcode reads out-of-bounds.
+  if (offset < 2)
+    return {};
```
quic-seaswara: checking for this again ?

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -98,6 +103,190 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
...
+    assert(overflowSect &&
+           "RegionFragment in relax candidate must have an owning section");
+    location = overflowSect->getLocation(offset, config().options());
+    return std::make_unique<plugin::DiagnosticEntry>(plugin::DiagnosticEntry(
+        Diag::error_x86_gotpcrelx_relax_overflow,
+        {std::move(location), "R_X86_64_GOTPCRELX", llvm::itostr(val),
+         llvm::itostr(std::numeric_limits<int32_t>::min()),
+         llvm::itostr(std::numeric_limits<int32_t>::max()),
+         std::string(rsym->name())}));
+  }
```
quic-seaswara, line 229: can we use Relocator::checkSignedRange here ?

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -98,6 +103,190 @@ bool x86_64LDBackend::initBRIslandFactory() { return true; }
...
+    // jmp *foo@GOTPCREL(%rip)  ->  jmp foo; nop
+    // "ff 25" indirect jmp (6 bytes) becomes "e9 rel32" direct jmp (5 bytes)
+    // followed by a 0x90 nop for padding. The displacement shifts one byte
+    // earlier (no ModR/M), so it is written at out_off-1 with val+1 to account
+    // for the one-byte-shorter instruction prefix.
+    buf[out_off - 2] = 0xe9;
+    llvm::support::endian::write32le(buf + out_off - 1,
+                                     static_cast<uint32_t>(val + 1));
+    buf[out_off + 3] = 0x90;
+    kind = "jmp->direct";
```
quic-seaswara, line 267: Missing map file information for relaxation.

lib/Target/X86/x86_64Relocator.cpp
```diff
@@ -288,6 +288,25 @@ void x86_64Relocator::scanLocalReloc(InputFile &pInputFile, Relocation &pReloc,
...
+    // non-preemptible; skip the GOT slot for the relaxable reference and record
+    // it so postProcessing can patch it without re-walking.
+    if (config().options().getRelax() &&
+        m_Target.isGOTPCRELXRelaxable(&pReloc)) {
+      m_Target.recordGOTPCRELXRelaxCandidate(&pReloc);
+      return;
+    }
+    if (rsym->reserved() & ReserveGOT)
+      return;
+    CreateGOT(Obj, pReloc, !config().isCodeStatic(), m_Target);
```
quic-seaswara, line 310: Can you test for -pie mode ?

test/x86_64/linux/GOTPCRELXRelaxation/GOTPCRELXRelax_BadInput.test
```diff
@@ -0,0 +1,12 @@
+#--GOTPCRELXRelax_BadInput.test----------Executable--------#
+# A GOTPCRELX relocation placed at offset 0 (via .reloc) means the opcode
+# bytes at loc[-2]/loc[-1] would be before the fragment buffer. The linker
+# must not crash; it must skip relaxation for this reference.
+#START_TEST
+RUN: %clang %clangopts -c %p/Inputs/badinput.s -o %t.o
+RUN: %link %linkopts --relax -shared -o %t.so %t.o
```
quic-seaswara, line 7: objdump and check

quic-seaswara (changes requested):

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: Have created the issue.

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: Have made it into an assert, instead of if condition. What would you suggest removing the assert or keep it?

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: checkSignedRange is not compatible because of the diagnostic: it uses the central result_overflow_moreinfo which lacks the --no-relax hint.

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: Have created an issue, to have it as a follow up. https://github.com/qualcomm/eld/issues/1477

lib/Target/X86/x86_64Relocator.cpp
bsaharsh: Have improved the existing Relax_PIE.test, please have a look.

test/x86_64/linux/GOTPCRELXRelaxation/GOTPCRELXRelax_BadInput.test
bsaharsh: have updated the test to check objdump before and after.

quic-seaswara (approved):

quic-seaswara:
@bsaharsh  please rebase the commit and resolve conflicts. We should merge this.

parth-07:
@bsaharsh Can you please make the clang-format workflow happy?

PR #1400, opened 2026-07-06
Reland "Make version script pattern matching GNU-compatible""

### LINKED ISSUE #1132: Version script symbol descriptions matching needs to be improved

parth-07:
This commit makes version script pattern matching GNU-compatible. Uptil now, eld version script pattern matching was very intuitive, and straightforward. Iterate the version nodes and blocks in order and match them with the global symbols, and the first match wins. However, GNU version script pattern matching rules are a little complex. The updated version script pattern matching follows the below rules:

Phase 1: Iterate the version nodes and blocks in order, match symbols with the non-wildcard patterns. The first match wins. Phase 2: Iterate the version nodes and blocks in order, match unmatched symbols with the non-* wildcard patterns. The last match wins. 
Phase 3: Iterate the version nodes and blocks in order, match unmatched symbols with the '*' wildcard pattern. The last match wins.

Additionally, now symbols from shared libraries do not take part in the version script pattern matching because it does not make sense to assign a version to a symbol from a shared library.

Resolves #1132

### DIFF
--- a/lib/Object/ObjectLinker.cpp
+++ b/lib/Object/ObjectLinker.cpp
@@ -407,65 +407,124 @@ bool ObjectLinker::parseVersionScript() {
       ThisModule->addVersionScriptNode(VersionScriptNode);
     }
   }
-  auto &SymbolScopes = getTargetBackend().symbolScopes();
+  assignVersionNodesToSymbols();
+  return true;
+}
+
+void ObjectLinker::assignVersionNodesToSymbols() {
   auto &NP = ThisModule->getNamePool();
+  auto &VersionNodes = ThisModule->getVersionScriptNodes();
+
+  if (VersionNodes.empty())
+    return;
+
 #ifdef ELD_ENABLE_SYMBOL_VERSIONING
-  DemangledNamesMap DemangledNames;
+  DemangledNamesMap demangledNames;
 #endif
+
+  auto canAssignVersionNode = [](const ResolveInfo &R) {
+    return (R.isDefine() || R.isCommon()) && !R.isDyn();
+  };
+
+  auto getVersionDesc = [](const VersionSymbol *VS) -> std::string {
+    auto *node = VS->getBlock()->getNode();
+    if (node->isAnonymous())
+      return VS->isGlobal() ? "VER_NDX_GLOBAL" : "VER_NDX_LOCAL";
+    return (node->getName() + (VS->isGlobal() ? "(global)" : "(local)")).str();
+  };
+
+  // Try assigning version node VS to the symbol R. It only assigns a
+  // version node to the symbol if the symbol does not already have an
+  // assigned version node. It emits version node reassign warning if
+  // warnOnReassing is true.
+  auto tryAssign = [&](ResolveInfo *R, VersionSymbol *VS, bool warnOnReassign) {
+    VersionSymbol *existing = getTargetBackend().getSymbolScope(R);
+    InputFile *verSymInputFile =
+        VS->getBlock()->getNode()->getVersionScript().getInputFile();
+    if (existing != nullptr) {
+      if (warnOnReassign && ThisConfig.showVersionScriptWarnings()) {
+        ThisConfig.raise(Diag::warn_version_script_reassign)
+            << verSymInputFile->getInput()->decoratedPath() << R->name()
+            << getVersionDesc(existing) << getVersionDesc(VS);
+      }
+      return false;
+    }
+
+    getTargetBackend().addSymbolScope(R, VS);
+
+#ifdef ELD_ENABLE_SYMBOL_VERSIONING
+    if (ThisConfig.getPrinter()->traceSymbolVersioning()) {
+      ThisConfig.raise(Diag::trace_version_script_matched_scope)
+          << R->name() << getVersionDesc(VS);
+    }
+#endif
+    return true;
+  };
+
+  using PatternFilter = std::function<bool(const WildcardPattern &)>;
+
+  std::vector<ResolveInfo *> VSApplicableSymbols;
   for (auto &G : NP.getGlobals()) {
     ResolveInfo *R = G.getValue();
-    for (auto &VersionScriptNode : ThisModule->getVersionScriptNodes()) {
-      if (VersionScriptNode->getGlobalBlock()) {
-        for (auto *Sym : VersionScriptNode->getGlobalBlock()->getSymbols()) {
-#ifdef ELD_ENABLE_SYMBOL_VERSIONING
-          bool isMatched = Sym->matched(*R, NP, DemangledNames);
-#else
-          bool isMatched = Sym->getSymbolPattern()->matched(*R);
-#endif
-          if (isMatched) {
-            getTargetBackend().addSymbolScope(R, Sym);
-#ifdef ELD_ENABLE_SYMBOL_VERSIONING
-            if (ThisConfig.getPrinter()->traceSymbolVersioning())
-              ThisConfig.raise(Diag::trace_version_script_matched_scope)
-                  << "global" << R->name();
-            break;
-#endif
-          } // end Symbol Match
-        } // end Symbols
-      } // end Global
-      if (SymbolScopes.find(R) != SymbolScopes.end()) {
-#ifdef ELD_ENABLE_SYMBOL_VERSIONING
-        break;
-#else
+    if (canAssignVersionNode(*R))
+      VSApplicableSymbols.push_back(R);
+  }
+
+  auto processBlock = [&](VersionScriptBlock *block, PatternFilter filter,
+                          bool warnOnReassign) {
+    if (!block)
+      return;
+
+    for (auto *sym : block->getSymbols()) {
+      auto *pattern = sym->getSymbolPattern();
+      if (!filter(*pattern))
         continue;
-#endif
-      }
-      if (VersionScriptNode->getLocalBlock()) {
-        for (auto *Sym : VersionScriptNode->getLocalBlock()->getSymbols()) {
+
+      for (auto *R : VSApplicableSymbols) {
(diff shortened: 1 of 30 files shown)


lib/Object/ObjectLinker.cpp
```diff
@@ -407,65 +407,121 @@ bool ObjectLinker::parseVersionScript() {
...
+  auto getVersionName = [](const VersionSymbol *VS) -> llvm::StringRef {
+    auto *block = VS->getBlock();
+    auto *node = block->getNode();
+    if (node->isAnonymous()) {
+      if (block->isGlobal())
+        return "VER_NDX_GLOBAL";
+      else
+        return "VER_NDX_LOCAL";
+    }
+    return (block->isLocal() ? "VER_NDX_LOCAL" : node->getName());
```
quic-seaswara: Why is the anonymous node handled differently ?

lib/Object/ObjectLinker.cpp
```diff
@@ -407,65 +407,121 @@ bool ObjectLinker::parseVersionScript() {
...
+  };
+
+  auto tryAssign = [&](ResolveInfo *R, VersionSymbol *VS, bool warnOnReassign) {
+    VersionSymbol *existing = getTargetBackend().getSymbolScope(R);
+    InputFile *verSymInputFile =
+        VS->getBlock()->getNode()->getVersionScript().getInputFile();
+    if (existing != nullptr) {
```
quic-seaswara, line 444: comments would be helpful on what is going on.

lib/Object/ObjectLinker.cpp
```diff
@@ -407,65 +407,121 @@ bool ObjectLinker::parseVersionScript() {
...
+
+  auto processBlock = [&](VersionScriptBlock *block, PatternFilter filter,
+                          bool warnOnReassign) {
+    if (!block)
+      return;
+
+    for (auto *sym : block->getSymbols()) {
+      auto *pattern = sym->getSymbolPattern();
+      if (!filter(*pattern))
         continue;
```
quic-seaswara, line 481: Why are we skipping here ? is it a bad wildcard pattern ?

lib/Object/ObjectLinker.cpp
```diff
@@ -407,65 +407,121 @@ bool ObjectLinker::parseVersionScript() {
...
-#endif
-      }
-      if (VersionScriptNode->getLocalBlock()) {
-        for (auto *Sym : VersionScriptNode->getLocalBlock()->getSymbols()) {
+
+      for (auto &G : NP.getGlobals()) {
```
quic-seaswara: Instead of iterating through all the global for each symbol in the block , can we filter it before ? 

What happens with linker script defined symbols / --defsym / PROVIDE Symbols ? Can we have a test for this ?

lib/Object/ObjectLinker.cpp
```diff
@@ -407,65 +407,121 @@ bool ObjectLinker::parseVersionScript() {
...
+    processNodeFirstWins(N, isExact, true);
+  }
+
+  for (auto It = VersionNodes.rbegin(); It != VersionNodes.rend(); ++It) {
+    processNodeLastWins(*It, isNonStarWildcard, false);
+  }
+
+  for (auto It = VersionNodes.rbegin(); It != VersionNodes.rend(); ++It) {
+    processNodeLastWins(*It, isMatchAll, false);
+  }
```
quic-seaswara, line 527: This definitely needs to be multithreaded.

lib/Object/ObjectLinker.cpp
quic-seaswara: How are symbols from LTO handled ? 

Before and after LTO ?

lib/Object/ObjectLinker.cpp
```diff
@@ -407,65 +407,121 @@ bool ObjectLinker::parseVersionScript() {
...
 }
```
quic-seaswara, line 528: Is this function called pre garbage collection or post garbage collection ?

test/Common/standalone/SymbolVersioning/BuildingSharedLibsWithSymbolVersioning/BuildingSharedLibsWithSymbolVersioningTrace.test
```diff
@@ -9,9 +9,9 @@ RUN: %clang %clangopts -o %t1.3.o %p/Inputs/3.c -c -fPIC
 RUN: %link %linkopts -o %t1.lib1.so %t1.3.o -shared --version-script %p/Inputs/vs.2.t \
 RUN:   --trace=symbol-versioning 2>&1 | %filecheck %s --check-prefix=TRACE1
 #END_TEST
-TRACE1: Trace: Version script node matched global symbol 'baz'
-TRACE1: Trace: Version script node matched global symbol 'foo@V1'
-TRACE1: Trace: Version script node matched global symbol 'bar@@V1'
+TRACE1-DAG: Trace: Version script node matched global symbol 'baz'
+TRACE1-DAG: Trace: Version script node matched global symbol 'foo@V1'
+TRACE1-DAG: Trace: Version script node matched global symbol 'bar@@V1'
```
quic-seaswara: Need to spell out which version script node is being matched.

test/Common/standalone/VersionScriptPatternMatching/VersionScriptPatternMatching.test
```diff
@@ -0,0 +1,71 @@
+#---VersionScriptPatternMatching.test--------------------- Exe------------------#
+
```
quic-seaswara: Remove empty line ?

quic-seaswara (changes requested):
Please document version script support as part of this change.

lib/Object/ObjectLinker.cpp
parth-07: Anonymous node is handled differently here because it does not have a name like non-anonymous nodes. The return value here is used in `warn_version_script_reassign` diagnostic.

lib/Object/ObjectLinker.cpp
parth-07: filter here is used for performing the version script pattern matching in the following phases:

1) In the first phase, only non-wildcard patterns are processed.
2) In the second phase, only non-star wildcard patterns are processed.
3) In the third phase, only star-only wildcard patterns are processed.

lib/Object/ObjectLinker.cpp
parth-07: Agreed, I will make it multithreaded as a follow-up.

lib/Object/ObjectLinker.cpp
parth-07: Version script pattern matching happens pre garbage-collection.

lib/Object/ObjectLinker.cpp
parth-07: > Instead of iterating through all the global for each symbol in the block , can we filter it before ?

Thanks for the suggestion. We are now filtering the symbol before.

parth-07: > What happens with linker script defined symbols / --defsym / PROVIDE Symbols ? Can we have a test for this ?

Version script pattern matching happens before we add linker script symbols to the symbol resolver. Hence, these symbols do not have a version currently. I have created a follow-up issue for this: #1442

lib/Object/ObjectLinker.cpp
parth-07: I have created a follow-up issue for this: #1443

lib/Object/ObjectLinker.cpp
parth-07: It seems that version script pattern matching is run before the LTO-generated objects are created. However, I believe that this task is beyond the scope of this PR. I have created a follow-up issue for it: #1441

test/Common/standalone/VersionScriptPatternMatching/VersionScriptPatternMatching.test
parth-07: Done.

lib/Object/ObjectLinker.cpp
parth-07: I have added a comment for this function.

parth-07:
> Please document version script support as part of this change.

I have added version script documentation as part of this patch.

parth-07:
@quic-seaswara can you please review the PR again?

test/Common/standalone/SymbolVersioning/BuildingSharedLibsWithSymbolVersioning/BuildingSharedLibsWithSymbolVersioningTrace.test
parth-07: I have updated the diagnostic to include the version script node.

quic-seaswara (approved):

Steven6798 (approved):

PR #1404, opened 2026-07-06
[RISC-V] Don't retarget pcrel_lo/tlsdesc_lo symbol in partial links

quic-areg:
2c6503d2 rewrote the pcrel_lo/tlsdesc_lo symbol to its paired pcrel_hi/tlsdesc_hi symbol unconditionally. Under -r that rewrite is emitted to the output, where the pcrel_lo must instead reference the local label on its hi20 instruction so the pair can be re-resolved. Skip the retarget for partial links.

### DIFF
--- a/lib/Target/RISCV/RISCVLDBackend.cpp
+++ b/lib/Target/RISCV/RISCVLDBackend.cpp
@@ -1953,6 +1953,11 @@ bool RISCVLDBackend::handlePendingRelocations(ELFSection *section) {
                      return A->getOffset() < B->getOffset();
                    });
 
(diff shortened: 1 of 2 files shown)


quic-seaswara (approved):

quic-akaryaki:
Thanks! Can we skip the everything what `handlePendingRelocations()` does for partial linking? At least, starting from relocation sorting, not sure about LEB relocations.

quic-seaswara (approved):

quic-akaryaki (approved):

PR #1405, opened 2026-07-07
[CI] Add nightly ELD code coverage workflow with regression detection

bsaharsh:
Added a workflow which runs nightly and computes the code coverage, and when ran at first creates a baseline text file on orphan branch called code-coverage, and from the next time, it checks with the previous coverage numbers and if any metric goes down, it fails, and if no regression is detected then we push the html report and update the baseline text file.

### DIFF
--- a/.github/workflows/nightly-eld-code-coverage.yml
+++ b/.github/workflows/nightly-eld-code-coverage.yml
@@ -1,0 +1,220 @@
+name: Nightly ELD code coverage
+
+on:
+  schedule:
+    - cron: '0 9 * * *'
+  workflow_dispatch: {}
+  pull_request: {} # Uncomment only to test this WF file update.
+
+jobs:
+  coverage:
+    if: github.repository == 'qualcomm/eld'
+    runs-on: ubuntu-latest
+    concurrency:
+      group: ${{ github.workflow }}
+      cancel-in-progress: true
+    permissions:
+      contents: write
+
+    env:
+      CCACHE_DIR: ${{ github.workspace }}/.ccache
+      CCACHE_BASEDIR: ${{ github.workspace }}
+      CCACHE_COMPILERCHECK: content
+      CCACHE_NOHASHDIR: true
+
+    steps:
+      - name: Checkout llvm-project
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          repository: llvm/llvm-project
+          path: llvm-project
+
+      - name: Checkout eld
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          repository: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
+          ref: ${{ github.event.pull_request.head.ref || github.ref_name }}
+          path: llvm-project/llvm/tools/eld
+
+      - name: Install cross-compilation toolchains
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y \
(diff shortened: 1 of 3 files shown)


.github/workflows/nightly-eld-code-coverage.yml
```diff
@@ -0,0 +1,207 @@
...
+
+      - name: Deploy coverage report to code-coverage branch
+        run: |
+          REPORT_SRC="${{ github.workspace }}/obj/coverage-report"
+
+          cd llvm-project/llvm/tools/eld
+          git config user.name  'github-actions'
+          git config user.email 'github-actions@github.com'
+
+          git checkout --orphan coverage-update
```
quic-seaswara: can we have this branch as code-coverage-reports ?

quic-seaswara (changes requested):

.github/workflows/nightly-eld-code-coverage.yml
```diff
@@ -0,0 +1,207 @@
...
+    - cron: '0 9 * * *'
+  workflow_dispatch: {}
+
+permissions:
+  contents: write
+
+jobs:
+  coverage:
+    if: github.repository == 'qualcomm/eld'
+    runs-on: self-hosted
```
parth-07: @quic-seaswara is it okay to run this workflow on self-hosted? self-hosted here will make adding this workflow easier because we would not need to install/setup the cross-toolchains and qemu.

.github/workflows/nightly-eld-code-coverage.yml
```diff
@@ -0,0 +1,207 @@
...
+          repository: llvm/llvm-project
+          path: llvm-project
+
+      - name: Checkout eld
+        uses: actions/checkout@v7
+        with:
+          repository: ${{ github.repository }}
+          ref: ${{ github.ref_name }}
+          path: llvm-project/llvm/tools/eld
+          fetch-depth: 0
```
parth-07: We can likely remove `fetch-depth: 0` from here.

.github/workflows/nightly-eld-code-coverage.yml
```diff
@@ -0,0 +1,207 @@
...
+          ccache --set-config "max_size=0"
+
+          rm -rf obj
+          mkdir -p obj
+          cd obj
+          cmake -G Ninja \
+            ../llvm-project/llvm \
+            -DCMAKE_BUILD_TYPE=Debug \
+            -DCMAKE_DISABLE_PRECOMPILE_HEADERS=ON \
+            -DLLVM_ENABLE_PROJECTS="llvm;clang;clang-tools-extra" \
```
parth-07: Would it be possible to use the clang and the other tools from the nightly eld toolchain and only build eld with code-coverage enabled?

.github/workflows/nightly-eld-code-coverage.yml
bsaharsh: Updated the name

.github/workflows/nightly-eld-code-coverage.yml
bsaharsh: Updated

.github/workflows/nightly-eld-code-coverage.yml
bsaharsh: Updated, but need to test it as fork won't be having the self hosted, may be we can use the workflow dispatch.

Steven6798 (commented):
Make sure your code doesn't have any Semgrep Findings if it is preventable. Most of these issues will be pointed out by the github-advanced-security bot. Look at other workflows to see how tags for GitHub actions are written.

parth-07 (approved):

PR #1406, opened 2026-07-07
Fix stats-file-option.ll output check

svs-quic:
https://github.com/llvm/llvm-project/pull/207870 renamed prologpeilog to prolog-epilog leading to the test failure. Modify the check line in the test to match this.


parth-07 (approved):

pgodeq (approved):
Approved.

PR #1416, opened 2026-07-07
[AArch64] Add overflow check for ADR_PREL_LO21/LD_PREL_LO19

quic-seaswara:
The ADR instruction encodes (S + A) - P in a 21-bit signed immediate field. Add a signed 21-bit range check for R_AARCH64_ADR_PREL_LO21.

The LDR (literal) instruction encodes (S + A) - P in a 19-bit signed field scaled by 4. Add a signed 21-bit range check on the raw byte offset for R_AARCH64_LD_PREL_LO19.

Add a lit test that verifies a link within range succeeds and a link that overflows the 21-bit field is caught with an overflow diagnostic.

### DIFF
--- a/lib/Target/AArch64/AArch64Relocator.cpp
+++ b/lib/Target/AArch64/AArch64Relocator.cpp
@@ -932,6 +932,14 @@ Relocator::Result call(Relocation &pReloc, AArch64Relocator &pParent) {
 
   Relocator::DWord X = S + A - P;
 
+  if (X & 0x3) {
+    DiagEngine->raise(Diag::error_reloc_not_aligned)
+        << pParent.getName(pReloc.type()) << pReloc.symInfo()->name()
+        << pReloc.getSourcePath(pParent.config().options()) << "4"
+        << ("PC-relative offset 0x" + llvm::utohexstr(X));
+    return Relocator::BadReloc;
+  }
+
   pReloc.target() = helper_reencode_branch_offset_26(pReloc.target(), X >> 2);
 
   return Relocator::OK;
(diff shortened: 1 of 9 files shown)


lib/Target/AArch64/AArch64Relocator.cpp
```diff
@@ -884,7 +887,10 @@ Relocator::Result ld_prel_lo19(Relocation &pReloc, AArch64Relocator &pParent) {
   Relocator::DWord P = pReloc.place(pParent.module());
 
   Relocator::DWord X = S + A - P;
-  // TODO:: check overflow
+
+  Relocator::Result R = checkSignedRange(pReloc, pParent, X, 21);
+  if (R != Relocator::OK)
```
efriedma-quic, line 900: Do we also need to check that X is divisible by 4?

quic-seaswara: Sure. I can add that.

lib/Target/AArch64/AArch64Relocator.cpp
```diff
@@ -917,7 +931,6 @@ Relocator::Result call(Relocation &pReloc, AArch64Relocator &pParent) {
             ->getAddr(DiagEngine);
 
   Relocator::DWord X = S + A - P;
-  // TODO: check overflow..
 
   pReloc.target() = helper_reencode_branch_offset_26(pReloc.target(), X >> 2);
```
efriedma-quic, line 943: I think we also need an alignment check here.

efriedma-quic (approved):
LGTM

lib/Target/AArch64/AArch64Relocator.cpp
quic-seaswara: Sure, thanks for catching that.

PR #1424, opened 2026-07-08
Consolidate constructor/destructor sections in H2 picolibc template

quic-k:
Merge .preinit_array, .ctors, and .init_array into a single .init_array output section, and .dtors and .fini_array into a single .fini_array output section in the static-executable-h2-picolibc linker script template.

This prevents the linker from placing orphan sections between the entries within the [__bothinit_array_start, __bothinit_array_end) window. An orphan landing inside that window would be walked by __libc_init_array() and called as a bogus constructor function pointer, causing a fault at startup.

### DIFF
--- a/templates/staticExecutable/static-executable-h2-picolibc.lcs.template
+++ b/templates/staticExecutable/static-executable-h2-picolibc.lcs.template
@@ -32,24 +32,28 @@ SECTIONS
   /* Need to pre-align so that the symbols come after padding */
   . = ALIGN(8);
   /* __bothinit_array_start/end pointers are used by Picolibc to
-   * call constructor functions, don't add unwanted sections in between
+   * call constructor functions, don't add unwanted sections in between.
+   *
+   * All constructor input sections (.preinit_array, .ctors, .init_array) are
+   * collected into a SINGLE output section so the linker cannot place an
+   * orphan section between them.  An orphan landing inside the
+   * [__bothinit_array_start, __bothinit_array_end) window would be walked by
+   * __libc_init_array() and called as a bogus constructor function pointer,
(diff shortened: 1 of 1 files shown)


parth-07 (approved):

templates/staticExecutable/static-executable-h2-picolibc.lcs.template
```diff
@@ -32,40 +32,45 @@ SECTIONS
...
     KEEP (*( SORT_BY_INIT_PRIORITY(.init_array.*)))
     KEEP (*(.init_array))
     PROVIDE_HIDDEN ( __init_array_end = . );
   }
   PROVIDE_HIDDEN ( __bothinit_array_end = . );
 
-  .dtors          :
+  /* All destructor input sections (.dtors, .fini_array) are likewise
+   * collected into a SINGLE output section so no orphan section can be
+   * placed between the entries that the runtime walks at exit. */
```
quic-seaswara, line 66: I am not sure if you can put this into one section. 

Can you add a test that includes ctors and dtors with init array / fini array + ctros/dtors and see if they run properly and match expections if they are separate ? 

`eld` now supports run tests so you can definitely add a run test and verify.

quic-seaswara (changes requested):

templates/staticExecutable/static-executable-h2-picolibc.lcs.template
quic-k: we don't have run tests for Hexagon yet
I ran Picolibc testsuite locally with this change and we have no regressions, so this should be good to merge

quic-seaswara (approved):

PR #1425, opened 2026-07-08
Fix --wrap option when dealing with archives

### LINKED ISSUE #936: `--wrap=<sym>`: archive member not pulled when wrapper calls `__real_<sym>`

Steven6798:
Ensure wrapped symbols are pulled from archives when the wrapper references __real_<sym>.

Fix: qualcomm#936

### DIFF
--- a/docs/userguide/documentation/linker_faq.rst
+++ b/docs/userguide/documentation/linker_faq.rst
@@ -3076,6 +3076,34 @@ The wrapper function prints the number of bytes that is requested by user using
   # '4' bytes requested using my_malloc
   # p: 11
 
+Archive Member Extraction
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+When '--wrap=<sym>' is specified, three names form the *wrap chain* for that
+symbol: 'sym' (the original name, any reference to which is renamed to
+'__wrap_sym'), '__wrap_sym' (the wrapper definition supplied by the user),
+and '__real_sym' (any reference to which is renamed to 'sym', letting the
+wrapper call through to the original definition).
+
+Archive member extraction still follows the ordinary rule: a member is only
+pulled from an archive to resolve an undefined reference that actually
+exists after wrap-renaming has been applied.
+
+Debugging Wrap Symbols
+^^^^^^^^^^^^^^^^^^^^^^
+
+To trace symbol wrapping operations, use '--trace=wrap-symbols':
+
+.. code-block:: bash
+
+  eld ... --wrap=add --trace=wrap-symbols
+
+This will report:
+
+- Which symbols are renamed by the wrap mechanism
+- Which archive members are pulled for wrapped symbols
+- The complete symbol resolution chain
+
 Build time issues
 ==================
 
--- a/include/eld/Diagnostics/DiagSymbolResolutions.inc
+++ b/include/eld/Diagnostics/DiagSymbolResolutions.inc
@@ -42,6 +42,8 @@ DIAG(insert_wrapper, DiagnosticEngine::Note,
      "Inserting wrapper `%0' for bitcode symbol")
 DIAG(real_sym_reference, DiagnosticEngine::Note,
      "Found real symbol `%0' for wrapped symbol `%1'")
+DIAG(trace_wrap_archive_pull, DiagnosticEngine::Trace,
+     "Pulling archive member from `%0' for wrapped symbol `%1'")
 DIAG(comm_override_by_definition, DiagnosticEngine::Warning,
      "common symbol %0 in file %1 is overridden by definition from %2")
 DIAG(common_symbol_doesnot_override_definition, DiagnosticEngine::Warning,
--- a/lib/Core/Module.cpp
+++ b/lib/Core/Module.cpp
@@ -667,8 +667,19 @@ LDSymbol *Module::addSymbolFromBitCode(
                                getPrinter());
     if (!ThisConfig.options().renameMap().empty() &&
         Desc == ResolveInfo::Undefined) {
-      if (ThisConfig.options().renameMap().count(Name))
+      auto RenameSym = ThisConfig.options().renameMap().find(Name);
+      if (RenameSym != ThisConfig.options().renameMap().end()) {
         saveWrapReference(Name);
+        // Also insert the rename target as undefined so archive scanning
+        // can find it and pull the archive member that defines it.
+        const std::string &RenamedName = RenameSym->getValue();
+        Resolver::Result RenameResult = {nullptr, false, false};
+        getNamePool().insertSymbol(
+            &CurInput, RenamedName, /*IsSymbolInDynamicLibrary=*/false, Type,
+            Desc, Binding, Size, /*Value=*/0, Visibility,
+            /*OldSymbolInfo=*/nullptr, RenameResult, /*IsLtoPhase=*/false,
+            /*IsBitCode=*/true, PIdx, getPrinter());
+      }
     }
   }
 
--- a/lib/Readers/ArchiveParser.cpp
+++ b/lib/Readers/ArchiveParser.cpp
@@ -501,9 +501,19 @@ ArchiveParser::shouldIncludeSymbol(const ArchiveFile &archive,
   // If there is a wrap option specified for that symbol, we only pull from
   // an archive, if the real symbol is still undefined.
   std::string realSymbol = ("__real_" + SymName).str();
-  ResolveInfo *RealSymbol = m_Module.getNamePool().findInfo(realSymbol);
-  if (RealSymbol && RealSymbol->isUndef())
-    return ArchiveFile::Symbol::Include;
+  if (m_Module.hasWrapReference(realSymbol)) {
+    ResolveInfo *RealSymbol = m_Module.getNamePool().findInfo(realSymbol);
+    if (!RealSymbol || RealSymbol->isUndef()) {
+      // Emit a trace message when pulling archive members for wrapped symbols
+      // to aid in debugging the wrap chain and archive resolution process.
+      if (m_Module.getPrinter()->traceWrapSymbols()) {
+        m_Module.getConfig().raise(Diag::trace_wrap_archive_pull)
+            << archive.getInput()->decoratedPath() // archive filename
+            << SymName;                            // wrapped symbol name
+      }
+      return ArchiveFile::Symbol::Include;
+    }
+  }
 
   return ArchiveFile::Symbol::Exclude;
 }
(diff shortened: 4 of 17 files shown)


Steven6798:
This PR replaces/supersedes #1187 
The original implementation by @vijayvarghese
contained several pieces that are being retained in this solution, including:

lit tests
Diagnostic to track archive pulls

However, the original PR did not fully resolve the issue requirements, and this has become a priority item. Credit to @vijayvarghese for the portions that were incorporated into the final solution.

Steven6798:
Added more test cases: Garbage collection and ThinLTO

docs/userguide/documentation/linker_faq.rst
```diff
@@ -3076,6 +3076,28 @@ The wrapper function prints the number of bytes that is requested by user using
   # '4' bytes requested using my_malloc
   # p: 11
 
+Archive Member Extraction
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+When using '--wrap=<sym>', archive members containing any part of the wrap chain
```
parth-07, line 3082: Can you please explain what do you mean by wrap chain here?

Will an archive member that provides `foo` definition be pulled into the link if the wrap functionality is used as `--wrap=foo`, and there are no references to `__real_foo` in the image?

include/eld/Diagnostics/DiagSymbolResolutions.inc
```diff
@@ -42,6 +42,8 @@ DIAG(insert_wrapper, DiagnosticEngine::Note,
      "Inserting wrapper `%0' for bitcode symbol")
 DIAG(real_sym_reference, DiagnosticEngine::Note,
      "Found real symbol `%0' for wrapped symbol `%1'")
+DIAG(trace_wrap_archive_pull, DiagnosticEngine::Note,
```
parth-07, line 45: Should the diagnostic severity be `DiagnosticEngine::Trace` here?

lib/Readers/ArchiveParser.cpp
```diff
@@ -496,9 +496,19 @@ ArchiveParser::shouldIncludeSymbol(const ArchiveFile &archive,
   // If there is a wrap option specified for that symbol, we only pull from
   // an archive, if the real symbol is still undefined.
   std::string realSymbol = ("__real_" + SymName).str();
-  ResolveInfo *RealSymbol = m_Module.getNamePool().findInfo(realSymbol);
-  if (RealSymbol && RealSymbol->isUndef())
-    return ArchiveFile::Symbol::Include;
+  if (m_Module.hasWrapReference(realSymbol)) {
+    ResolveInfo *RealSymbol = m_Module.getNamePool().findInfo(realSymbol);
+    if (!RealSymbol || RealSymbol->isUndef()) {
```
parth-07, line 501: If there is a wrap reference, then can `RealSymbol` be a nullptr here?

lib/Core/Module.cpp
```diff
@@ -663,8 +663,18 @@ LDSymbol *Module::addSymbolFromBitCode(
...
+      auto RenameSym = ThisConfig.options().renameMap().find(Name);
+      if (RenameSym != ThisConfig.options().renameMap().end()) {
         saveWrapReference(Name);
+        // Also insert the rename target as undefined so archive scanning
+        // can find it and pull the archive member that defines it.
+        const std::string &RenamedName = RenameSym->getValue();
+        Resolver::Result RenameResult = {nullptr, false, false};
+        getNamePool().insertSymbol(&CurInput, RenamedName, false, Type, Desc,
+                                   Binding, Size, 0, Visibility, nullptr,
+                                   RenameResult, false /*isPostLTOPhase*/, true,
```
parth-07, line 675:
```suggestion
                                   RenameResult,  /*isPostLTOPhase=*/false, /*isBitCode=*/true,
```

lib/Core/Module.cpp
```diff
@@ -663,8 +663,18 @@ LDSymbol *Module::addSymbolFromBitCode(
...
-      if (ThisConfig.options().renameMap().count(Name))
+      auto RenameSym = ThisConfig.options().renameMap().find(Name);
+      if (RenameSym != ThisConfig.options().renameMap().end()) {
         saveWrapReference(Name);
+        // Also insert the rename target as undefined so archive scanning
+        // can find it and pull the archive member that defines it.
+        const std::string &RenamedName = RenameSym->getValue();
+        Resolver::Result RenameResult = {nullptr, false, false};
+        getNamePool().insertSymbol(&CurInput, RenamedName, false, Type, Desc,
+                                   Binding, Size, 0, Visibility, nullptr,
```
parth-07, line 674:
```suggestion
                                   Binding, Size, /*Value=*/0, Visibility, nullptr,
```

lib/Core/Module.cpp
parth-07, line 660: Should we call `insertSymbol` here even if the symbol is present in the rename map?

For example, let's say a bitcode file has an undefined reference to `foo`, and the link has option `--wrap=foo`. Now, we have an archive that contains a member with `foo` definition. If we add `foo` from bitcode here, then would we incorrectly pull the archive member that contains the `foo` definition?

test/Common/standalone/WrapSymbolArchive/WrapSymbolWrapInArchive.test
```diff
@@ -0,0 +1,60 @@
...
+
+# Archive contains both wrapper and original implementation
+RUN: %ar cr %aropts %t.2.libfoo.a %t.1.wrap.o %t.1.add.o
+
+# Link: main.o directly, both wrap and add from archive
+RUN: %link %linkopts %t.1.main.o %t.2.libfoo.a \
+RUN:   --wrap=add -o %t.2.out
+
+# Verify it linked successfully
+RUN: test -f %t.2.out
```
parth-07, line 20: Why are we using `test` command here? Isn't it enough to check that the `%link` does not return a failure?

parth-07 (changes requested):
There's a regression with this patch when a bitcode file contains both references and definition for a wrap symbol. In this case, the reference should bind to the actual definition, instead of the wrap symbol. Reproducer:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
__attribute__((noinline)) int add(int a, int b) { return a + b; }
__attribute__((noinline)) int caller(void) { return add(1, 2); }
int main(void) { return caller(); }
int __wrap_add(int a, int b) { return 100 + a + b; }
EOF

clang -o 1.o 1.c -flto -c 
ld.eld -o 1.out 1.o --wrap=add -e main
```

Here `add(1, 2)` call in the `caller` is now incorrectly replaced by `__wrap_add(1, 2)` call.

Steven6798:
> There's a regression with this patch when a bitcode file contains both references and definition for a wrap symbol. In this case, the reference should bind to the actual definition, instead of the wrap symbol. Reproducer:

Thanks for all the feedback. This patch still requires some work and will remain in draft. If anything else comes to mind let me know.

include/eld/Diagnostics/DiagSymbolResolutions.inc
Steven6798: Thanks!

Steven6798:
> There's a regression with this patch when a bitcode file contains both references and definition for a wrap symbol. In this case, the reference should bind to the actual definition, instead of the wrap symbol. Reproducer:

This matches what lld does. Is that incorrect?

test/Common/standalone/WrapSymbolArchive/WrapSymbolWrapInArchive.test
Steven6798: Thanks!

parth-07:
> > There's a regression with this patch when a bitcode file contains both references and definition for a wrap symbol. In this case, the reference should bind to the actual definition, instead of the wrap symbol. Reproducer:
> 
> This matches what lld does. Is that incorrect?

wrapping should happen when there is an undefined reference. In this case, the symbol is defined. Why should wrapping happen here?

Steven6798:
> wrapping should happen when there is an undefined reference. In this case, the symbol is defined. Why should wrapping happen here?

I was using lld as a reference. Thanks for the correction.

Steven6798:
> > > There's a regression with this patch when a bitcode file contains both references and definition for a wrap symbol. In this case, the reference should bind to the actual definition, instead of the wrap symbol. Reproducer:

I spotted the issue. I am saving a wrap reference unconditionally. Thanks for the help.

lib/Core/Module.cpp
Steven6798: I added some test cases related to this, and it fails with and without the fix. I'll look into this. Thanks!

lib/Core/Module.cpp
Steven6798: Thanks!

lib/Core/Module.cpp
Steven6798: Thanks!

docs/userguide/documentation/linker_faq.rst
Steven6798: > Can you please explain what do you mean by wrap chain here?

The original symbol, the wrap symbol and the real symbol. I'll update this doc.

> Will an archive member that provides `foo` definition be pulled into the link if the wrap functionality is used as `--wrap=foo`, and there are no references to `__real_foo` in the image?

No. Archives must still follow the rules:  member is only pulled from an archive to resolve an undefined reference that actually exists after wrap-renaming has been applied.

lib/Readers/ArchiveParser.cpp
Steven6798: Yes. This can be tested using the tests added in this PR. They fail if the check is not included. Thanks for keeping an eye on things like this.

lib/Core/Module.cpp
Steven6798: Update: the test case was wrong. After fixing it this is not an issue before or after the fix. After I update this PR let me know if the test is correct. I'll let you know the name of the test here.

Steven6798: WrapSymbolWrapNotDefined.test

PR #1426, opened 2026-07-08
[ARM] Honor dot assignments before *(.ARM.exidx*) in sortEXIDX

parth-07:
This commit fixes ARMGNULDBackend::sortEXIDX so that a dot assignment such as '. = ALIGN(0x8)' appearing before the `*(.ARM.exidx*)` rule is honored when the EXIDX fragments are laid out.


parth-07:
Closing in favor of https://github.com/qualcomm/eld/pull/1439

PR #1428, opened 2026-07-08
[RISCV] Add tests for all relocs that ignore non-zero addends

jonathonpenix:
As far as I can tell these weren't tested elsewhere. Add tests similar to what was already done for PCREL_LO12_I.

Also fixup a stale comment after the relocation handling was refactored recently.


quic-areg (approved):

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -1812,16 +1807,15 @@ bool RISCVLDBackend::handleRelocation(ELFSection *pSection,
...
+  // R_RISCV_PCREL_LO* and TLSDESC relocations must have an addend of zero.
+  // Ignore any non-zero addends and warn.
   case llvm::ELF::R_RISCV_PCREL_LO12_I:
   case llvm::ELF::R_RISCV_PCREL_LO12_S:
   case llvm::ELF::R_RISCV_TLSDESC_LOAD_LO12:
   case llvm::ELF::R_RISCV_TLSDESC_ADD_LO12:
   case llvm::ELF::R_RISCV_TLSDESC_CALL: {
     if (pAddend) {
-      config().raise(Diag::warn_ignore_pcrel_lo_addend)
+      config().raise(Diag::warn_ignore_reloc_addend)
```
quic-seaswara, line 1818: Why only for these relocations ?

jonathonpenix: Didn't check on the original history, but right now at least this set looks about right to me still as these are the relocations I see called out in the psabi as required to have a zero addend ("the addend must be 0" in this [table](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#relocations)).

The exception is `R_RISCV_TLSDESC_CALL` which doesn't have that note. But I left it as:
* The already existing behavior was to warn (I'm just changing comments/naming/adding tests in this PR, otherwise it should be NFC)
* IIUC, this relocation is basically just for informational purposes and the addend shouldn't do anything, so warning about it doesn't seem that bad

~I didn't check the history (need to go back to before eld-in-github times 😄~)*, but my guess would be that `R_RISCV_TLSDESC_CALL` was mostly here for doing the lo -> hi reloc association that was recently refactored out.

But I might be wrong/misunderstanding.

If folks have strong feelings either way, I'm happy to change it.

*Edit: I don't know why I thought it was added earlier but it was added [here](https://github.com/qualcomm/eld/commit/c945f3bda380f44d8bc381fb63e292055c9a5624). My guess is still that it was mostly there for the lo -> hi reloc mapping, but I don't think the addend checking is necessarily _harmful_. @quic-akaryaki do you have thoughts on this?

quic-seaswara (approved):

quic-seaswara (changes requested):
Update the developer docs accordingly

jonathonpenix:
What sort of updates are you looking for?

This is just adding tests/updating comments/renaming the diag--it should be NFC

If you mean documenting addends or something, this info is already clearly marked in the psABI--I'm a bit hesitant to create a downstream copy of all that info that we have to maintain.

quic-seaswara:
sorry I was trying to add to this PR.

https://github.com/qualcomm/eld/pull/1534

By mistake it landed on this PR.

Ignore my comment for this PR.

quic-seaswara (approved):

PR #1429, opened 2026-07-08
[RISCV] Ignore addend in R_RISCV_GOT_HI20

### LINKED ISSUE #1413: [risc-v] Add diagnostic for non zero addends
Yes please can it be a follow-up.

Compilers should only produce addend=0 code, but hand-assembly might do something funky and wrong. bfd ld gives an error when the addend is non-zero, which might be the sort of thing ELD wants because I know you love a good diagnostic @quic-seaswara.

_Originally posted by @lenary in https://github.com/qualcomm/eld/pull/1270#discussion_r3538889646_

jonathonpenix:
This was recently changed in the psABI to require the addend be non-zero [1].

Handle this similarly to other relocations that must have zero addends by just
ignoring the addend (setting it to zero) and producing a warning.

Note that the calculation in `applyGOT` is left unchanged (the `+ A` is still
there). `applyGOT` is shared with some other GOT-relative relocations that have
no explicit zero addend requirement and eld/lld/bfd all seem to a) accept those
addends b) honor those addends and c) not produce a diagnostic.

So for now, I think it is a bit clearer to keep `R_RISCV_GOT_HI20` using
`applyGOT` and just make sure we zero the addend ahead of time.

Since we're 0-ing the addend ahead of time, switch the GOT relaxations
to asserting on non-0 addends rather than returning. It's a bit unrelated,
but the assert was also extended to cover both the got_hi and pcrel_lo
(as both must have 0 addends).

[1] https://github.com/riscv-non-isa/riscv-elf-psabi-doc/commit/eb794e8ccfabfbb370db2070565b588252c4f8b7

Fixes: #1413

### DIFF
--- a/lib/Target/RISCV/RISCVLDBackend.cpp
+++ b/lib/Target/RISCV/RISCVLDBackend.cpp
@@ -1422,11 +1422,11 @@ bool RISCVLDBackend::doRelaxationGOT(Relocation &Reloc) {
 
   // Until recently, the calculation for R_RISCV_GOT_HI20 was `G + GOT + A - P`
   // where non-zero addends were allowed. The addend must now be zero and has
-  // been removed from calculation, but not all tools reflect this yet. It's
-  // unclear how this relaxation should work in the presence of a non-zero
-  // addend so avoid doing so to be safe.
-  if (BaseReloc->addend())
-    return false;
+  // been removed from the calculation. eld should ignore any non-zero addends
+  // for the relocations involved here, but assert to be sure as it is unclear
+  // how non-zero addends should be handled in this relaxation.
+  assert(!Reloc->addend() && !BaseReloc->addend() &&
+         "Unexpected non-zero addend!");
 
   ResolveInfo *SymInfo = BaseReloc->symInfo();
   // The psABI only includes "it’s bound at link time to be within the object"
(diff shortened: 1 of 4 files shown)


lenary:
I did a bit of background research and it's likely all the other `applyGOT` calls should also ignore their addends, but we didn't cover those in the psABI update yet, so I think it's reasonable to limit this to those covered by the psABI update (just GOT_HI20)

quic-seaswara:
> I did a bit of background research and it's likely all the other `applyGOT` calls should also ignore their addends, but we didn't cover those in the psABI update yet, so I think it's reasonable to limit this to those covered by the psABI update (just GOT_HI20)

Thanks @lenary. Can we raise an issue to update the ABI ?

lenary:
> > I did a bit of background research and it's likely all the other `applyGOT` calls should also ignore their addends, but we didn't cover those in the psABI update yet, so I think it's reasonable to limit this to those covered by the psABI update (just GOT_HI20)
> 
> Thanks Sam Elliott (@lenary). Can we raise an issue to update the ABI ?

Yes, I intend to. I just need to finish reading about how tls gd works in LLD to make sure my understanding is right.

lenary:
https://github.com/riscv-non-isa/riscv-elf-psabi-doc/issues/510

jonathonpenix:
Thanks Sam for filing the issue in the psABI!

As I don't think there's been movement in the psABI, does anyone have concerns about keeping this patch limited to just `R_RISCV_GOT_HI20`?

I'll update this PR as the GOT relaxations have since been merged (which, there's some interaction here I think), but waiting on #1428 to save some rebasing.

But if we'd rather wait for further changes in the psABI, etc. I can put this on hold.

jonathonpenix:
Rebased to unstack after merging #1428 + added some fixups of some of the GOT relaxation stuff.

jonathonpenix:
Ping, I think we still want something like this in.

lenary (approved):
LGTM

quic-seaswara (approved):

PR #1431, opened 2026-07-08
[github-workflow] Gate PR ELD toolset upload on `zephyr-check` label

quic-areg:
97ea3ddc disabled the upload because it ran on every CI build and the storage utilization ended up being too high (~25 MB * ~30 runs/day at 1 day retention).

Re-enable, gated on a `zephyr-check` PR label.

Requires the `zephyr-check` label to exist on qualcomm/eld.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,8 @@ name: CI
 
 on:
   pull_request:
+    # `labeled` so adding `zephyr-check` mid-PR triggers a fresh CI upload.
+    types: [opened, synchronize, reopened, labeled]
     paths-ignore:
       - '.github/workflows/sanitize.yml'
       - '.github/workflows/reviter.yml'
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -183,13 +193,15 @@ jobs:
...
-      #  if: steps.file-check.outputs.skip_build == 'false'
-      #  uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a #v7.0.1
-      #  with:
-      #    name: pr-eld-toolset
-      #    retention-days: 1
-      #    path: pr-${{ env.PR_NUMBER }}/install-pr-toolset.tar.xz
+      - name: Upload PR ELD toolset
+        if: |
+          steps.file-check.outputs.skip_build == 'false' &&
+          contains(github.event.pull_request.labels.*.name, 'zephyr-check')
```
quic-seaswara, line 199: nit : this can be converted to a step that evaluates a condition and that variable can be used across the board, that will reduce repeating this across the workflow.

quic-seaswara (approved):

Steven6798 (approved):

.github/workflows/ci.yml
quic-areg: Considered it, but it ends up adding more code and an extra workflow step so I think it's not worth it. I will keep it in mind for later patches if this wf uses this check more times.

PR #1432, opened 2026-07-08
Reland "Make symbol table order deterministic (#1263)"

### LINKED ISSUE #1261: fix non-deterministic order of entries in .dynsym

quic-areg:
Multithreaded relocation scanning made the order in which symbols were appended to the module symbol vector dependent on thread timing. Both .dynsym and .symtab inherited that order: existing sorts had address ties everywhere, and stable_sort fell back to the unpredictable insertion order.

Sort each table deterministically by (input ordinal, input .symtab index).

Also fixes a latent bug where an absolute symbol was not counted as
global for computing sh_info.

Fixes #1261.

This relands 257cefcbf8c1271a38af7e1e9b0211aacc405eec, which was reverted in 3e0b10c6d82bced29cac51a00cbf3db975562beb.


quic-seaswara (approved):

parth-07 (approved):

lib/Fragment/DynSymFragment.cpp
```diff
@@ -48,7 +48,9 @@ eld::Expected<void> DynSymFragment::emit(MemoryRegion &Mr, Module &M) {
       M.getBackend().emitSymbol64(Symtab64[SymIdx], DynSym->outSymbol(),
                                   nullptr, 0, SymIdx,
                                   /*isDynSymTab=*/true);
-    if ((DynSym->isGlobal() || DynSym->isWeak()) && !FirstNonLocal)
+    if (M.getBackend().getSymbolBinding(DynSym->outSymbol()) !=
+            llvm::ELF::STB_LOCAL &&
+        !FirstNonLocal)
```
quic-seaswara, line 53: Why do we want to call backend.getSymbolBinding()  

can this be simplified to DynSym->isLocal() ?

lib/Target/GNULDBackend.cpp
```diff
@@ -941,6 +941,25 @@ void GNULDBackend::sizeDynNamePools() {
...
+    // within each group by (input ordinal, input .symtab index).
+    auto Cmp = [](const ResolveInfo *A, const ResolveInfo *B) -> bool {
+      bool UndA = A->isUndef() || A->isDyn();
+      bool UndB = B->isUndef() || B->isDyn();
+      if (UndA != UndB)
+        return UndA;
+      auto OrdA = A->resolvedOrigin()->getInput()->getInputOrdinal();
+      auto OrdB = B->resolvedOrigin()->getInput()->getInputOrdinal();
+      if (OrdA != OrdB)
+        return OrdA < OrdB;
```
quic-seaswara, line 956: Is there an ordering around protected and global symbols ?

lib/Fragment/DynSymFragment.cpp
quic-areg: For absolute symbols, where DynSym->isLocal() is false

lib/Target/GNULDBackend.cpp
quic-areg: There is no separate ordering for protected symbols; they are ordered (deterministically) like any other symbol.

lib/Fragment/DynSymFragment.cpp
quic-seaswara: It would be great if the commit message described the order of symbols that we use now, and also please document symbol order in our eld docs in this section under

https://qualcomm.github.io/eld/documentation/layout.html

quic-seaswara (changes requested):

quic-seaswara (approved):
please raise tickets for follow up.

lib/Fragment/DynSymFragment.cpp
quic-areg: https://github.com/qualcomm/eld/pull/1580

PR #1194, opened 2026-05-18
ARM: replace ARMEXIDXSection with EXIDX fragment/piece model

quic-seaswara:

PR #1438, opened 2026-07-09
[LinkerScript] Fix orphan section placement with scripts

### LINKED ISSUE #1408: Executable orphan placed ahead of .init when .init contains data-flagged input
On ARM Cortex-M targets, when a linker script gathers a data-flagged (WA) input section into an output section named .init, as picolibc-based bare-metal toolchains do for the interrupt vector table, ELD places an executable orphan section at the load-segment origin, ahead of .init. GNU ld and LLD honour .init's positional convention and place executable orphans after it; ELD does not.

### Environment

- ELD: eld 23.0 — commit 9e74efb3b
- Clang / llvm-readelf: 19.1.7.
- Host: Linux x86_64.

### Impact

.init is a long-standing positional convention: on architectures whose reset behaviour is to fetch from the beginning of the load segment (ARM Cortex-M reads initial SP and reset PC from flash address 0; equivalent conventions exist for m68k, PowerPC embedded, and others), the linker is expected to keep .init at that origin regardless of the flag bits of its contents. GNU ld and LLD both do this. When ELD places an executable orphan section in front of .init, the bytes at flash origin are no longer the ones the runtime expects, on Cortex-M the CPU reads the orphan's instructions as SP/PC and HardFaults on reset.

### Reproducer

```
  ; repro.s
      .syntax unified
      .thumb

      ; Data-flagged input section that the script gathers into .init.
      ; On Cortex-M / picolibc this is the interrupt vector table
      ; (initial SP + reset PC + IRQ handlers), stored as data.
      .section .data.init.enter, "aw", %progbits
      .p2align 2
  vector_table:
      .long 0x20001000     ; initial SP
      .long 0x00000001     ; reset PC (thumb bit set)

      ; Script-placed executable section.
      .section .text, "ax", %progbits
      .thumb_func
      .globl _start
  _start:
      bx lr

      ; Executable orphan — script does not name it.
      .section orphan_code, "ax", %progbits
      .thumb_func
      .globl orphan_fn
  orphan_fn:
      bx lr
```

```
  ; link.t
  MEMORY { flash (rx) : ORIGIN = 0x00000000, LENGTH = 0x00100000 }
  SECTIONS {
      .init : { KEEP(*(.data.init.enter)) } >flash    /* becomes WA */
      .text : { *(.text) }                  >flash    /* AX */
  }
```

```
$ clang -target armv7em-none-eabi -mfloat-abi=soft -mthumb -c repro.s -o repro.o
$ ld.eld -march arm repro.o -T link.t -o repro.out
$ llvm-readelf -W -S repro.out
  ...
    [Nr] Name              Type            Address  Off    Size   ES Flg Lk Inf Al
    [ 0]                   NULL            00000000 000000 000000 00      0   0  0
    [ 1] orphan_code       PROGBITS        00000000 000094 000002 00  AX  0   0  1
    [ 2] .init             PROGBITS        00000004 000098 000008 00  WA  0   0  4
    [ 3] .text             PROGBITS        0000000c 0000a0 000002 00  AX  0   0  4
  ...
```

### Expected vs. observed

The linker script assigns .init and .text to the flash region starting at 0x00000000; orphan_code is an orphan and its position is up to the linker. GNU ld and LLD both place .init at flash origin and put the AX orphan after .text.

```
  Expected (GNU ld / LLD behaviour):

  [ 1] .init         PROGBITS   Address: 00000000   Flg: WA
  [ 2] .text         PROGBITS   Address: 00000008   Flg: AX
  [ 3] orphan_code   PROGBITS   Address: 0000000a   Flg: AX

  Observed (ELD 23.0):

  [ 1] orphan_code   PROGBITS   Address: 00000000   Flg: AX   ← at flash origin, ahead of .init
  [ 2] .init         PROGBITS   Address: 00000004   Flg: WA   ← displaced from flash origin
  [ 3] .text         PROGBITS   Address: 0000000c   Flg: AX
```

The AX orphan orphan_code is placed at flash origin 0x00000000, and .init is pushed to 0x00000004. The first bytes of the image are therefore orphan_code's instructions, not the vector table the runtime expects at reset. Linking the same input with ld.bfd or ld.lld produces the "Expected" layout.

quic-areg:
Orphan insertion scan stopped at the first section with a greater rank than the orphan, placing the orphan before that section. This groups sections with like permissions together when the script is sorted by rank, but when a
  script declares a lower ranked section first, the scan can insert a higher ranked orphan too early:

```ld
  SECTIONS {
      .init : { KEEP((.data.init.enter)) }  /* WA (picolibc) */
      .text : { (.text) }                   /* AX */
  }
```
  Note: AX/.text has a lower rank than WA/.init.

  A third (orphan) section `orphan_code` (AX) is placed before `.init`, the first section with a higher rank, but `.init` is expected to come first. lld/bfd instead group the orphan with its like ranked neighbor (`.text`) rather
  than taking eld's single pass insertion shortcut, so the orphan lands after `.text` and `.init` keeps its place.

  Scan the whole list and insert after the last section with rank `<=` the orphan's. For scripts sorted by rank, this is equivalent. The no linker script path is unchanged.

  Also update Hexagon TrampolineMissingSymbol: the fix moved its orphaned `.init` next to `.txt.foo`, removing the long call. Add an explicit `.init` rule to restore the call distance and check that the expected trampoline is
  emitted.

Fixes #1408

### DIFF
--- a/lib/Target/GNULDBackend.cpp
+++ b/lib/Target/GNULDBackend.cpp
@@ -2732,6 +2732,29 @@ bool GNULDBackend::checkCrossReferences() {
   return true;
 }
 
+// True if a section is bound only to non-PT_LOAD segments. A section with no
+// phdr binding stays in the default PT_LOAD, so it is loadable.
+static bool isInNonLoadSegment(const OutputSectionEntry *Sec,
+                               LinkerScript &Script) {
+  if (Sec->getSection()->isNullKind())
+    return true;
+  if (!Sec->epilog().hasPhdrs())
+    return false;
+  bool sawPhdr = false;
+  for (const StrToken *Tok : Sec->epilog().phdrs()->tokens()) {
+    uint32_t type = llvm::ELF::PT_NULL;
+    for (const auto *Phdr : Script.phdrList())
+      if (Phdr->spec().name() == Tok->name()) {
+        type = Phdr->spec().type();
+        break;
+      }
+    if (type == llvm::ELF::PT_LOAD)
+      return false;
+    sawPhdr = true;
+  }
+  return sawPhdr;
+}
+
 /// placeOutputSections - place output sections based on SectionMap
 bool GNULDBackend::placeOutputSections() {
   typedef std::vector<ELFSection *> Orphans;
@@ -2996,15 +3019,37 @@ bool GNULDBackend::placeOutputSections() {
         // later
         // though. The value is initialized to -1, to skip the Null section
         // inserted.
+        //
+        // With a SECTIONS command sectionMap is in declaration order, not
+        // rank order, so stopping at the first higher ranked entry can insert
+        // the orphan too early. Track the last position it may follow instead.
+        //
+        // An orphan inherits the segment of the section before it. Keep an
+        // alloc orphan out of a non-load segment by preferring the last
+        // loadable segment. Fall back to the last fit if there is none (eg
+        // NONE segment layouts).
         int numSections = -1;
+        SectionMap::iterator lastFit = outEnd, lastLoadFit = outEnd;
         for (out = outBegin; out != outEnd; ++out) {
           if (numSections && (*out)->isDiscard() && (order <= SHO_UNDEFINED))
             break;
           if ((*out)->order() > order) {
-            break;
+            if (!linkerScriptHasSectionsCommand)
+              break;
+          } else if (linkerScriptHasSectionsCommand) {
+            lastFit = out;
+            if (orphan->isAlloc() &&
+                !isInNonLoadSegment(*out, m_Module.getScript()))
+              lastLoadFit = out;
           }
           ++numSections;
         }
+        if (lastLoadFit != outEnd)
+          lastFit = lastLoadFit;
+        // lastFit points at the entry to follow; insert after it. No fit
+        // (outEnd) means the orphan outranks everything, so insert at front.
+        if (linkerScriptHasSectionsCommand)
+          out = (lastFit == outEnd) ? outBegin : ++lastFit;
         if (orphan->getOutputSection())
           out = sectionMap.insert(out, orphan->getOutputSection());
         else {
--- a/test/Common/RunTests/OrphanNoLoadSegmentPhdr/OrphanNoLoadSegmentPhdr.test
+++ b/test/Common/RunTests/OrphanNoLoadSegmentPhdr/OrphanNoLoadSegmentPhdr.test
@@ -1,0 +1,77 @@
+REQUIRES: run_test
+
+## Verify that orphan section placement respects segments. An orphan writable
+## section must not be placed into a non-loadable  segment even when a same rank
+## script mentioned section is bound there; doing so would leave
+## the orphan unmapped at run time.
+
+RUN: split-file %s %t
+RUN: %run_cc %clangopts -c -fPIC %t/orphan.c -o %t.o
+RUN: %run_cc %clangopts -o %t.out %t.o -Wl,-dy -Wl,-z,max-page-size=0x1000 -T %t/script.t
+RUN: %run %t.out | %filecheck %s
+RUN: %readelf -l %t.out | %filecheck %s --check-prefix=PHDRS
+
+CHECK: OK
+
+PHDRS:      LOAD
+PHDRS:      LOAD
+PHDRS:      NULL
+
+#--- orphan.c
+#include <stdio.h>
+
+__attribute__((section(".orphan_data"))) volatile int sentinel = 0xC0FFEE;
+
+/// Force the overlay output section to exist so the orphan placement scan
+/// actually considers it as a candidate anchor.
+__attribute__((section(".overlay"), used)) volatile int overlay_marker;
+
+int main(void) {
+  if (sentinel != 0xC0FFEE) {
+    fprintf(stderr, "orphan stranded: sentinel=0x%x\n", sentinel);
+    return 1;
+  }
+  puts("OK");
+  return 0;
+}
+
+#--- script.t
+PHDRS {
+  ph_phdr    PT_PHDR    PHDRS;
+  ph_interp  PT_INTERP;
+  ph_text    PT_LOAD    FILEHDR PHDRS FLAGS(5);
+  ph_data    PT_LOAD    FLAGS(6);
+  ph_dyn     PT_DYNAMIC;
+  ph_overlay PT_NULL    FLAGS(6);
+}
+SECTIONS {
+  .interp        : { *(.interp) }         :ph_interp :ph_text
+  .note.gnu.build-id : { *(.note.gnu.build-id) } :ph_text
+  .dynsym        : { *(.dynsym) }         :ph_text
+  .dynstr        : { *(.dynstr) }         :ph_text
+  .gnu.version   : { *(.gnu.version) }    :ph_text
+  .gnu.version_r : { *(.gnu.version_r) }  :ph_text
+  .gnu.hash      : { *(.gnu.hash) }       :ph_text
+  .rela.dyn      : { *(.rela.dyn) }       :ph_text
+  .rela.plt      : { *(.rela.plt) }       :ph_text
+  .rel.dyn       : { *(.rel.dyn) }        :ph_text
+  .rel.plt       : { *(.rel.plt) }        :ph_text
+  .init          : { *(.init) }           :ph_text
+  .plt           : { *(.plt*) }           :ph_text
+  . = ALIGN(4K);
+  .text          : { *(.text*) }          :ph_text
+  .fini          : { *(.fini) }           :ph_text
+  .rodata        : { *(.rodata*) }        :ph_text
+  .eh_frame      : { *(.eh_frame*) }      :ph_text
+  . = ALIGN(4K);
+  .init_array : { KEEP(*(.init_array*)) } :ph_data
+  .fini_array : { KEEP(*(.fini_array*)) } :ph_data
+  .dynamic    : { *(.dynamic) }           :ph_data :ph_dyn
+  .got        : { *(.got) }               :ph_data
+  .got.plt    : { *(.got.plt) }           :ph_data
+  __global_pointer$ = .;
+  .mydata     : { *(.mydata) *(.data*) }  :ph_data
+  .bss        : { *(.bss*) *(COMMON) }    :ph_data
+  . = 0x40000000;
+  .overlay (NOLOAD) : { *(.overlay) }     :ph_overlay
+}
--- a/test/Common/standalone/OrphanBeforeLowerRankedSection/OrphanBeforeLowerRankedSection.test
+++ b/test/Common/standalone/OrphanBeforeLowerRankedSection/OrphanBeforeLowerRankedSection.test
@@ -1,0 +1,53 @@
+## When a linker script declares a lower ranked section before a higher ranked
+## one (a WA .vectors before an AX .text), the output section list is not
+## sorted by rank. An executable orphan must still be placed after .text, not
+## spliced in ahead of .vectors.
+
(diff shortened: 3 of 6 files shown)


lib/Target/GNULDBackend.cpp
```diff
@@ -2930,15 +2930,25 @@ bool GNULDBackend::placeOutputSections() {
...
+        SectionMap::iterator lastFit = outEnd;
         for (out = outBegin; out != outEnd; ++out) {
           if (numSections && (*out)->isDiscard() && (order <= SHO_UNDEFINED))
             break;
           if ((*out)->order() > order) {
-            break;
+            if (!linkerScriptHasSectionsCommand)
+              break;
+          } else {
+            lastFit = out;
```
quic-seaswara, line 3040: This algorithm will make the linker assign the orphan to the last group of sections that are grouped together ? 

For example :

```
cat > script.t << \!
SECTIONS {
  .rw : { *(.rw1) }
  .rx : { *(.rx1) }
  .rw : { *(.rw2) }
  .rx : { *(.rx2) }
}
!

cat > a.s << \!
.section .rw1, "aw", @progbits
foo1:
.word 20
.section .rx1, "ax", @progbits
foo2:
.word 20
.section .rw2, "aw", @progbits
foo3:
.word 20
.section .rx2, "ax", @progbits
foo4:
.word 20
.section .orphan.rx3, "ax", @progbits
foo5:
.word 20
!

clang -c -target riscv32 a.s
ld.eld -T script.t a.o
```

On the other hand should we just change the condition to return SHO_INIT when linker scripts are used ?

quic-seaswara (commented):
Can we change and assign SHO_INIT rank to the init section instead ?

lib/Target/GNULDBackend.cpp
quic-areg: >  should we just change the condition to return SHO_INIT when linker scripts are used

I did consider this, but I don't think it's the right move.

It is special-casing on section names, which is error prone (I think you even mentioned this in the issue thread). It also only fixes the literal section `.init` and does not generalize to other cases (like the `.vectors` test case in this patch). 
> This algorithm will make the linker assign the orphan to the last group of sections that are grouped together ?

Your example is actually very interesting: with pre-patch eld this example reproduces the #1408 bug; the orphan lands at address 0 ahead of everything, including the first `.rw`. With this patch, it  happens to match lld: both linkers produce the order `.rw .rx .rw .rx .orphan.rx3` (bfd happens to merge the duplicate .rw/.rx sections).

quic-seaswara: Its good that we are matching lld now :) Can we document orphan placements and integrate the test as well.

quic-seaswara (approved):

Steven6798 (approved):

parth-07 (approved):
LGTM.

As a follow-up, it would be nice to formally document the orphan placement algorithm used by eld.

bojle:
LGTM. This PR fixes the issue I am seeing. See https://github.com/qualcomm/eld/pull/1194#issuecomment-4977586201.

lib/Target/GNULDBackend.cpp
```diff
@@ -2732,6 +2732,25 @@ bool GNULDBackend::checkCrossReferences() {
...
+  bool sawNonLoad = false;
+  for (const StrToken *Tok : Sec->epilog().phdrs()->tokens()) {
+    uint32_t type = llvm::ELF::PT_NULL;
+    for (const auto *Phdr : Script.phdrList())
+      if (Phdr->spec().name() == Tok->name()) {
+        type = Phdr->spec().type();
+        break;
+      }
+    if (type == llvm::ELF::PT_LOAD)
+      return false;
```
quic-seaswara, line 2752: For each output section you already know if this is part of a load segment by calling getLoadSegment().

lib/Target/GNULDBackend.cpp
```diff
@@ -2996,15 +3015,32 @@ bool GNULDBackend::placeOutputSections() {
...
         int numSections = -1;
+        SectionMap::iterator lastFit = outEnd;
         for (out = outBegin; out != outEnd; ++out) {
           if (numSections && (*out)->isDiscard() && (order <= SHO_UNDEFINED))
             break;
           if ((*out)->order() > order) {
-            break;
+            if (!linkerScriptHasSectionsCommand)
+              break;
+          } else if (!(scriptHasPhdrs &&
```
quic-seaswara: The function checks epilog.hasPhdrs() which means linker script has phdrs.

lib/Target/GNULDBackend.cpp
```diff
@@ -2732,6 +2732,25 @@ bool GNULDBackend::checkCrossReferences() {
   return true;
 }
 
+static bool isInNonLoadPhdrOnly(const OutputSectionEntry *Sec,
```
quic-seaswara: make this a private member function ?

lib/Target/GNULDBackend.cpp
```diff
@@ -2732,6 +2732,25 @@ bool GNULDBackend::checkCrossReferences() {
...
+                                LinkerScript &Script) {
+  if (!Sec->epilog().hasPhdrs())
+    return false;
+  bool sawNonLoad = false;
+  for (const StrToken *Tok : Sec->epilog().phdrs()->tokens()) {
+    uint32_t type = llvm::ELF::PT_NULL;
```
quic-seaswara, line 2745: can we add a build/run test to catch this early in CI ?

quic-seaswara (changes requested):

lib/Target/GNULDBackend.cpp
quic-seaswara: sorry my bad, this is set only during layout, when assigning virtual addresses. 

It might be useful to add a function to return the type of segment per output section entry ?

lib/Target/GNULDBackend.cpp
quic-areg: removed the extra check

lib/Target/GNULDBackend.cpp
quic-areg: I think the existing test is already sufficient and the runtime test is overkill, but I did add one.

lib/Target/GNULDBackend.cpp
quic-areg: static function is already private

lib/Target/GNULDBackend.cpp
```diff
@@ -2732,6 +2732,29 @@ bool GNULDBackend::checkCrossReferences() {
   return true;
 }
 
+// True if a section is bound only to non-PT_LOAD segments. A section with no
+// phdr binding stays in the default PT_LOAD, so it is loadable.
+static bool isInNonLoadSegment(const OutputSectionEntry *Sec,
+                               LinkerScript &Script) {
+  if (Sec->getSection()->isNullKind())
+    return true;
```
quic-seaswara, line 2740: Why this condition ? NullKind sections are not loadable anyway.

lib/Target/GNULDBackend.cpp
```diff
@@ -2996,15 +3019,35 @@ bool GNULDBackend::placeOutputSections() {
...
+        // rank order, so stopping at the first higher ranked entry can insert
+        // the orphan too early. Track the last position it may follow instead.
+        //
+        // An orphan inherits the segment of the section before it. Keep an
+        // alloc orphan out of a non-load segment by preferring the last
+        // loadable segment. Fall back to the last fit if there is none (eg
+        // NONE segment layouts).
         int numSections = -1;
+        SectionMap::iterator lastFit = outEnd, lastLoadFit = outEnd;
+        bool avoidNonLoad = linkerScriptHasSectionsCommand && orphan->isAlloc();
```
quic-seaswara: avoidNonLoad is inside the if condition that checks linkerScriptHasSectionsCommand, this can be removed and you can just add

orphan->isAlloc() && !isInNonLoadSegment

lib/Target/GNULDBackend.cpp
```diff
@@ -2996,15 +3019,35 @@ bool GNULDBackend::placeOutputSections() {
...
+            lastFit = out;
+            if (avoidNonLoad && !isInNonLoadSegment(*out, m_Module.getScript()))
+              lastLoadFit = out;
           }
           ++numSections;
         }
+        if (avoidNonLoad && lastLoadFit != outEnd)
+          lastFit = lastLoadFit;
+        if (linkerScriptHasSectionsCommand)
+          out = (lastFit == outEnd) ? outBegin : ++lastFit;
```
quic-seaswara, line 3052: What is an example when lastFit will reach upto outEnd ?

lib/Target/GNULDBackend.cpp
```diff
@@ -2996,15 +3019,35 @@ bool GNULDBackend::placeOutputSections() {
...
+          } else if (linkerScriptHasSectionsCommand) {
+            lastFit = out;
+            if (avoidNonLoad && !isInNonLoadSegment(*out, m_Module.getScript()))
+              lastLoadFit = out;
           }
           ++numSections;
         }
+        if (avoidNonLoad && lastLoadFit != outEnd)
+          lastFit = lastLoadFit;
+        if (linkerScriptHasSectionsCommand)
```
quic-seaswara, line 3051: Remove avoidNonLoad condition here ?

quic-seaswara, line 3051: comments will be useful on what is being done here.

quic-seaswara (changes requested):

lib/Target/GNULDBackend.cpp
quic-areg: when the orphan ranks ahead of all sections

quic-seaswara: Could you add a test ?

lib/Target/GNULDBackend.cpp
```diff
@@ -2732,6 +2732,29 @@ bool GNULDBackend::checkCrossReferences() {
...
+      if (Phdr->spec().name() == Tok->name()) {
+        type = Phdr->spec().type();
+        break;
+      }
+    if (type == llvm::ELF::PT_LOAD)
+      return false;
+    sawPhdr = true;
+  }
+  return sawPhdr;
+}
```
quic-seaswara, line 2756: I was trying to remember that we had something to iterate through and figure out if the segment for a section is loadable or not, can you use this when PHDR's are present.

`GNULDBackend::getSegmentsForSection`

quic-seaswara (changes requested):

lib/Target/GNULDBackend.cpp
quic-areg: ok, sure

lib/Target/GNULDBackend.cpp
quic-areg: sure, done

quic-areg: added

lib/Target/GNULDBackend.cpp
quic-areg: That's exactly why we have this condition?

lib/Target/GNULDBackend.cpp
quic-areg: I don't think that can be used here since the ` _segmentsForSection` map is populated after this code runs.

lib/Target/GNULDBackend.cpp
quic-areg: done

lib/Target/GNULDBackend.cpp
quic-seaswara: oh, I see. thanks for letting me know.

quic-seaswara (approved):

PR #1450, opened 2026-07-12
Fix reproduce functionality for shared libraries with sysroot marker

### LINKED ISSUE #972: Reproduce functionality does not properly support sysroot markers
The reproduce functionality does not properly support sysroot markers.

Reproducer:

```bash
#!/usr/bin/env bash

mkdir -p A/lib64

cat >1.c <<\EOF
int foo() { return 1; }
EOF

cat >main.c <<\EOF
int foo();
int main() { return foo(); }
EOF

clang -target x86_64-linux-gnu main.c -c -ffunction-sections
clang -o lib1.so 1.c -fPIC -shared
mv main.o lib1.so A/lib64
ld.eld -o main.eld.out =/main.o =/lib1.so --reproduce a.tar --sysroot A/lib64
mkdir reproduce -p
tar -xvf a.tar -C reproduce --strip-components=1
cd reproduce
bash -x response.txt 
```

**Output:**
```
+ cd reproduce
+ bash -x response.txt
+ ld.eld -o main.eld.out Object/main.o.6932507498210797127 lib1.so --sysroot A/lib64 --mapping-file=mapping.ini
Fatal: cannot read file 'lib1.so'
Fatal: Linking had errors.
```

Cause, the `lib1.so` mapping file entry is incorrect:

```
$ cat reproduce/mapping.ini
[Object Files]
=/main.o=Object/main.o.6932507498210797127

[Shared Libraries]
A/lib64/lib1.so=SharedLibrary/lib1.so.9024421968594623665
```

parth-07:
This commit fixes the --reproduce functionality so that command-line shared libraries specified using a sysroot marker (e.g. =/lib.so) works correctly.

The root cause was incorrect mapped path for non-namespec shared libraries when the shared library was specified using the sysroot marker (=). For a shared library, '=/lib1.so', the mapped path was the expanded path '${SYSROOT}/lib1.so'. However, response.txt file contains the lib1.so entry using 'rewritePath('=lib1.so')'. The rewritePath calls returns 'lib1.so' as fallback because there is no entry for '=lib1.so'. This is because '=lib1.so' entry was never added because we used incorrect mapped path for non-namespec shared libraries (decoratedPath instead of the command-line name used for everything else).

Resolves #972

### DIFF
--- a/test/Common/standalone/CommandLine/Reproduce/SysrootMarker.test
+++ b/test/Common/standalone/CommandLine/Reproduce/SysrootMarker.test
@@ -1,0 +1,22 @@
+#BEGIN_COMMENT
+# This test verifies that the reproduce functionality works correctly
+# with shared libraries specified using a sysroot marker (=lib1.so)
+#END_COMMENT
+RUN: %rm -rf %t.sysroot && %mkdir -p %t.sysroot/lib64
+RUN: %clang %clangopts -c -fPIC %p/Inputs/1.c -o %t.sysroot/lib64/lib.o
+RUN: %clang %clangopts -c %p/Inputs/2.c -o %t.sysroot/lib64/main.o
+RUN: %link %linkopts -shared %t.sysroot/lib64/lib.o -o %t.sysroot/lib64/lib.so
+RUN: %link %linkopts --dynamic =/main.o =/lib.so --sysroot %t.sysroot/lib64 \
+RUN:   --reproduce %t.tar --dump-mapping-file %t.mapping \
+RUN:   --dump-response-file %t.response -o %t.out
+# The shared library must be keyed by its marker form, not the resolved path.
+RUN: %filecheck %s --check-prefix=MAPPING < %t.mapping
+
+MAPPING-DAG: =/main.o=Object/
+MAPPING-DAG: =/lib.so=SharedLibrary/
+
+# Replaying the response file must succeed.
+RUN: %rm -rf %t.rep && %mkdir %t.rep
+RUN: %tar %gnutaropts -xf %t.tar -C %t.rep --strip-components=1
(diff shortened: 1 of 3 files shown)


quic-areg (approved):
LGTM

Steven6798 (approved):

test/Common/standalone/CommandLine/Reproduce/SysrootMarker.test
```diff
@@ -0,0 +1,22 @@
+#BEGIN_COMMENT
+# This test verifies that the reproduce functionality works correctly
+# with shared libraries specified using a sysroot marker (=lib1.so)
+#END_COMMENT
+RUN: %rm -rf %t.sysroot && %mkdir -p %t.sysroot/lib64
+RUN: %clang %clangopts -c -fPIC %p/Inputs/1.c -o %t.sysroot/lib64/lib.o
+RUN: %clang %clangopts -c %p/Inputs/2.c -o %t.sysroot/lib64/main.o
+RUN: %link %linkopts -shared %t.sysroot/lib64/lib.o -o %t.sysroot/lib64/lib.so
+RUN: %link %linkopts --dynamic =/main.o =/lib.so --sysroot %t.sysroot/lib64 \
```
quic-seaswara, line 9: This is not testing namespec. 

Namespec is -l:<library name> 

Can you please check the behavior ?

parth-07: This patch is not affecting namespec usage with reproduce. We already have a namespec test with reproduce here: test/Common/standalone/CommandLine/Reproduce/Namespec.test

quic-seaswara (approved):

PR #1459, opened 2026-07-13
Add iplt_(start|end) symbols, if referenced, in initStandardSymbols

### LINKED ISSUE #1342: Incorrect GOT behavior for __rela_iplt_start/__rela_iplt_end in statically linked HelloWorld.test run test
I'm not entirely sure what the cause is/I'm a bit fuzzy on how some of this works so I might get some things wrong here.

But broadly, I think this test is currently failing in the statically linked case because eld isn't correctly populating the GOT for `__rela_iplt_start`/`__rela_iplt_end` so usages of ifuncs end up jumping to 0x0 (and segfaulting).

In glibc, `__libc_start_main_impl` contains the following snippet to fetch addrs of `__rela_iplt_start`/`__rela_iplt_end`:

```
     152: 00000417      auipc   s0, 0x0
                        00000152:  R_RISCV_GOT_HI20     __rela_iplt_start
                        00000152:  R_RISCV_RELAX        *ABS*
     156: 00042403      lw      s0, 0x0(s0)
                        00000156:  R_RISCV_PCREL_LO12_I .L0
                        00000156:  R_RISCV_RELAX        *ABS*
     15a: 00000b97      auipc   s7, 0x0
                        0000015a:  R_RISCV_GOT_HI20     __rela_iplt_end
                        0000015a:  R_RISCV_RELAX        *ABS*
     15e: 000bab83      lw      s7, 0x0(s7)
                        0000015e:  R_RISCV_PCREL_LO12_I .L0
                        0000015e:  R_RISCV_RELAX        *ABS*
```

Note that these are GOT-relative accesses in the builds of glibc that are being used.

For some reason, when actually executing, these loads produce 0 when linked with eld.

I do see that eld created the symbols and they look to be about the right value:

```
0001014c r __rela_iplt_end
00010134 r __rela_iplt_start
```

But, since the loads out of the GOT are both 0, we skip over the code that does the `R_RISCV_IRELATIVE` handling, so I think the resolver never actually runs and updates the entry in the .got.plt.

Then, eventually when we try to use one of these ifuncs, from the .plt we end up loading 0 out of the .got.plt and jumping to 0x0, which naturally segfaults.

When tracing through the execution of this test when built with lld, the GOT-relative loads above correctly point to the `__rela_iplt_start`/`__rela_iplt_end` syms, we go through the whole ifunc resolving process as expected, etc.

I _think_ when this issue gets fixed we can un-XFAIL the HelloWorld.test test for RISC-V targets.

parth-07:
This commit modifies where `__rel(a)_iplt_(start|end)` symbols are added.
Previously, they were added during scan relocation phase on encountering
an ifunc symbol. However, it causes issues when a GOT to
`__rel(a)_iplt_(start|end)` symbol is encountered before an ifunc symbol.
Additionally, these symbols should also be made available for GNU
compatibility even if there is no ifunc symbol.
Now these symbols are added, if referenced, in initStandardSymbolsphase.
This phase happens before scan relocation.

Additionally, this PR refactors creating and defining of rela_iplt
symbols such that it happens in GNULDBackend, instead of separately in
each target.

This PR changes the rela_iplt symbols to have global binding and hidden
visibility. Previously, these properties were not uniform across targets:

- AArch64 used to use the binding and visibility of the IFunc symbol
  which triggered the creation of these symbols.
- ARM and RISCV already used to be GNU-compatible here with local
  binding and default visibility.
- x86-64 used local binding but with hidden visibillity.

Resolves #1342

### DIFF
--- a/lib/Target/AArch64/AArch64LDBackend.cpp
+++ b/lib/Target/AArch64/AArch64LDBackend.cpp
@@ -54,12 +54,10 @@ using namespace llvm;
 //===----------------------------------------------------------------------===//
 // AArch64LDBackend
 //===----------------------------------------------------------------------===//
-AArch64LDBackend::AArch64LDBackend(eld::Module &pModule,
-                                                 TargetInfo *pInfo)
+AArch64LDBackend::AArch64LDBackend(eld::Module &pModule, TargetInfo *pInfo)
     : GNULDBackend(pModule, pInfo), m_pErrata843419Factory(nullptr),
       m_pAArch64ErrataIslandFactory(nullptr), m_pRelocator(nullptr),
-      m_pDynamic(nullptr), m_pIRelativeStart(nullptr), m_pIRelativeEnd(nullptr),
-      m_ptdata(nullptr), m_ptbss(nullptr) {}
+      m_pDynamic(nullptr), m_ptdata(nullptr), m_ptbss(nullptr) {}
 
 AArch64LDBackend::~AArch64LDBackend() {}
 
@@ -520,54 +518,7 @@ void AArch64LDBackend::createErratum843419Stub(Fragment *frag,
   } // for all relocation section
 }
 
-void AArch64LDBackend::defineIRelativeRange(ResolveInfo &pSym) {
-  // It is up to linker script to define those symbols.
-  if (m_Module.getScript().linkerScriptHasSectionsCommand())
-    return;
-
-  // Define the copy symbol in the bss section and resolve it
-  auto SymbolName = "__rela_iplt_start";
-  if (!m_pIRelativeStart && !m_pIRelativeEnd) {
-    m_pIRelativeStart =
-        m_Module.getIRBuilder()
-            ->addSymbol<IRBuilder::Force, IRBuilder::Resolve>(
-                m_Module.getInternalInput(Module::Script), SymbolName,
-                ResolveInfo::Object, ResolveInfo::Define,
-                (ResolveInfo::Binding)pSym.binding(),
-                0,   // size
-                0x0, // value
-                FragmentRef::null(), (ResolveInfo::Visibility)pSym.other());
-    if (m_Module.getConfig().options().isSymbolTracingRequested() &&
-        m_Module.getConfig().options().traceSymbol(SymbolName)) {
-      config().raise(Diag::target_specific_symbol) << SymbolName;
-    }
-    m_pIRelativeStart->setShouldIgnore(false);
-    SymbolName = "__rela_iplt_end";
-    m_pIRelativeEnd =
-        m_Module.getIRBuilder()
-            ->addSymbol<IRBuilder::Force, IRBuilder::Resolve>(
-                m_Module.getInternalInput(Module::Script), SymbolName,
-                ResolveInfo::Object, ResolveInfo::Define,
-                (ResolveInfo::Binding)pSym.binding(),
-                pSym.size(), // size
-                0x0,         // value
-                FragmentRef::null(), (ResolveInfo::Visibility)pSym.other());
-    if (m_Module.getConfig().options().isSymbolTracingRequested() &&
-        m_Module.getConfig().options().traceSymbol(SymbolName)) {
-      config().raise(Diag::target_specific_symbol) << SymbolName;
-    }
-    m_pIRelativeEnd->setShouldIgnore(false);
-  }
-}
-
 bool AArch64LDBackend::finalizeTargetSymbols() {
-  if (m_pIRelativeStart && m_pIRelativeEnd) {
-    m_pIRelativeStart->setValue(
-        getRelaPLT()->getOutputSection()->getSection()->addr());
-    m_pIRelativeEnd->setValue(
-        getRelaPLT()->getOutputSection()->getSection()->addr() +
-        getRelaPLT()->getOutputSection()->getSection()->size());
-  }
   if (m_pGOTSymbol) {
     m_pGOTSymbol->setValue(getGOT()->getOutputSection()->getSection()->addr());
   }
--- a/lib/Target/ARM/ARMLDBackend.cpp
+++ b/lib/Target/ARM/ARMLDBackend.cpp
@@ -58,10 +58,9 @@ using namespace llvm;
 //===----------------------------------------------------------------------===//
 ARMGNULDBackend::ARMGNULDBackend(eld::Module &pModule, TargetInfo *pInfo)
     : GNULDBackend(pModule, pInfo), m_pRelocator(nullptr), m_pDynamic(nullptr),
-      m_pEXIDXStart(nullptr), m_pEXIDXEnd(nullptr), m_pIRelativeStart(nullptr),
-      m_pIRelativeEnd(nullptr), m_pEXIDX(nullptr), m_pRWPIBase(nullptr),
-      m_pSBRELSegment(nullptr), m_pARMAttributeSection(nullptr),
-      AttributeFragment(nullptr) {}
+      m_pEXIDXStart(nullptr), m_pEXIDXEnd(nullptr), m_pEXIDX(nullptr),
+      m_pRWPIBase(nullptr), m_pSBRELSegment(nullptr),
+      m_pARMAttributeSection(nullptr), AttributeFragment(nullptr) {}
 
 ARMGNULDBackend::~ARMGNULDBackend() {}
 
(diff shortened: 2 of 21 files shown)


lib/Target/AArch64/AArch64LDBackend.cpp
```diff
@@ -127,6 +125,9 @@ void AArch64LDBackend::initTargetSymbols() {
     if (m_pGOTSymbol)
       m_pGOTSymbol->setShouldIgnore(false);
   }
+
+  if (config().isCodeStatic())
+    defineIRelativeRange();
```
quic-seaswara: I would like this hook to be called by GNULDBackend since the entire support has moved to GNULDBackend.

lib/Target/AArch64/AArch64LDBackend.cpp
```diff
@@ -520,54 +521,8 @@ void AArch64LDBackend::createErratum843419Stub(Fragment *frag,
...
-
 bool AArch64LDBackend::finalizeTargetSymbols() {
-  if (m_pIRelativeStart && m_pIRelativeEnd) {
-    m_pIRelativeStart->setValue(
-        getRelaPLT()->getOutputSection()->getSection()->addr());
-    m_pIRelativeEnd->setValue(
-        getRelaPLT()->getOutputSection()->getSection()->addr() +
-        getRelaPLT()->getOutputSection()->getSection()->size());
-  }
+  finalizeIRelativeRange();
```
quic-seaswara: Same as previous comment, finalizeIRelativeRange would be nice to be called by finalizeTargetSymbols() in GNULDBackend

lib/Target/ARM/ARMLDBackend.cpp
```diff
@@ -138,6 +137,12 @@ void ARMGNULDBackend::initTargetSymbols() {
       m_pGOTSymbol->setShouldIgnore(false);
   }
 
+  // Define __rel_iplt_start/__rel_iplt_end (used by glibc to process
+  // R_ARM_IRELATIVE entries in static executables) if they are referenced.
+  // This must happen before scanRelocations creates GOT entries for them.
+  if (config().isCodeStatic())
```
quic-seaswara: Why only isCodeStatic()

lib/Target/X86/x86_64Relocator.cpp
```diff
@@ -270,6 +270,18 @@ void x86_64Relocator::scanLocalReloc(InputFile &pInputFile, Relocation &pReloc,
...
+  case llvm::ELF::R_X86_64_GOTPCREL:
+  case llvm::ELF::R_X86_64_GOTPCRELX:
+  case llvm::ELF::R_X86_64_REX_GOTPCRELX: {
+    // A GOT access to a local symbol (e.g. a linker-defined local symbol such
+    // as __rela_iplt_start) still needs a GOT entry.
+    std::lock_guard<std::mutex> relocGuard(m_RelocMutex);
+    if (rsym->reserved() & ReserveGOT)
+      return;
+    CreateGOT(Obj, pReloc, !config().isCodeStatic(), m_Target);
+    rsym->setReserved(rsym->reserved() | ReserveGOT);
```
quic-seaswara: This is the problem with changing the symbol from Global to Local.

We should make it remain Global, as the semantics change. 

What if there is a call to __rela_iplt_start, and then we need a PLT for local symbols too. 

We should preserve incoming types.

lib/Target/GNULDBackend.cpp
```diff
@@ -583,6 +583,64 @@ bool GNULDBackend::defineStandardAndSectionMagicSymbols() {
...
+  const char *EndName = IsRela ? "__rela_iplt_end" : "__rel_iplt_end";
+
+  m_pIRelativeStart =
+      m_Module.getIRBuilder()
+          ->addSymbol<IRBuilder::AsReferred, IRBuilder::Resolve>(
+              m_Module.getInternalInput(Module::Script), StartName,
+              ResolveInfo::NoType, ResolveInfo::Define, ResolveInfo::Local,
+              0,   // size
+              0x0, // value
+              FragmentRef::null(), ResolveInfo::Default);
```
quic-seaswara: ResolveInfo::Hidden

lib/Target/GNULDBackend.cpp
```diff
@@ -583,6 +583,64 @@ bool GNULDBackend::defineStandardAndSectionMagicSymbols() {
...
+  // match.
+  const bool IsRela = getRelocator()->relocType() == llvm::ELF::SHT_RELA;
+  const char *StartName = IsRela ? "__rela_iplt_start" : "__rel_iplt_start";
+  const char *EndName = IsRela ? "__rela_iplt_end" : "__rel_iplt_end";
+
+  m_pIRelativeStart =
+      m_Module.getIRBuilder()
+          ->addSymbol<IRBuilder::AsReferred, IRBuilder::Resolve>(
+              m_Module.getInternalInput(Module::Script), StartName,
+              ResolveInfo::NoType, ResolveInfo::Define, ResolveInfo::Local,
```
quic-seaswara: ResolveInfo::Default

lib/Target/GNULDBackend.cpp
```diff
@@ -583,6 +583,64 @@ bool GNULDBackend::defineStandardAndSectionMagicSymbols() {
...
+
+  if (m_pIRelativeStart || m_pIRelativeEnd)
+    return;
+
+  // The relocation section that holds R_*_IRELATIVE entries is `.rela.plt` on
+  // RELA targets and `.rel.plt` on REL targets; the range symbols are named to
+  // match.
+  const bool IsRela = getRelocator()->relocType() == llvm::ELF::SHT_RELA;
+  const char *StartName = IsRela ? "__rela_iplt_start" : "__rel_iplt_start";
+  const char *EndName = IsRela ? "__rela_iplt_end" : "__rel_iplt_end";
```
quic-seaswara: llvm:::StringRef

lib/Target/GNULDBackend.cpp
```diff
@@ -583,6 +583,64 @@ bool GNULDBackend::defineStandardAndSectionMagicSymbols() {
...
+  if (m_Module.getScript().linkerScriptHasSectionsCommand())
+    return;
+
+  if (m_pIRelativeStart || m_pIRelativeEnd)
+    return;
+
+  // The relocation section that holds R_*_IRELATIVE entries is `.rela.plt` on
+  // RELA targets and `.rel.plt` on REL targets; the range symbols are named to
+  // match.
+  const bool IsRela = getRelocator()->relocType() == llvm::ELF::SHT_RELA;
```
quic-seaswara, line 606: Can we have a backend function to return the symbol name, instead of checking in the core backend for RELA and REL.

Steven6798 (approved):
Nice refactoring.

quic-seaswara (changes requested):

lib/Target/X86/x86_64Relocator.cpp
```diff
@@ -270,6 +270,18 @@ void x86_64Relocator::scanLocalReloc(InputFile &pInputFile, Relocation &pReloc,
     if (config().isCodeIndep())
       checkDynamicRelocAllowed(pReloc, pSection, true);
     return;
+  case llvm::ELF::R_X86_64_GOTPCREL:
+  case llvm::ELF::R_X86_64_GOTPCRELX:
+  case llvm::ELF::R_X86_64_REX_GOTPCRELX: {
+    // A GOT access to a local symbol (e.g. a linker-defined local symbol such
```
quic-areg: This new section is doing something broader than the rest of the patch: we create a GOT entry for any local symbol accessed via GOTPCREL. not just the iplt range symbols, and we don't relax them. I think we should include some x86-specific tests with this patch, or handle x86 separately.

lib/Target/GNULDBackend.cpp
```diff
@@ -583,6 +583,64 @@ bool GNULDBackend::defineStandardAndSectionMagicSymbols() {
...
+  if (m_pIRelativeEnd) {
+    m_pIRelativeEnd->setShouldIgnore(false);
+    if (m_Module.getConfig().options().isSymbolTracingRequested() &&
+        m_Module.getConfig().options().traceSymbol(EndName))
+      config().raise(Diag::target_specific_symbol) << EndName;
+  }
+}
+
+void GNULDBackend::finalizeIRelativeRange() {
+  if (m_pIRelativeStart && m_pIRelativeEnd && getRelaPLT() &&
```
quic-areg, line 642: I don't think this could happen in practice, but what if only one of the symbols is referenced?

lib/Target/GNULDBackend.cpp
quic-areg: It looks like symbol binding/visibility changed from the per target definitions we had before. lld and bfd emit these as local symbols, with lld making them `HIDDEN` and bfd emitting them `DEFAULT`. We match bfd and that looks good to me, but I think the change should be mentioned in the commit message.

lib/Target/ARM/ARMLDBackend.cpp
parth-07: This patch is maintaining the previous behavior. We only created these symbols for static executables. Additionally, these symbols are only used with static executables and both bfd and lld do not provide these symbols for dynamic executables (and PIE).

lib/Target/X86/x86_64Relocator.cpp
parth-07: > This is the problem with changing the symbol from Global to Local.

> We should make it remain Global, as the semantics change.

This is not due to changing binding from Global to Local. The binding was always Local for `__rel(a)_iplt_(start|end)` symbols for all targets, including x86-64. This change is required because now the symbol is defined before the scan relocation phase. Previously, the symbol was undefined weak in scan relocation phase and hence it was handled in `scanGlobalReloc`, but now it is handled in `scanLocalReloc` because the symbol is now defined before the scan relocation phase and has local binding. However, `scanLocalReloc` lacks the support for handling `GOTPLCRELX` relocations. I will remove the x86-64 change from this patch and mark the failing test XFAIL for now.

> What if there is a call to __rela_iplt_start, and then we need a PLT for local symbols too.

It is invalid code to call to __rela_iplt_start/end symbols. These symbols point to a relocation. I don't think we need anything special in this case.

lib/Target/GNULDBackend.cpp
parth-07: We are matching GNU ld currently. GNU ld has `local` binding and `default` visibility for rela_iplt symbols.

lib/Target/GNULDBackend.cpp
parth-07: We are matching GNU ld currently. GNU ld has local binding and default visibility for rela_iplt symbols.

lib/Target/GNULDBackend.cpp
parth-07: Done.

lib/Target/GNULDBackend.cpp
parth-07: We can do this, but what is the benefit of adding two functions in each target for returning rela_ipt symbols when we can easily deduce them and it is used only at one place?

lib/Target/GNULDBackend.cpp
parth-07: I thought about this. I believe we should define the missing symbol in that case instead of only providing the symbol if both of them are undefined. However, it is an old behavior and this PR is simply preserving it. I plan to handle this separately.

lib/Target/X86/x86_64Relocator.cpp
parth-07: I agree. For now, I have removed the x86-64 change, and mark the test which are now failing as XFAIL. I will fix them in a follow-up patch.

lib/Target/AArch64/AArch64LDBackend.cpp
parth-07: Thank you for the suggestion. I have moved this code to GNULDBackend::initStandardSymbol and finalizeStandardSymbol. Now we only call these hooks in GNULDBackend instead of per target. I feel that StandardSymbol is a better fit than target symbol here because these symbols are not specific to a target.

lib/Target/AArch64/AArch64LDBackend.cpp
parth-07: Done.

lib/Target/GNULDBackend.cpp
parth-07: Totally agreed. I have updated the commit message with the details of how these properties are changed per target.

quic-areg (approved):
LGTM

lib/Target/GNULDBackend.cpp
quic-seaswara: Hidden visibility will accomplish the same and let us keep symbol binding to be Global.

We can choose not to model GNU here, and it is ok in this case.

lib/Target/GNULDBackend.cpp
quic-seaswara: Linker core would be much cleaner.

lib/Target/GNULDBackend.cpp
parth-07: But why should we use Global binding here? Both lld and bfd use local binding for these symbols.

parth-07: I have changed the rela_iplt symbols to have global binding and default visibility.

lib/Target/GNULDBackend.cpp
parth-07: It is in the linker core, `GNULDBackend`...

parth-07: It is in the linker core, `GNULDBackend`. I don't see how duplication of this in each target would be beneficial.

lib/Target/GNULDBackend.cpp
parth-07: Sorry, hidden visibility**.

quic-seaswara (approved):

PR #1468, opened 2026-07-14
Squash sbss and bss in one section for picolibc H2 template

SuperChamp234:
This results in calculation of the bss & sbss size in a cleaner fashion, which also prevents any orphan sections from being placed in between sbss and bss sections.

### DIFF
--- a/templates/staticExecutable/static-executable-h2-picolibc.lcs.template
+++ b/templates/staticExecutable/static-executable-h2-picolibc.lcs.template
@@ -167,9 +167,9 @@ SECTIONS
     *(.sdata.hot .sdata.hot.* .gnu.linkonce.s.hot.*)
     *(.sdata .sdata.* .gnu.linkonce.s.*)
   }
-  .sbss           :
+  . = ALIGN (64);
+  .bss :
   {
-    __bss_start = .;
     PROVIDE (__sbss_start = .);
     PROVIDE (___sbss_start = .);
     *(.dynsbss)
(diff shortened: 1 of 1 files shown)


SuperChamp234:
CC @quic-k

templates/staticExecutable/static-executable-h2-picolibc.lcs.template
```diff
@@ -167,9 +167,9 @@ SECTIONS
     *(.sdata.hot .sdata.hot.* .gnu.linkonce.s.hot.*)
     *(.sdata .sdata.* .gnu.linkonce.s.*)
   }
-  .sbss           :
+  . = ALIGN (64);
```
quic-k, line 170: why do we need to align it now?

quic-k (approved):
nit

templates/staticExecutable/static-executable-h2-picolibc.lcs.template
```diff
@@ -167,9 +167,9 @@ SECTIONS
...
+  .bss (NOLOAD)   :
```
parth-07: Previously, we did not need to explicitly specify `NOLOAD`. Why do we need it now?

SuperChamp234: I'm not sure exactly how it was automatically inferred by ELD. However, specifying as NOBITS should be cleaner, right?

templates/staticExecutable/static-executable-h2-picolibc.lcs.template
SuperChamp234: Previously bss section was aligned for 64-byte boundary so this does the same. 

I guess this might be better?

```
.bss (NOLOAD) :
{
  __bss_start = .;
  PROVIDE(__sbss_start = .);

  *(.sbss*)
  *(.scommon*)

  PROVIDE(__sbss_end = .);

  . = ALIGN(64);

  *(.bss*)
  *(COMMON)

  __bss_end = .;
}
```

SuperChamp234: I've resolved this by flat out aligning entire bss section by 64 bytes.

SuperChamp234:
@quic-k and @parth-07 please have a look and let me know if anything else is needed

templates/staticExecutable/static-executable-h2-picolibc.lcs.template
parth-07: `(NOLOAD)` is generally used for non-bss sections. bss sections already has the same semantics as `(NOLOAD)`. I do not think it is cleaner because it makes the linker script reader think why was the `(NOLOAD)` needed here.

SuperChamp234: Got it, I've made this change.

parth-07: @SuperChamp234 Can you please remove `(NOLOAD)` from here as well?

SuperChamp234: Done

parth-07 (approved):
LGTM.

quic-k (approved):

quic-seaswara (approved):

PR #1471, opened 2026-07-14
Add support for /dev/null input file

### LINKED ISSUE #1464: Add support for /dev/null input file
## Expected Behavior

Passing `/dev/null` as the input file should be treated as no input file.

### Current Behavior
$ ld.eld /dev/null
Note: Input file /dev/null has no contents
Error: target emulation unknown: -m or at least one .o file required
Fatal: Linking had errors (a.out)

### Expectec Behavior
$ ld.eld /dev/null
Error: target emulation unknown: -m or at least one .o file required
Fatal: Linking had errors (a.out)

Steven6798:
Fix: qualcomm#1464

### DIFF
--- a/lib/Input/InputAction.cpp
+++ b/lib/Input/InputAction.cpp
@@ -19,6 +19,7 @@
 #include "eld/Input/SearchDirs.h"
 #include "eld/Support/FileSystem.h"
 #include "eld/Support/MsgHandling.h"
+#include "eld/Support/Utils.h"
 #include <fstream>
 
 using namespace eld;
@@ -44,6 +45,10 @@ InputFileAction::InputFileAction(std::string Name,
     : InputAction(K, Printer), Name(Name), I(nullptr) {}
 
 bool InputFileAction::activate(InputBuilder &PBuilder) {
+  // Skip null file.
+  if (eld::utility::isNullDevice(Name))
+    return true;
+
   const LinkerConfig &Config = PBuilder.getLinkerConfig();
   std::string ExpandedName = Input::expandSysrootMarkers(
       Name, Config.directories(), *Config.getDiagEngine());
--- a/test/Common/standalone/devnull/devnull.test
+++ b/test/Common/standalone/devnull/devnull.test
@@ -1,2 +1,18 @@
+#UNSUPPORTED: aarch64
+#---devnull.test--------------------- Executable--------------------#
+#BEGIN_COMMENT
+#This tests that /dev/null is supported by the linker
+#END_COMMENT
+
+#Test output files
 RUN: %clang %clangopts -c %p/Inputs/1.c -o %t1.o
 RUN: %link %linkopts %t1.o -o /dev/null
+
+#Test input files. Should throw an error because no emulation can be determined.
+RUN: %not %link %linkopts /dev/null 2>&1 | %filecheck %s -check-prefix=NOINPUT
+#Test that the linker behaves normally when multiple files are linked
+#and /dev/null is included
+RUN: %link %linkopts /dev/null %t1.o
+
+#NOINPUT-NOT: Note: Input file /dev/null has no contents
+#NOINPUT: Error: target emulation unknown: -m or at least one .o file required
--- a/test/AArch64/standalone/devnull/devnull.test
+++ b/test/AArch64/standalone/devnull/devnull.test
@@ -1,3 +1,17 @@
-#UNSUPPORTED: windows
+#---devnull.test--------------------- Executable--------------------#
+#BEGIN_COMMENT
+#This tests that /dev/null is supported by the linker
+#END_COMMENT
+
+#Test output files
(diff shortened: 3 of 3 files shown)


lib/LinkerWrapper/GnuLdDriver.cpp
```diff
@@ -1497,9 +1497,11 @@ bool GnuLdDriver::createInputActions(llvm::opt::InputArgList &Args,
     } break;
 
     case T::INPUT: {
-      actions.push_back(eld::make<eld::InputFileAction>(arg->getValue(),
-                                                        Config.getPrinter()));
-      ++input_num;
+      if (!eld::utility::isNullDevice(arg->getSpelling().str())) {
+        actions.push_back(eld::make<eld::InputFileAction>(arg->getValue(),
```
quic-seaswara: Can this be moved to InputFileAction instead of GnuLdDriver ?

quic-seaswara: How does reproduce functionality work with this ? What does map file show ?

quic-seaswara (changes requested):

lib/LinkerWrapper/GnuLdDriver.cpp
Steven6798: I'll address those separately.

Steven6798: I'll look into that.

Steven6798: Calling InputFileAction when there is no file feels off. I think the driver should not call it if there is no input file. This matches the current behavior when no file is present. What do you think?

quic-seaswara: Driver should be very simple. We have InputFileAction::activate which can skip this file from being added to the input tree.

Steven6798: Got it. Thanks!

Steven6798: First follow up: https://github.com/qualcomm/eld/issues/1473

lib/Input/InputAction.cpp
```diff
@@ -44,6 +45,11 @@ InputFileAction::InputFileAction(std::string Name,
     : InputAction(K, Printer), Name(Name), I(nullptr) {}
 
 bool InputFileAction::activate(InputBuilder &PBuilder) {
+  if (eld::utility::isNullDevice(Name)) {
+    PBuilder.getDiagEngine()->raise(Diag::err_no_inputs) << Name;
```
quic-seaswara: You should not error out but skip the input, return true. 

ld.eld /dev/null 1.o 

would be a good test.

Steven6798: I missed that. Thanks!

Steven6798: I noticed that we currently throw the same error as lld so I will keep it that way. I will update the issue to reflect that.

Steven6798: I looked at lld and they allow linking /dev/null if you specify the emulation (-m somearch). I don't think this is the right approach. I think we should error out if there is no file regardless of any emulation specified. What do you think?

Steven6798: GNU ld allows linking even without the emulation. Not sure why. I'll look into this.

quic-areg (approved):

test/Common/standalone/devnull/devnull.test
```diff
@@ -1,2 +1,18 @@
+#UNSUPPORTED: aarch64
```
quic-seaswara, line 1: why unsupported for aarch64 ?

quic-seaswara (changes requested):

test/Common/standalone/devnull/devnull.test
Steven6798: The %link option for aarch64 includes the emulation. This is probably why the original test had this unsupported option and there was an existing devnull test just for aarch64

lib/Input/InputAction.cpp
```diff
@@ -44,6 +45,10 @@ InputFileAction::InputFileAction(std::string Name,
     : InputAction(K, Printer), Name(Name), I(nullptr) {}
 
 bool InputFileAction::activate(InputBuilder &PBuilder) {
+  // Skip null file.
+  if (eld::utility::isNullDevice(Name))
```
parth-07, line 49: For a linker, is there any difference in `/dev/null` and an empty file, assuming that `/dev/null` is provided by the runtime/OS?

Steven6798: Yes. both GNU ld and lld behave different in both cases.

```
$ ld.lld 
ld.lld: error: no input files
$ ld.lld /dev/null
ld.lld: error: target emulation unknown: -m or at least one .o file required
```

```
$ ld
ld: no input files
$ ld /dev/null
ld: warning: cannot find entry symbol _start; not setting start address

```

PR #1474, opened 2026-07-14
[RISCV] Fix unnecessary .got.plt section creation

deepakshirkem:
ELD was unnecessarily creating .got.plt section when only .got is required. The .got.plt section should only be created when PLT entries are needed.

Note: Similar fix is needed for AArch64, ARM, Hexagon and x86_64 backends and will be addressed in follow-up PRs.

Part of #1030

### DIFF
--- a/lib/Target/RISCV/RISCVLDBackend.cpp
+++ b/lib/Target/RISCV/RISCVLDBackend.cpp
@@ -2493,16 +2493,13 @@ RISCVGOT *RISCVLDBackend::createGOT(GOT::GOTType T, ELFObjectFile *Obj,
                                     ResolveInfo *R) {
 
   traceGOTCreation(T, R);
-  // If we are creating a GOT, always create a .got.plt.
-  if (!getGOTPLT()->hasFragments()) {
-    LDSymbol *Dynamic = m_Module.getNamePool().findSymbol("_DYNAMIC");
-    // TODO: This should be GOT0, not GOTPLT0.
-    RISCVGOT::CreateGOT0(getGOT(), Dynamic ? Dynamic->resolveInfo() : nullptr,
-                         config().targets().is32Bits());
-    RISCVGOT::CreateGOTPLT0(getGOTPLT(), nullptr,
-                            config().targets().is32Bits());
-  }
-
+  // Create GOT0 in .got when first GOT entry is needed.
+  if (auto *GOTSec = getGOT())
+    if (!GOTSec->hasFragments()) {
+      LDSymbol *Dynamic = m_Module.getNamePool().findSymbol("_DYNAMIC");
+      RISCVGOT::CreateGOT0(GOTSec, Dynamic ? Dynamic->resolveInfo() : nullptr,
+                           config().targets().is32Bits());
+    }
   RISCVGOT *G = nullptr;
   bool GOT = true;
   switch (T) {
@@ -2510,6 +2507,10 @@ RISCVGOT *RISCVLDBackend::createGOT(GOT::GOTType T, ELFObjectFile *Obj,
     G = RISCVGOT::Create(Obj->getGOT(), R, config().targets().is32Bits());
     break;
   case GOT::GOTPLT0:
+    // Create GOTPLT0 in .got.plt only when PLT is actually needed.
+    if (!getGOTPLT()->hasFragments())
+      RISCVGOT::CreateGOTPLT0(getGOTPLT(), nullptr,
+                              config().targets().is32Bits());
     G = llvm::dyn_cast<RISCVGOT>(*getGOTPLT()->getFragmentList().begin());
     GOT = false;
     break;
--- a/test/RISCV/standalone/UnnecessaryGOTPLT/UnnecessaryGOTPLT.test
+++ b/test/RISCV/standalone/UnnecessaryGOTPLT/UnnecessaryGOTPLT.test
@@ -1,0 +1,29 @@
+#---UnnecessaryGOTPLT.test--------------------------- Executable -----------------#
+#BEGIN_COMMENT
+# Test that ELD does not create unnecessary .got.plt section when only
+# .got section is required.
+# Also verify that .got.plt IS created when PLT entries are needed.
+#END_COMMENT
+#START_TEST
+
+# Test 1: Static executable with only GOT - should have .got but NOT .got.plt
+RUN: %clang %clangopts -target %linuxtarget -o %t1.o %p/Inputs/got-only.c -c -fPIC
+RUN: %link %linkopts -o %t1.static.out %t1.o
+RUN: %readelf -S %t1.static.out | %filecheck %s --check-prefix=STATIC --implicit-check-not=.got.plt
+
+# Test 2: Shared library with no function calls - should have .got but NOT .got.plt
+RUN: %link %linkopts -shared -o %t1.shared.out %t1.o
+RUN: %readelf -S %t1.shared.out | %filecheck %s --check-prefix=STATIC --implicit-check-not=.got.plt
+
+# Test 3: Shared library with external function call - should have both .got AND .got.plt
+RUN: %clang %clangopts -target %linuxtarget -o %t1.plt.o %p/Inputs/plt-needed.c -c -fPIC
+RUN: %clang %clangopts -target %linuxtarget -o %t1.bar.o %p/Inputs/bar.c -c -fPIC
+RUN: %link %linkopts -shared -o %t1.plt.shared.out %t1.plt.o %t1.bar.o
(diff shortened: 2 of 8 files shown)


test/RISCV/standalone/UnnecessaryGOTPLT/UnnecessaryGOTPLT.test
```diff
@@ -0,0 +1,37 @@
+UNSUPPORTED: riscv32
```
quic-seaswara: Why riscv32 unsupported ? 

Please add a build/run test to make sure that the change runs a test that you have mentioned.

quic-seaswara (changes requested):

test/RISCV/standalone/UnnecessaryGOTPLT/UnnecessaryGOTPLT.test
deepakshirkem: > Why riscv32 unsupported?

I originally thought this test would only work on `riscv64`, but I updated it and it now supports both 

> Please add a build/run test to make sure that the change runs a test that you have mentioned.

I believe Test 4 is the run test for this change. For now, I referred to PR #1157 while adding it. If you give me a referring test, I'd really appreciate it.

deepakshirkem:
@quic-seaswara Please take another cycle of review for this PR?

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -2466,23 +2466,24 @@ RISCVGOT *RISCVLDBackend::createGOT(GOT::GOTType T, ELFObjectFile *Obj,
...
-    LDSymbol *Dynamic = m_Module.getNamePool().findSymbol("_DYNAMIC");
-    // TODO: This should be GOT0, not GOTPLT0.
-    RISCVGOT::CreateGOT0(getGOT(), Dynamic ? Dynamic->resolveInfo() : nullptr,
-                         config().targets().is32Bits());
-    RISCVGOT::CreateGOTPLT0(getGOTPLT(), nullptr,
-                            config().targets().is32Bits());
-  }
-
+  // Create GOT0 in .got when first GOT entry is needed.
+  if (auto *GOTSec = getGOT())
```
parth-07: We are now adding `.got` entry even if it is not required, for example, when `createGOT` is called with `GOT::GOTPLT0`/`GOT::GOTPLTN`. Can you please change it such that `.got` is also only created when it is required?

parth-07 (changes requested):

lib/Target/RISCV/RISCVLDBackend.cpp
deepakshirkem: @parth-07, thank you. Addressed this in the latest commit. The GOT0 creation in .got is now guarded with T != GOT::GOTPLT0 && T != GOT::GOTPLTN condition, so it is only created when a real GOT entry is needed (e.g., Regular, TLS_IE, TLS_GD, etc.) and not for PLT-related GOT entries.

parth-07 (approved):
LGTM.

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -2466,23 +2466,27 @@ RISCVGOT *RISCVLDBackend::createGOT(GOT::GOTType T, ELFObjectFile *Obj,
...
     G = RISCVGOT::Create(Obj->getGOT(), R, config().targets().is32Bits());
     break;
   case GOT::GOTPLT0:
+    // Create GOTPLT0 in .got.plt only when PLT is actually needed.
+    if (!getGOTPLT()->hasFragments())
+      RISCVGOT::CreateGOTPLT0(getGOTPLT(), nullptr,
+                              config().targets().is32Bits());
     G = llvm::dyn_cast<RISCVGOT>(*getGOTPLT()->getFragmentList().begin());
     GOT = false;
     break;
```
quic-seaswara, line 2492: Thanks for the patch!, LGTM.

follow up : 

- add some tracing when GOTPLT0 and when GOTPLTN is created. 
- Is this something that needs to be done for other targets as well ?

quic-seaswara (approved):

lib/Target/RISCV/RISCVLDBackend.cpp
deepakshirkem: Thank you @quic-seaswara!. 

>add some tracing when GOTPLT0 and when GOTPLTN is created.

Will add tracing for GOTPLT0 and GOTPLTN creation as a follow-up.

>Is this something that needs to be done for other targets as well ?

Yes, similar fix is needed for AArch64, ARM, Hexagon and x86_64 backends and will be addressed in follow-up PRs.

deepakshirkem:
@parth-07 Can you run the CI when you get a chance?

deepakshirkem:
@quic-seaswara @quic-areg Please run the CI for me? I tried to check the failing tests but wasn’t able to verify them locally. Sorry for this regression.

deepakshirkem:
@quic-seaswara  Can you run the CI when you get a chance ?

deepakshirkem:
@parth-07, Need Help

We are updating riscv-tlsdesc.test and riscv-tlsdesc-relax.test to reflect the new GOT layout. We are unable to verify the RISC-V32 values locally since our llvm-mc doesn't support RISC-V32 TLSDESC instructions. Could you give me the exact RISC-V32 expected values for these tests?

Based on the CI output and our calculations, we believe the correct values are:

riscv-tlsdesc.test:

GD4: .got {{0*}}20 {{0*}}2880 DATA

riscv-tlsdesc-relax.test:

GD4: .got {{0*}}0c {{0*}}20c0 DATA
GD-REL4: {{0*}}20c4
lw a3, 0xc0(a2)
addi a0, a2, 0xc0
lw a5, 0xa4(a4)
addi a0, a4, 0xa4
lw a7, 0x92(a6)
addi a0, a6, 0x92

Could you please confirm if these values are correct? I don't want to run CI again with incorrect values and have it fail.

deepakshirkem:
@parth-07 Can you please run the CI when you get a chance ?

deepakshirkem:
[Shankar Easwaran (@quic-seaswara)](https://github.com/quic-seaswara) Can you run the CI when you get a chance ?

lib/Target/RISCV/RISCVLDBackend.cpp
```diff
@@ -2553,11 +2557,16 @@ RISCVPLT *RISCVLDBackend::createPLT(ELFObjectFile *Obj, ResolveInfo *R,
...
   if (!config().options().hasNow()) {
-    // For lazy binding, create GOTPLT0 and PLT0, if they don't exist.
+    // For lazy binding, create PLT0 if it doesn't exist.
     if (!getPLT()->hasFragments())
-      RISCVPLT::CreatePLT0(*this, createGOT(GOT::GOTPLT0, Obj, nullptr),
-                           getPLT(), is32Bits);
+      RISCVPLT::CreatePLT0(
+          *this,
+          llvm::dyn_cast<RISCVGOT>(*getGOTPLT()->getFragmentList().begin()),
+          getPLT(), is32Bits);
```
quic-seaswara, line 2569: The default mode is bind lazy, and it looks like we are creating a .got.plt anyway. 

Can we just do a !config().options().hasNow() in the original code ? 

What about if you a TLSDESC with -z now ? Do you need PLT0 ?

quic-seaswara (changes requested):

lib/Target/RISCV/RISCVLDBackend.cpp
deepakshirkem: @quic-seaswara 

>The default mode is bind lazy, and it looks like we are creating a .got.plt anyway.

Yes, we do create .got.plt in default lazy binding mode, but only when createPLT() is called. Before this patch, .got.plt was created unconditionally in createGOT() even for static executables with no PLT entries at all. Our fix moves GOTPLT0 creation to createPLT() so .got.plt is only created when PLT entries actually exist.

>Can we just do a !config().options().hasNow() in the original code ?

I think this is not sufficient. With -z now, hasNow() is true so the block is skipped and GOTPLT0 is never created, leaving .got.plt empty. The dynamic linker then crashes with a segfault. We verified this locally with riscv32 toolchain upstream/main segfaults with -z now + TLS, but this patch fix correctly outputs 42.

>What about if you a TLSDESC with -z now ? Do you need PLT0 ?

TLSDESC uses createGOT(GOT::TLS_DESC) which goes directly to .got and does not need PLT0. However, with -z now, .got.plt is merged into .got and GOTPLT0 slot is still needed for the dynamic linker. Without it, the dynamic linker crashes.

Is this analysis correct?

deepakshirkem:
@quic-seaswara Any feedback.

PR #1475, opened 2026-07-15
[X86] Fix RELATIVE relocation addend for merge-string data pointers

BhoumikPatidar:
### Problem

In a PIE, a pointer into a merged `SHF_MERGE|SHF_STRINGS` section (`.rodata.str1.1`) is emitted as an `R_X86_64_RELATIVE` dynamic relocation whose addend must be the string's final, post-merge address. Since relocation scanning runs before string merging, the dynamic relocation's target fragment and addend must be corrected afterwards, once the surviving string's real location is known.

`x86_64LDBackend` used to shadow the non-virtual `GNULDBackend::recordRelativeReloc()` with its own version writing into its own `m_RelativeRelocMap`, so x86_64's relative relocations never reached `GNULDBackend::doPostLayout()`'s shared fixup loop, which iterates `GNULDBackend::m_RelativeRelocMap`. That map stayed empty for x86_64, so the fixup never ran and merge-string RELATIVE addends kept their pre-merge input offset.

### Fix

Removes x86_64's shadow `recordRelativeReloc()`/`m_RelativeRelocMap` entirely, so x86_64's relative relocations flow through the shared `GNULDBackend::recordRelativeReloc()` into `GNULDBackend::m_RelativeRelocMap` like every other backend, and are corrected by `GNULDBackend::doPostLayout()` like every other backend.

`GNULDBackend::doPostLayout()`'s fixup loop now needs a `isMergeStr()` guard: once x86_64's relative relocations reach this shared map, it also holds x86_64's GOT-related RELATIVE relocations. Those are recorded with a deliberately different addend (0, since the writer computes the final value as just `S`) than the addend already on the original `R_X86_64_GOTPCREL(X)` relocation (`-4`, the RIP end-of-instruction bias for the 4-byte displacement field). Copying that `-4` onto the GOT slot's RELATIVE relocation unconditionally would corrupt it. The guard restricts the fixup to relocations whose target is an actual merge-string fragment, leaving GOT-related relocations on every backend untouched.

### DIFF
--- a/lib/Target/GNULDBackend.cpp
+++ b/lib/Target/GNULDBackend.cpp
@@ -4741,6 +4741,9 @@ void GNULDBackend::doPostLayout() {
     for (auto &r : m_RelativeRelocMap) {
       const Relocation *N = r.first;
       Relocation *R = r.second;
+      const FragmentRef *NFragRef = N->targetFragRef();
+      if (!NFragRef || !NFragRef->frag() || !NFragRef->frag()->isMergeStr())
+        continue;
       R->modifyRelocationFragmentRef(N->targetFragRef());
       R->setAddend(N->addend());
     }
--- a/lib/Target/X86/x86_64LDBackend.h
+++ b/lib/Target/X86/x86_64LDBackend.h
@@ -96,10 +96,6 @@ public:
 
   void doPreLayout() override;
 
-  void recordRelativeReloc(Relocation *DynRel, const Relocation *OrigRel) {
-    m_RelativeRelocMap[DynRel] = OrigRel;
-  }
-
   void sortRelocation(ELFSection &pSection) override;
 
   DynRelocType getDynRelocType(const Relocation *X) const override {
@@ -160,7 +156,6 @@ private:
   llvm::DenseMap<ResolveInfo *, x86_64GOT *> m_GOTMap;
   llvm::DenseMap<ResolveInfo *, x86_64GOT *> m_GOTPLTMap;
   llvm::DenseMap<ResolveInfo *, x86_64PLT *> m_PLTMap;
-  llvm::DenseMap<Relocation *, const Relocation *> m_RelativeRelocMap;
 };
 } // namespace eld
 
--- a/test/x86_64/RunTests/MergeStringRelativeReloc/Inputs/main.c
+++ b/test/x86_64/RunTests/MergeStringRelativeReloc/Inputs/main.c
@@ -1,0 +1,31 @@
+/* main.c -- compiled and linked SECOND.
+ *
+ * Its .rodata.str1.1 starts with "a"(2) + "c"(2) + "fF"(3) = 7 bytes, followed
+ * by the escape strings. Those first three strings are duplicates of pad.c's
+ * canonical copies, so they are excluded during string merging. As a result
+ * the first escape string "  \n", input offset 7 in main.o, shifts to output
+ * offset 0 within main.o's surviving fragment.
+ *
+ * The g_* pointers live in .data and, in a PIE, are emitted as
+ * R_X86_64_RELATIVE dynamic relocations whose addend is the escape string's
+ * final address. If the post-merge fixup does not run for x86_64, the addend
+ * keeps the pre-merge input offset (7) and points 7 bytes past "  \n", into
+ * the following string, so g_nl resolves to the wrong string at runtime. */
+#include <stdio.h>
+#include <string.h>
+
+const char *g_a = "a";
+const char *g_c = "c";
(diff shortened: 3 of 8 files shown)


test/x86_64/linux/MergeStringRelativeReloc/Inputs/main.c
```diff
@@ -0,0 +1,41 @@
...
+ * first escape string "  \n", input offset 7 in main.o, shifts to output
+ * offset 0 within main.o's surviving fragment.
+ *
+ * The g_* pointers live in .data and, in a PIE, are emitted as
+ * R_X86_64_RELATIVE dynamic relocations whose addend is the escape string's
+ * final address. If the post-merge fixup does not run for x86_64, the addend
+ * keeps the pre-merge input offset (7) and points 7 bytes past "  \n", into the
+ * following string, so g_nl resolves to the wrong string at runtime and the
+ * strcmp checks below fail. */
+#include <string.h>
```
quic-seaswara: Dont include string.h and make the test complex for a lit test. 

These headers can be used for run tests.

That said, please add a run test also.

quic-seaswara (changes requested):

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -135,6 +135,30 @@ void x86_64LDBackend::doPreLayout() {
...
+void x86_64LDBackend::doPostLayout() {
+  {
+    eld::RegisterTimer T("Update Relative Relocations", "Do Post Layout",
+                         m_Module.getConfig().options().printTimingStats());
+    for (auto &r : m_RelativeRelocMap) {
+      Relocation *R = r.first;
+      const Relocation *N = r.second;
+      const FragmentRef *NFragRef = N->targetFragRef();
+      if (!NFragRef || !NFragRef->frag() || !NFragRef->frag()->isMergeStr())
+        continue;
```
quic-seaswara: The only difference between the GNULDBackend function and this one is the check for MergeStr. Why do you need that ? 

Calling the doPostLayout function later will trigger the function to be called, that does not have this condition. 

I am confused.

BhoumikPatidar: The two loops iterate over different maps, not the same one twice.
`x86_64LDBackend` declares its own `m_RelativeRelocMap`, separate from `GNULDBackend::m_RelativeRelocMap`. All of x86_64's relative-reloc bookkeeping goes into the x86_64-local map. GNULDBackend::m_RelativeRelocMap is empty for x86_64.

So when `x86_64LDBackend::doPostLayout()` calls `GNULDBackend::doPostLayout()` afterward, that base-class loop iterates zero entries on x86_64. It's a no-op for relocations there, it only runs the other post-layout work which x86_64 still needs.

BhoumikPatidar: The `isMergeStr()` guard is needed because the x86_64-local map isn't merge-string-only. It also holds GOT-related RELATIVE relocations. Without the guard, the fixup loop would overwrite the correct GOT addend, corrupting it.

BhoumikPatidar: Before the fix this was the exact bug - 
`helper_DynRel_init` in `x86_64Relocator.cpp` called `recordRelativeReloc` which wrote the (dynamic reloc, original reloc) pairs into `x86_64LDBackend::m_RelativeRelocMap`. But `GNULDBackend::doPostLayout()`, the function that runs after merging and iterates these pairs to propagate the correct post-merge targetFragRef and addend onto each dynamic reloc was iterating `GNULDBackend::m_RelativeRelocMap`, a completely separate map that x86_64 never populated. So for x86_64, doPostLayout's fixup loop always iterated zero entries, the dynamic RELATIVE relocations were never updated.

Earlier fix -
The fix I initially did was to remove the x86_64 `recordRelativeReloc` and `RelativeRelocMap` so that the GNULDBackend version is populated and used in doPostLayout but this also required adding the guard in the common backend GNULDBackend's doPostLayout. 

Current fix - 
But as we discussed on teams to move this to x86_64 target, we now keep the x86_64 `recordRelativeReloc` and `RelativeRelocMap` and add `x86_64LDBackend::doPostLayout()` to use this correctly populated map.

quic-seaswara: Remove the RelativeRelocMap from X86_64 and use GNULDBackend.

quic-seaswara (approved):

quic-areg (commented):
LGTM

quic-areg (approved):

PR #1493, opened 2026-07-16
Address FIXME in FragmentRef.h

### LINKED ISSUE #1480: Address FIXME and TODO's in FragmentRef.h

Steven6798:
Offset now use SizeTraits<64> instead of uint64_t. Created SizeTraits.h file to reuse similar options that rely on bitwidth.

Fix: qualcomm#1480

### DIFF
--- a/include/eld/Fragment/FragmentRef.h
+++ b/include/eld/Fragment/FragmentRef.h
@@ -15,6 +15,7 @@
 #define ELD_FRAGMENT_FRAGMENTREF_H
 
 #include "eld/Config/Config.h"
+#include "eld/Support/SizeTraits.h"
 #include "llvm/Support/DataTypes.h"
 
 namespace eld {
@@ -32,7 +33,7 @@ class OutputSectionEntry;
  */
 class FragmentRef {
 public:
-  typedef uint64_t Offset; // FIXME: use SizeTraits<T>::Offset
+  typedef SizeTraits<64>::Offset Offset;
 
 public:
   static FragmentRef *null();
--- a/include/eld/Support/SizeTraits.h
+++ b/include/eld/Support/SizeTraits.h
@@ -1,0 +1,28 @@
+//===- SizeTraits.h--------------------------------------------------------===//
+// Part of the eld Project, under the BSD License
+// See https://github.com/qualcomm/eld/LICENSE.txt for license information.
+// SPDX-License-Identifier: BSD-3-Clause
+//===----------------------------------------------------------------------===//
+#ifndef ELD_SUPPORT_SIZETRAITS_H
+#define ELD_SUPPORT_SIZETRAITS_H
+
+#include "llvm/Support/DataTypes.h"
+
+namespace eld {
+
+/// SizeTraits - maps a target bit class (32 or 64, as returned by
+/// TargetOptions::bitclass()) to the integer types wide enough to hold an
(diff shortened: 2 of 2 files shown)


quic-areg:
Typo in PR title `FragmentedRef.h`

Steven6798:
> Typo in PR title `FragmentedRef.h`

Thanks!

include/eld/Support/SizeTraits.h
```diff
@@ -0,0 +1,28 @@
+//===- SizeTraits.h--------------------------------------------------------===//
```
quic-areg, line 1: Is there nothing existing in LLVM Support for this that we could reuse?

include/eld/Fragment/FragmentRef.h
```diff
@@ -32,7 +33,7 @@ class OutputSectionEntry;
  */
 class FragmentRef {
 public:
-  typedef uint64_t Offset; // FIXME: use SizeTraits<T>::Offset
+  typedef SizeTraits<64>::Offset Offset;
```
quic-areg, line 36: Why is this better? I think the `FIXME` intended for FragmentRef to be templated and hold a 32 or 64 bit type to potentially save size, but isn't this still a 64 bit type?

include/eld/Support/SizeTraits.h
Steven6798: I'll look into this. Thanks!

include/eld/Fragment/FragmentRef.h
Steven6798: If this is the case, then the class would need to be templated too, right? that would require more changes, but I can look into that.

Just making sure I understand this correctly.

quic-areg: Right, to actually save size `FragmentRef` itself would need to be templated. That's a bigger change, and I don't think it's worth doing right now when there are higher value changes for reducing memory usage.

Without templating `FragmentRef` this patch is effectively a no-op (that potentially introduces a new class we have to maintain) . I think it's pretty common  across eld and lld to just use `uint64_t` for both 32 and 64 bit, and I don't think the churn caused by templating `FragmentRef` is worth it. The best way to resolve the FIXME imo is to just remove it. @quic-seaswara @parth-07 thoughts?

Steven6798: That's what I initially noticed. I went with this solution to prevent making so many changes, but you are right. My current solution just over-complicates the initial problem without any meaningful change.

Thanks for the help!

quic-seaswara: > Right, to actually save size `FragmentRef` itself would need to be templated. That's a bigger change, and I don't think it's worth doing right now when there are higher value changes for reducing memory usage.
> 
> Without templating `FragmentRef` this patch is effectively a no-op (that potentially introduces a new class we have to maintain) . I think it's pretty common across eld and lld to just use `uint64_t` for both 32 and 64 bit, and I don't think the churn caused by templating `FragmentRef` is worth it. The best way to resolve the FIXME imo is to just remove it. Shankar Easwaran (@quic-seaswara) Parth (@parth-07) thoughts?

I agree.

parth-07 (changes requested):
Please improve the PR title and the commit title to describe the change.

PR #1519, opened 2026-07-18
Redesign --print-timing-stats as a hierarchical report

quic-seaswara:
Replace the flat per-group LLVM timer output with a custom report that reflects the actual linker pipeline. Timers are now grouped under six top-level phases in execution order/categories:

  Initialize -> Read -> Resolve -> Layout -> Emit -> Plugins

Each phase shows total wall, user, and system time plus its percentage of the total link time. Sub-timers appear in linker flow order (first seen) under their phase so the report reads as a trace of the pipeline. The name column width is computed from the longest timer name present so no names are truncated.

Previously, --print-timing-stats emitted one flat table per TimerGroup using LLVM's default printer. There were ~25 group names with no relationship to linker phases, numbers in only one column (wall time), and no sense of which phase dominated the link.

### DIFF
--- a/include/eld/Config/GeneralOptions.h
+++ b/include/eld/Config/GeneralOptions.h
@@ -878,7 +878,10 @@ public:
   // --print-stats
   bool printTimingStats(const char *TimeRegion = nullptr) const;
 
-  void setPrintTimingStats() { BPrintTimeStats = true; }
+  void setPrintTimingStats() {
+    BPrintTimeStats = true;
+    BPrintAllUserPluginTimeStats = true;
+  }
 
   bool allUserPluginStatsRequested() const {
     return BPrintAllUserPluginTimeStats;
--- a/lib/Support/RegisterTimer.cpp
+++ b/lib/Support/RegisterTimer.cpp
@@ -6,10 +6,18 @@
 
 #include "eld/Support/RegisterTimer.h"
 #include "llvm/ADT/StringMap.h"
+#include "llvm/ADT/StringRef.h"
+#include "llvm/Support/Format.h"
 #include <memory>
 #include <mutex>
+#include <vector>
 
 namespace {
+
+// Pipeline order for the hierarchical report.
+static const llvm::StringRef PipelineOrder[] = {
+    "Initialize", "Read", "Resolve", "Layout", "Emit", "Plugins",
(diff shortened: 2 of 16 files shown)


include/eld/Config/GeneralOptions.h
```diff
@@ -878,7 +878,10 @@ class GeneralOptions {
   // --print-stats
   bool printTimingStats(const char *TimeRegion = nullptr) const;
 
-  void setPrintTimingStats() { BPrintTimeStats = true; }
+  void setPrintTimingStats() {
+    BPrintTimeStats = true;
+    BPrintAllUserPluginTimeStats = true;
```
parth-07, line 883: We are now enabling the printing of plugin time stats without explicit request using `--time-region`. Can we please document this change in the commit-message?

lib/Support/RegisterTimer.cpp
```diff
@@ -36,10 +47,54 @@ class RegisterTimerRegistry {
...
+    double Wall = 0.0;
+    double User = 0.0;
+    double Sys = 0.0;
+    std::vector<TimerData> Timers; // in first-seen (linker flow) order
+  };
+
+  // Snapshot all timer data under a single lock and return phases in pipeline
+  // order with timers in first-seen order. Callers must not hold the mutex.
+  std::vector<PhaseData> snapshot() {
+    std::lock_guard<std::mutex> Lock(Mu);
```
parth-07, line 68: Why do we need mutex here? `snapshot()` only seems to have one caller `RegisterTimer::printTimingReport` and we should not be having concurrent instances of this function call?

parth-07 (commented):
Thank you for this change. The new timing reports are significantly more readable. 

It seems that some register timers are now getting ignored because now we only print
the timers whose phase is one of:

```cpp
// Pipeline order for the hierarchical report.
static const llvm::StringRef PipelineOrder[] = {
    "Initialize", "Read", "Resolve", "Layout", "Emit", "Plugins",
};
```

For example, the below timer now does not contribute to the timing report:

```cpp
eld::RegisterTimer T("Allocate Common Symbols",
                       "Hexagon Allocate Common Symbols",
                       m_Module.getConfig().options().printTimingStats());
```

Issue #513, opened 2025-10-31
Incorrect linker script symbol value due to incorrect order of parsing/reading linker scripts


parth-07:
We read linker scripts provided using `-T <script>` at the time of activating inputs, but read other non-`T` linker scripts later when reading all inputs. This discrepancy in the order of reading linker scripts can introduce cases where linker script assignments behave incorrectly.

Reproducer:

```bash
cat > 1.c << '!'
int foo() { return 1; }
!

cat > script.t << '!'
SECTIONS {
  u = v;
}
v = 0x500;
!

cat >another_script.t <<\EOF
v = 0x200;
EOF

LDs=(ld.eld)
SFs=(eld)

clang -o 1.o --target=aarch64-unknown-elf 1.c -c -ffunction-sections
clang -o 2.o --target=aarch64-unknown-elf 2.s -c 

for i in "${!SFs[@]}"; do
  ${LDs[$i]} -o 1.${SFs[$i]}.out 1.o another_script.t -T script.t -Map 1.${SFs[$i]}.map.txt
done

llvm-readelf -s 1.eld.out | grep -i 'ABS u'
```

**Output**
```
 grep -i 'ABS u'
     6: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT   ABS u
```

Correct value of `u` should be 0x200.

PR #1523, opened 2026-07-20
Reland "Fix evaluation order of (before/in/after)-SECTIONS assignments"

parth-07:
This PR relands #1203. #1203 was reverted because it fixed the issue #430, and fixing this issue causes a regression for a user. The root cause of the regression was an independent issue #513, however, the issue #430 was hiding this regression. This PR lands the fixes to both #430 and #513.


quic-seaswara (changes requested):
I would really want to change the design here for maintenance consideration.

- Each script command can be considered to be activated early / activated late. 
- Each script command will have a activation kind that can be set to activate early.
- The activate function that already exists can activate and set the activation flag, that is activated. 

This will benefit the linker map file to describe what commands were activated early, and what commands were activated late. 

Adding a new command to early activation is easy with this design. 

Can you please change the patch to incorporate this ?

parth-07:
> I would really want to change the design here for maintenance consideration.
>
> Each script command can be considered to be activated early / activated late.
> Each script command will have a activation kind that can be set to activate early.
> The activate function that already exists can activate and set the activation flag, that is activated.

Thank you for the comment. I have improved the code based on the offline discussion.

quic-seaswara (approved):

PR #1528, opened 2026-07-20
[x86] Implement R_X86_64_TLSGD GD→LE TLS relaxation

### LINKED ISSUE #1490: TLS GD→LE relaxation not implemented for x86
When it should fire
- Relocation: `R_X86_64_TLSGD` (type 19)
- Building an executable (static or dynamic, incl. `-pie`) and the symbol is
  non-preemptible / defined in the same binary. In a `-shared` output the symbol
  stays GD even with `-Bsymbolic`.
- Under `-static`, `__tls_get_addr()` is not in `libc.a`, so a non-relaxed GD
  reference fails to link.

Current behavior
Static link fails; the `R_X86_64_PLT32` against `__tls_get_addr` is left in place:
```
Error: tls_gd.o(.text+0x10): undefined reference to `__tls_get_addr'
Fatal: Linking had errors
```
In shared lib mode ELD keeps the GD sequence unchanged and emits
`R_X86_64_DTPMOD64` + `R_X86_64_DTPOFF64` plus a `JUMP_SLOT` for `__tls_get_addr`.

Expected behavior (lld reference)
```
Before (16 bytes):
  0x66
  leaq  sym@tlsgd(%rip), %rdi    ; 48 8d 3d [rel32]
  0x66 0x66
  rex64
  call  __tls_get_addr@PLT       ; e8 [rel32]

After (16 bytes):
  movq  %fs:0x0, %rax            ; 64 48 8b 04 25 00 00 00 00  (9 bytes)
  leaq  sym@tpoff(%rax), %rax    ; 48 8d 80 [imm32]            (7 bytes)
```
Relaxed result: 0 dynamic relocs, 0 `.got` bytes, 0 `__tls_get_addr` references.

What is eliminated
- 2-slot GD GOT pair (16 bytes in `.got`)
- The `R_X86_64_DTPMOD64` dynamic relocation in a PIC output (a static exe has
  none). GD→LE targets non-preemptible symbols, so the offset slot is a link-time
  constant, not a `DTPOFF64` dynamic reloc.
- `__tls_get_addr()` call 

Relocation change
`R_X86_64_TLSGD` (19) + `R_X86_64_PLT32` → no output relocation
(TP-relative offset is embedded as an `imm32` constant in the `leaq` instruction)

Reproducer
```bash
cat > tls_gd.c << 'EOF'
__thread int tls_var = 42;              /* same-binary, non-preemptible under -static */
EOF
clang -target x86_64-linux-gnu tls_gd.c -c -fPIC -ftls-model=global-dynamic -o tls_gd.o
ld.eld  -static -e get_tls -o tls_gd.eld tls_gd.o   # FAILS: undefined reference to __tls_get_addr
ld.lld  -static -e get_tls -o tls_gd.lld tls_gd.o   # Works: movq %fs:0 + leaq tpoff
```

Implementation note
Size-preserving (16 → 16 bytes). Can use `postProcessing` direct buffer write.
Unlike LD, the GD static path does not reach the `getTLSModuleID()` assert; it
fails later as an undefined reference to `__tls_get_addr`. The relaxation must be
added in the `R_X86_64_TLSGD` scan/apply paths in `x86_64Relocator.cpp`.

bsaharsh:
Implements GD→LE TLS relaxation for R_X86_64_TLSGD relocations.

  For non-preemptible symbols in executables, the 16-byte GD sequence is
  rewritten in-place to the LE form when:

  - The symbol is non-preemptible and the output is an executable (static,
    dynamic, or PIE)
  - The link is not partial (-r)

  GD→LE is mandatory and fires regardless of --relax, matching lld and
  bfd. Without this relaxation, executables fail with "undefined reference
  to __tls_get_addr".

  The companion call relocation at TLSGD_offset+8 may be either
  R_X86_64_PLT32 (standard) or R_X86_64_GOTPCRELX (-fno-plt). Both are
  handled.

  Note: This PR contains two commits. The first ([x86] Support
  R_X86_64_GOTPCRELX relaxation) is a dependency 
  Tests:

  - TLSGDRelax_StaticHidden
  - TLSGDRelax_DynExeHidden
  - TLSGDRelax_PIE_Hidden
  - TLSGDRelax_SharedPreemptible_NoRelax
  - TLSGDRelax_SharedHidden_NoRelax
  - TLSGDRelax_PartialRelink
  - TLSGDRelax_Trace
  - TLSGDRelax_NoPLT
  
Fixes #1490

### DIFF
--- a/include/eld/Diagnostics/DiagVerbose.inc
+++ b/include/eld/Diagnostics/DiagVerbose.inc
@@ -36,6 +36,8 @@ DIAG(relax_to_compress, DiagnosticEngine::Verbose,
      "%3 in section %4+0x%5 file %6")
 DIAG(trace_relax_gotpcrelx, DiagnosticEngine::Trace,
      "%0: relaxed GOTPCRELX (%1) for symbol '%2' to PC-relative access")
+DIAG(trace_relax_tlsgd_le, DiagnosticEngine::Trace,
+     "relaxed TLSGD for symbol '%0' to LE access in section %1+0x%2 file %3")
 DIAG(verbose_ehframe_remove_fde, DiagnosticEngine::Verbose, "EhFrame %0")
 DIAG(verbose_ehframe_read_fde, DiagnosticEngine::Verbose, "EhFrame %0")
 DIAG(verbose_ehframe_read_cie, DiagnosticEngine::Verbose, "EhFrame %0")
--- a/lib/Target/X86/x86_64LDBackend.cpp
+++ b/lib/Target/X86/x86_64LDBackend.cpp
@@ -151,7 +151,7 @@ bool x86_64LDBackend::isGOTPCRELXRelaxable(const Relocation *reloc) const {
 }
 
 bool x86_64LDBackend::shouldIgnoreRelocSync(Relocation *reloc) const {
-  return isGOTPCRELXRelaxCandidate(reloc);
+  return isGOTPCRELXRelaxCandidate(reloc) || isTLSGDRelaxCandidate(reloc);
 }
 
 eld::Expected<void>
@@ -160,7 +160,7 @@ x86_64LDBackend::postProcessing(llvm::FileOutputBuffer &pOutput) {
 
   // Relaxation rewrites bytes in the laid-out output image; it is not
   // applicable to partial links, which have no final layout.
-  if (config().options().getRelax() && !config().isLinkPartial())
+  if (!config().isLinkPartial())
     ELDEXP_RETURN_DIAGENTRY_IF_ERROR(doRelax(pOutput));
 
   return {};
@@ -168,21 +168,24 @@ x86_64LDBackend::postProcessing(llvm::FileOutputBuffer &pOutput) {
 
 eld::Expected<void> x86_64LDBackend::doRelax(llvm::FileOutputBuffer &pOutput) {
   uint8_t *buf = pOutput.getBufferStart();
-  // The scan phase already identified every relaxation candidate (skipping
-  // debug relocations and internal files, which never carry a relaxable
-  // relocation). Iterate that cached set instead of re-walking all
-  // relocations, and dispatch on the relocation type. Future relaxable types
-  // (e.g. R_X86_64_REX_GOTPCRELX, TLS relaxations) add a case here.
-  for (Relocation *reloc : m_GOTPCRELXRelaxCandidates) {
-    switch (reloc->type()) {
-    case llvm::ELF::R_X86_64_GOTPCRELX:
-      ELDEXP_RETURN_DIAGENTRY_IF_ERROR(relaxGOTPCRELXReloc(reloc, buf));
-      break;
-    default:
-      // Only relaxable types are recorded as candidates; skip anything else.
-      break;
+  // The scan phase already identified every relaxation candidate
+  // GOTPCRELX relaxation is optional (--relax); GD→LE is
+  // mandatory (otherwise we get an undefined reference for __tls_get_addr in
+  // static links).
+  if (config().options().getRelax()) {
+    for (Relocation *reloc : m_GOTPCRELXRelaxCandidates) {
+      switch (reloc->type()) {
+      case llvm::ELF::R_X86_64_GOTPCRELX:
+        ELDEXP_RETURN_DIAGENTRY_IF_ERROR(relaxGOTPCRELXReloc(reloc, buf));
+        break;
+      default:
+        // Only relaxable types are recorded as candidates; skip anything else.
+        break;
+      }
     }
   }
+  for (Relocation *reloc : m_TLSGDLERelaxCandidates)
+    ELDEXP_RETURN_DIAGENTRY_IF_ERROR(relaxTLSGDToLEReloc(reloc, buf));
   return {};
 }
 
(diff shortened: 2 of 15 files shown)


include/eld/Diagnostics/DiagVerbose.inc
```diff
@@ -36,6 +36,8 @@ DIAG(relax_to_compress, DiagnosticEngine::Verbose,
      "%3 in section %4+0x%5 file %6")
 DIAG(trace_relax_gotpcrelx, DiagnosticEngine::Trace,
      "%0: relaxed GOTPCRELX (%1) for symbol '%2' to PC-relative access")
+DIAG(trace_relax_tlsgd_le, DiagnosticEngine::Trace,
+     "%0: relaxed TLSGD for symbol '%1' to LE access")
```
parth-07: Can we please also add section information here as well? For reference, you can see how the section information is added for the diagnostic `relax_to_compress`.

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -152,7 +152,7 @@ bool x86_64LDBackend::isGOTPCRELXRelaxable(const Relocation *reloc) const {
 }
 
 bool x86_64LDBackend::shouldIgnoreRelocSync(Relocation *reloc) const {
-  return isGOTPCRELXRelaxCandidate(reloc);
+  return isGOTPCRELXRelaxCandidate(reloc) || isHandledByTLSGDLERelax(reloc);
```
parth-07: What do you think about renaming `isHandledByTLSGDLERelax` to `isTLSGDRelaxCandidate` for consistency?

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -161,29 +161,31 @@ x86_64LDBackend::postProcessing(llvm::FileOutputBuffer &pOutput) {
...
-    switch (reloc->type()) {
-    case llvm::ELF::R_X86_64_GOTPCRELX:
-      ELDEXP_RETURN_DIAGENTRY_IF_ERROR(relaxGOTPCRELXReloc(reloc, buf));
-      break;
-    default:
-      // Only relaxable types are recorded as candidates; skip anything else.
-      break;
+  // The scan phase already identified every relaxation candidate
+  // GOTPCRELX relaxation is optional (--relax); GD→LE is
+  // mandatory (undefined reference for __tls_get_addr static links).
```
parth-07:
```suggestion
  // mandatory (otherwise we get an undefined reference for __tls_get_addr in static links).
```

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -286,6 +288,61 @@ eld::Expected<void> x86_64LDBackend::relaxGOTPCRELXReloc(Relocation *reloc,
   return {};
 }
 
+eld::Expected<void> x86_64LDBackend::relaxTLSGDToLEReloc(Relocation *reloc,
+                                                         uint8_t *buf) {
+  auto *RF = llvm::dyn_cast<RegionFragment>(reloc->targetRef()->frag());
+  assert(RF && "TLSGD relax candidate must reference a RegionFragment");
```
parth-07:
```suggestion
  auto *RF = llvm::cast<RegionFragment>(reloc->targetRef()->frag());
```

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -286,6 +288,61 @@ eld::Expected<void> x86_64LDBackend::relaxGOTPCRELXReloc(Relocation *reloc,
...
+
+  uint32_t offset = reloc->targetRef()->offset();
+  // loc is at byte 4 of the 16-byte GD sequence; need at least 4 bytes before.
+  if (offset < 4)
```
parth-07: Do we check for this condition in the scanning phase when deciding if a TLSGD instruction pair is relaxable?

lib/Target/X86/x86_64LDBackend.cpp
```diff
@@ -286,6 +288,61 @@ eld::Expected<void> x86_64LDBackend::relaxGOTPCRELXReloc(Relocation *reloc,
...
+    config().raise(Diag::no_pt_tls_segment);
+    return {};
+  }
+  // Use finalizeTLSSymbol to get the TLS-segment-relative offset (S).
+  // Relocation::symValue() ASSERTs on TLS-type symbols; the correct path
+  // for TLS is through finalizeTLSSymbol(), mirroring what
+  // Relocator::getSymValue does for thread-local symbols.
+  uint64_t S = finalizeTLSSymbol(reloc->symInfo()->outSymbol());
+  int64_t A = static_cast<int64_t>(reloc->addend());
+  int64_t tpoff =
```
parth-07, line 314: The result of this computation is written as a signed 32-bit value. Do we need an overflow check here?

parth-07 (changes requested):

include/eld/Diagnostics/DiagVerbose.inc
bsaharsh: Have added the section information.

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: renamed

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: updated the comment

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: updated

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: Have added the check in the scanning phase

lib/Target/X86/x86_64LDBackend.cpp
bsaharsh: Currently lld, mold, or bfd doesn't check for the overflow, have created an issue referencing this.

parth-07 (approved):

bsaharsh:
Have fixed the clang format and added the fixes comment.

Steven6798 (approved):

quic-seaswara:
the changes here also have GOTPCRELX, now since GOTPCRELX is merged, please rebase

PR #1536, opened 2026-07-21
Add nightly workflow to test x86_64 backend against llvm-test-suite

BhoumikPatidar:
Add nightly workflow to test `x86_64` backend against `llvm-test-suite`

Runs `llvm-test-suite` nightly for `x86_64` backend, uses the toolchain built by `nightly-test.yml`.
`Bitcode`, `External`, `MicroBenchmarks`, `MultiSource`, `SingleSource` and `tools/test` directories are tested.
This is with the `TEST_SUITE_SUBDIRS` left unset, so the suite's own autodiscovery builds the
directories it can.

### DIFF
--- a/.github/workflows/nightly-llvm-test-suite.yml
+++ b/.github/workflows/nightly-llvm-test-suite.yml
@@ -1,0 +1,127 @@
+name: Nightly LLVM Test Suite (x86_64)
+
+on:
+  pull_request:
+    paths:
+      - '.github/workflows/nightly-llvm-test-suite.yml'
+  schedule:
+    - cron: '0 11 * * *'  # 5:00 AM CST
+  workflow_dispatch:
+
+
+permissions:
+  contents: write
+  actions: read
+jobs:
+  test-llvm-test-suite-x86_64:
+    if: github.repository == 'qualcomm/eld' || github.event_name == 'workflow_dispatch'
+    runs-on: ubuntu-latest
+    env:
+      ELD_SOURCE_DIR: ${{ github.workspace }}/llvm-project/llvm/tools/eld
+
+    steps:
+      - name: Checkout eld
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          path: llvm-project/llvm/tools/eld
+          persist-credentials: false
+
+      - name: Configure workflow environment variables
+        uses: ./llvm-project/llvm/tools/eld/.github/workflows/ConfigureBuildWorkflowENV
+
+      - name: Record pre-build entry
+        if: github.event_name == 'schedule'
+        uses: ./llvm-project/llvm/tools/eld/.github/workflows/BuildStatusDataRecorder
+        with:
+          enabled: true
+          project: "llvm-project/llvm/tools/eld"
+          stage: "pre-"
+          workflow-name: "llvm-test-suite"
+          record-mode: "record"
+          status: "fail"
+          run-id: ${{github.run_id}}
+          arch-name: "x86_64"
+          branch-name: "${{env.BUILD_BRANCH}}"
+
+      - name: Install dependencies
+        run: pip3 install lit
+
+      - name: Fetch nightly ELD
+        uses: ./llvm-project/llvm/tools/eld/.github/workflows/FetchNightlyToolset
+
+      - name: Clone LLVM test suite
+        run: |
+          git clone --depth 1 https://github.com/llvm/llvm-test-suite.git
+
+      - name: Configure test suite
+        run: |
+          ELD_BIN=$(which ld.eld)
+          LIB_DIR="$(dirname "$ELD_BIN")/../lib"
+          # TEST_SUITE_SUBDIRS is left unset, so llvm-test-suite/CMakeLists.txt
+          # runs its own auto-discovery.
+          # That resolves to:
+          #   Bitcode;External;MicroBenchmarks;MultiSource;SingleSource;tools/test
+          #
+          # What runs:
+          #   - Bitcode: runs. Needs -DELD_ENABLE_SYMBOL_VERSIONING=ON, which the
+          #     fetched nightly toolchain is built with.
+          #   - MicroBenchmarks, MultiSource, SingleSource, tools/test: run.
+          #
+          # What is discovered but contributes nothing on this runner:
+          #   - External: configures, but every benchmark under it is gated on
+          #     its own TEST_SUITE_<NAME>_ROOT pointing at licensed/proprietary
+          #     sources (e.g. TEST_SUITE_SPEC2000_ROOT) that are not supplied
+          #     here, so none of it builds.
+          #
+          # What auto-discovery drops before we ever see it:
+          #   - CTMark: removed by the suite itself; it is just symlinks into
+          #     MultiSource, so keeping it would double-build those tests.
+          #   - Fortran: removed because TEST_SUITE_FORTRAN defaults OFF unless
+          #     Fortran is already in TEST_SUITE_SUBDIRS, which it isn't since
+          #     we leave that variable unset.
+          #   - litsupport, tools: removed; these are the lit harness and its
+          #     support utilities, not tests (tools/test is added back above).
+          #
+          # Not part of the CMake-driven build:
+          #   ABI-Testsuite, ClangAnalyzer, LNTBased, litsupport-tests,
+          #   test-suite-externals, utils.
+
+          cmake -G Ninja \
+            -S llvm-test-suite \
+            -B build-llvm-test-suite \
+            -DCMAKE_C_COMPILER=$(which clang) \
+            -DCMAKE_CXX_COMPILER=$(which clang++) \
+            -DCMAKE_C_FLAGS="--ld-path=$ELD_BIN" \
+            -DCMAKE_CXX_FLAGS="--ld-path=$ELD_BIN -stdlib=libc++" \
+            -DCMAKE_EXE_LINKER_FLAGS="--ld-path=$ELD_BIN -stdlib=libc++ -Wl,-rpath,$LIB_DIR" \
+            -DTEST_SUITE_RUN_TYPE=nightly
+
+      - name: Build test suite
+        run: |
+          ninja -C build-llvm-test-suite -j$(nproc) -k 0
+
+      - name: Run tests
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-llvm-test-suite.yml
```diff
@@ -0,0 +1,116 @@
...
+          status: "fail"
+          run-id: ${{github.run_id}}
+          arch-name: "x86_64"
+          branch-name: "${{env.BUILD_BRANCH}}"
+
+      - name: Set up Clang 20
+        uses: egor-tensin/setup-clang@23bc15cd4207e45f1566447deb48f4ff3ef932cb #v2.3
+        with:
+          version: "20"
+          platform: x64
```
quic-seaswara: clang 22 to mimic the nightly builder.

.github/workflows/nightly-llvm-test-suite.yml
```diff
@@ -0,0 +1,116 @@
...
+
+      - name: Install dependencies
+        run: |
+          sudo apt update
+          sudo apt install -y libc++-dev libc++abi-dev \
+            libclang-rt-20-dev cmake ninja-build llvm-20 llvm-20-tools
+          pip3 install lit
+          echo "PATH=/usr/lib/llvm-20/bin:$PATH" >> $GITHUB_ENV
```
quic-seaswara: follow up : Can we make the nightly builder use a common action to install things, so that whatever LLVM/clang is used is at one place ?

.github/workflows/nightly-llvm-test-suite.yml
```diff
@@ -0,0 +1,116 @@
...
+            exit 1
+          fi
+
+      - name: Upload test output
+        if: always()
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a #v7.0.1
+        with:
+          name: llvm-test-suite-results
+          path: test-output.txt
+          retention-days: 2
```
quic-seaswara: retetion days 1?

.github/workflows/nightly-llvm-test-suite.yml
```diff
@@ -0,0 +1,116 @@
...
+            build-llvm-test-suite/SingleSource \
+            build-llvm-test-suite/MultiSource \
+            build-llvm-test-suite/MicroBenchmarks \
+            2>&1 | tee test-output.txt
+
+      - name: Validate test results
+        run: |
+          if grep -q '^FAIL: ' test-output.txt; then
+            grep '^FAIL: ' test-output.txt
+            exit 1
```
quic-seaswara: Do we need this ?

.github/workflows/nightly-llvm-test-suite.yml
```diff
@@ -0,0 +1,116 @@
...
+      - name: Configure test suite
+        run: |
+          ELD_BIN=$(which ld.eld)
+          cmake -G Ninja \
+            -S llvm-test-suite \
+            -B build-llvm-test-suite \
+            -DCMAKE_C_COMPILER=$(which clang) \
+            -DCMAKE_CXX_COMPILER=$(which clang++) \
+            -DCMAKE_C_FLAGS="-fuse-ld=$ELD_BIN" \
+            -DCMAKE_CXX_FLAGS="-fuse-ld=$ELD_BIN -stdlib=libc++" \
```
quic-seaswara: I think you need to use --ld-path. Not sure if -fuse-ld respects eld.

.github/workflows/nightly-llvm-test-suite.yml
```diff
@@ -0,0 +1,116 @@
...
+      - name: Build test suite
+        run: |
+          ninja -C build-llvm-test-suite -j$(nproc)
+
+      - name: Run tests
+        run: |
+          lit -j$(nproc) \
+            build-llvm-test-suite/SingleSource \
+            build-llvm-test-suite/MultiSource \
+            build-llvm-test-suite/MicroBenchmarks \
```
quic-seaswara: Add a comment on why other llvm test suites are not enabled. Add everything and comment it out.

.github/workflows/nightly-llvm-test-suite.yml
```diff
@@ -0,0 +1,116 @@
...
+          ELD_BIN=$(which ld.eld)
+          cmake -G Ninja \
+            -S llvm-test-suite \
+            -B build-llvm-test-suite \
+            -DCMAKE_C_COMPILER=$(which clang) \
+            -DCMAKE_CXX_COMPILER=$(which clang++) \
+            -DCMAKE_C_FLAGS="-fuse-ld=$ELD_BIN" \
+            -DCMAKE_CXX_FLAGS="-fuse-ld=$ELD_BIN -stdlib=libc++" \
+            -DCMAKE_EXE_LINKER_FLAGS="-fuse-ld=$ELD_BIN -stdlib=libc++" \
+            -DTEST_SUITE_SUBDIRS="SingleSource;MultiSource;MicroBenchmarks" \
```
quic-seaswara: Do you need this as you are saying lit to run specific folders ?

.github/workflows/nightly-llvm-test-suite.yml
```diff
@@ -0,0 +1,116 @@
+name: Nightly LLVM Test Suite (x86_64)
```
quic-seaswara, line 1: Follow up : enable for other targets as well.

.github/workflows/nightly-llvm-test-suite.yml
```diff
@@ -0,0 +1,116 @@
...
+
+on:
+  schedule:
+    - cron: '0 3 * * *'  # 9:00 PM CST
```
quic-seaswara: This needs to run after the nightly completes, now it overlaps I believe.

.github/workflows/nightly-llvm-test-suite.yml
```diff
@@ -0,0 +1,116 @@
...
+
+permissions:
+  contents: write
+  actions: read
+jobs:
+  test-llvm-test-suite-x86_64:
+    if: github.repository == 'qualcomm/eld' || github.event_name == 'workflow_dispatch'
+    runs-on: ubuntu-latest
+    env:
+      ELD_SOURCE_DIR: ${{ github.workspace }}/llvm-project/llvm/tools/eld
```
quic-seaswara, line 20: We can improve this to run on a PR as well, which would be great.

quic-seaswara (changes requested):

.github/workflows/nightly-llvm-test-suite.yml
BhoumikPatidar: okay, added to tasks

.github/workflows/nightly-llvm-test-suite.yml
BhoumikPatidar: Not needed, removed it.

.github/workflows/nightly-llvm-test-suite.yml
BhoumikPatidar: `-fuse-ld` does respect eld but emits warning `warning: '-fuse-ld=' taking a path is deprecated; use '--ld-path=' instead`.
I will switch to `--ld-path`.

.github/workflows/nightly-llvm-test-suite.yml
BhoumikPatidar: okay, added to tasks

.github/workflows/nightly-llvm-test-suite.yml
BhoumikPatidar: changed to 5:00 AM CST

.github/workflows/nightly-llvm-test-suite.yml
BhoumikPatidar: currently its needed because we are restricting the folders because of the BitCode failures. 
Without this CMake compiles/builds all default that we are not currently executing.
After the BitCode fix, we dont restrict folders so this wont be needed then.

parth-07 (approved):

Steven6798 (approved):
Great addition to ELD

.github/workflows/nightly-llvm-test-suite.yml
jonathonpenix: If the goal is just to always run it after the nightly, would it be worth using `workflow_run` (with `types` as `completed`) rather than scheduling it?

BhoumikPatidar: Yes, using workflow_run instead of scheduling should be better. This can also prevent cases where the nightly fails and the artifact expires and leads to these workflows failing. @parth-07 @quic-seaswara what do you suggest?

jonathonpenix: FWIW, since originaly suggesting this, we've had the zizmor checks turned on which will complain about `workflow_run` being insecure (we're seeing these errors in CPULLVM right now). So, maybe `workflow_run` wasn't such a good suggestion in hindsight 😅

quic-k: I think `workflow_run` is a good suggestion for jobs that depend upon nightly artifact for the cases @BhoumikPatidar pointed out
I am using it here : https://github.com/qualcomm/eld/pull/1611

who added zizmor? is it really needed?

jonathonpenix: My understanding from what happened with cpullvm is it was turned on at an organization level (github.com/qualcomm).

Again only based on what we saw in cpullvm, most of the things reported are/were legitimate issues and the fixes follow recommended practices so it's probably good from that perspective.

For workflow_run specifically, I didn't look into why it is/was considered insecure, so can't comment on that part.

.github/workflows/nightly-llvm-test-suite.yml
BhoumikPatidar: using clang built by nightly-test.yml now

.github/workflows/nightly-llvm-test-suite.yml
BhoumikPatidar: with the BitCode issue resolved, TEST_SUITE_SUBDIRS is left unset, so the suite's own autodiscovery builds the
directories it can.
Bitcode, External, MicroBenchmarks, MultiSource, SingleSource and tools/test directories are tested now.

quic-seaswara (approved):

PR #1542, opened 2026-07-21
[ELD] Intialize backend in ArchiveParser using ELF member

### LINKED ISSUE #1526: eld does not extract archive members when the archive is the first input file in the link command-line.
eld does not extract archive members when the archive is the first input file in the link command-line. The root cause is that archive parsing is skipped (and is not revisited later) if the backend is not initialized (due to unknown emulation before visiting an object file). 

Reproducer:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
int bar();

int foo() { return bar(); }
EOF

cat >2.c <<\EOF
int baz() { return 1; }
int bar() { return 3; }
EOF

cat >3.c <<\EOF
int car() { return 5; }
EOF

cat >script.t <<\EOF
ENTRY(bar);
EOF


TARGET="x86_64-linux-gnu"
llvm-ar cr lib2.a 2.o

LDs=(ld.eld ld.lld ld.bfd)
SFs=(eld lld bfd)

for i in "${!SFs[@]}"; do
  ${LDs[$i]} -o a.${SFs[$i]}.out lib2.a 1.o -T script.t -Map a.${SFs[$i]}.map.txt -e bar
done

for i in "${!SFs[@]}"; do
  ${LDs[$i]} -o a.${SFs[$i]}.out -e bar lib2.a 1.o -Map a.${SFs[$i]}.map.txt
done

for i in "${!SFs[@]}"; do
  ${LDs[$i]} -o a.${SFs[$i]}.out -u bar lib2.a 1.o -Map a.${SFs[$i]}.map.txt
done
```

Output:

```
+ ld.eld -o a.eld.out lib2.a 1.o -T script.t -Map a.eld.map.txt -e bar
Error: 1.o(.text.foo+0x7): undefined reference to `bar'
Fatal: Linking had errors.
+ for i in "${!SFs[@]}"
+ ld.lld -o a.lld.out lib2.a 1.o -T script.t -Map a.lld.map.txt -e bar
+ for i in "${!SFs[@]}"
+ ld.bfd -o a.bfd.out lib2.a 1.o -T script.t -Map a.bfd.map.txt -e bar
+ for i in "${!SFs[@]}"
+ ld.eld -o a.eld.out -e bar lib2.a 1.o -Map a.eld.map.txt
Error: 1.o(.text.foo+0x7): undefined reference to `bar'
Fatal: Linking had errors.
+ for i in "${!SFs[@]}"
+ ld.lld -o a.lld.out -e bar lib2.a 1.o -Map a.lld.map.txt
+ for i in "${!SFs[@]}"
+ ld.bfd -o a.bfd.out -e bar lib2.a 1.o -Map a.bfd.map.txt
+ for i in "${!SFs[@]}"
+ ld.eld -o a.eld.out -u bar lib2.a 1.o -Map a.eld.map.txt
Error: 1.o(.text.foo+0x7): undefined reference to `bar'
Fatal: Linking had errors.
+ for i in "${!SFs[@]}"
+ ld.lld -o a.lld.out -u bar lib2.a 1.o -Map a.lld.map.txt
+ for i in "${!SFs[@]}"
+ ld.bfd -o a.bfd.out -u bar lib2.a 1.o -Map a.bfd.map.txt
ld.bfd: warning: cannot find entry symbol _start; defaulting to 0000000000401000
```

Steven6798:
The ArchiveParser now initialize the backend based on target
information in ELF members.

Fix: qualcomm#1526

### DIFF
--- a/lib/Readers/ArchiveParser.cpp
+++ b/lib/Readers/ArchiveParser.cpp
@@ -103,9 +103,6 @@ eld::Expected<uint32_t> ArchiveParser::parseFile(InputFile &inputFile) const {
     return numObjects;
   }
 
-  if (!m_Module.isBackendInitialized())
-    return 0;
-
   if (!hasAFI) {
     auto expInitAF = initArchiveFile(archiveFile);
     DG_ELDEXP_REPORT_AND_RETURN_FALSE_IF_ERROR(*config.getDiagEngine(),
@@ -285,6 +282,14 @@ eld::Expected<bool> ArchiveParser::computeSymInfoTableForELFMember(
   ASSERT(ELFObjFile.isRelocatableObject(),
          "Only relocatable object files are supported as archive members!");
   using Elf_Sym = typename ELFT::Sym;
+
+  if (!m_Module.isBackendInitialized()) {
+    auto *ELFObj = llvm::dyn_cast<llvm::object::ELFObjectFileBase>(&ELFObjFile);
+    uint16_t machine = ELFObj->getEMachine();
+    bool is64Bit = ELFObj->is64Bit();
+    if (!m_Module.getLinker()->initializeTarget(machine, is64Bit))
+      return false;
+  }
 
   for (const llvm::object::ELFSymbolRef &sym : ELFObjFile.symbols()) {
     llvm::Expected<const Elf_Sym *> expRawSym =
--- a/test/Common/standalone/Sniffing/Sniff.test
+++ b/test/Common/standalone/Sniffing/Sniff.test
@@ -17,7 +17,7 @@ RUN: %ar cr %t.lib1.a %t1.1.o
 # whole-archive sniffing (ELF)
 RUN: %eld --whole-archive %t.lib1.a -o %t2.out --verbose 2>&1 | %filecheck %s -check-prefix=INFERRED
 # only archive sniffing
-RUN: %not %eld %t.lib1.a -o %t2.out 2>&1 | %filecheck %s -check-prefix=CANNOTINFER
+RUN: %eld %t.lib1.a -o %t2.out --verbose 2>&1 | %filecheck %s -check-prefix=INFERRED
 # Create archive (LTO)
 RUN: %ar cr %t.lib1.lto.a %t1.1.lto.o
 # whole-archive sniffing (ELF)


lib/Readers/ArchiveParser.cpp
```diff
@@ -41,6 +41,14 @@
 
 using namespace eld;
 
+static ResolveInfo::Desc getSymDesc(uint32_t shndx) {
+  if (shndx >= llvm::ELF::SHN_HEXAGON_SCOMMON &&
```
quic-seaswara: `ArchiveParser::computeSymInfoTableForELFMember` uses the object file to read symbols. So we can set the backend right ?

quic-seaswara (changes requested):

lib/Readers/ArchiveParser.cpp
Steven6798: I'll look into that and update the patch.

test/Common/standalone/Sniffing/Sniff.test
```diff
@@ -17,7 +17,7 @@ RUN: %ar cr %t.lib1.a %t1.1.o
 # whole-archive sniffing (ELF)
 RUN: %eld --whole-archive %t.lib1.a -o %t2.out --verbose 2>&1 | %filecheck %s -check-prefix=INFERRED
 # only archive sniffing
-RUN: %not %eld %t.lib1.a -o %t2.out 2>&1 | %filecheck %s -check-prefix=CANNOTINFER
+RUN: %eld %t.lib1.a -o %t2.out --verbose 2>&1 | %filecheck %s -check-prefix=INFERRED
```
quic-seaswara, line 20: follow up : support LTO bitcode archives.

quic-seaswara (approved):

test/Common/standalone/Sniffing/Sniff.test
quic-seaswara: can we also add %link instead of %eld ?

Steven6798: Sounds good. Thanks for the feedback

quic-areg (approved):

parth-07 (commented):
Can we please add a test where there is an archive file that has intermixed bitcode and native object files as members, and the first member is a bitcode file.

Steven6798:
> Can we please add a test where there is an archive file that has intermixed bitcode and native object files as members, and the first member is a bitcode file.

I can add that as a follow up. What should the expected behavior be? Still get the info from the ELF members?

PR #1546, opened 2026-07-21
[Zephyr] Fix PR workflow permission error

quic-areg:
The zephyr job granted `contents: read` but calls zephyr-run.yml which
declares `contents: write`.


quic-seaswara (approved):

quic-seaswara (commented):
follow up : Update ci.yml to not run when changes to zephyr-pr.yml

PR #1548, opened 2026-07-22
[LinkerScript] Support ENTRY command in the SECTIONS command

### LINKED ISSUE #1527: eld does not support `ENTRY` command in the `SECTIONS` command
GNU ld supports `ENTRY` command in the `SECTIONS` command. We do not. We should support this for GNU-compatibility.

reproducer:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
int bar();

int foo() { return bar(); }
EOF

cat >2.c <<\EOF
int baz() { return 1; }
int bar() { return 3; }
EOF

cat >3.c <<\EOF
int car() { return 5; }
EOF

cat >script.t <<\EOF
SECTIONS {
  ENTRY(bar)
}
EOF


TARGET="x86_64-linux-gnu"
llvm-ar cr lib2.a 2.o

LDs=(ld.eld ld.lld ld.bfd)
SFs=(eld lld bfd)

for i in "${!SFs[@]}"; do
  ${LDs[$i]} -o a.${SFs[$i]}.out lib2.a 1.o -T script.t -Map a.${SFs[$i]}.map.txt
done
```

Output:

```
+ for i in "${!SFs[@]}"
+ ld.eld -o a.eld.out lib2.a 1.o -T script.t -Map a.eld.map.txt
Error: script.t:3: : expected, but got }
>>> }
>>> ^
Fatal: Linker script script.t has errors, Please check
Fatal: Linking had errors.
+ for i in "${!SFs[@]}"
+ ld.lld -o a.lld.out lib2.a 1.o -T script.t -Map a.lld.map.txt
ld.lld: error: script.t:3: : expected, but got }
>>> }
>>> ^
+ for i in "${!SFs[@]}"
+ ld.bfd -o a.bfd.out lib2.a 1.o -T script.t -Map a.bfd.map.txt
```

Both eld and lld errors out, whereas bfd links fine.

Steven6798:
Add support for ENTRY command in the SECTIONS command.

Fix: qualcomm#1527

### DIFF
--- a/lib/ScriptParser/ScriptParser.cpp
+++ b/lib/ScriptParser/ScriptParser.cpp
@@ -561,6 +561,8 @@ void ScriptParser::readSections() {
     } else if (readAssignment(Tok)) {
     } else if (Tok == "OVERLAY") {
       readOverlay();
+    } else if (Tok == "ENTRY") {
+      readEntry();
     } else {
       readOutputSectionDescription(Tok);
     }
--- a/test/Common/LinkerScriptParsing/SECTIONS/Entry/Entry.test
+++ b/test/Common/LinkerScriptParsing/SECTIONS/Entry/Entry.test
@@ -1,0 +1,5 @@
+#---SimpleSections.test--------------------- Executable------------------#
+#BEGIN_COMMENT
+# This test verifies linker script parsing for an ENTRY command inside a SECTION
+#END_COMMENT
+RUN: %lsparserverifier %lsparserverifier_opts %p/Inputs/script.t
--- a/test/Common/LinkerScriptParsing/SECTIONS/Entry/Inputs/script.t
+++ b/test/Common/LinkerScriptParsing/SECTIONS/Entry/Inputs/script.t
@@ -1,0 +1,3 @@
+SECTIONS {
+  ENTRY(bar)
+}
--- a/test/Common/standalone/linkerscript/EntryCmdTopLevelOverridesInSections/EntryCmdTopLevelOverridesInSections.test
+++ b/test/Common/standalone/linkerscript/EntryCmdTopLevelOverridesInSections/EntryCmdTopLevelOverridesInSections.test
@@ -1,0 +1,14 @@
+#---EntryCmdTopLevelOverridesInSections.test---------- Executable,LS-----------#
+#BEGIN_COMMENT
+# This test checks that a top-level ENTRY command (outside SECTIONS) takes
+# precedence over an ENTRY command nested inside the SECTIONS command, since
+# the top-level command activates first.
+#END_COMMENT
+#START_TEST
+RUN: %clang %clangopts -c %p/Inputs/1.c -o %t1.o -ffunction-sections
+RUN: %link %linkopts %t1.o -T %p/Inputs/script.t -o %t2.out
+RUN: %readelf -h %t2.out | %filecheck %s
+# .bar is placed at 0x1000 and .foo right after; if ENTRY(foo) had won the
+# entry point would be .foo's address instead.
+#CHECK: Entry point address: 0x1000
+#END_TEST
--- a/test/Common/standalone/linkerscript/EntryCmdInSections/EntryCmdInSections.test
+++ b/test/Common/standalone/linkerscript/EntryCmdInSections/EntryCmdInSections.test
@@ -1,0 +1,15 @@
+#---EntryCmdInSections.test--------------------- Executable,LS-----------------#
+#BEGIN_COMMENT
+# This test checks that when ENTRY is specified inside the SECTIONS command,
+# the entry symbol's section is treated as a garbage-collection root and is
+# kept, while other unreferenced sections are still collected as usual.
+#END_COMMENT
+#START_TEST
+RUN: %clang %clangopts -c %p/Inputs/1.c -o %t1.o -ffunction-sections
+RUN: %link %linkopts %t1.o -T %p/Inputs/script.t -o %t2.out --gc-sections \
+RUN:   --print-gc-sections 2>&1 | %filecheck %s
+RUN: %readelf -h %t2.out | %filecheck %s --check-prefix=ENTRY
+#CHECK: GC : {{.*}}1.o[.text.unused]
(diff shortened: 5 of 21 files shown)


parth-07 (approved):

test/Common/LinkerScriptParsing/SECTIONS/Entry/Entry.test
```diff
@@ -0,0 +1,6 @@
+#---SimpleSections.test--------------------- Executable------------------#
+#BEGIN_COMMENT
+# This test verifies linker script parsing for an ENTRY command inside a SECTION
+#END_COMMENT
+RUN: %lsparserverifier %lsparserverifier_opts %p/Inputs/script.t
+
```
quic-seaswara: remove extra new line.

test/Common/LinkerScriptParsing/SECTIONS/Entry/Inputs/script.t
```diff
@@ -0,0 +1,3 @@
+SECTIONS {
+  ENTRY(bar)
+}
```
quic-seaswara, line 3: Test it with building with --gc-sections and entry symbol was kept.

lib/ScriptParser/ScriptParser.cpp
```diff
@@ -561,6 +561,8 @@ void ScriptParser::readSections() {
     } else if (readAssignment(Tok)) {
     } else if (Tok == "OVERLAY") {
       readOverlay();
+    } else if (Tok == "ENTRY") {
+      readEntry();
```
quic-seaswara, line 565: not correct. this test passes.

```
cat > 1.c << \!
int foo() { return 0; }
!

cat > script.t << \!
PHDRS {
  A PT_LOAD;
}

SECTIONS {
  .foo : { ENTRY(foo) *(.text.foo) } :A
}
!

ld.eld 1.o -T script.t -Map x
```

```
$ riscv-ld 1.o -T script.t
/opt/riscv/bin/riscv-ld: cannot find ENTRY: No such file or directory
```

quic-seaswara (changes requested):

lib/ScriptParser/ScriptParser.cpp
Steven6798: not sure i understand what you mean. Is this test supposed to pass or fail?

quic-seaswara: riscv-ld is the GNU linker. 

There is a secondary bug that ENTRY is parased incorrectly.

parth-07: Hi @quic-seaswara, This is not a bug. `ENTRY(foo)` fails for GNU ld because GNU ld attempts to open the file specified in an input section description if the file pattern does not contain any wildcard. In this case, GNU ld parses `ENTRY(foo)` as an input section description, it then attempts to open a file named `ENTRY` and fails because it cannot find such a file.

quic-seaswara: We parse ENTRY inside output section description.

parth-07: Both eld and bfd are parsing `ENTRY(foo)` as an input section description, where `ENTRY` is the file pattern and `foo` is the section pattern.

quic-seaswara: ah I see. We should provide some kind of warning with -Wlinker-script. Let us pick that as a follow up.

test/Common/standalone/linkerscript/EntryCmdTopLevelOverridesInSections/EntryCmdTopLevelOverridesInSections.test
```diff
@@ -0,0 +1,14 @@
+#---EntryCmdTopLevelOverridesInSections.test---------- Executable,LS-----------#
+#BEGIN_COMMENT
+# This test checks that a top-level ENTRY command (outside SECTIONS) takes
```
parth-07: @Steven6798 Would a top-level ENTRY command takes precedence even if it is after the SECTIONS command such as below?

```
SECTIONS {
  ENTRY(foo)
}
ENTRY(bar)
```

Steven6798: Currently it does not, I'll fix that. Thanks!

Steven6798: I updated GeneralOptions.h to track the origin of the entry. The priority is:

1. Command line
2. Top level ENTRY
3. ENTRY in SECTIONS

Let me know what you think of the tracking implementation.

parth-07: @Steven6798 Should the top-level ENTRY have higher precedence than `ENTRY` in SECTIONS? I only asked because the test explicitly mentioned this.

Steven6798: Ohhhh I see. I misunderstood. The wording I used might have been confusing. Let me double check against GNU ld

Steven6798: So, what I meant in the original description of the test was that the top-level ENTRY activates because it appears first. If the top-level entry is after SECTIONS, then no. I thought that your feedback was implying that it should also take precedence if it comes after. I confirmed that this is not the case by using GNU ld. I will undo these changes.

Steven6798: @parth-07 I'm seeing a discrepancy between eld and GNU ld outside of this patch. When there are multiple ENTRY commands, ELD chooses the first one while GNU ld chooses the last one.

parth-07 (approved):
LGTM.

quic-seaswara (approved):

PR #1552, opened 2026-07-22
[x86] Implement R_X86_64_TLSLD LD→LE TLS relaxation

### LINKED ISSUE #1510: TLS LD→LE relaxation not implemented in x86
When it should fire
- Relocation: `R_X86_64_TLSLD` (type 20)
- All TLS variables referenced are in the same binary (static link)
- `__tls_get_addr()` is not in `libc.a`, so static links fail today

Current behavior
Static link aborts with an assertion. The `R_X86_64_TLSLD` scan routes through
`getTLSModuleID()`, which cannot allocate a module-ID GOT entry without `__tls_get_addr`:
```
LLVM ERROR: Assertion `!isStatic` failed in .../x86_64Relocator.cpp line 237:
We always need to relax if -static because libc.a doesn't contain__tls_get_addr().
Relaxations are currently unsupported
```

In shared lib mode ELD keeps the LD sequence unchanged (emits `R_X86_64_DTPMOD64` and a `JUMP_SLOT` for `__tls_get_addr`).

Expected behavior (lld reference)
```
Before (12 bytes):
  leaq  sym@tlsld(%rip), %rdi    ; 48 8d 3d [rel32]  (7 bytes)
  callq __tls_get_addr@PLT       ; e8 [rel32]          (5 bytes)

After (12 bytes):
  0x66 0x66 0x66                 ; 3 padding prefixes 
  movq  %fs:0x0, %rax            ; 64 48 8b 04 25 00 00 00 00  (9 bytes)
```
Per-variable `@dtpoff` accesses below this are applied as embedded `leaq`
displacements; 0 `R_X86_64_DTPOFF32` dynamic relocations remain.

What is eliminated
- LD GOT pair (16 bytes in `.got`)
- The `R_X86_64_DTPMOD64` dynamic relocation in a PIC output (a static exe has none)
-  `__tls_get_addr()` call for the module base

Relocation change
`R_X86_64_TLSLD` (20) + `R_X86_64_PLT32` → no relocation (constant bytes, no fixup needed)

Reproducer
```bash
cat > tls_ld.c << 'EOF'
__thread int tls_var = 42;
EOF
clang -target x86_64-linux-gnu tls_ld.c -c -fPIC -ftls-model=local-dynamic -o tls_ld.o
ld.eld -static -e get_tls -o tls_ld.eld tls_ld.o   # CRASH (assertion)
ld.lld -static -e get_tls -o tls_ld.lld tls_ld.o   # Works: 0x66 0x66 0x66 + movq %fs:0
```

Implementation note
Size-preserving (12 → 12 bytes).

bsaharsh:
Implements LD→LE TLS relaxation for R_X86_64_TLSLD relocations.

  For non-preemptible local-dynamic TLS in executables, the LD sequence is
  rewritten in-place to the LE form when:

  - The output is an executable (static, dynamic, or PIE)
  - The link is not partial (-r)

  Both the standard PLT call variant (12 bytes) and the -fno-plt indirect
  call variant (13 bytes) are handled. R_X86_64_DTPOFF32/64 relocations
  that follow a relaxed TLSLD in the same fragment are automatically
  switched to the TP-relative formula.

  LD→LE is mandatory and fires regardless of --relax, matching lld and
  bfd. Without it, executables fail with "undefined reference to
  __tls_get_addr".

  Note: This PR contains two commits. The first ([x86] Support
  R_X86_64_GOTPCRELX relaxation) is a dependency.

  Tests:

  TLSLDRelax_Static
  TLSLDRelax_PIE
  TLSLDRelax_Dynamic
  TLSLDRelax_TwoVars
  TLSLDRelax_IndirectCall
  TLSLDRelax_NoPLT
  TLSLDRelax_DTPOFF
  TLSLDRelax_BadInput
  TLSLDRelax_Truncated
  TLSLDRelax_PartialRelink
  TLSLDRelax_SharedNoRelax
  
Fixes #1510

### DIFF
--- a/lib/Target/X86/x86_64Relocator.cpp
+++ b/lib/Target/X86/x86_64Relocator.cpp
@@ -136,7 +136,12 @@ void x86_64Relocator::scanRelocation(Relocation &pReloc,
   // symbol
   if (rsym->isUndef() || rsym->isBitCode()) {
     std::lock_guard<std::mutex> relocGuard(m_RelocMutex);
-    if (!m_Target.canProvideSymbol(rsym)) {
+    // A companion call reloc in an LD→LE relaxed sequence has its bytes
+    // overwritten by postProcessing; the symbol never needs to be resolved.
+    // isTLSLDCompanion identifies companions positionally (independent of
+    // reloc type or symbol name), so this suppression is precise.
+    if (!m_Target.isTLSLDCompanion(&pReloc) &&
+        !m_Target.canProvideSymbol(rsym)) {
       if (m_Target.canIssueUndef(rsym)) {
         if (rsym->visibility() != ResolveInfo::Default)
           issueInvisibleRef(pReloc, pInputFile);
@@ -236,10 +241,6 @@ x86_64GOT *x86_64Relocator::getTLSModuleID(ResolveInfo *R, bool isStatic) {
 
   G = m_Target.createGOT(GOT::TLS_LD, nullptr, nullptr);
 
-  ASSERT(!isStatic,
-         "We always need to relax if -static because libc.a doesn't "
-         "contain__tls_get_addr(). Relaxations are currently unsupported");
-
   if (!isStatic)
     helper_DynRel_init(m_Target.getDynamicSectionHeadersInputFile(), nullptr,
                        nullptr, G, 0x0, llvm::ELF::R_X86_64_DTPMOD64, m_Target);
@@ -253,6 +254,18 @@ void x86_64Relocator::scanLocalReloc(InputFile &pInputFile, Relocation &pReloc,
   ELFObjectFile *Obj = llvm::dyn_cast<ELFObjectFile>(&pInputFile);
   // rsym - The relocation target symbol
   ResolveInfo *rsym = pReloc.symInfo();
+  // The companion call reloc of a LD→LE relaxed sequence (PLT32/PC32 for the
+  // direct call, GOTPCRELX/GOTPCREL for the -fno-plt indirect call) has its
+  // bytes overwritten by postProcessing, so no GOT/PLT entry is needed for it.
+  // A single positional check here subsumes the per-type guards that would
+  // otherwise be scattered across the relevant cases. Locked because
+  // isTLSLDCompanion reads m_TLSLDFragOffsets, which other scan threads write
+  // under the same mutex.
+  {
+    std::lock_guard<std::mutex> relocGuard(m_RelocMutex);
+    if (m_Target.isTLSLDCompanion(&pReloc))
+      return;
+  }
   switch (pReloc.type()) {
   case llvm::ELF::R_X86_64_64: {
     if (config().isCodeIndep()) {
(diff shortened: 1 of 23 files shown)


lib/Target/X86/x86_64Relocator.cpp
```diff
@@ -253,6 +254,18 @@ void x86_64Relocator::scanLocalReloc(InputFile &pInputFile, Relocation &pReloc,
...
   ResolveInfo *rsym = pReloc.symInfo();
+  // The companion call reloc of a LD→LE relaxed sequence (PLT32/PC32 for the
+  // direct call, GOTPCRELX/GOTPCREL for the -fno-plt indirect call) has its
+  // bytes overwritten by postProcessing, so no GOT/PLT entry is needed for it.
+  // A single positional check here subsumes the per-type guards that would
+  // otherwise be scattered across the relevant cases. Locked because
+  // isTLSLDCompanion reads m_TLSLDFragOffsets, which other scan threads write
+  // under the same mutex.
+  {
+    std::lock_guard<std::mutex> relocGuard(m_RelocMutex);
```
parth-07, line 265: Why do we need mutex here?

parth-07 (commented):
Should we report an error if `R_X86_64_TLSLD` relocation is not followed by a `R_X86_64_GOTPCRELX` or `R_X86_64_PLT32` relocations?

```
#!/usr/bin/env bash
set -euo pipefail

cat > "tmp/tls_ld_malformed.s" <<'EOF'
	.text
	.globl bad
	.type bad,@function
bad:
	leaq	tls_var@tlsld(%rip), %rdi
	movl	tls_var@dtpoff(%rax), %eax
	retq

	.section .tbss,"awT",@nobits
tls_var:
	.zero 4
EOF

clang -target x86_64-linux-gnu -c "tmp/tls_ld_malformed.s" -o "tmp/tls_ld_malformed.o"

echo "ELD result:"
if ld.eld -m elf_x86_64 -static -e bad -o "tmp/a.eld" \
  "tmp/tls_ld_malformed.o"; then
  echo "BUG: ELD linked malformed TLSLD sequence successfully"
  llvm-objdump -d "tmp/a.eld" 
else
  echo "OK: ELD rejected malformed TLSLD sequence"
fi

echo
echo "LLD result:"
if ld.lld -m elf_x86_64 -static -e bad -o "tmp/a.lld" \
  "tmp/tls_ld_malformed.o"; then
  echo "LLD linked successfully"
else
  echo "LLD rejected malformed TLSLD sequence"
fi
```

This outputs:

```
ELD result:
BUG: ELD linked malformed TLSLD sequence successfully

tmp/a.eld:	file format elf64-x86-64

Disassembly of section .text:

00000000004000b0 <bad>:
  4000b0: 48 8d 3d 00 00 00 00         	leaq	(%rip), %rdi            # 0x4000b7 <bad+0x7>
  4000b7: 8b 80 00 00 00 00            	movl	(%rax), %eax
  4000bd: c3                           	retq

LLD result:
ld.lld: error: tmp/tls_ld_malformed.o:(.text+0x0): expected R_X86_64_PLT32 or R_X86_64_GOTPCRELX after R_X86_64_TLSLD
LLD rejected malformed TLSLD sequence
```

lib/Target/X86/x86_64Relocator.cpp
```diff
@@ -313,9 +326,25 @@ void x86_64Relocator::scanLocalReloc(InputFile &pInputFile, Relocation &pReloc,
...
     getTLSModuleID(pReloc.symInfo(), config().isCodeStatic());
     return;
   }
+  case llvm::ELF::R_X86_64_DTPOFF32:
+  case llvm::ELF::R_X86_64_DTPOFF64: {
+    // Tag DTPOFF relocs that follow a relaxed TLSLD in the same fragment so
+    // relocDTPOFF can apply the TP-relative formula only for those.
+    // Lock held for both the read of m_TLSLDFragOffsets and the insert.
+    std::lock_guard<std::mutex> relocGuard(m_RelocMutex);
+    if (m_Target.isDTPOFFInRelaxedLDSequence(&pReloc))
```
parth-07, line 344: We scan relocations parallelly and the check `isDTPOFFInRelaxedLDSequence` seems to depend upon whether we have already seen a `TLSLD` relocation in this fragment. Can this cause any issues?

PR #1556, opened 2026-07-22
[all] Accept weak hidden/protected undef symbols when building shared…

### LINKED ISSUE #1555: handle undefined weak hidden symbols when building shared objects
cat > wk.c <<  \!
__attribute__((weak)) __attribute__((visibility("hidden")))  extern int foo;
int bar() { return foo; }
!

clang -c -target riscv32 -fPIC wk.c
ld.eld wk.o -shared

This test produces an error about symbol foo when building a shared object.

Linker should not raise an error for this case.

quic-seaswara:
A weak undefined symbol with hidden or protected visibility is valid in any output type: the dynamic linker cannot preempt it so it always resolves to zero at runtime. The linker should accept it silently and leave the GOT entry at zero with no dynamic relocation.

Fixes #1555

### DIFF
--- a/lib/Target/AArch64/AArch64Relocator.cpp
+++ b/lib/Target/AArch64/AArch64Relocator.cpp
@@ -90,6 +90,11 @@ AArch64GOT &CreateGOT(ELFObjectFile *Obj, Relocation &pReloc, bool pHasRel,
     return *G;
   }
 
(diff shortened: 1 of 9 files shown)


lenary (approved):
LGTM. Thanks for fixing this quickly.

We will need this on the 23.x branch in due course.

PR #1578, opened 2026-07-24
[Zephyr] Reduce CI disk usage and quarantine rv32 FPU test

quic-areg:
We can sometimes hit "No space left on device" mid-setup since the runs sit right at the runner disk limit. Trim usage in zephyr-run.yml (shared by PR and nightly):
- west update: drop 30 vendor HALs no QEMU board we build uses (~4-5 GB saved).
- twister: pass -M pass to delete passing tests' build dirs as they run.

Also quarantine kernel.fpu_sharing.generic.riscv32, which forces hard-float rv32 that cpullvm picolibc has no multilib for.

The PR workflow should pass for all targets after this change.

### DIFF
--- a/.github/zephyr-quarantine.yml
+++ b/.github/zephyr-quarantine.yml
@@ -1,0 +1,9 @@
+# Twister quarantine list for Zephyr-on-ELD runs (PR and nightly), passed via
+# `west twister --quarantine-list`. Entries are skipped, not failed.
+
+- scenarios:
+  - kernel.fpu_sharing.generic.riscv32
+  comment: >-
+    Forces -mabi=ilp32d -march=rv32imafdc (hard double-float); cpullvm picolibc
+    has no matching multilib, so the build fails with 'string.h' not found.
+
--- a/.github/workflows/zephyr-run.yml
+++ b/.github/workflows/zephyr-run.yml
@@ -139,10 +139,17 @@ jobs:
           git apply eld-support.patch
 
       - name: West setup
+        shell: bash
         run: |
           cd zephyr
           west init -l .
-          west update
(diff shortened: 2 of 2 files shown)


quic-seaswara:
@jonathonpenix  can you comment on the rv32 FPU / CPULLVM toolchain ?

quic-seaswara (approved):

.github/zephyr-quarantine.yml
```diff
@@ -0,0 +1,9 @@
+# Twister quarantine list for Zephyr-on-ELD runs (PR and nightly), passed via
+# `west twister --quarantine-list`. Entries are skipped, not failed.
+
+- scenarios:
+  - kernel.fpu_sharing.generic.riscv32
+  comment: >-
+    Forces -mabi=ilp32d -march=rv32imafdc (hard double-float); cpullvm picolibc
```
jonathonpenix, line 7: FWIW, there's one in 23 when that releases

jonathonpenix (approved):
Not that familiar with these options, but seems reasonable to me.

Out of curiosity, how often are you hitting these issues?

We only run 1x per week in cpullvm but I don't think we've hit space issues recently

jonathonpenix:
> Jonathon Penix (@jonathonpenix) can you comment on the rv32 FPU / CPULLVM toolchain ?

Oops, left a comment but just to respond this directly: there's already one in cpullvm's mainline, should be in 23.x when that is out

quic-areg:
> Not that familiar with these options, but seems reasonable to me.
> 
> Out of curiosity, how often are you hitting these issues?
> 
> We only run 1x per week in cpullvm but I don't think we've hit space issues recently

Almost never. I think I've only seen it once.

PR #1585, opened 2026-07-25
Refactor: Replace custom getOrigin() method with existing getOldInput…

### LINKED ISSUE #1300: `CommonELFSection` should override `getOldInputFile`
Currently, we assign common symbols an internal input file and store the original input in `CommonELFSection::Origin`. We then use `CommonELFSection::getOrigin` for rule-matching, printing the filename in the diagnostics, and so on. Instead of creating a new member function in `CommonELFSection` and adding an if-condition on every path which requires the input filename of the section, we can reuse the already existing virtual function in the base class `ELFSection`, `getOldInputFile`, and use this instead for retrieving the original input file of the `CommonELFSection`s. Wen can (and should) reuse `getOldInputFile` because it follows the same/similar semantics as of `CommonELFSection::getOrigin`. For example, old input file must be used for rule-matching and printing the input file name in the diagnostics.

rachitmeht:
CommonELFSection had a custom getOrigin() method while other sections
used the virtual getOldInputFile(). This forced type-specific code:

    if (auto *CS = dyn_cast<CommonELFSection>(sect))
      input = CS->getOrigin();
    if (sect->getOldInputFile())
      input = sect->getOldInputFile();

Now CommonELFSection overrides getOldInputFile(), enabling uniform code:

    if (sect->hasOldInputFile())
      input = sect->getOldInputFile();

Changes:
- Add hasOldInputFile/getOldInputFile overrides to CommonELFSection
- Remove redundant getOrigin method
- Simplify ObjectBuilder, SectionMap, GarbageCollection call sites
- Replace dyn_cast + type-specific calls with polymorphic interface

Resolves #1300

### DIFF
--- a/include/eld/Readers/CommonELFSection.h
+++ b/include/eld/Readers/CommonELFSection.h
@@ -28,7 +28,9 @@ public:
     return S->getSectionKind() == Section::Kind::CommonELF;
   }
 
-  InputFile *getOrigin() const { return Origin; }
+  bool hasOldInputFile() const override { return Origin != nullptr; }
+
+  InputFile *getOldInputFile() const override { return Origin; }
 
   static constexpr uint32_t DefaultType = llvm::ELF::SHT_NOBITS;
   static constexpr uint32_t DefaultFlags =
--- a/lib/GarbageCollection/GarbageCollection.cpp
+++ b/lib/GarbageCollection/GarbageCollection.cpp
@@ -733,13 +733,13 @@ void GarbageCollection::stripSections(SectionSetTy &S,
       if (!mayProcessGC(*Section))
         continue;
       if (MReferencedSections.find(Section) == MReferencedSections.end()) {
-        bool IsCommonSection = false;
+        bool IsCommonSection = llvm::isa<CommonELFSection>(Section);
         Input *I = ObjFile->getInput();
-        if (CommonELFSection *CommonSection =
-                dyn_cast<CommonELFSection>(Section)) {
-          I = CommonSection->getOrigin()->getInput();
-          IsCommonSection = true;
+
+        if (IsCommonSection && Section->hasOldInputFile()) {
+          I = Section->getOldInputFile()->getInput();
         }
+
         // Print the GC collected input section if Tracing is enabled.
         if (ThisConfig.options().printGCSections() ||
             ThisModule.getPrinter()->traceGC()) {
--- a/lib/LinkerWrapper/PluginADT.cpp
+++ b/lib/LinkerWrapper/PluginADT.cpp
@@ -710,9 +710,6 @@ plugin::InputFile plugin::Section::getRuleMatchingInput() const {
     return plugin::InputFile(nullptr);
   if (m_Section->hasOldInputFile())
     return plugin::InputFile(m_Section->originalInput());
-  if (CommonELFSection *commonSect =
-          llvm::dyn_cast<CommonELFSection>(m_Section))
-    return plugin::InputFile(commonSect->getOrigin());
   return plugin::InputFile(m_Section->originalInput());
 }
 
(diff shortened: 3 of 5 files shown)


include/eld/Readers/CommonELFSection.h
```diff
@@ -28,6 +28,14 @@ class CommonELFSection : public ELFSection {
...
 
+  bool hasOldInputFile() const override {
+      return Origin != nullptr;
+  }
+
+  InputFile *getOldInputFile() const override {
+      return Origin;
+  }
+
   InputFile *getOrigin() const { return Origin; }
```
parth-07: Do we need `getOrigin` now?

parth-07 (commented):
Can you please improve the commit message here? This resource might be helpful: https://cbea.ms/git-commit/

include/eld/Readers/CommonELFSection.h
rachitmeht: No, we don't need it, may be confusing later (fixed).

lib/GarbageCollection/GarbageCollection.cpp
```diff
@@ -733,13 +733,13 @@ void GarbageCollection::stripSections(SectionSetTy &S,
...
       if (MReferencedSections.find(Section) == MReferencedSections.end()) {
-        bool IsCommonSection = false;
+        bool IsCommonSection = llvm::isa<CommonELFSection>(Section);
         Input *I = ObjFile->getInput();
-        if (CommonELFSection *CommonSection =
-                dyn_cast<CommonELFSection>(Section)) {
-          I = CommonSection->getOrigin()->getInput();
-          IsCommonSection = true;
+
+        if (Section->hasOldInputFile()) {
```
parth-07: Were we previously using old input file for non-common sections?

rachitmeht: No, we were not, but as I variable is only used for diagnostic outputs, so it is better to use origin file only irrespective of common sections.

parth-07: A refactoring patch should not introduce functionality change, even a diagnostic change. If you think this change is helpful, can you open a separate issue to discuss it?

rachitmeht: I have fixed and opened a seperate issue for the change.

quic-areg (approved):
LGTM

Steven6798:
There are multiple commits and some of them are not signed. Please add all changes into a single commit since the fix is small enough.

include/eld/Readers/CommonELFSection.h
```diff
@@ -28,7 +28,9 @@ class CommonELFSection : public ELFSection {
     return S->getSectionKind() == Section::Kind::CommonELF;
   }
 
-  InputFile *getOrigin() const { return Origin; }
+  bool hasOldInputFile() const override { return Origin != nullptr; }
+
+  InputFile *getOldInputFile() const override { return Origin; }
```
quic-seaswara, line 33: probably another TODO is, rename `getOldInputFile` function to 

`getPreLtoInputFile` or `getOriginalInputFile` ?

Steven6798 (approved):

include/eld/Readers/CommonELFSection.h
parth-07: I feel that this would be against the motivation of this change. This change key motivation was to have same named function in CommonELFSection and ELFSection to get the original input, so that at places where we need to use the old input, we will not need different branches for `CommonELFSection` and `ELFSection`.

parth-07 (approved):

quic-seaswara:
@rachitmeht please resolve the comments if they have been addressed.

PR #1587, opened 2026-07-27
Warn when --version-script has no effect

### LINKED ISSUE #1444: Report warning if a version script is used in cases where it is not meaningful
We should report a warning if version script is used in cases where it is not meaningful, for example, when creating a static executables.

rachitmeht:
Version scripts control symbol visibility in the dynamic symbol table (`.dynsym`). When output types don't create `.dynsym`, version scripts are silently ignored, So, analyzed when `.dynsym` is created: `isShared() || isPIE() || (isStatic() && forceDynamic())`

Negated these conditions to warn when version scripts have no effect:
- Static executable: `isCodeStatic() && !isPIE() && !forceDynamic()`
- Relocatable output: `isLinkPartial()`  (in this case also no `.dynsym` is created)

Implemented in `ObjectLinker::parseVersionScript()` after version script is recorded but before parsing.

 Changes:
- `DiagLDScript.inc`: Added 2 warning diagnostics
- `ObjectLinker.cpp`: Added warning checks 
- `VersionScriptNoEffect.test`: 6 test cases

Resolves #1444

### DIFF
--- a/include/eld/Diagnostics/DiagLDScript.inc
+++ b/include/eld/Diagnostics/DiagLDScript.inc
@@ -65,12 +65,8 @@ DIAG(reading_dynamic_list, DiagnosticEngine::Verbose, "Dynamic List[%0] : %1")
 DIAG(reading_extern_list, DiagnosticEngine::Verbose, "Extern List[%0] : %1")
 DIAG(error_parsing_version_script, DiagnosticEngine::Error,
      "Error parsing version script %0")
-DIAG(warn_version_script_no_dynsym, DiagnosticEngine::Warning,
-     "%0: warning: version script has no effect: static executable does not "
-     "emit a dynamic symbol table")
-DIAG(warn_version_script_no_effect_relocatable, DiagnosticEngine::Warning,
-     "%0: warning: version script has no effect: relocatable output (-r) does "
-     "not emit a dynamic symbol table")
+DIAG(warn_version_script_no_effect, DiagnosticEngine::Warning,
+     "%0: version script has no effect")
 DIAG(fatal_divide_by_zero, DiagnosticEngine::Fatal,
      "%0: division by zero in expression %1")
 DIAG(fatal_modulo_by_zero, DiagnosticEngine::Fatal,
--- a/include/eld/Diagnostics/DiagLDScript.inc
+++ b/include/eld/Diagnostics/DiagLDScript.inc
@@ -65,6 +65,12 @@ DIAG(reading_dynamic_list, DiagnosticEngine::Verbose, "Dynamic List[%0] : %1")
 DIAG(reading_extern_list, DiagnosticEngine::Verbose, "Extern List[%0] : %1")
 DIAG(error_parsing_version_script, DiagnosticEngine::Error,
      "Error parsing version script %0")
+DIAG(warn_version_script_no_dynsym, DiagnosticEngine::Warning,
+     "%0: warning: version script has no effect: static executable does not "
+     "emit a dynamic symbol table")
+DIAG(warn_version_script_no_effect_relocatable, DiagnosticEngine::Warning,
+     "%0: warning: version script has no effect: relocatable output (-r) does "
+     "not emit a dynamic symbol table")
 DIAG(fatal_divide_by_zero, DiagnosticEngine::Fatal,
      "%0: division by zero in expression %1")
 DIAG(fatal_modulo_by_zero, DiagnosticEngine::Fatal,
--- a/lib/Object/ObjectLinker.cpp
+++ b/lib/Object/ObjectLinker.cpp
@@ -352,37 +352,48 @@ bool ObjectLinker::normalize() {
 // FIXME: We should maybe parse version script after reading LTO-generated
 // object files.
 bool ObjectLinker::parseVersionScript() {
-  if (ThisConfig.options().hasVersionScript()) {
-    LayoutInfo *layoutInfo = ThisModule->getLayoutInfo();
-    for (const auto &List : ThisConfig.options().getVersionScripts()) {
-      Input *VersionScriptInput =
-          eld::make<Input>(List, ThisConfig.getDiagEngine(), Input::Script);
-      if (!VersionScriptInput->resolvePath(ThisConfig))
-        return false;
-      // Create an Input file and set the input file to be of kind DynamicList
-      InputFile *VersionScriptInputFile =
-          InputFile::create(VersionScriptInput, InputFile::GNULinkerScriptKind,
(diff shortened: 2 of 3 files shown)


parth-07 (approved):

test/Common/standalone/VersionScript/VersionScriptNoEffect.test
```diff
@@ -0,0 +1,34 @@
+# Test that version scripts emit warnings when they have no effect
+# (i.e., when no dynamic symbol table will be created)
```
quic-seaswara: nit : use the test format for new tests.

lib/Object/ObjectLinker.cpp
```diff
@@ -367,6 +367,18 @@ bool ObjectLinker::parseVersionScript() {
     // Record the dynamic list script in the Map file.
     if (layoutInfo)
       layoutInfo->recordVersionScript(List);
+    // Warn if the version script will have no effect because no dynamic
+    // symbol table will be emitted.
+    if (ThisConfig.isLinkPartial()) {
+      // Relocatable output (-r) never emits .dynsym.
+      ThisConfig.raise(Diag::warn_version_script_no_effect_relocatable)
+          << VersionScriptInput->decoratedPath();
```
quic-seaswara: Should this be a warning under -Wlinker-script category ?

test/Common/standalone/VersionScript/VersionScriptNoEffect.test
```diff
@@ -0,0 +1,34 @@
...
+# RUN 3: PIE executable should NOT warn (has .dynsym)
+RUN: %clang %clangopts -c -fpic -O2 %p/Inputs/1.c -o %t1_pic.o
+RUN: %link %linkopts -pie --entry=main --version-script=%p/Inputs/vs1 -o %t_pie.out %t1_pic.o 2>&1 | %filecheck --allow-empty --check-prefix=NO-WARN-PIE %s
+
+NO-WARN-PIE-NOT: warning: version script has no effect
+
+# RUN 4: Shared library should NOT warn (has .dynsym)
+RUN: %link %linkopts -shared --version-script=%p/Inputs/vs1 -o %t_shared.so %t1_pic.o 2>&1 | %filecheck --allow-empty --check-prefix=NO-WARN-SHARED %s
+
+NO-WARN-SHARED-NOT: warning: version script has no effect
```
quic-seaswara: no version script filename is printed ? 

what about if you have a linker script that has a version block ?

include/eld/Diagnostics/DiagLDScript.inc
```diff
@@ -65,6 +65,12 @@ DIAG(reading_dynamic_list, DiagnosticEngine::Verbose, "Dynamic List[%0] : %1")
 DIAG(reading_extern_list, DiagnosticEngine::Verbose, "Extern List[%0] : %1")
 DIAG(error_parsing_version_script, DiagnosticEngine::Error,
      "Error parsing version script %0")
+DIAG(warn_version_script_no_dynsym, DiagnosticEngine::Warning,
+     "%0: warning: version script has no effect: static executable does not "
+     "emit a dynamic symbol table")
+DIAG(warn_version_script_no_effect_relocatable, DiagnosticEngine::Warning,
+     "%0: warning: version script has no effect: relocatable output (-r) does "
+     "not emit a dynamic symbol table")
```
quic-seaswara: do we need does not emit a dynamic symbol table ? remove the reason ?

include/eld/Diagnostics/DiagLDScript.inc
```diff
@@ -65,6 +65,12 @@ DIAG(reading_dynamic_list, DiagnosticEngine::Verbose, "Dynamic List[%0] : %1")
 DIAG(reading_extern_list, DiagnosticEngine::Verbose, "Extern List[%0] : %1")
 DIAG(error_parsing_version_script, DiagnosticEngine::Error,
      "Error parsing version script %0")
+DIAG(warn_version_script_no_dynsym, DiagnosticEngine::Warning,
+     "%0: warning: version script has no effect: static executable does not "
+     "emit a dynamic symbol table")
```
quic-seaswara: not sure about static executables and version script. We should check 

`ld.eld <object> .... <object> -static -pie `

test/Common/standalone/VersionScript/VersionScriptNoEffect.test
```diff
@@ -0,0 +1,34 @@
...
+
+# RUN 5: Static with --force-dynamic should NOT warn (forces .dynsym)
+RUN: %link %linkopts -static --force-dynamic --entry=main --version-script=%p/Inputs/vs1 -o %t_force.out %t1.o 2>&1 | %filecheck --allow-empty --check-prefix=NO-WARN-FORCE %s
+
+NO-WARN-FORCE-NOT: warning: version script has no effect
+
+# RUN 6: Verify the warnings include the version script path
+RUN: %link %linkopts -static --entry=main --version-script=%p/Inputs/vs1 -o %t_path.out %t1.o 2>&1 | %filecheck --check-prefix=CHECK-PATH %s
+
+CHECK-PATH: {{.*}}/Inputs/vs1: warning: version script has no effect
```
quic-seaswara: will this test pass on windows ?

quic-seaswara (changes requested):

quic-areg (approved):

lib/Object/ObjectLinker.cpp
rachitmeht: Yes, it should, but there is -Wversion-script also ( not documented ), should it be warning under both category ?

include/eld/Diagnostics/DiagLDScript.inc
rachitmeht: Verified: static pie do have .dynsym table but it doesn't contain any user defined symbols, so, version-script has no effect on it. We should warn in this case also.

rachitmeht: Actually, for every pie combination, it is not affected by version script.
except for -pie --export-dynamic.

test/Common/standalone/VersionScript/VersionScriptNoEffect.test
rachitmeht: I think, the case of version block in linkerscript will be solved automatically if this PR #1672 is correct. Otherwise we have to add same check in two different paths.

include/eld/Diagnostics/DiagLDScript.inc
rachitmeht: removed.

PR #1588, opened 2026-07-27
Report error when version script node has local block before the global

### LINKED ISSUE #1582: [version-script] Improve version script diagnostics
```
cat > 1.c << \!
int foo() { return 0; }
int bar() { return 0; }
!

cat > vs.t << \!
V1 {
  local:
    *;
  global:
    foo;
};
!

clang -c 1.c -fPIC
ld.eld 1.o -shared --version-script vs.t
```

This produces an error that 

`Error: Error parsing version script vs.t`

It does not show the line number or what the error is. 

Please improve.

parth-07:
This commit updates ScriptParser to report error when a version script node contains local block before the global block.

Resolves #1582

### DIFF
--- a/docs/userguide/documentation/linker_script.rst
+++ b/docs/userguide/documentation/linker_script.rst
@@ -1522,6 +1522,9 @@ A version script is a sequence of *version nodes*. Each node has an optional
 name and contains ``global:`` and/or ``local:`` blocks. Each block lists symbol
 patterns terminated by ``;``.
 
+It is an error for a ``local:`` block to appear before a ``global:`` block in
+the same version node.
+
 .. code-block::
 
   /* Anonymous (unnamed) version node. */
--- a/lib/ScriptParser/ScriptParser.cpp
+++ b/lib/ScriptParser/ScriptParser.cpp
@@ -1568,6 +1568,10 @@ void ScriptParser::readVersionSymbols(VersionScriptNode &VSN) {
         continue;
       }
       if (Tok == "global:" || (Tok == "global" && consume(":"))) {
+        if (VSN.getLocalBlock()) {
+          setError("global scope must appear before local scope", Tok);
+          return;
+        }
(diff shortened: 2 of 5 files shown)


docs/userguide/documentation/linker_script.rst
```diff
@@ -1522,6 +1522,9 @@ A version script is a sequence of *version nodes*. Each node has an optional
 name and contains ``global:`` and/or ``local:`` blocks. Each block lists symbol
 patterns terminated by ``;``.
 
+It is an error for a ``local:`` block to appear before a ``global:`` block in
+the same version node.
+
```
quic-seaswara, line 1527: can we add a attention syntax here 

`.. attention:: `

quic-seaswara (changes requested):

docs/userguide/documentation/linker_script.rst
quic-seaswara: is there any other kinds of errors that version script parser detects and silently errors out?

quic-shaksuma (approved):

docs/userguide/documentation/linker_script.rst
parth-07: > can we add a attention syntax here

The file has now been changed to markdown file. I have added the markdown-alternative `:::{note}` syntax here.

parth-07: > is there any other kinds of errors that version script parser detects and silently errors out?

If version script node is getting parsed and there has been any other error in the script parsing then the version script node parsing also unnecessarily fails due to the below highlighted code:

```diff
VersionScriptBlock *VersionScriptNode::switchToGlobal() {
  if (MLocal && !MGlobal) {
    MHasErrorDuringParsing = true;
    return nullptr;
  }
+  if (MHasErrorDuringParsing)
+    return nullptr;
...
```

I believe that this should be addresses separately.

PR #1593, opened 2026-07-27
[Workflow] Fix GitHub Actions Security Scan (zizmor) errors

Steven6798:
- template-injection: move ${{ ... }} expansions into env vars and
    reference as quoted shell variables instead of interpolating
    directly into run: blocks (ci.yml, clang-format-pr.yml,
    cpp-linter-pr.yml, nightly-musl-builder.yml,
    BuildStatusDataRecorder/action.yaml)
  - overly-broad-permissions: drop workflow-level `contents: write` to
    `permissions: {}` and scope `contents: write` (and `actions: read`
    where needed) down to the individual jobs that require it
    (busybox-eld.yml, ci.yml, clang-cross-schedule.yml,
    nightly-musl-builder.yml, zephyr-nightly.yml, zephyr-run.yml)
  - reusable-workflow-permissions: grant `permissions: contents: write`
    on the calling job when it invokes a reusable workflow that
    requires it, since callers no longer inherit elevated permissions
    from the workflow level (busybox-eld.yml, clang-cross-schedule.yml)
  - dangerous-github-env: write PATH via GITHUB_PATH instead of
    GITHUB_ENV, and annotate one remaining GITHUB_ENV write as a false
    positive (literal $(pwd)-derived value, not attacker-controlled)
    (FetchEldToolset/action.yaml, FetchNightlyToolset/action.yaml,
    MuslBuilderBuildELD/action.yaml)
  - unsound-condition: fix `if: ${{ inputs.enabled }} == "true"` (always
    true, comparing a rendered string to a literal outside the
    expression) to `if: ${{ inputs.enabled == 'true' }}`
    (BuildStatusDataRecorder/action.yaml)
  - dangerous-triggers: suppress workflow_run trigger warning with an
    explanation comment; the alternative (pull_request_target) can't
    exercise this path pre-merge (zephyr-pr.yml)

### DIFF
--- a/.github/workflows/busybox-eld.yml
+++ b/.github/workflows/busybox-eld.yml
@@ -6,8 +6,7 @@ on:
   schedule:
    - cron: "0 5 * * *"
 
-permissions:
-  contents: write
+permissions: {}
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
@@ -36,6 +35,8 @@ jobs:
             toolchain_asset: aarch64-unknown-linux-musl.tar.xz
             enabled: 1
     uses: ./.github/workflows/clang-cross-download.yml
+    permissions:
+      contents: write
     with:
       runner: ubuntu-latest
       workspace: /home/runner/work/${{ github.event.repository.name }}/${{ github.event.repository.name }}
@@ -48,6 +49,8 @@ jobs:
     if: github.repository == 'qualcomm/eld'
     needs: fetch-clang-cross-toolchain
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/BuildStatusDataRecorder/action.yaml
+++ b/.github/workflows/BuildStatusDataRecorder/action.yaml
@@ -34,15 +34,23 @@ runs:
   using: "composite"
   steps:
     - name: Record ${{ inputs.stage }}build entry
-      if: ${{ inputs.enabled }} == "true"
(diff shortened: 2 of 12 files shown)


Steven6798:
Example of the errors: https://github.com/qualcomm/eld/actions/runs/30285028797/job/90040586561?pr=1592

quic-shaksuma (approved):
LGTM. Please verify in your fork

quic-areg:
It would be nice to mention which audits were fixed and with what resolution in the commit message, eg "template-injection: move ${{...}} into env vars"

Steven6798:
> It would be nice to mention which audits were fixed and with what resolution in the commit message, eg "template-injection: move ${{...}} into env vars"

Thanks. will do.

.github/workflows/busybox-eld.yml
```diff
@@ -2,12 +2,11 @@ name: BusyBox with ELD
...
 on:
   workflow_dispatch:
-  # pull_request: {}
+  pull_request: {}
   schedule:
    - cron: "0 5 * * *"
 
-permissions:
-  contents: write
+permissions: {}
```
quic-areg, line 9: I think this could fail since it calls a workflow with more restrictive permissions?

Steven6798: Yes. There where 2 instances of this failure. Thanks.

quic-areg (approved):

quic-seaswara (approved):

Steven6798:
In draft until I rebase and revert any change made for testing the workflows

PR #1612, opened 2026-07-29
[Command Line][Driver] Add -v option to match GNU ld/ld.lld

### LINKED ISSUE #1610: [Command Line] Add support for -v option
both GNU ld and lld support the -v option. We should add this for compatibility and functionality purposes. This option is slightly different from --version, as it does not short circuit right after processing it. It is a weird known behavior of GNU ld documented in lld.

Steven6798:
Add support for -v option. The behavior is the following:

  - Bare `-v` (no inputs) prints the version banner and exits successfully, instead of erroring with "no inputs". This matters for tools like the Linux kernel's scripts/mkcompile_h, which runs `$LD -v` to capture the linker version string.
  - `-v` combined with real inputs prints the banner but does not short-circuit: the link still proceeds and produces output.
  - `-v` and `--version` conflict with each other and the last one on the command line wins.

Fix: qualcomm#1610

### DIFF
--- a/include/eld/Input/InputAction.h
+++ b/include/eld/Input/InputAction.h
@@ -16,6 +16,7 @@
 #include "eld/Input/Input.h"
 #include "eld/Support/Path.h"
 #include <string>
+#include <vector>
 
 namespace eld {
 
@@ -51,6 +52,7 @@ public:
     StartLib,
     WholeArchive,
     JustSymbols,
+    Version,
   };
 
 protected:
@@ -364,6 +366,24 @@ public:
 
 private:
   const std::string InputFormat;
+};
+
+/// VersionAction handles '-v', which prints the version banner and lets
+/// linking continue (unlike '--version', which always short-circuits and
+/// is handled directly in parseOptions, not as an InputAction).
+class VersionAction : public InputAction {
+public:
+  explicit VersionAction(const std::vector<std::string> &SupportedTargets,
+                         DiagnosticPrinter *Printer);
+
+  bool activate(InputBuilder &) override;
+
+  static bool classof(const InputAction *I) {
+    return I->getInputActionKind() == InputActionKind::Version;
+  }
+
+private:
+  std::vector<std::string> SupportedTargets;
 };
 
 } // namespace eld
--- a/lib/LinkerWrapper/ARMLinkDriver.cpp
+++ b/lib/LinkerWrapper/ARMLinkDriver.cpp
@@ -106,9 +106,12 @@ ARMLinkDriver::parseOptions(ArrayRef<const char *> Args,
                      /*ShowAllAliases=*/true);
     return LINK_SUCCESS;
   }
-  if (ArgList.hasArg(OPT_ARMLinkOptTable::version)) {
-    printVersionInfo();
-    return LINK_SUCCESS;
+  if (llvm::opt::Arg *Arg = ArgList.getLastArg(OPT_ARMLinkOptTable::v,
+                                               OPT_ARMLinkOptTable::version)) {
+    if (Arg->getOption().matches(OPT_ARMLinkOptTable::version)) {
+      printVersionInfo();
+      return LINK_SUCCESS;
+    }
   }
   // --about
   if (ArgList.hasArg(OPT_ARMLinkOptTable::about)) {
--- a/lib/LinkerWrapper/GnuLdDriver.cpp
+++ b/lib/LinkerWrapper/GnuLdDriver.cpp
@@ -196,10 +196,7 @@ void GnuLdDriver::printVersionInfo() const {
   // linker by parsing --version output (e.g. the Linux kernel's
   // scripts/ld-version.sh) look for a first line whose first two words are
   // exactly "GNU"/"ld" and take the last word on that line as the version;
-  // without this, such scripts reject eld as an "unknown linker". This is
-  // an interim identification (agreed on until eld is proposed/added to the
-  // kernel community's own linker-detection logic) -- not a claim that eld
-  // is GNU ld.
+  // without this, such scripts reject eld as an "unknown linker".
   outs() << "GNU ld compatible linker - eld " << eld::getELDVersion() << "\n";
   outs() << "Supported Targets: ";
   for (const auto &x : m_SupportedTargets)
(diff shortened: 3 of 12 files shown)


Steven6798:
Example of the fix working:

Before the fix: https://github.com/qualcomm/eld/actions/runs/30389664625/job/90377743483?pr=1608
```
  SYSHDR  arch/riscv/include/generated/uapi/asm/unistd_64.h
Error: no inputs
Fatal: Linking had errors (a.out)
  UPD     include/generated/compile.h
```

After the fix: https://github.com/qualcomm/eld/actions/runs/30479138389/job/90668284429?pr=1608
```
  SYSHDR  arch/riscv/include/generated/uapi/asm/unistd_64.h
  UPD     include/generated/uapi/linux/version.h
  HOSTCC  scripts/dtc/fstree.o
```

lib/LinkerWrapper/ARMLinkDriver.cpp
```diff
@@ -106,9 +106,22 @@ ARMLinkDriver::parseOptions(ArrayRef<const char *> Args,
...
+  // for compatibility with GNU linkers (see ld.lld's Driver.cpp for the
+  // same rationale): -v prints the version banner and, if there are no
+  // input files, exits immediately without erroring on "no input files"
+  // If inputs are present, -v just prints the banner as a side effect and the
+  // link proceeds normally. --version always prints and exits, regardless
+  // of inputs.
+  if (llvm::opt::Arg *Arg = ArgList.getLastArg(OPT_ARMLinkOptTable::v,
+                                               OPT_ARMLinkOptTable::version)) {
+    if (!Config.options().hasPrintedVersionInfo()) {
+      printVersionInfo();
```
quic-seaswara, line 112: print the version info and continue processing means it needs to be an InputAction.

This will reduce the amount of code repetitions as well.

quic-seaswara (changes requested):

lib/LinkerWrapper/ARMLinkDriver.cpp
Steven6798: I'll look into this. Thanks for the feedback.

Steven6798: If we move the logic of -V to where InputActions are handled, then -v and --version will behave differently when there is no input.

quic-seaswara: for --version we will not create VersionAction, only for -v we would.

lib/LinkerWrapper/ARMLinkDriver.cpp
```diff
@@ -106,9 +106,12 @@ ARMLinkDriver::parseOptions(ArrayRef<const char *> Args,
                      /*ShowAllAliases=*/true);
     return LINK_SUCCESS;
   }
-  if (ArgList.hasArg(OPT_ARMLinkOptTable::version)) {
-    printVersionInfo();
-    return LINK_SUCCESS;
+  if (llvm::opt::Arg *Arg = ArgList.getLastArg(OPT_ARMLinkOptTable::v,
+                                               OPT_ARMLinkOptTable::version)) {
```
quic-seaswara, line 110: Why this change ?

quic-seaswara (changes requested):

lib/LinkerWrapper/ARMLinkDriver.cpp
Steven6798: -v and --version conflict with each other. Following the ELD convention, the last option win.

Steven6798: Not necessary, but good to have to be consistent with our convention.

quic-seaswara: let us reduce the diff. We can add it if needed.

Steven6798: Sounds good

quic-areg: Then how do we handle `-v` and `--version` together? I don't think it makes the diff meaningfully larger or this patch any harder to review. It seems strange to have a unified convention for everything else (the whole point of  96172db3560c60391f26df8e00fd15e2516d1166) but not this pair.

include/eld/Input/InputAction.h
```diff
@@ -366,6 +368,24 @@ class InputFormatAction : public InputAction {
   const std::string InputFormat;
 };
 
+/// VersionAction handles '-v', which prints the version banner and lets
```
quic-areg, line 371: Why do we need new `InputAction` classes for this? Can this be handled in `parseOptions`?. The only change is whether we continue the link or stop. The classes seem overkill.

lib/LinkerWrapper/GnuLdDriver.cpp
```diff
@@ -1962,6 +1964,13 @@ eld::Module *GnuLdDriver::ThisModule = nullptr;
 template <class T>
 bool GnuLdDriver::doLink(llvm::opt::InputArgList &Args,
                          std::vector<eld::InputAction *> &actions) {
+  // Bare -v (no real inputs): print the banner and exit immediately.
```
quic-areg, line 1967: Does this rely on every target's `OptTable::v` having the same value? should this be `T::v` instead?

Steven6798: Yes, this should be T. I have a fix for this. This is why the CI crashed.

include/eld/Input/InputAction.h
Steven6798: Handling everything in parseOptions was my initial solution. Shankar suggested using InputAction.

https://github.com/qualcomm/eld/commit/d1f1c7570a9f0a42c6d5a1642e9e65cd029be123

quic-areg: Sorry, I missed that conversation. I did not mean to reopen something that was already resolved.

lib/LinkerWrapper/ARMLinkDriver.cpp
quic-seaswara: -v and --version are different right ? both print the version but they should not be handled together IMO

quic-seaswara (approved):

quic-areg (approved):

PR #1615, opened 2026-07-29
[Layout] Fix LMA alignment when section-description ALIGN() is used w…

### LINKED ISSUE #1614: [linker-script] LMA should be aligned when explicitly aligning sections
```
cat > 1.c << \!
int foo() { return 0; }
int bar() { return 0; }
int baz() { return 0; }
!

cat > script.t << \!
PHDRS {
  A PT_LOAD;
  B PT_LOAD;
  C PT_LOAD;
}

MEMORY {
  M1 : ORIGIN = 0x1000, LENGTH = 1M
  M1_LMA : ORIGIN = 0x1000, LENGTH = 1M
}

SECTIONS {
  .foo : { *(.text.foo) } >M1 AT>M1_LMA :A
  .bar ALIGN(0x1000) : { *(.text.bar) } >M1 AT>M1_LMA :B
  .baz : ALIGN(0x1000) { *(.text.baz) } >M1 AT>M1_LMA :C
}
!

ld.eld 1.o -T script.t -Map x.eld -o eld.out
llvm-readelf -l -W eld.out
```

The test would show that the the LMA of .baz is not aligned to 0x1000.

Linker needs to honor that the ALIGN in the output description prologue will bump the section to the alignment 0x1000, and should honor LMA to be aligned.

quic-seaswara:
When a section uses ALIGN(n) in the section description (after the colon, e.g. `.sec : ALIGN(n) { }`) alongside AT>REGION, the LMA was not being aligned to match the VMA.

- Prolog ALIGN (before colon, e.g. `.sec ALIGN(n) : { }`): this sets the VMA only; the LMA stays at the LMA region cursor. No change in behavior.
- Section-description ALIGN (after colon, e.g. `.sec : ALIGN(n) { }`): this is an explicit alignment request that should apply to both VMA and LMA. With AT>, the LMA was previously left unaligned; now it is aligned to match the VMA.

Fixes #1614

### DIFF
--- a/docs/userguide/documentation/linker_script.rst
+++ b/docs/userguide/documentation/linker_script.rst
@@ -1182,8 +1182,40 @@ In GNU linker scripts, the AT command is used to control the Load Memory Address
 
 .. important::
 
-   When an AT command is specified as part of the output section, the linker
-   will not automatically align the load memory address of the section.
+   When ``AT>REGION`` is specified on an output section, the linker does **not**
+   automatically align the LMA to match the natural alignment of the input
+   sections. The LMA region cursor is authoritative in that case.
+
+   However, if the output section description contains an explicit ``ALIGN(n)``
+   keyword *after* the colon (the section address expression form, e.g.
+   ``.sec : ALIGN(n) { }``), ELD applies that alignment to both the VMA and the
+   LMA, even when ``AT>REGION`` is present.
+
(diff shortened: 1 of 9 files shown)


quic-seaswara:
Zephyr check

https://github.com/qualcomm/eld/actions/runs/30496594468/job/90726792749

Steven6798 (approved):

PR #1619, opened 2026-07-30
Honor --export-dynamic for PIE executables

### LINKED ISSUE #1620: --export-dynamic is ignored for PIE executables
For PIE output, `GNULDBackend::canSkipSymbolFromExport()` skips exporting every global symbol except the entry symbol, even when `--export-dynamic` is passed. Only main ends up in `.dynsym`. 
Impact: a plugin `dlopen()`'d at runtime cannot resolve symbols defined in the host executable. This causes 5 LLVM unit test failures  (`PluginsTests.LoadPlugin`, `PluginsTests.LoadMultiplePlugins`, `DynamicLibrary.Overload`, `PluginInlineAdvisorTest.PluginLoad`, `PluginInlineOrderTest.NoInlineFoo`).

Reproducer:
```bash
LINKER=$1
cat > main.c << 'EOF'
int exported_var = 42;
int exported_func(void) { return exported_var; }
int main(void) { return exported_func(); }
EOF

clang -c main.c -o main.o -fPIC
clang -fuse-ld=$LINKER -pie main.o -o main.out -Wl,--export-dynamic

readelf --dyn-syms main.out | grep exported_func || echo "BUG: exported_func missing from .dynsym"
```

BhoumikPatidar:
For PIE output, `canSkipSymbolFromExport()` skipped exporting every global symbol except the entry symbol, even when `--export-dynamic` was passed. As a result, a plugin `dlopen()`'d at runtime could not resolve symbols defined in the host executable.

Let `--export-dynamic` override that restriction.

This PR fixes failing tests in llvm-unit-tests for x86_64 backend

Resolves #1620

### DIFF
--- a/lib/Target/GNULDBackend.cpp
+++ b/lib/Target/GNULDBackend.cpp
@@ -812,7 +812,10 @@ bool GNULDBackend::canSkipSymbolFromExport(ResolveInfo *R, bool isEntry) const {
     return false;
   // For PIE, only symbols that really need to be exported are the only ones
   // that can be exported. Dynamic List will control this as well.
-  if (config().options().isPIE() && !isEntry)
+  // --export-dynamic is an explicit request to export all globals, so it
+  // overrides this default restriction.
+  if (config().options().isPIE() && !isEntry &&
+      !config().options().exportDynamic())
     return true;
   if (R->isAbsolute())
     return true;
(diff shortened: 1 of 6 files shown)


parth-07 (commented):
Can you please create an issue for this and link the issue to be resolved by this PR?

parth-07 (changes requested):
Can you please add `Fixes #1620` in the commit message as well?

quic-seaswara (approved):

quic-seaswara (approved):

parth-07 (approved):
LGTM.

Please create an issue to track that we need to decide upon the precedence of version-script,  dynamic list, and options such as --export-dynamic.

PR #1622, opened 2026-07-30
[Command Line][Driver] Change version info to be compatible with linux kernel

### LINKED ISSUE #1621: [Command Line][Driver] Change version info to be compatible with Linux kernel
The Linux kernel only supports LLD and GNU LD as recognized linkers. For now, the version information of ELD will be changed to be compatible with GNU. Eventually, this will be reverted after ELD is recognized as a valid linker for the Linux kernel.

Steven6798:
The Linux kernel only recognizes LLD and GNU LD as valid linkers. For now, the version information of ELD will be changed to be compatible with GNU. Eventually, this will be reverted after ELD is recognized as a valid linker for the Linux kernel.

Fix: qualcomm#1621

### DIFF
--- a/lib/LinkerWrapper/GnuLdDriver.cpp
+++ b/lib/LinkerWrapper/GnuLdDriver.cpp
@@ -191,8 +191,16 @@ void GnuLdDriver::printAboutInfo() const {
 }
 
 void GnuLdDriver::printVersionInfo() const {
-  outs() << "eld " << eld::getELDVersion() << " (GNU Compatible linker)"
-         << "\n";
+  // The first line's first two words are deliberately "GNU"/"ld", ending in
+  // nothing but the bare version number. Build systems that identify the
+  // linker by parsing --version output (e.g. the Linux kernel's
+  // scripts/ld-version.sh) look for a first line whose first two words are
+  // exactly "GNU"/"ld" and take the last word on that line as the version;
+  // without this, such scripts reject eld as an "unknown linker". This is
+  // an interim identification (agreed on until eld is proposed/added to the
+  // kernel community's own linker-detection logic) -- not a claim that eld
+  // is GNU ld.
+  outs() << "GNU ld compatible linker - eld " << eld::getELDVersion() << "\n";
   outs() << "Supported Targets: ";
   for (const auto &x : m_SupportedTargets)
     outs() << x << " ";
--- a/test/Common/standalone/printVersion/printVersion.test
+++ b/test/Common/standalone/printVersion/printVersion.test
@@ -6,5 +6,5 @@
 RUN: %link --version 2>&1 | %filecheck %s
 RUN: %link --about 2>&1 | %filecheck %s -check-prefix=ABOUT
 
-#CHECK: eld {{.*}} (GNU Compatible linker)
+#CHECK: GNU ld compatible linker - eld {{.*}}
 #ABOUT: Linker Plugin Interface Version {{[[:xdigit:]]+}}.{{[[:xdigit:]]+}}


quic-seaswara (approved):

lib/Script/OutputSectData.cpp
```diff
@@ -30,9 +30,8 @@ using namespace eld;
 namespace {
 
 std::string buildLinkerVersionString() {
-  std::string Version = "eld ";
+  std::string Version = "GNU ld compatible linker - eld ";
```
quic-seaswara: We dont need to change this.

quic-seaswara (changes requested):

lib/Script/OutputSectData.cpp
Steven6798: Sounds good

quic-seaswara: @quic-k  will this affect `meson` ?

parth-07 (approved):

quic-shaksuma (approved):

quic-seaswara (approved):

lib/Script/OutputSectData.cpp
quic-k: @quic-seaswara 
meson now thinks ELD is GNU ld because of GNU prefix in --version output

Picolibc has a check, if the linker is ELD, don't use multiple PT_TLS sections
but now since meson says ELD is GNU ld, Picolibc uses multiple PT_TLS sections breaking Picolibc nightly for Hexagon
we actually do support mutiple PT_TLS segments ,but test fails due to actual bug in ELD for Hexagon


but, instead of changing the --version output of ELD and then reverting it to build linux kernel, can we not maintain a patch in linux kernel that identifies ELD (this can be part of ELD workflow) @Steven6798

Steven6798: > Shankar Easwaran (@quic-seaswara) meson now thinks ELD is GNU ld because of GNU prefix in --version output
> 
> Picolibc has a check, if the linker is ELD, don't use multiple PT_TLS sections but now since meson says ELD is GNU ld, Picolibc uses multiple PT_TLS sections breaking Picolibc nightly for Hexagon we actually do support mutiple PT_TLS segments ,but test fails due to actual bug in ELD for Hexagon
> 
> but, instead of changing the --version output of ELD and then reverting it to build linux kernel, can we not maintain a patch in linux kernel that identifies ELD (this can be part of ELD workflow) Steven Ramirez Rosa (@Steven6798)

Since this is a temporary solution for one problem, it should not cause other problems. I can make a temporary fix in the workflow itself like you suggested, and this change can be reverted. That way we don't have to either modify ELD or the Linux kernel. @quic-seaswara what do you think?

quic-k: it actually exposed a bug in ELD (I have a fix here : https://github.com/qualcomm/eld/pull/1632)
so this is not causing other problems per se

I would prefer having a temporary patch in the workflow rather than changing --version output of ELD

but since this is already merged, maybe we can revert this once linux kerel thing is done
but is there a timeline for that? do we know when we are going to revert to the original --version output?

Steven6798: > it actually exposed a bug in ELD (I have a fix here : #1632) so this is not causing other problems per se.

Ohh i see. Thanks for the fix.

> but is there a timeline for that? do we know when we are going to revert to the original --version output?

We don't have a timeline for this, as this was just started last week as an experiment. Also, the Linux kernel workflow is meant to be permanent for testing purposes, so we have 2 options until we submit an official patch to the Linux kernel.
1. Put a temporary fix in ELD (what we have)
2. Put a temporary fix in the workflow (possible solution)

I chose ELD because it was not causing any issues and keeps the workflow closer to what the user will run.

quic-seaswara: Now since the bug is fixed, do you want to still revert the change ?

quic-seaswara: > Shankar Easwaran (@quic-seaswara) meson now thinks ELD is GNU ld because of GNU prefix in --version output
> 
> Picolibc has a check, if the linker is ELD, don't use multiple PT_TLS sections but now since meson says ELD is GNU ld, Picolibc uses multiple PT_TLS sections breaking Picolibc nightly for Hexagon we actually do support mutiple PT_TLS segments ,but test fails due to actual bug in ELD for Hexagon
> 
> but, instead of changing the --version output of ELD and then reverting it to build linux kernel, can we not maintain a patch in linux kernel that identifies ELD (this can be part of ELD workflow) Steven Ramirez Rosa (@Steven6798)

Great, now it simulates Gnu ld in picolibc. I was not aware of this path where picolibc will use a single segment for TLS.

quic-k: I like option 2 more but if everyone thinks option 1 is better, we can work with that as well

PR #1625, opened 2026-07-30
[Zephyr] PR workflow cleanups

### LINKED ISSUE #1616: [github-workflows] Zephyr PR check improvements
(1) For the Zephyr check, I always find when added as part of the CI, the Zephyr check passes without doing anything. 

Example : [Zephyr on ELD PR · qualcomm/eld@bf33904](https://github.com/qualcomm/eld/actions/runs/30494180271) 

Question : Can we fix this or not even run this workflow as part of CI if there is a zephyr-check label ? 

(2) Also if I add the run number to be added in the inputs, like here 

[Zephyr on ELD PR · qualcomm/eld@bf33904](https://github.com/qualcomm/eld/actions/runs/30496294643)

Question : The PR check workflow, fails to download, can this be fixed or this option removed ? 

(3) XIP tests do not run 

Example : [Zephyr on ELD PR · qualcomm/eld@bf33904](https://github.com/qualcomm/eld/actions/runs/30496594468/job/90726792769) 

I am not sure if I am running it as designed but I feel there is some more work to be done here. 

(4) Also since the manual workflow run, needs a PR, it should probably go and add a comment in the PR for this run automatically, otherwise hunting for the run is something becomes very hard to search through. 

Question : Expecting something like this here : [[Layout] Fix LMA alignment when section-description ALIGN() is used w… by quic-seaswara · Pull Reque…](https://github.com/qualcomm/eld/pull/1615#issuecomment-5124140888)

Can you address these (4) requests ?

quic-areg:
Previously zephyr-pr.yml ran on every CI completion and reported a green "skipped" run on PRs without the zephyr-check label, doing no actual work. Now ci.yml dispatches the workflow directly only after a labeled build, so unlabeled PRs produce no run at all.

Also drop the manual ci_run_id dispatch input, which was easy to misuse (run id vs run number) and could fail to download. Manual runs now take just a PR number and discover the right CI run automatically.

Fixes #1616

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,7 @@ jobs:
     permissions:
       contents: write
       statuses: write
+      actions: write      # needed to dispatch zephyr-pr.yml for labeled PRs
 
     env:
       PR_NUMBER: ${{ github.event.number }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -203,6 +204,25 @@ jobs:
...
+            // uploaded above is what its resolve job gates on, so by the time
+            // it runs the pr-eld-toolset is present on this CI run.
+            await github.rest.actions.createWorkflowDispatch({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              workflow_id: 'zephyr-pr.yml',
+              ref: context.payload.pull_request.base.ref,
+              inputs: { pr_number: String(context.payload.pull_request.number) },
+            });
+
```
quic-seaswara, line 225: delete the pr toolset after this is complete ?

quic-areg: This turned out to be more trouble than it's worth imo. We would have to add a new workflow step and make sure that all 5 jobs have finished downloading the toolset before we delete it. The step would need its own set of permissions and code just for a delete.

Retention days is set to 1, and only a few patches use the `zephyr-check` label. I don't think the delete step would save any significant amount of storage.

quic-seaswara (approved):

Steven6798 (approved):

PR #1628, opened 2026-07-30
[command-line] Add --[no-]warn-rwx-segments

### LINKED ISSUE #1627: Add support for --warn-rwx-segments
```
cat > 1.c << \!
int data = 10;
int foo() { return 0; }
int bar() { return 0; }
!

# Linker script: one explicit RWX PT_LOAD, one normal RX PT_LOAD.
cat > script.t << \!
PHDRS {
  rwx  PT_LOAD FLAGS(7);
  text PT_LOAD FLAGS(5);
}
SECTIONS {
  .rwx  0x1000 : { *(.text.foo) } :rwx
  .text 0x2000 : { *(.text.bar) } :text
}
!

echo "=== riscv-ld (GNU ld, default: warning on) ==="
/opt/riscv/bin/riscv-ld 1.o -T script.t -o gnu.out 2>&1

echo "=== riscv-ld --no-warn-rwx-segments ==="
/opt/riscv/bin/riscv-ld 1.o -T script.t --no-warn-rwx-segments -o gnu.out 2>&1

llvm-readelf -l -W gnu.out 2>/dev/null | grep -E "LOAD|Flg"

echo ""
echo "=== -N / --omagic: GNU ld warns even with -N ==="
echo "--- riscv-ld -N ---"
/opt/riscv/bin/riscv-ld 1.o -N -o gnu.omagic.out 2>&1
/opt/riscv/bin/riscv-ld -l -W gnu.omagic.out 2>/dev/null | grep -E "LOAD|Flg"
```

quic-seaswara:
Warn by default when a PT_LOAD (or any) segment has both write and execute permissions (RWX). The warning can be silenced with --no-warn-rwx-segments and re-enabled with --warn-rwx-segments.

Fixes #1627

### DIFF
--- a/include/eld/Driver/GnuLinkerOptions.td
+++ b/include/eld/Driver/GnuLinkerOptions.td
@@ -657,6 +657,14 @@ def no_warn_shared_textrel : FF<"no-warn-shared-textrel">,
                              HelpText<"Don't Warn if the linker adds a "
                                       "DT_TEXTREL">,
                              Group<grp_dynlib>;
+def warn_rwx_segments : FF<"warn-rwx-segments">,
+                        HelpText<"Warn if the linker creates a segment with "
+                                 "RWX permissions (default)">,
+                        Group<grp_general>;
+def no_warn_rwx_segments : FF<"no-warn-rwx-segments">,
+                           HelpText<"Silence warnings for segments with RWX "
+                                    "permissions">,
+                           Group<grp_general>;
 defm Bsymbolic : mDashComp<"Bsymbolic", "Bind references to global symbols to "
                            "the definition within the shared library">,
                  Group<grp_dynlib>;
--- a/test/Common/standalone/WarnRWXSegments/WarnRWXSegments.test
+++ b/test/Common/standalone/WarnRWXSegments/WarnRWXSegments.test
@@ -1,0 +1,35 @@
+#---WarnRWXSegments.test--------------------------------- Executable -----------------#
+#BEGIN_COMMENT
+# Test --[no-]warn-rwx-segments option.
+# By default the linker warns when a PT_LOAD segment has RWX permissions.
+# --no-warn-rwx-segments suppresses the warning.
+# --warn-rwx-segments restores it.
+# Non-PT_LOAD segments with RWX (e.g. PT_GNU_STACK via -z execstack) must
+# not trigger the warning.
+#END_COMMENT
+#START_TEST
+RUN: %clang %clangopts -c %p/Inputs/1.c -o %t.o
+
+# Default: warning is emitted for a PT_LOAD with RWX permissions.
+RUN: %link %linkopts %t.o -T %p/Inputs/rwx.t -o %t.out 2>&1 | %filecheck %s --check-prefix=WARN
+
+# --no-warn-rwx-segments: warning is suppressed.
+RUN: %link %linkopts %t.o -T %p/Inputs/rwx.t --no-warn-rwx-segments -o %t.out 2>&1 | %filecheck %s --allow-empty --check-prefix=NOWARN
+
+# --warn-rwx-segments: warning is emitted (explicit enable).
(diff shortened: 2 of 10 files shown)


Steven6798 (approved):

include/eld/Driver/GnuLinkerOptions.td
```diff
@@ -657,6 +657,14 @@ def no_warn_shared_textrel : FF<"no-warn-shared-textrel">,
                              HelpText<"Don't Warn if the linker adds a "
                                       "DT_TEXTREL">,
                              Group<grp_dynlib>;
+def warn_rwx_segments : FF<"warn-rwx-segments">,
+                        HelpText<"Warn if the linker creates a segment with "
+                                 "RWX permissions (default)">,
```
quic-shaksuma, line 662: Why is "(default)" added here?

quic-shaksuma: nvm :)

test/Common/standalone/WarnRWXSegments/WarnRWXSegments.test
```diff
@@ -0,0 +1,22 @@
...
+# Test --[no-]warn-rwx-segments option.
+# By default the linker warns when a segment has RWX permissions.
+# --no-warn-rwx-segments suppresses the warning.
+# --warn-rwx-segments restores it.
+#END_COMMENT
+#START_TEST
+RUN: %clang %clangopts -c %p/Inputs/1.c -o %t.o
+
+# Default: warning is emitted.
+RUN: %link %linkopts %t.o -T %p/Inputs/rwx.t -o %t.out 2>&1 | %filecheck %s --check-prefix=WARN
```
quic-shaksuma, line 14: can we add a check for no warning emitted when no RWX segment present

test/Common/standalone/WarnRWXSegments/WarnRWXSegments.test
```diff
@@ -0,0 +1,22 @@
...
+
+# --no-warn-rwx-segments: warning is suppressed.
+RUN: %link %linkopts %t.o -T %p/Inputs/rwx.t --no-warn-rwx-segments -o %t.out 2>&1 | %filecheck %s --allow-empty --check-prefix=NOWARN
+
+# --warn-rwx-segments: warning is emitted (explicit enable).
+RUN: %link %linkopts %t.o -T %p/Inputs/rwx.t --warn-rwx-segments -o %t.out 2>&1 | %filecheck %s --check-prefix=WARN
+
+#WARN: Warning: PT_LOAD has RWX permissions
```
quic-shaksuma: Can the segment name be emitted as well not just type?

test/Common/standalone/WarnRWXSegments/WarnRWXSegments.test
```diff
@@ -0,0 +1,22 @@
+#---WarnRWXSegments.test--------------------------------- Executable -----------------#
+#BEGIN_COMMENT
+# Test --[no-]warn-rwx-segments option.
+# By default the linker warns when a segment has RWX permissions.
```
quic-shaksuma: having "-z execstack" would also show this warning for GNU_STACK, should this be noted here?

test/Common/standalone/WarnRWXSegments/WarnRWXSegments.test
quic-seaswara: This handles PHDRS and non PHDR's use case. 

I was thinking to replace this message with output elf has RWX segments instead of per PT_LOAD segment.

test/Common/standalone/WarnRWXSegments/WarnRWXSegments.test
quic-seaswara: interesting usecase. let me get back to you on this.

quic-seaswara: it looks like `bfd` only warns for loadable segments :)

quic-seaswara: nice catch!, thanks!

PR #1629, opened 2026-07-30
[LinkerScript] Support VERSION{} block embedded in a -T linker script

### LINKED ISSUE #1626: VERSION{} block embedded in a -T linker script is silently ignored
eld correctly applies symbol versioning when a version script is passed via --version-script=<file> but silently drops a VERSION { ... } block that is embedded directly inside a normal -T linker script (mixed with SECTIONS/PHDRS in one file). No error/warning, exit 0, but no .gnu.version/.gnu.version_d sections are produced and local: *; never demotes non-matching symbols.

Example
```
cat > foo.c <<'EOF'
int exported_symbol(void) { return 42; }
int internal_symbol(void) { return 1; }
EOF
 
cat > script.lds <<'EOF'
SECTIONS
{
  .text : { *(.text) }
}
 
VERSION
{
  V1 {
  global:
    exported_symbol;
  local:
    *;
  };
}
EOF
 
clang -fPIC -c foo.c -o foo.o
 
# ld.eld
echo "ld.eld output"
ld.eld -shared -T script.lds foo.o -o out.eld.so
llvm-readelf --dyn-syms out.eld.so
llvm-readelf --version-info out.eld.so
 
# ld.lld
echo ""
echo "ld.lld output"
ld.lld -shared -T script.lds foo.o -o out.lld.so
llvm-readelf --dyn-syms out.lld.so
llvm-readelf --version-info out.lld.so
```

Output
```
ld.eld output

Symbol table '.dynsym' contains 3 entries:
   Num:    Value          Size Type    Bind   Vis       Ndx Name
     0: 0000000000000000     0 NOTYPE  LOCAL  DEFAULT   UND 
     1: 0000000000000090    11 FUNC    GLOBAL DEFAULT     4 exported_symbol
     2: 00000000000000a0    11 FUNC    GLOBAL DEFAULT     4 internal_symbol

ld.lld output

Symbol table '.dynsym' contains 2 entries:
   Num:    Value          Size Type    Bind   Vis       Ndx Name
     0: 0000000000000000     0 NOTYPE  LOCAL  DEFAULT   UND 
     1: 0000000000000130    11 FUNC    GLOBAL DEFAULT     8 exported_symbol@@V1
Version symbols section '.gnu.version' contains 2 entries:
 Addr: 0000000000000030  Offset: 0x001030  Link: 1 (.dynsym)
  000:   0 (*local*)       2 (V1)         

Version definition section '.gnu.version_d' contains 2 entries:
 Addr: 0000000000000034  Offset: 0x001034  Link: 6 (.dynstr)
  0x0000: Rev: 1  Flags: BASE  Index: 1  Cnt: 1  Name: out.lld.so
  0x001c: Rev: 1  Flags: none  Index: 2  Cnt: 1  Name: V1
```

Steven6798:
ld.eld only applied symbol versioning when passed via --version-script=<file>; a VERSION{} block embedded directly inside a -T linker script (mixed with SECTIONS/PHDRS) was silently ignored -- no .gnu.version/.gnu.version_d sections, and local: *; never demoted non-matching symbols. ld.lld already handles this. This is the idiom used by VDSO linker scripts in glibc, musl, and the Linux kernel.

Track each -T script's parsed VersionScript on Module as it's activated, and register its nodes from parseVersionScript() alongside --version-script= files, via a shared helper. Registration happens later than parsing because the target backend isn't guaranteed to be initialized yet when a -T script is first read.

Add VersionInLinkerScript test covering a -T script with an embedded VERSION{} block.

Fix: qualcomm#1626

### DIFF
--- a/include/eld/Core/Module.h
+++ b/include/eld/Core/Module.h
@@ -607,6 +607,19 @@ public:
     return VersionScripts;
   }
 
+  // Tracks VersionScript objects parsed from a VERSION{} block embedded
+  // directly inside a -T linker script, as opposed to a standalone
+  // --version-script= file. These still need their nodes registered via
+  // ObjectLinker::parseVersionScript(), just not re-parsed from scratch.
+  void addLinkerScriptVersionScript(const VersionScript *VerScr) {
+    LinkerScriptVersionScripts.push_back(VerScr);
+  }
+
+  const llvm::SmallVectorImpl<const VersionScript *> &
+  getLinkerScriptVersionScripts() const {
+    return LinkerScriptVersionScripts;
+  }
+
   bool isLinkStateBeforeLayout() const {
     return getState() == LinkState::BeforeLayout;
   }
@@ -719,6 +732,7 @@ private:
       DynamicListFileToScriptSymbolsMap;
   llvm::StringSet<> OutputSectDescNameSet;
   llvm::SmallVector<const VersionScript *> VersionScripts;
+  llvm::SmallVector<const VersionScript *> LinkerScriptVersionScripts;
   llvm::DenseMap<Fragment *, uint64_t> FragmentPaddingValues;
   PluginManager PM;
   NamePool SymbolNamePool;
--- a/lib/Object/ObjectLinker.cpp
+++ b/lib/Object/ObjectLinker.cpp
@@ -247,6 +247,12 @@ bool ObjectLinker::readLinkerScript(InputFile *Input) {
       return false;
   }
 
+  // A -T script may embed its own VERSION{} block. Record it now so
+  // parseVersionScript() can register its nodes later, once the target
+  // backend is guaranteed to be initialized, which is not yet the case here.
+  if (S->getVersionScript())
+    ThisModule->addLinkerScriptVersionScript(S->getVersionScript());
+
   Input->setUsed(true);
 
   return true;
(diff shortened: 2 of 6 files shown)


quic-seaswara (approved):

Steven6798:
Example of the fix in action:

Before the fix: https://github.com/qualcomm/eld/actions/runs/30499885360/job/90737005902
```
  VDSOLD  arch/riscv/kernel/vdso/vdso.so.dbg
Warning: Unrecognized option `--no-warn-rwx-segments'
arch/riscv/kernel/vdso/vdso.so.dbg: dynamic relocations are not supported
make[2]: *** [arch/riscv/kernel/vdso/Makefile:81: arch/riscv/kernel/vdso/vdso.so.dbg] Error 1
make[1]: *** [arch/riscv/Makefile:162: vdso_prepare] Error 2
Error: Process completed with exit code 2.
```

After the fix: https://github.com/qualcomm/eld/actions/runs/30572034438/job/90970877516

```
  VDSOLD  arch/riscv/kernel/vdso/vdso.so.dbg
Warning: Unrecognized option `--no-warn-rwx-segments'
  VDSOSYM include/generated/vdso-offsets.h
```

quic-shaksuma (approved):

include/eld/Core/Module.h
```diff
@@ -607,6 +607,19 @@ class Module {
...
 
+  // Tracks VersionScript objects parsed from a VERSION{} block embedded
+  // directly inside a -T linker script, as opposed to a standalone
+  // --version-script= file. These still need their nodes registered via
+  // ObjectLinker::parseVersionScript(), just not re-parsed from scratch.
+  void addLinkerScriptVersionScript(const VersionScript *VerScr) {
+    LinkerScriptVersionScripts.push_back(VerScr);
+  }
+
+  const llvm::SmallVectorImpl<const VersionScript *> &
```
parth-07, line 618: Why are we using `SmallVectorImpl` instead of `SmallVector`?

lib/Object/ObjectLinker.cpp
```diff
@@ -350,63 +356,79 @@ bool ObjectLinker::normalize() {
...
+                                      VersionScriptInput->decoratedPath()))
+        return false;
+    }
+  }
+
+  // VersionScript objects parsed from a VERSION{} block embedded directly
+  // inside a -T linker script (recorded by readLinkerScript(), which runs
+  // before the target backend is guaranteed to exist). Process them here,
+  // where registerVersionScriptNodes() can safely touch the backend.
+  for (const VersionScript *VS : ThisModule->getLinkerScriptVersionScripts()) {
```
parth-07, line 396: We are breaking the input order of the version scripts here. We are unconditionally first reading version script specified using `--version-script`, and the reading the VERSION nodes specified in the other linker scripts. Can you please check and update how GNU ld behaves here?

lib/Object/ObjectLinker.cpp
```diff
@@ -350,63 +356,79 @@ bool ObjectLinker::normalize() {
...
+    if (!registerVersionScriptNodes(
```
parth-07, line 397: What does `register` mean here? What do you think about renaming it to `addVersionScriptNodes` based on the `Module::addVersionScriptNode`?

parth-07 (changes requested):
We report version script nodes in the map-file as well. It seems that we do not report the VERSION nodes that came from the ordinary linker scripts. Can you please take a look into it?

include/eld/Core/Module.h
Steven6798: Following the implementation of the similar existing function "getVersionScripts()"

lib/Object/ObjectLinker.cpp
Steven6798: I'll look into it.

lib/Object/ObjectLinker.cpp
Steven6798: I can do that. I tried not to have the same name to prevent confusion, that's why I used a similar name.

lib/Object/ObjectLinker.cpp
Steven6798: This needs more research as the behavior is inconsistent between linkers. Since this is related to symbol versioning and out of the scope of this PR, can you handle this as a follow up? This fix is needed for the Linux kernel workflow and should not break anything as this is something that was not supported in the first place.

parth-07: Sure. Sounds good.

parth-07: Can you please create an issue to track this?

Steven6798:
Added 2 follow ups:
https://github.com/qualcomm/eld/issues/1636
https://github.com/qualcomm/eld/issues/1635

parth-07 (approved):
LGTM.

PR #1630, opened 2026-07-31
Add --emit-symbol-resolution-report option for a standalone JSON report

parth-07:
This commit moves the symbol resolution report out of the text map file into a
dedicated `--emit-symbol-resolution-report=<file>` option that emits the report
as standalone JSON. It also removes the old `--MapDetail show-symbol-resolution`
sub-option that previously embedded the report inside the text map file.

The emitted JSON is versioned and records `Symbols` as a flat array in the same
sequence that the linker observes symbols. Each array element describes one
symbol record with its input file, section, size, binding, type, visibility,
bitcode origin, selection state, and plugin information when applicable.
Duplicate names are expected in the array, so consumers can inspect the actual
symbol resolution directly.

For bitcode symbols, the report includes the corresponding post-LTO object symbol
when available. Binding and visibility are always present so consumers see a
complete picture instead of the abbreviated form the text report used to print.

A minimal standalone python inspector tool is added under `tools/SymbolResolutionInspector`
to query the report by symbol name or by rendered symbol description, and to re-emit a
filtered report as JSON.

This is the schema used for symbol resolution report:

```
{
"SymbolResolutionReportVersion": 1,
"Symbols": [
  {
    "Name": "foo",
    "InputFile": "libc.a(malloc.o)",
    "Section": ".text",    // only when there is an associated section
    "Plugin": "MyPlugin",  // only for plugin-created symbol
    "Size": 4,
    "Bitcode": false,
    "SectionIndexKind": "Def",
    "Binding": "Global",
    "Type": "Object",
    "Visibility": "Default",
    "IsSelected": true,
    // only if post-LTO symbol exists. Additionally, it requires
    // LTO + linker script support to work.
    "LTOObjectSymbol": { ...same structure as top-level symbol... },
    ...
    ...
  }
 ]
}
```

### DIFF
--- a/include/eld/Config/GeneralOptions.h
+++ b/include/eld/Config/GeneralOptions.h
@@ -522,6 +522,19 @@ public:
     return ArchiveMemberReportFile;
   }
 
+  // --emit-symbol-resolution-report
+  void setSymbolResolutionReportFile(llvm::StringRef File) {
+    SymbolResolutionReportFile = File.str();
+  }
+
+  const std::optional<std::string> &getSymbolResolutionReportFile() const {
+    return SymbolResolutionReportFile;
+  }
+
+  bool shouldEmitSymbolResolutionReport() const {
+    return SymbolResolutionReportFile.has_value();
+  }
+
   // --ld-generated-unwind-info
   void setGenUnwindInfo(bool PEnable = true) { BGenUnwindInfo = PEnable; }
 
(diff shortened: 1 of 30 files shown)


quic-seaswara (changes requested):
SymbolResolutionInspector needs to be in utils. 

Also add documentation of how to use it.

parth-07:
> SymbolResolutionInspector needs to be in utils.
>
> Also add documentation of how to use it.

Done!

quic-shaksuma (changes requested):
The eld documentation has migrated to markdown, please update the new docs

parth-07:
> The eld documentation has migrated to markdown, please update the new docs

Done!

quic-shaksuma (approved):

quic-seaswara (approved):

PR #1632, opened 2026-08-03
[TLS] Fix TLS template size with multiple padded PT_TLS segments

quic-k:
setupProgramHdrs() summed each PT_TLS segment's p_memsz to get the TLS template size. When the TLS block spans multiple PT_TLS segments (e.g. picolibc's separate .tdata/.tbss segments) and one is over-aligned, summing drops the alignment padding between them, producing a size that is too small.

That size is subtracted from a symbol's TLS offset when resolving TP-relative relocations (R_HEX_TPREL_*), so an undersized value shifts every thread-local access by the missing padding and disagrees with the runtime thread-pointer setup.

Compute the size as the span across all PT_TLS segments instead: max(vaddr + memsz) - min(vaddr), which includes inter-segment padding.

### DIFF
--- a/lib/Target/GNULDBackend.cpp
+++ b/lib/Target/GNULDBackend.cpp
@@ -99,6 +99,7 @@
 #include <chrono>
 #include <climits>
 #include <cstring>
+#include <limits>
 #include <map>
 #include <string>
 #include <vector>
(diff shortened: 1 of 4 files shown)


quic-seaswara (approved):

lib/Target/GNULDBackend.cpp
```diff
@@ -2599,10 +2600,21 @@ bool GNULDBackend::setupProgramHdrs() {
...
+    // drop that inter-segment padding and produce a TLS template size that
+    // disagrees with the runtime thread-pointer setup, corrupting every
+    // TP-relative (TPREL) relocation.
+    uint64_t lo = std::numeric_limits<uint64_t>::max();
+    uint64_t hi = 0;
+    for (auto &seg : tls_segs) {
+      lo = std::min(lo, seg->vaddr());
+      hi = std::max(hi, seg->vaddr() + seg->memsz());
+    }
+    setTLSTemplateSize(hi - lo);
```
quic-seaswara, line 2617: What if there are other sections in between ? 

The picolibc linker script defines other sections like `.data` between TLS sections.

quic-seaswara (changes requested):

quic-seaswara (approved):

lib/Target/GNULDBackend.cpp
quic-k: I don't think its valid for a TLS template to have other sections in between
looking at Picolibc's default linker, I don't see any other sections between tdata and tbss
https://github.com/picolibc/picolibc/blob/main/picolibc.ld.in#L155-L175

PR #1634, opened 2026-08-04
[Patching] Remove --patch-enable and entire patch infrastructure

quic-seaswara:
The runtime symbol patching feature (--patch-enable / --patch-base) is no longer needed and is being removed to reduce linker complexity.


lenary (approved):
LGTM.

I don't recall the status of this work in clang, but I don't think anything landed there.

quic-shaksuma (approved):

PR #1642, opened 2026-08-05
linkerscript: fix ENTRY() last-wins 

### LINKED ISSUE #1538: Last `ENTRY` command in the linker script should take affect
The general programming convention for non-appending allowed repeatable commands and options is that the last option takes effect. It is helpful in during development and debugging to quickly override an option/command. We do not have this conventional behavior for the `ENTRY` linker script command.

Reproducer:

```bash
#!/usr/bin/env bash

cat >1.c <<\EOF
int bar();

int foo() { return bar(); }
EOF

cat >2.c <<\EOF
int baz() { return 1; }
int bar() { return 3; }
EOF

cat >3.c <<\EOF
int car() { return 5; }
EOF

cat >script.t <<\EOF
ENTRY(bar);
ENTRY(foo); 
EOF


TARGET="x86_64-linux-gnu"
llvm-ar cr lib2.a 2.o

LDs=(ld.eld ld.lld ld.bfd)
SFs=(eld lld bfd)

for i in "${!SFs[@]}"; do
  ${LDs[$i]} -o a.${SFs[$i]}.out 3.o lib2.a 1.o -T script.t -Map a.${SFs[$i]}.map.txt
done
```

Output:

```bash
+ for i in "${!SFs[@]}"
+ ld.eld -o a.eld.out 3.o lib2.a 1.o -T script.t -Map a.eld.map.txt
+ for i in "${!SFs[@]}"
+ ld.lld -o a.lld.out 3.o lib2.a 1.o -T script.t -Map a.lld.map.txt
+ for i in "${!SFs[@]}"
+ ld.bfd -o a.bfd.out 3.o lib2.a 1.o -T script.t -Map a.bfd.map.txt
ld.bfd: 1.o: in function `foo':
1.c:(.text.foo+0x7): undefined reference to `bar'
```

eld and lld takes the (conventionally incorrect) value `bar` for `ENTRY` command here, whereas bfd takes the (conventionally correct) value `foo` for `ENTRY` command.

rachitmeht:
EntryCmd::activate() guarded with hasEntry(), making the first ENTRY() win over later ones and allowing script ENTRY() to overwrite a -e value set on the command line.

Similar to GNU ld, replace the guard with !isEntryFromCmdLine() so each ENTRY() overwrites the previous (last wins), while -e is always preserved regardless of script processing order.

Add EntryCmdPrecedence.test covering: last-wins within one script, -e override, last -e wins, raw-address -e 0, cross-script last-wins, and -e placed before -T.

Resolves #1538

### DIFF
--- a/include/eld/Config/GeneralOptions.h
+++ b/include/eld/Config/GeneralOptions.h
@@ -759,6 +759,10 @@ public:
 
   bool hasEntry() const;
 
+  void setEntryFromCmdLine();
+
+  bool isEntryFromCmdLine() const;
+
   llvm::ArrayRef<std::string> mapStyle() const { return MapStyles; }
 
   bool setMapStyle(llvm::StringRef MapStyle);
@@ -1395,6 +1399,7 @@ private:
   std::vector<std::string> LTOOutputFile;
   std::optional<uint64_t> ImageBase; // --image-base=value
   std::string Entry;
+  bool EntryFromCmdLine = false;
   SymbolRenameMap SymbolRenames;
   AddressMapType AddressMap;
   std::vector<const char *> CommandLineArgs;
--- a/include/eld/Diagnostics/DiagLDScript.inc
+++ b/include/eld/Diagnostics/DiagLDScript.inc
@@ -39,6 +39,8 @@ DIAG(assert_failed, DiagnosticEngine::Error, "Assertion failed %0")
 DIAG(error_printcmd, DiagnosticEngine::Error, "%0: PRINT: %1")
 DIAG(linker_script_uses_phdrs_no_sections, DiagnosticEngine::Error,
      "Linker Script is using PHDR's but not using SECTIONS command")
+DIAG(warn_multiple_entry, DiagnosticEngine::Warning,
+     "multiple entry points specified; the last one takes effect")
 DIAG(cannot_set_at_address, DiagnosticEngine::Error,
      "Address for section %0 specified with AT cannot be set, as there is a "
      "dependency issue, check the map file for more analysis")
--- a/lib/LinkerWrapper/GnuLdDriver.cpp
+++ b/lib/LinkerWrapper/GnuLdDriver.cpp
@@ -722,6 +722,7 @@ bool GnuLdDriver::processOptions(llvm::opt::InputArgList &Args) {
   // set up entry point from -e
   if (llvm::opt::Arg *arg = Args.getLastArg(T::entrypoint)) {
     Config.options().setEntry(arg->getValue());
+    Config.options().setEntryFromCmdLine();
     Config.addCommandLine(Table->getOptionName(T::entrypoint), arg->getValue());
   }
 
--- a/lib/LinkerWrapper/GnuLdDriver.cpp
+++ b/lib/LinkerWrapper/GnuLdDriver.cpp
@@ -719,7 +719,7 @@ bool GnuLdDriver::processOptions(llvm::opt::InputArgList &Args) {
       T::fatal_internal_errors, T::no_fatal_internal_errors, /*default=*/false);
   Config.options().setFatalInternalErrors(enableFatalInternalErrors);
 
-  // set up entry point from -e
+  // set up entry point from -e;
   if (llvm::opt::Arg *arg = Args.getLastArg(T::entrypoint)) {
     Config.options().setEntry(arg->getValue());
     Config.options().setEntryFromCmdLine();
--- a/lib/Script/EntryCmd.cpp
+++ b/lib/Script/EntryCmd.cpp
@@ -32,7 +32,8 @@ void EntryCmd::dump(llvm::raw_ostream &Outs) const {
 }
 
 eld::Expected<void> EntryCmd::activate(Module &CurModule) {
-  if ((!EntrySymbol.empty()) && (!(CurModule.getConfig().options().hasEntry())))
-    CurModule.getConfig().options().setEntry(EntrySymbol);
+  GeneralOptions &Options = CurModule.getConfig().options();
+  if (!EntrySymbol.empty() && !Options.isEntryFromCmdLine())
+    Options.setEntry(EntrySymbol);
   return eld::Expected<void>();
 }
--- a/lib/Script/EntryCmd.cpp
+++ b/lib/Script/EntryCmd.cpp
@@ -33,7 +33,11 @@ void EntryCmd::dump(llvm::raw_ostream &Outs) const {
 
 eld::Expected<void> EntryCmd::activate(Module &CurModule) {
   GeneralOptions &Options = CurModule.getConfig().options();
-  if (!EntrySymbol.empty() && !Options.isEntryFromCmdLine())
-    Options.setEntry(EntrySymbol);
+  if (!EntrySymbol.empty()) {
+    if (Options.hasEntry() && CurModule.getConfig().showLinkerScriptWarnings())
+      CurModule.getConfig().raise(Diag::warn_multiple_entry);
+    if (!Options.isEntryFromCmdLine())
+      Options.setEntry(EntrySymbol);
+  }
   return eld::Expected<void>();
 }
--- a/test/Common/standalone/linkerscript/EntryCmd/EntryCmdPrecedence.test
+++ b/test/Common/standalone/linkerscript/EntryCmd/EntryCmdPrecedence.test
@@ -1,0 +1,50 @@
+#---EntryCmdPrecedence.test-------------- Executable,LS ----------------------#
+#BEGIN_COMMENT
+# Tests GNU ld entry point precedence rules:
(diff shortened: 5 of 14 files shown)


include/eld/Config/GeneralOptions.h
```diff
@@ -759,6 +754,10 @@ class GeneralOptions {
 
   bool hasEntry() const;
 
+  void setEntryFromCmdLine();
+
+  bool isEntryFromCmdLine() const;
+
```
quic-seaswara, line 760: Why do we need this ?

lib/Script/EntryCmd.cpp
```diff
@@ -32,7 +32,8 @@ void EntryCmd::dump(llvm::raw_ostream &Outs) const {
 }
 
 eld::Expected<void> EntryCmd::activate(Module &CurModule) {
-  if ((!EntrySymbol.empty()) && (!(CurModule.getConfig().options().hasEntry())))
-    CurModule.getConfig().options().setEntry(EntrySymbol);
+  GeneralOptions &Options = CurModule.getConfig().options();
+  if (!EntrySymbol.empty() && !Options.isEntryFromCmdLine())
+    Options.setEntry(EntrySymbol);
```
quic-seaswara: how do we handle 

ld.eld -e <entry symbol> -T <script> 

and the script has ENTRY ?

Can we add a warning with -Wlinker-script when we see multiple ENTRY commands mentioned in the link line either in command line or linker script ?

test/Common/standalone/linkerscript/EntryCmd/EntryCmdPrecedence.test
```diff
@@ -0,0 +1,50 @@
+#---EntryCmdPrecedence.test-------------- Executable,LS ----------------------#
+#BEGIN_COMMENT
+# Tests GNU ld entry point precedence rules:
+#   Case 1: Last ENTRY() in a single script wins over earlier ENTRY() commands.
+#   Case 2: Command-line -e always overrides any script ENTRY().
+#   Case 3: Last -e on command line wins over earlier -e.
+#   Case 4: -e with a raw address overrides script ENTRY().
+#   Case 5: Across multiple -T scripts, the last ENTRY() activated wins.
+#   Case 6: -e overrides script ENTRY() regardless of command-line order
+#           (i.e. -e placed before -T still wins).
```
quic-seaswara, line 10: Please document this as well.

quic-seaswara (changes requested):

lib/Script/EntryCmd.cpp
rachitmeht: It is handled similarly, how GNU ld handles, the command line -e will always take precedence, irrespective of linker script's ENTRY(). To keep track of this we need bool EntryFromCmdLine.  
For multiple -e currently last one is considered

include/eld/Config/GeneralOptions.h
rachitmeht: To maintain the precedence of -e over linker Script's ENTRY()

lib/LinkerWrapper/GnuLdDriver.cpp
```diff
@@ -719,9 +710,16 @@ bool GnuLdDriver::processOptions(llvm::opt::InputArgList &Args) {
       T::fatal_internal_errors, T::no_fatal_internal_errors, /*default=*/false);
   Config.options().setFatalInternalErrors(enableFatalInternalErrors);
 
-  // set up entry point from -e
+  // set up entry point from -e; warn if specified more than once
+  unsigned NumEntryArgs = 0;
+  for (auto *Arg : Args.filtered(T::entrypoint))
+    ++NumEntryArgs;
+  if (NumEntryArgs > 1 && Config.showLinkerScriptWarnings())
```
parth-07, line 717: Why is multiple `-e` command-line options part of the linker script warnings?

lib/LinkerWrapper/GnuLdDriver.cpp
```diff
@@ -719,9 +710,16 @@ bool GnuLdDriver::processOptions(llvm::opt::InputArgList &Args) {
...
+    Config.raise(Diag::warn_multiple_entry);
+
```
parth-07, line 719: I do not think we should warn for multiple -e options, and if we should, then it should be under some command-line warning category.

lib/Script/EntryCmd.cpp
```diff
@@ -32,7 +32,12 @@ void EntryCmd::dump(llvm::raw_ostream &Outs) const {
 }
 
 eld::Expected<void> EntryCmd::activate(Module &CurModule) {
-  if ((!EntrySymbol.empty()) && (!(CurModule.getConfig().options().hasEntry())))
-    CurModule.getConfig().options().setEntry(EntrySymbol);
+  GeneralOptions &Options = CurModule.getConfig().options();
+  if (!EntrySymbol.empty()) {
```
parth-07, line 36: I know it was part of the original code, but do you know if there is any valid case where `EntrySymbol` may be empty?

lib/LinkerWrapper/GnuLdDriver.cpp
rachitmeht: I understood the warning should be under -Wcommand-line, but if we are warning on multiple ENTRY() then we should also warn for multiple -e. aren't they the same case?

lib/Script/EntryCmd.cpp
rachitmeht: Syntactically we can pass ENTRY("") it will result in  EntrySymbol.empty(),  

lld do gives warning in this case but ld.eld and ld.bfd silently ignores it 
```
  Entry point address:               0x0
```

lib/LinkerWrapper/GnuLdDriver.cpp
parth-07: > but if we are warning on multiple ENTRY() then we should also warn for multiple -e. aren't they the same case?

They are similar for sure, but passing same options multiple times is well-defined (and often well-used) behavior. And if we are giving warning on multiple `-e` options, then we should likely do the same for other options as well which can be repeated and last one takes effect. This being said, I personally do not think we should report warning on repeated command-line options. @quic-seaswara what do you think about it?

parth-07: @rachitmeht for now can you please remove warning from here?

lib/Script/EntryCmd.cpp
parth-07: This is interesting. We should also give warning for it. Can you please raise an issue for this?

rachitmeht: ENTRY("") is treated as having no ENTRY() at all.  
lld gives this warning if we have not specified an entry point at all.

lib/LinkerWrapper/GnuLdDriver.cpp
```diff
@@ -719,9 +710,10 @@ bool GnuLdDriver::processOptions(llvm::opt::InputArgList &Args) {
       T::fatal_internal_errors, T::no_fatal_internal_errors, /*default=*/false);
   Config.options().setFatalInternalErrors(enableFatalInternalErrors);
 
-  // set up entry point from -e
+  // set up entry point from -e;
```
parth-07:
```suggestion
  // set up entry point from -e
```

parth-07:
@rachitmeht Can you please reply/mark-resolved the comments which are now resolved?

include/eld/Diagnostics/DiagLDScript.inc
```diff
@@ -39,6 +39,8 @@ DIAG(assert_failed, DiagnosticEngine::Error, "Assertion failed %0")
 DIAG(error_printcmd, DiagnosticEngine::Error, "%0: PRINT: %1")
 DIAG(linker_script_uses_phdrs_no_sections, DiagnosticEngine::Error,
      "Linker Script is using PHDR's but not using SECTIONS command")
+DIAG(warn_multiple_entry, DiagnosticEngine::Warning,
+     "multiple entry points specified; the last one takes effect")
```
parth-07: This diagnostic is incorrect/misleading for this case: `ld.eld -e main script.t`, if `script.t` defines multiple ENTRY commands, then the diagnostic will be emitted. However, the first entry pointer specified (`-e main`) is taking effect instead of the last.

parth-07 (changes requested):

parth-07:
@rachitmeht I hate to ask the same thing multiple times. But can you please reply/mark-resolve the comment which you have resolved/made-progress on and re-request the review after making changes?

rachitmeht:
> Rachit Mehta (@rachitmeht) I hate to ask the same thing multiple times. But can you please reply/mark-resolve the comment which you have resolved/made-progress on and re-request the review after making changes?

I apologize, I will take care next time.

parth-07 (approved):

parth-07:
@rachitmeht Can you please rebase the changes on top of the current main?

PR #1646, opened 2026-08-06
[X86] Fix R_X86_64_TPOFF against weak undefined TLS symbols

### LINKED ISSUE #1540: Weak undefined symbols causing wrong tpoff in Local Exec model
Reproduce :
```bash
cat > weak_tls.c <<\EOF
  extern __thread int weak_tls __attribute__((weak));
  __attribute__((visibility("hidden"))) __thread int real_tls;
  int *get_weak(void) { return &weak_tls; }
EOF

clang -target x86_64-linux-gnu -ftls-model=local-exec -c weak_tls.c -o weak_tls.o
LDs=(ld.eld ld.lld ld.bfd ld.mold ld.gold)
SFs=(eld    lld    bfd    mold    gold)

for i in "${!SFs[@]}"; do
  ${LDs[$i]} -static -e get_weak -o ${SFs[$i]}.out weak_tls.o
done
```

Results (real_tls = 4 bytes, templateSize = 4):
    eld:   lea    -0x3(%rax),%rax       tpoff = -3       
    lld:   lea    0x0(%rax),%rax        tpoff = 0
    gold:  lea    -0x4(%rax),%rax       tpoff = -4
    bfd:   lea    -0x404004(%rax),%rax  tpoff = -tp_addr
    mold:  lea    -0x2022ac(%rax),%rax  tpoff = -tp_addr

rachitmeht:
A weak undefined TLS symbol has no slot in the PT_TLS segment, so the thread-pointer-offset formula (S - templateSize) is meaningless for it. Previously relocTPOFF applied that formula with a bogus symbol value, producing an incorrect tpoff (e.g. -3) instead of the correct 0.

Two fixes, comparing lld's R_TPREL handling:

- relocTPOFF (x86_64Relocator.cpp): short-circuit weak undefined symbols in a static executable, returning the addend directly instead of running the tpoff formula.

Add x86_64/linux/WeakUndefTLS covering TPOFF32/64 and DTPOFF32/64 against weak-undef symbols, a defined-symbol regression guard, and a non-zero addend case.

Resolves #1540

### DIFF
--- a/lib/Target/X86/x86_64Relocator.cpp
+++ b/lib/Target/X86/x86_64Relocator.cpp
@@ -752,6 +752,13 @@ Relocator::Result eld::relocTPOFF(Relocation &pReloc, x86_64Relocator &pParent,
 
   uint64_t TLSTemplateSize = pParent.getTarget().getTLSTemplateSize();
 
+  ResolveInfo *rsym = pReloc.symInfo();
+  if (rsym && rsym->isWeakUndef() &&
+      (pParent.config().codeGenType() == LinkerConfig::Exec)) {
+    Relocator::DWord A = pReloc.addend();
+    return ApplyReloc(pReloc, A, pRelocDesc, DiagEngine, options, pParent);
+  }
+
   if (TLSTemplateSize == 0) {
     pParent.config().raise(Diag::no_pt_tls_segment);
     return Relocator::BadReloc;
--- a/lib/Target/X86/x86_64Relocator.cpp
+++ b/lib/Target/X86/x86_64Relocator.cpp
@@ -754,7 +754,7 @@ Relocator::Result eld::relocTPOFF(Relocation &pReloc, x86_64Relocator &pParent,
 
   ResolveInfo *rsym = pReloc.symInfo();
   if (rsym && rsym->isWeakUndef() &&
-      (pParent.config().codeGenType() == LinkerConfig::Exec)) {
+      pParent.config().isBuildingExecutable()) {
     Relocator::DWord A = pReloc.addend();
     return ApplyReloc(pReloc, A, pRelocDesc, DiagEngine, options, pParent);
   }
--- a/test/x86_64/linux/WeakUndefTLS/Inputs/weak_tls_ld.s
+++ b/test/x86_64/linux/WeakUndefTLS/Inputs/weak_tls_ld.s
@@ -1,0 +1,6 @@
+.weak weak_tls
+.section .data
+.align 4
+result:
+  .long 0xdeadbeef
+  .reloc result, R_X86_64_DTPOFF32, weak_tls
--- a/test/x86_64/linux/WeakUndefTLS/WeakUndefTLS.test
+++ b/test/x86_64/linux/WeakUndefTLS/WeakUndefTLS.test
@@ -1,0 +1,61 @@
+#--WeakUndefTLS.test----------Executable--------#
+BEGIN_COMMENT
+# Test R_X86_64_TPOFF32/64 and R_X86_64_DTPOFF32/64 relocations against
+# weak undefined TLS symbols. A weak undefined TLS symbol has no TLS slot,
+# so the tpoff formula must not run; the result is the addend only (= 0
+# for the common case). Mirrors lld's R_TPREL guard:
+#   if (sym.isUndefined()) return a;
+# Also verifies that defined TLS symbols are unaffected (regression guard).
+#END_COMMENT
+
+#START_TEST
+
+# Case 1: TPOFF32 + weak-undef -> 0x00000000
+RUN: %clang %clangopts -c %p/Inputs/weak_tls_le.s -o %t.le.o
+RUN: %link %linkopts -static %t.le.o -o %t.le.exe
+RUN: %objdump -s -j .data %t.le.exe | %filecheck %s --check-prefix=LE32
+
+# Case 2: TPOFF64 + weak-undef -> 0x0000000000000000
+RUN: %clang %clangopts -c %p/Inputs/weak_tls_le64.s -o %t.le64.o
+RUN: %link %linkopts -static %t.le64.o -o %t.le64.exe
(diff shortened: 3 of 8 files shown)


lib/Target/GNULDBackend.cpp
```diff
@@ -760,11 +759,8 @@ uint64_t GNULDBackend::finalizeTLSSymbol(LDSymbol *pSymbol) {
   }
 
   // ignore if symbol has no fragRef
-  // FIXME: This is probably wrong. If a symbol does not have a fragment ref
-  // that implies the symbol is undefined, in that case, the symbol index should
-  // be 0 instead of 1 (true).
   if (!pSymbol->hasFragRef())
-    return true;
+    return false;
```
quic-seaswara: This should be a seperate change.

lib/Target/X86/x86_64Relocator.cpp
```diff
@@ -752,6 +752,13 @@ Relocator::Result eld::relocTPOFF(Relocation &pReloc, x86_64Relocator &pParent,
 
   uint64_t TLSTemplateSize = pParent.getTarget().getTLSTemplateSize();
 
+  ResolveInfo *rsym = pReloc.symInfo();
+  if (rsym && rsym->isWeakUndef() &&
+      (pParent.config().codeGenType() == LinkerConfig::Exec)) {
+    Relocator::DWord A = pReloc.addend();
+    return ApplyReloc(pReloc, A, pRelocDesc, DiagEngine, options, pParent);
+  }
```
quic-seaswara, line 760: What about shared/pie ? 

what is the difference for other targets ?

test/x86_64/linux/WeakUndefTLS/WeakUndefTLS.test
```diff
@@ -0,0 +1,61 @@
...
+
+# Case 5: TPOFF32 + defined symbol -> -4 = 0xfffffffc (regression guard)
+RUN: %clang %clangopts -c %p/Inputs/weak_tls_defined.s -o %t.def.o
+RUN: %link %linkopts -static %t.def.o -o %t.def.exe
+RUN: llvm-objdump -s -j .data %t.def.exe | FileCheck %s --check-prefix=DEFINED
+
+# Case 6: TPOFF32 + weak-undef + addend=2 -> 0x00000002
+RUN: %clang %clangopts -c %p/Inputs/weak_tls_le_addend.s -o %t.addend.o
+RUN: %link %linkopts -static %t.addend.o -o %t.addend.exe
+RUN: llvm-objdump -s -j .data %t.addend.exe | FileCheck %s --check-prefix=ADDEND
```
quic-seaswara: %objdump

quic-seaswara (changes requested):

lib/Target/X86/x86_64Relocator.cpp
rachitmeht: TPOFF handling for Weak undef is not done for other architectures also.

rachitmeht: I copied the check from relocAbs(), it mentions that  weak-undef symbol resolving to 0 is only the correct link-time behavior in a static executable.  So, for a shared/PIE, an unresolved symbol should be handled via dynamic relocations.

parth-07 (changes requested):
We are fixing TPOFF64 in this PR as well. Can you please update the commit title and the PR title to indicate that?

Additionally, eld only errors out on encountering TPOFF relocation in shared library because there is no PT_TLS segment. This is misleading and not a good error message. LLD properly reports that the relocation is unsupported for shared library. Can you please create an issue regarding this?

Reproducer:

```
#!/usr/bin/env bash

# Assembly is needed to force exact local-exec TPOFF relocations in data.
#cat > tmp/tpoff32-shared.s <<\EOF
#.weak weak_tls
#.section .data
#.align 4
#result:
#  .long 0xdeadbeef
#  .reloc result, R_X86_64_TPOFF32, weak_tls
#EOF

cat > tmp/tpoff64-shared.s <<\EOF
.weak weak_tls
.section .data
.align 8
result:
  .quad 0xdeadbeefdeadbeef
  .reloc result, R_X86_64_TPOFF64, weak_tls
EOF

"$BD/bin/clang" -target x86_64-linux-gnu -c tmp/tpoff32-shared.s -o tmp/tpoff32-shared.o
"$BD/bin/clang" -target x86_64-linux-gnu -c tmp/tpoff64-shared.s -o tmp/tpoff64-shared.o

LDs=(ld.eld ld.lld ld.bfd)
SFs=(eld lld bfd)

for obj in tpoff64-shared; do
  for i in "${!SFs[@]}"; do
    ${LDs[$i]} -m elf_x86_64 -shared tmp/$obj.o -o tmp/$obj.${SFs[$i]}.so
  done
done
```

lib/Target/X86/x86_64Relocator.cpp
rachitmeht: PIE must also handle weak-undefined TPOFF relocations same as static excecutable. When the PIE contains only a weak-undefined TLS reference, no PT_TLS segment is created; computeTLSOffsets() leaves TLSTemplateSize as zero, causing relocTPOFF() to emit the PT_TLS segment not found error instead of resolving the relocation to its addend, as lld does.

parth-07:
@rachitmeht Can you please add a comment/mark-comment-as-resolved in the PR whenever you address/resolve one of the comments?

parth-07 (approved):

test/x86_64/linux/WeakUndefTLS/Inputs/weak_tls_ld.s
```diff
@@ -0,0 +1,6 @@
+.weak weak_tls
+.section .data
```
BhoumikPatidar: `.weak` has symbol type  NOTYPE, not TLS. 
Should we add `.type weak_tls, @tls_object`? With this change the DTPOFF check fails. Is DTPOFF intended to be covered in this PR?

Also we should have a separate defined TLS symbol in the test case so as to reproduce the -3 bug identically

Issue #1004, opened 2026-04-01
Implement TLS relaxations


quic-seaswara:
Implement TLS relaxations for 

- AArch64
- RISC-V 
- ARM

Create sub tickets as necessary

jonathonpenix:
FWIW the RISC-V tp-rel (non tlsdesc) relaxations were started here: https://github.com/qualcomm/eld/pull/1534

There's a bit of follow-up work still needed around either erroring or preventing out of range relaxations though.

PR #1658, opened 2026-08-07
[AArch64] Fix IE->LE TLS relaxation for global symbols in static executables

deepakshirkem:
IE->LE TLS relaxation was only applied to local symbols. Global TLS symbols in static executables were incorrectly doing a GOT-relative  load (adrp + ldr) instead of relaxing to direct TP-relative instructions (movz + movk).

part of #1004

### DIFF
--- a/lib/Target/AArch64/AArch64Relocator.cpp
+++ b/lib/Target/AArch64/AArch64Relocator.cpp
@@ -543,11 +543,12 @@ void AArch64Relocator::scanGlobalReloc(InputFile &pInput, Relocation &pReloc,
     // return if we already create GOT for this symbol
     if (rsym->reserved() & ReserveGOT)
       return;
-
+    if (config().isCodeStatic())
+      return;
     // set up the got and the corresponding rel entry
     AArch64GOT *G = m_Target.createGOT(GOT::TLS_IE, Obj, rsym);
-    if (config().isCodeStatic() || (config().isBuildingExecutable() &&
-                                    !m_Target.isSymbolPreemptible(*rsym))) {
+    if (config().isBuildingExecutable() &&
+        !m_Target.isSymbolPreemptible(*rsym)) {
       rsym->setReserved(rsym->reserved() | ReserveGOT);
       G->setValueType(GOT::TLSStaticSymbolValue);
       return;
--- a/lib/Target/AArch64/AArch64Relocator.cpp
+++ b/lib/Target/AArch64/AArch64Relocator.cpp
@@ -543,7 +543,9 @@ void AArch64Relocator::scanGlobalReloc(InputFile &pInput, Relocation &pReloc,
     // return if we already create GOT for this symbol
     if (rsym->reserved() & ReserveGOT)
       return;
-
+   
+    if (config().isCodeStatic())
+      return;
     // set up the got and the corresponding rel entry
     AArch64GOT *G = m_Target.createGOT(GOT::TLS_IE, Obj, rsym);
     if (config().isCodeStatic() || (config().isBuildingExecutable() &&
@@ -1185,7 +1187,8 @@ Relocator::Result tls_gottprel_page(Relocation &pReloc,
   DiagnosticEngine *DiagEngine = pParent.config().getDiagEngine();
   Relocator::DWord A = pReloc.addend();
 
-  if (!(pReloc.symInfo()->reserved() & Relocator::ReserveGOT)) {
+  if (!(pReloc.symInfo()->reserved() & Relocator::ReserveGOT) ||
+      pParent.config().isCodeStatic()) {
     Relocator::DWord X =
         pParent.getSymValue(&pReloc) + AArch64LDBackend::getStaticTCBSize();
     // Convert to movz
@@ -1210,7 +1213,8 @@ Relocator::Result tls_gottprel_lo(Relocation &pReloc,
                                   AArch64Relocator &pParent) {
   Relocator::DWord A = pReloc.addend();
 
-  if (!(pReloc.symInfo()->reserved() & Relocator::ReserveGOT)) {
+  if (!(pReloc.symInfo()->reserved() & Relocator::ReserveGOT) ||
+      pParent.config().isCodeStatic()) {
     Relocator::DWord X =
         pParent.getSymValue(&pReloc) + AArch64LDBackend::getStaticTCBSize();
     // Convert to movk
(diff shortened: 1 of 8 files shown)


deepakshirkem:
CC: @quic-areg @quic-seaswara

test/AArch64/RunTests/lit.local.cfg
```diff
@@ -0,0 +1,2 @@
+if not 'run_test' in config.available_features:
```
quic-areg: This is already registered globally, I don't think you need this file.

quic-areg (commented):
Current changes look correct to me. The runtime test looks good; I think it could use a static test as well. 

Do we still reserve a GOT entry for IE?

test/AArch64/RunTests/lit.local.cfg
deepakshirkem: >Do we still reserve a GOT entry for IE?

Yes, for global IE symbols in static exec, a GOT entry is still  created in the scan phase. Should I add this scan phase fix to this PR or as a follow-up?

deepakshirkem:
@quic-areg Please run the CI one more time when you get a chance?

deepakshirkem:
@quic-areg I think this CI failure is not related to this patch.

lib/Target/AArch64/AArch64Relocator.cpp
```diff
@@ -1185,7 +1187,8 @@ Relocator::Result tls_gottprel_page(Relocation &pReloc,
   DiagnosticEngine *DiagEngine = pParent.config().getDiagEngine();
   Relocator::DWord A = pReloc.addend();
 
-  if (!(pReloc.symInfo()->reserved() & Relocator::ReserveGOT)) {
+  if (!(pReloc.symInfo()->reserved() & Relocator::ReserveGOT) ||
```
parth-07: We are still creating a GOT slot for the symbol in the `isCodeStatic()` path, even though now it may not be used at all due to relaxation. Can we improve the code such that we do not create the GOT slot when we know that it will not be used due to relaxation?

parth-07 (changes requested):

lib/Target/AArch64/AArch64Relocator.cpp
deepakshirkem: @parth-07 Updated the patch. Please take one more cycle of review?

deepakshirkem:
Hi @parth-07, I’ve fixed the issues as suggested. Please take another look and trigger the CI when you get a chance?

deepakshirkem:
@parth-07 Thank you. I’ve fixed the clang-format issue. If you think there’s anything else we can improve here, I’ll work on it. I see that the build already succeeded in the last CI run, but if possible, trigger it again when you get a chance?

lib/Target/AArch64/AArch64Relocator.cpp
```diff
@@ -543,7 +543,8 @@ void AArch64Relocator::scanGlobalReloc(InputFile &pInput, Relocation &pReloc,
     // return if we already create GOT for this symbol
     if (rsym->reserved() & ReserveGOT)
       return;
-
+    if (config().isCodeStatic())
+      return;
     // set up the got and the corresponding rel entry
     AArch64GOT *G = m_Target.createGOT(GOT::TLS_IE, Obj, rsym);
     if (config().isCodeStatic() || (config().isBuildingExecutable() &&
```
parth-07:
```suggestion
    if ((config().isBuildingExecutable() &&
```

lib/Target/AArch64/AArch64Relocator.cpp
```diff
@@ -1185,7 +1186,8 @@ Relocator::Result tls_gottprel_page(Relocation &pReloc,
   DiagnosticEngine *DiagEngine = pParent.config().getDiagEngine();
   Relocator::DWord A = pReloc.addend();
 
-  if (!(pReloc.symInfo()->reserved() & Relocator::ReserveGOT)) {
+  if (!(pReloc.symInfo()->reserved() & Relocator::ReserveGOT) ||
+      pParent.config().isCodeStatic()) {
```
parth-07: Do we need `isCodeStatic()` check here?

lib/Target/AArch64/AArch64Relocator.cpp
```diff
@@ -1210,7 +1212,8 @@ Relocator::Result tls_gottprel_lo(Relocation &pReloc,
                                   AArch64Relocator &pParent) {
   Relocator::DWord A = pReloc.addend();
 
-  if (!(pReloc.symInfo()->reserved() & Relocator::ReserveGOT)) {
+  if (!(pReloc.symInfo()->reserved() & Relocator::ReserveGOT) ||
+      pParent.config().isCodeStatic()) {
```
parth-07: Do we need `isCodeStatic()` check here?

parth-07 (changes requested):

lib/Target/AArch64/AArch64Relocator.cpp
deepakshirkem: Done.

lib/Target/AArch64/AArch64Relocator.cpp
deepakshirkem: No need. We’ve added isCodeStatic() in the scan phase now. @parth-07 Thank You.

lib/Target/AArch64/AArch64Relocator.cpp
deepakshirkem: >No need. We’ve added isCodeStatic() in the scan phase now. [Parth (@parth-07)](https://github.com/parth-07) Thank You.

Same.

deepakshirkem:
@parth-07 Resolved the suggested comment. I believe we are close to landing this PR. If you get a chance, Please also run the CI? I’ve verified it locally as well.

parth-07 (commented):
@deepakshirkem Can you please check whether `--relax` and `--no-relax` options are supported by eld for AArch64 target? If not, can you please create an issue for that and add support for `--relax`/`--no-relax` before this PR? I would like to keep the new relaxations disabled (`--no-relax` as default instead of `--relax`) by default for now.

deepakshirkem:
>Can you please check whether --relax and --no-relax options are supported by eld for AArch64 target? If not, can you please create an issue for that and add support for --relax/--no-relax before this PR?

@parth-07 Not supported. I’m creating an issue for both flags, and for now I’ll mark this PR as draft until those flags land.

PR #1672, opened 2026-08-10
[SymbolVersioning] fix version-script input ordering

### LINKED ISSUE #1636: [SymbolVersioning] Investigate input order of the version scripts
We are unconditionally first reading version script specified using --version-script, and the reading the VERSION nodes specified in the other linker scripts.

rachitmeht:
ELD's version-script node order did not follow command-line order; GNU ld processes both -T and --version-script inline in a single option loop, so node order always equals CLI order.

Move --version-script into createInputActions() as a ScriptAction (VersionScript) at its exact CLI position. Add readVersionScriptFile() to parse it during activateInputs(), mark it setToSkip() to prevent reprocessing in normalize(), and append to Module::VersionScripts.

readLinkerScript() appends embedded VERSION{} blocks to the same list via addVersionScript(). Module::LinkerScriptVersionScripts, GeneralOptions::VersionScripts, and their accessors are removed. Module::VersionScripts is now the single ordered list for both sources.

parseVersionScript() collapses to one loop over getVersionScripts() calling registerVersionScriptNodes(). Registration stays deferred as getTargetBackend() is unsafe during activateInputs().

applyVersionScriptScopes() guard changed from hasVersionScript() to getVersionScriptNodes().empty(). readVersionScriptFile() sets setParsed()/setToSkip() only after successful parse and null-guards getVersionScript() before addVersionScript().

Add test VersionScriptInputOrder covering --version-script before/after -T, two --version-script files in both orders, and embedded-only VERSION{} local scope.

Resolves #1636

### DIFF
--- a/lib/Core/Linker.cpp
+++ b/lib/Core/Linker.cpp
@@ -268,9 +268,11 @@ bool Linker::activateInputs(std::vector<InputAction *> &Actions) {
     }
     if (!Action->activate(IR->getInputBuilder()))
       return false;
-    if (!ObjLinker->readAndActivateLinkerScript(
-            llvm::dyn_cast<eld::ScriptAction>(Action)->getLinkerScriptFile(),
-            ScriptFile::ScriptActivationKind::Early)) {
+    auto *SA = llvm::dyn_cast<eld::ScriptAction>(Action);
+    if (SA->kind() != eld::ScriptFile::VersionScript &&
+        !ObjLinker->readAndActivateLinkerScript(
+                   SA->getLinkerScriptFile(),
+                   ScriptFile::ScriptActivationKind::Early)) {
       ThisModule->setFailure(true);
       return false;
     }
--- a/lib/Core/Linker.cpp
+++ b/lib/Core/Linker.cpp
@@ -268,9 +268,15 @@ bool Linker::activateInputs(std::vector<InputAction *> &Actions) {
     }
     if (!Action->activate(IR->getInputBuilder()))
       return false;
-    if (!ObjLinker->readAndActivateLinkerScript(
-            llvm::dyn_cast<eld::ScriptAction>(Action)->getLinkerScriptFile(),
-            ScriptFile::ScriptActivationKind::Early)) {
+    auto *SA = llvm::dyn_cast<eld::ScriptAction>(Action);
+    if (SA->kind() == eld::ScriptFile::VersionScript) {
+      if (!ObjLinker->readVersionScriptFile(SA->getLinkerScriptFile())) {
+        ThisModule->setFailure(true);
+        return false;
+      }
+    } else if (!ObjLinker->readAndActivateLinkerScript(
+                   SA->getLinkerScriptFile(),
+                   ScriptFile::ScriptActivationKind::Early)) {
       ThisModule->setFailure(true);
       return false;
     }
--- a/lib/LinkerWrapper/GnuLdDriver.cpp
+++ b/lib/LinkerWrapper/GnuLdDriver.cpp
@@ -886,12 +886,6 @@ bool GnuLdDriver::processOptions(llvm::opt::InputArgList &Args) {
   if (Config.options().getDynList().size())
     Config.options().setDynamicList();
 
-  // --version-script
-  for (auto *Arg : Args.filtered(T::version_script))
-    Config.options().getVersionScripts().emplace(Arg->getValue());
-  if (Config.options().getVersionScripts().size())
-    Config.options().setVersionScript();
-
   // --extern-list
   for (auto *Arg : Args.filtered(T::extern_list))
     Config.options().getExternList().emplace(Arg->getValue());
@@ -1365,6 +1359,13 @@ bool GnuLdDriver::createInputActions(llvm::opt::InputArgList &Args,
           arg->getValue(), eld::ScriptFile::LDScript, Config,
           Config.getPrinter()));
       ++input_num;
+    } break;
+
+    case T::version_script: {
+      actions.push_back(eld::make<eld::ScriptAction>(
+          arg->getValue(), eld::ScriptFile::VersionScript, Config,
+          Config.getPrinter()));
+      Config.options().setVersionScript();
     } break;
 
     case T::R: {
--- a/lib/Object/GroupReader.cpp
+++ b/lib/Object/GroupReader.cpp
@@ -52,13 +52,25 @@ bool GroupReader::readGroup(InputBuilder::InputIteratorT &CurNode,
     if (!Input->resolvePath(PConfig))
       return false;
 
+    if (Input->getInputFile() &&
+        Input->getInputFile()->getKind() == InputFile::GNULinkerScriptKind) {
+      auto *LSFile =
+          llvm::dyn_cast<eld::LinkerScriptFile>(Input->getInputFile());
+      if (LSFile->isVersionScript()) {
+        if (!MObjLinker->readVersionScriptFile(LSFile))
+          return false;
+        ++CurNode;
+        continue;
+      }
+    }
     if (!MObjLinker->readAndProcessInput(Input, IsPostLtoPhase))
       return false;
-
-    if (Input->getInputFile()->getKind() == InputFile::GNULinkerScriptKind) {
-      if (!MObjLinker->readInputs(
-              llvm::dyn_cast<eld::LinkerScriptFile>(Input->getInputFile())
-                  ->getNodes()))
+    if (Input->getInputFile() &&
+        Input->getInputFile()->getKind() == InputFile::GNULinkerScriptKind) {
+      auto *LSFile =
+          llvm::dyn_cast<eld::LinkerScriptFile>(Input->getInputFile());
+      if (!LSFile->isVersionScript() &&
+          !MObjLinker->readInputs(LSFile->getNodes()))
         return false;
     }
 
--- a/lib/Object/LibReader.cpp
+++ b/lib/Object/LibReader.cpp
@@ -15,6 +15,7 @@
 #include "eld/Config/LinkerConfig.h"
 #include "eld/Input/Input.h"
 #include "eld/Input/InputTree.h"
+#include "eld/Input/LinkerScriptFile.h"
 #include "eld/Object/ObjectLinker.h"
 #include "eld/Support/MemoryArea.h"
 #include "eld/Support/RegisterTimer.h"
@@ -82,6 +83,29 @@ bool LibReader::readLib(InputBuilder::InputIteratorT &CurNode,
     Input *Input = Node->getInput();
     if (!Input->resolvePath(Config))
       return false;
+
+    if (Input->getInputFile() &&
+        Input->getInputFile()->getKind() == InputFile::GNULinkerScriptKind) {
+      auto *LSFile = llvm::cast<eld::LinkerScriptFile>(Input->getInputFile());
+      if (LSFile->isVersionScript()) {
+        if (!MObjLinker->readVersionScriptFile(LSFile))
+          return false;
+        ++CurNode;
+        continue;
+      }
+    }
+
+    if (Input->getInputFile() &&
+        Input->getInputFile()->getKind() == InputFile::GNULinkerScriptKind) {
+      auto *LSFile = llvm::cast<eld::LinkerScriptFile>(Input->getInputFile());
+      if (!MObjLinker->readAndProcessInput(Input, IsPostLtoPhase))
+        return false;
+      if (!LSFile->isVersionScript() &&
+          !MObjLinker->readInputs(LSFile->getNodes()))
+        return false;
+      ++CurNode;
+      continue;
+    }
 
     if (IsThin)
       MemberNames.push_back(Input->getResolvedPath().getFullPath());
--- a/lib/Object/ObjectLinker.cpp
+++ b/lib/Object/ObjectLinker.cpp
@@ -259,12 +259,38 @@ bool ObjectLinker::readAndActivateLinkerScript(
       return false;
   }
 
+  // Register an embedded VERSION{} block when the script is first activated
+  // so it remains ordered with standalone --version-script inputs.
+  if (S->getVersionScript() &&
+      (kind == ScriptFile::ScriptActivationKind::Early ||
+       !linkerScriptFile->isEarlyActivated()))
+    ThisModule->addVersionScript(S->getVersionScript());
+
   if (kind == ScriptFile::ScriptActivationKind::Early)
     linkerScriptFile->setEarlyActivated();
   else
     linkerScriptFile->setFullyActivated();
 
   input->setUsed(true);
+  return true;
+}
+
+bool ObjectLinker::readVersionScriptFile(InputFile *Input) {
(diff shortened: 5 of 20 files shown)


lib/LinkerWrapper/GnuLdDriver.cpp
quic-areg, line 1349: Should `--default-script` be evaluated positionally too?

quic-areg (commented):
it's a bit of a contrived case, but evaluating `--version-script` positionally made it become a --start-lib --end-lib member: 

```
ld.eld -shared --whole-archive --start-lib a.o \
       --version-script=vs --end-lib --no-whole-archive -o out.so
# Error: vs:1:1: unknown directive: V
# Fatal: Linker script <start-lib:1> has errors
```

rachitmeht:
> it's a bit of a contrived case, but evaluating `--version-script` positionally made it become a --start-lib --end-lib member:
> 
> ```
> ld.eld -shared --whole-archive --start-lib a.o \
>        --version-script=vs --end-lib --no-whole-archive -o out.so
> # Error: vs:1:1: unknown directive: V
> # Fatal: Linker script <start-lib:1> has errors
> ```

Scripts should not be part of input tree (--start-lib --end-lib member). is it correct? this will fix the error.

parth-07:
@rachitmeht Can you please rebase this PR on top of origin/main?

lib/Core/Linker.cpp
```diff
@@ -268,11 +268,17 @@ bool Linker::activateInputs(std::vector<InputAction *> &Actions) {
     }
     if (!Action->activate(IR->getInputBuilder()))
       return false;
-    if (!ObjLinker->readAndActivateLinkerScript(
-            llvm::dyn_cast<eld::ScriptAction>(Action)->getLinkerScriptFile(),
-            ScriptFile::ScriptActivationKind::Early)) {
-      ThisModule->setFailure(true);
-      return false;
+    auto *SA = llvm::dyn_cast<eld::ScriptAction>(Action);
+    if (SA->kind() == eld::ScriptFile::VersionScript) {
```
parth-07: Can we please read version script along with the other scripts in `ObjectLinker::readInputs`? We should only read `-T <script>` scripts early, and too because some nodes in `-T <script>` scripts has to be processed early. 

One of the previous patch added support for `ScriptFile::ScriptActivationKind::Early` and `ScriptFile::ScriptActivationKind::Full` which we can use in `ObjectLinker::readLinkerScript` while adding the version script to the `Module` object.

parth-07:
@rachitmeht Can you please check and report what is the behaviour of GNU ld and lld when a version script is present in between `--start-lib` and `--end-lib`?

rachitmeht:
GNU gold (bfd doesn't have --start-lib functionality)
 
  --start-lib creates an Input_file_lib and sets in_lib = true.
  Object files encountered afterwards are added to that library.
 
  --version-script is still treated as a linker option while inside the
  --start-lib/--end-lib range. Gold calls read_version_script() immediately
  during command-line processing. The version script is not added to the
  Input_file_lib and therefore cannot become a lazy archive member.
 
  --end-lib closes the library range. Later, gold uses the parsed version
  information while adding and resolving symbols from the selected lazy
  members.
 
  Gold processes -T scripts and --version-script options in the same
  left-to-right command-line loop. Consequently, their relative command-line
  order is preserved, including the order of embedded VERSION{} blocks in
  -T scripts.
 
  lld
 
  lld first calls readConfigs(), before createFiles(). In readConfigs(), it
  processes all standalone --version-script options:
 
  for (auto *arg : args.filtered(OPT_version_script))
    readVersionScript(...);
 
  These scripts are parsed before lld processes --start-lib, --end-lib,
  objects, or -T scripts. They are not added to the lazy library.
 
  createFiles() later processes --start-lib and marks object inputs as lazy.
  The --version-script options are not handled there because they were already
  consumed by readConfigs().
 
  Thus lld does not preserves the order between standalone version scripts and embedded VERSION{}
  blocks in -T scripts.
 
  For example:
 
  -T script-with-VERSION-B --version-script=A.map
 
  Gold processes B then A.
 
  Current lld processes A first in readConfigs(), then B while processing -T.

parth-07:
@rachitmeht Thanks for reporting the behavior here across linkers. Can you update the patch such that the `--version-script <file>` is not added as an archive member when used with `--start-lib`/`--end-lib`?

lib/Script/ScriptAction.cpp
```diff
@@ -59,7 +60,7 @@ bool ScriptAction::activate(InputBuilder &PBuilder) {
     Path = Res->native();
   }
   setFileName(Path);
-  InputFileAction::activate(PBuilder);
```
parth-07, line 62: Why did you remove `InputFileAction::activate` from here?

parth-07 (changes requested):
This PR introduces two regressions:

1) --remap-inputs does not work properly now with version scripts.

```
#!/usr/bin/env bash
set -eux

cat > tmp/pr1672_remap.c <<\EOF
int alpha(void) { return 1; }
EOF

cat > tmp/pr1672_actual.vs <<\EOF
V1 { global: alpha; local: *; };
EOF

clang -c -fPIC tmp/pr1672_remap.c -o tmp/pr1672_remap.o

ld.eld -shared tmp/pr1672_remap.o \
  --remap-inputs=tmp/pr1672_missing.vs=tmp/pr1672_actual.vs \
  --version-script=tmp/pr1672_missing.vs \
  -o tmp/pr1672_remap.so

llvm-readelf --dyn-syms tmp/pr1672_remap.so | grep alpha
```

This errors out with this patch.

2) `-L <dir> --version-script vs.t` breaks reproduce functionality with this patch
```
#!/usr/bin/env bash
set -eux

mkdir -p tmp/pr1672_vsdir

cat > tmp/pr1672_repro.c <<\EOF
int alpha(void) { return 1; }
EOF

cat > tmp/pr1672_vsdir/vs <<\EOF
V1 { global: alpha; local: *; };
EOF

clang -c -fPIC tmp/pr1672_repro.c -o tmp/pr1672_repro.o

ld.eld -shared tmp/pr1672_repro.o \
  -L tmp/pr1672_vsdir \
  --version-script=vs \
  --reproduce=tmp/pr1672_repro.tar \
  -o tmp/pr1672_repro.so

mkdir -p tmp/reproduce
tar -xf tmp/pr1672_repro.tar -C tmp/reproduce --strip-components=1
cd tmp/reproduce
bash -x response.txt
```

The 2nd issue will get resolved with the #1763, and the first issue currently also presents for linker scripts. 

@rachitmeht  Can we please resolve these two issues before merging this PR. I do not want to introduce temporary regressions if they can be avoided.

lib/Script/ScriptAction.cpp
rachitmeht: ScriptAction::activate() calls createInput() directly instead of InputFileAction::activate(), avoiding createInputNode() which adds to the tree. This prevents scripts from becoming lib members when inside --start-lib.

parth-07: Linker so far has an invariant that for all InputActions, InputAction::activate must always be called. I would like to keep that invariant. Otherwise, it is difficult to track when `activate()` should be called and when it should not be called. Can you please update the design such that the invariant that `InputAction::activate` is always called remains true?

lib/Object/ObjectLinker.cpp
```diff
@@ -352,49 +394,11 @@ bool ObjectLinker::normalize() {
 // FIXME: We should maybe parse version script after reading LTO-generated
 // object files.
 bool ObjectLinker::parseVersionScript() {
-  if (ThisConfig.options().hasVersionScript()) {
-    LayoutInfo *layoutInfo = ThisModule->getLayoutInfo();
-    for (const auto &List : ThisConfig.options().getVersionScripts()) {
-      Input *VersionScriptInput =
-          eld::make<Input>(List, ThisConfig.getDiagEngine(), Input::Script);
-      if (!VersionScriptInput->resolvePath(ThisConfig))
```
rachitmeht, line 360: @parth-07  --remap inputs mapping should be applied by `resolvePath()` which we were doing by creating an VersionScriptInput and then calling `resolvedPath()`. 

But, we also call resolvePath() during ScriptAction::activate() 
and we also create input during`I = PBuilder.createInputNode(Name)`.

lib/Script/ScriptAction.cpp
rachitmeht, line 39: The same check is done in ResolvedPath() also, and before doing --remap inputs mapping if we enter into ScriptAction::activate(). It discards the remap input before doing ResolvePath();

lib/Object/ObjectLinker.cpp
parth-07: Okay, I see. Thank you for the explanation.

lib/Script/ScriptAction.cpp
parth-07: Okay, I see. Thanks.

lib/Object/GroupReader.cpp
```diff
@@ -52,13 +52,25 @@ bool GroupReader::readGroup(InputBuilder::InputIteratorT &CurNode,
     if (!Input->resolvePath(PConfig))
       return false;
 
+    if (Input->getInputFile() &&
```
parth-07, line 55: Why do we need explicit if-block for version script? Wouldn't the version scripts be read through `MObjLinker->readAndProcessInput` (line below this if-block)?

lib/Object/LibReader.cpp
```diff
@@ -83,6 +84,29 @@ bool LibReader::readLib(InputBuilder::InputIteratorT &CurNode,
...
+      auto *LSFile = llvm::cast<eld::LinkerScriptFile>(Input->getInputFile());
+      if (!MObjLinker->readAndProcessInput(Input, IsPostLtoPhase))
+        return false;
+      if (!LSFile->isVersionScript() &&
+          !MObjLinker->readInputs(LSFile->getNodes()))
+        return false;
+      ++CurNode;
+      continue;
+    }
+
```
parth-07, line 109: Is this change related to version-script ordering fix?

lib/Object/ObjectLinker.cpp
parth-07: > --remap inputs mapping should be applied by resolvePath() which we were doing by creating an VersionScriptInput and then calling resolvedPath().
> 
> But, we also call resolvePath() during ScriptAction::activate()
> and we also create input duringI = PBuilder.createInputNode(Name).

With this patch, version scripts fail earlier in `ScriptAction::activate` before they have the chance to call `resovePath()` when remap inputs affect their names. The remap-inputs test case shared by me works without this patch and fails with this patch. Can you please verify it? 

This being said, I think both the test-cases that I have shared should be resolved separately (and before this PR).

lib/Script/ScriptAction.cpp
parth-07: I could not follow this comment. 

> before doing --remap inputs mapping if we enter into ScriptAction::activate().

We always enter `ScriptAction::activate` before remap inputs take effect as it takes effect during resolvePath which is called within `ScriptAction::activate`, right?

rachitmeht: Before this patch, we were calling `resolvePath()` in `parseVersionScript()` , that's why --remap mapping was working, But now I am calling `ScriptAction()` which checks if the file exists before calling `resolvePath()` and Fails.

rachitmeht: My thought was that is `!llvm::sys::fs::exists(Path)` check is redundant as  ResolvePath() already handles that kind of case?

lib/Object/ObjectLinker.cpp
```diff
@@ -278,18 +308,20 @@ bool ObjectLinker::readInputs(const std::vector<Node *> &InputVector) {
...
 
     if ((*Begin)->kind() == Node::LibStart) {
       eld::RegisterTimer T("Read Start Lib and End Lib", "Read all Input files",
                            ThisConfig.options().printTimingStats());
-      getLibReader()->readLib(Begin,
-                              ThisModule->getIRBuilder()->getInputBuilder(),
-                              ThisConfig, MPostLtoPhase);
+      if (!getLibReader()->readLib(
+              Begin, ThisModule->getIRBuilder()->getInputBuilder(), ThisConfig,
+              MPostLtoPhase))
```
rachitmeht, line 323: > Is this change related to version-script ordering fix?

Scripts were also getting a part of --start-lib --end-lib, So handled them in the `readLib()` and `readGroup()` seperately.

lib/Object/LibReader.cpp
rachitmeht: Scripts were also getting a part of --start-lib --end-lib, So handled them in the readLib() and readGroup() seperately.

lib/Object/GroupReader.cpp
rachitmeht: `readAndProcessInput()` only calls `readandactivateLinkerScripts()`, We also call `readInputs()` and I had added `readversionscript()` in `readInputs()` also, but I saw the implementation of processing `LinkerScripts` ( we first call `readAndProcessInput()` and then call recursively 'readInputs()`). So, I didn't changed that.

lib/Object/LibReader.cpp
parth-07: What do you mean by 'Scripts were also getting a part of --start-lib --end-lib'?

Were we previously handling linker scripts in `--start-lib ... --end-lib`? If it is a separate already-existing bug then it should not be fixed in this PR.

lib/Script/ScriptAction.cpp
parth-07: > My thought was that is !llvm::sys::fs::exists(Path) check is redundant as ResolvePath() already handles that kind of case?

It might be redundant, and refactoring should done eventually. However, I was only pointing out that the patch currently has a regression when version-scripts is used with remap-inputs.

parth-07 (changes requested):

lib/Object/GroupReader.cpp
parth-07: Why are we handling the processing of linker scripts within GroupReader if the patch is only about version scripts?

rachitmeht: I understood your point, I will see if there could be better design for this fix.

lib/Object/LibReader.cpp
rachitmeht: Yes, I verified , this was already exisiting bug for LinkerScripts
```
#!/usr/bin/env bash
  set -eux

  cat > main.c <<'EOF'
  extern int foo();
  int main() { return foo(); }
EOF

  cat > foo.c <<'EOF'
  int foo() { return 42; }
EOF

  cat > script.ld <<'EOF'
  SECTIONS
  {
    .foo_out : { *(.text.foo) }
    .text : { *(.text*) }
  }
EOF

  TARGET=x86_64-linux-gnu

  clang -target $TARGET -c main.c -o main.o
  clang -target $TARGET -c foo.c -ffunction-sections -o foo.o

  ld.eld \
    main.o \
    --start-lib foo.o \
    -T script.ld \
    --end-lib \
    -o eld.out

  llvm-readelf -S eld.out
```
Result :
.foo_out is missing and .text.foo remains, demonstrating that the linker script inside --start-lib was not fully processed.

PR #1683, opened 2026-08-11
Add a workflow to monitor failing workflows

quic-shaksuma:
A monitoring workflow is added that looks for failing workflow runs and creates an issue for it and avoids duplicate issue creation by looking for a passing run before marking a failed run as a new failure and triggering an issue creation.

### DIFF
--- a/.github/workflows/workflow_monitor.yml
+++ b/.github/workflows/workflow_monitor.yml
@@ -1,0 +1,125 @@
+name: Monitor workflow failures
+
+on:
+  # A workflow_run trigger lets this workflow inspect the completed run without
+  # checking out or executing code from the monitored workflow.
+  workflow_run:
+    workflows:
+      - BusyBox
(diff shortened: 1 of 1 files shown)


Steven6798:
Is it possible to assign someone to the issue based on the CODEOWNERS file? Other than that, it looks good to me.

quic-shaksuma:
> Is it possible to assign someone to the issue based on the CODEOWNERS file? Other than that, it looks good to me.

@Steven6798 I want to do that in a follow-up as this patch only has a workflow level view, not exact error/code level to auto-assign

Steven6798:
> Steven Ramirez Rosa (@Steven6798) I want to do that in a follow-up as this patch only has a workflow level view, not exact error/code level to auto-assign

Awesome!

Steven6798 (approved):

quic-areg (approved):
LGTM

PR #1700, opened 2026-08-13
[Reproduce] Capture --remap-inputs-file in --reproduce tarballs

### LINKED ISSUE #1698: [reproduce] tarball fails to replay when --remap-inputs-file is used
## Description:

Linking with --remap-inputs-file=<file> together with --reproduce=<tar> produces a tarball that cannot be replayed.

Expected: the replay link succeeds, since the tarball is supposed to be a self-contained reproducer.
Actual: the replay fails outright:
Error: cannot open --remap-inputs-file remap.txt: No such file or directory

```
#!/usr/bin/env bash
#
# Reproducer: ELD's --reproduce does not capture --remap-inputs-file's own
# file, so the resulting tarball cannot be replayed.
#
# What happens:
#   1. Link with --remap-inputs-file=<path> and --reproduce=<tar>.
#   2. Inspect the tarball: --remap-inputs-file's own file is never
#      packaged, and mapping.ini has no entry for it, even though every
#      other captured input (objects, scripts, plugin configs, ...) gets
#      one.
#   3. response.txt (the reconstructed command line inside the tarball)
#      still references the file, but only by a bare basename with no
#      category/hash prefix -- unlike every other captured input, which
#      gets a "<Category>/<basename>.<hash>" path.
#   4. Replaying the extracted tarball via `ld.eld @response.txt` then
#      fails outright: "cannot open --remap-inputs-file <name>: No such
#      file or directory".
#
# Usage:
#   ./reproduce_remap_inputs_file_gap.sh [path-to-ld.eld] [path-to-cc]
#
# Prints BUG REPRODUCED (and exits 0) if replay fails as described above.
# Prints NOT REPRODUCED (and exits 1) if replay succeeds, i.e. the remap
# file was captured and the bug is not present in this build.

set -uo pipefail

LD="${1:-ld.eld}"
CC="${2:-clang}"

command -v "$LD" >/dev/null 2>&1 || { echo "error: '$LD' not found" >&2; exit 2; }
command -v "$CC" >/dev/null 2>&1 || { echo "error: '$CC' not found" >&2; exit 2; }

WORKDIR="$(mktemp -d -t eld-remap-repro.XXXXXX)"
cleanup() { rm -rf "$WORKDIR"; }
trap cleanup EXIT

cd "$WORKDIR"
echo "== Working in $WORKDIR =="
echo "== Using LD=$LD  CC=$CC =="
echo

cat > main.c <<'EOF'
int main(void) { return 0; }
EOF
"$CC" -c main.c -o main.o

# The remap rule doesn't need to actually match anything -- the bug only
# depends on --remap-inputs-file being present on the command line, not
# on the rule taking effect.
echo "nonexistent-pattern.o=main.o" > remap.txt

echo "== Step 1: link with --remap-inputs-file + --reproduce =="
"$LD" main.o -o out --remap-inputs-file=remap.txt --reproduce repro.tar --dump-response-file response.txt
echo "-- response.txt --"
cat response.txt
echo
echo

echo "== Step 2: extract the reproduce tarball =="
mkdir extract
tar -xf repro.tar -C extract --strip-components=1
echo "-- extracted files --"
find extract -type f | sort
echo
echo "-- mapping.ini --"
cat extract/mapping.ini
echo
echo

echo "== Step 3: replay the link from inside the extracted tarball =="
if ( cd extract && "$LD" @response.txt -o rep.out ); then
  echo
  echo "Replay SUCCEEDED -- remap.txt was captured and replay worked."
  echo "RESULT: NOT REPRODUCED"
  exit 1
else
  echo
  echo "Replay FAILED -- the reproduce tarball is missing a file its own"
  echo "recorded command line depends on (remap.txt was never packaged)."
  echo "RESULT: BUG REPRODUCED"
  exit 0
fi
```

Steven6798:
Capture --remap-inputs-file in --reproduce tarballs. Previously, the remap file was not added to the archive, so reproduce
tarballs using this option could not be replayed. Register it with createAndAddConfigFile, mirroring the --plugin-config
handling. 
When a remap rule matched an input, response.txt also retained the original command-line path even though the tarball
contained only the replacement file. Generate response.txt using the resolved remapped paths for positional inputs and input-backed options, including:
  - linker scripts and just-symbols files
  - version scripts
  - dynamic and extern lists
  - trampoline list files

Add regression coverage that verifies the captured paths and successfully replays the extracted tarball.

Fix: qualcomm#1698

### DIFF
--- a/test/Common/standalone/CommandLine/Reproduce/RemapInputsFileCapture.test
+++ b/test/Common/standalone/CommandLine/Reproduce/RemapInputsFileCapture.test
@@ -1,0 +1,41 @@
+#---RemapInputsFileCapture.test---------------- Executable --------------------#
+#BEGIN_COMMENT
+# Confirms --reproduce correctly captures --remap-inputs-file. The response
+# uses the captured replacement object when a remap rule substitutes a
+# positional input, and uses the captured replacement file when remap rules
+# substitute option-backed inputs such as version scripts and symbol lists.
+# It also records the remap file's categorized/hashed tar path
+# (Config/remap.txt.<hash>) instead of a bare basename, mapping.ini lists the
+# remap file under [Config Files], and the extracted reproduce tarball can be
+# replayed in place via its own response.txt to produce a byte-identical link.
+#END_COMMENT
+#START_TEST
+RUN: %clang %clangopts -c %p/Inputs/1.c -o %t1.original.o
+RUN: %clang %clangopts -c %p/Inputs/1.c -o %t1.replacement.o
+RUN: %clang %clangopts -c %p/Inputs/2.c -o %t1.2.o
+RUN: %echo "%t1.original.o=%t1.replacement.o" > %t1.remap.txt
+RUN: %echo "%t1.original.version=%p/Inputs/versionscript" >> %t1.remap.txt
+RUN: %echo "%t1.original.dynamic=%p/Inputs/list1" >> %t1.remap.txt
+RUN: %echo "%t1.original.extern=%p/Inputs/externlist" >> %t1.remap.txt
+RUN: %link %linkopts %t1.original.o %t1.2.o -o %t1.out --force-dynamic --version-script=%t1.original.version --dynamic-list=%t1.original.dynamic --extern-list=%t1.original.extern --remap-inputs-file=%t1.remap.txt --reproduce %t0.tar --dump-response-file %t1.response
+RUN: %filecheck %s --check-prefix=RESPONSE < %t1.response
+RUN: %mkdir %t1.extract
+RUN: %tar %gnutaropts -xf %t0.tar -C %t1.extract --strip-components=1
+RUN: %filecheck %s --check-prefix=MAPPING < %t1.extract/mapping.ini
+
+RUN: cd %t1.extract
+RUN: %link @%t1.extract/response.txt -o %t1.extract/rep.out
+RUN: %diff %t1.extract/rep.out %t1.out
+
+RESPONSE: --thread-count
+RESPONSE-NOT: original
+RESPONSE: Object{{[^ ]*}}replacement.o{{[^ ]*}}
+RESPONSE: --version-script {{.*}}VersionScript{{.*}}versionscript{{[^ ]*}}
+RESPONSE: --dynamic-list {{.*}}DynamicList{{.*}}list1{{[^ ]*}}
+RESPONSE: --extern-list {{.*}}ExternList{{.*}}externlist{{[^ ]*}}
+RESPONSE-NOT: original
+RESPONSE: --remap-inputs-file {{.*}}Config{{.*}}remap.txt{{.*}}
+RESPONSE-NOT: original
+MAPPING: [Config Files]
+MAPPING-NEXT: {{.*}}remap.txt{{.*}}=Config{{.*}}remap.txt{{.*}}
+#END_TEST
--- a/lib/LinkerWrapper/GnuLdDriver.cpp
+++ b/lib/LinkerWrapper/GnuLdDriver.cpp
@@ -1680,6 +1680,36 @@ bool GnuLdDriver::processReproduceOption(
   if (!Config.options().getDumpResponse())
     os << getProgramName() << " ";
   size_t lastNamespecId = -1;
+  size_t lastInputFileId = -1;
+  size_t lastScriptId = -1;
+  size_t lastJustSymbolsId = -1;
+
+  auto getRewrittenActionInputPath =
+      [&](eld::InputAction::InputActionKind kind, size_t &lastActionId,
+          llvm::StringRef fallback) -> std::optional<std::string> {
+    for (size_t i = lastActionId + 1; i < actions.size(); ++i) {
+      auto action = actions[i];
+      if (action->getInputActionKind() != kind)
+        continue;
+      lastActionId = i;
+      auto ipt = action->getInput();
+      if (!ipt)
+        return std::nullopt;
+      std::string key = fallback.str();
+      if (ipt->getInputFile() && ipt->getInputFile()->hasMappedPath())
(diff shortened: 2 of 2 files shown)


test/Common/standalone/CommandLine/Reproduce/RemapInputsFileCapture.test
```diff
@@ -0,0 +1,25 @@
+UNSUPPORTED:
```
quic-seaswara: UNSUPPORTED: ?

Steven6798: I missed that. a Remnant of using a test template. Thanks for catching that

test/Common/standalone/CommandLine/Reproduce/RemapInputsFileCapture.test
```diff
@@ -0,0 +1,24 @@
...
+#END_COMMENT
+#START_TEST
+RUN: %clang %clangopts -c %p/Inputs/1.c -o %t1.1.o
+RUN: %clang %clangopts -c %p/Inputs/2.c -o %t1.2.o
+RUN: %echo "nonexistent-pattern.o=%t1.1.o" > %t1.remap.txt
+RUN: %link %linkopts %t1.1.o %t1.2.o -o %t1.out --remap-inputs-file=%t1.remap.txt --reproduce %t0.tar --dump-response-file %t1.response
+RUN: %filecheck %s --check-prefix=RESPONSE < %t1.response
+RUN: %mkdir %t1.extract
+RUN: %tar %gnutaropts -xf %t0.tar -C %t1.extract --strip-components=1
+RUN: %grep -q remap %t1.extract/mapping.ini
```
quic-areg: This should be a `CHECK` rather than a `grep`

quic-areg (approved):

test/Common/standalone/CommandLine/Reproduce/RemapInputsFileCapture.test
Steven6798: I'll make that change. Thanks!

SuperChamp234 (approved):

BhoumikPatidar:
Consider this repro:
```bash
cat > original.c <<'EOF'
int main(){ return 10; }
EOF
cat > replacement.c <<'EOF'
int main(){ return 20; }
EOF
clang -c original.c -o original.o
clang -c replacement.c -o replacement.o
echo "original.o=replacement.o" > remap.txt

clang -fuse-ld=eld original.o \
  -Wl,--remap-inputs-file=remap.txt \
  -Wl,--reproduce=repro.tar \
  -o remapped.out

mkdir extract && tar -xf repro.tar -C extract --strip-components=1
cd extract
bash response.txt
```
This results in - 
```bash
Fatal: cannot read file 'original.o'
```

The `--reproduce` still fails to replay when `--remap-inputs-file` matches an input and substitutes it. It seems remap.txt is not being considered. 

`response.txt` has plain original.o

```bash
ld.eld -z relro --hash-style=gnu --eh-frame-hdr -m elf_x86_64 -pie -dynamic-linker /lib64/ld-linux-x86-64.so.2 -o remapped.out Object/Scrt1.o.2137534658666283053 Object/crti.o.15849601514134208758 Object/crtbeginS.o.11488495536862619661 original.o --remap-inputs-file Config/remap.txt.3791689425116741856 Archives/gcc.1781399130461890127 --as-needed LinkerScript/gcc_s.9044401076882912980 --no-as-needed LinkerScript/c.7929735926503149735 Archives/gcc.1781399130461890127 --as-needed LinkerScript/gcc_s.9044401076882912980 --no-as-needed Object/crtendS.o.1971631304738010186 Object/crtn.o.18031929860169820072 --mapping-file=mapping.ini
```

Please correct me if I am doing some mistake here

Steven6798:
> Please correct me if I am doing some mistake here

No mistake here. This is indeed an issue that I did not consider. Thanks for the clear description and the repro!

Steven6798:
> The `--reproduce` still fails to replay when `--remap-inputs-file` matches an input and substitutes it. It seems remap.txt is not being considered.

This affects multiple options which I have addressed. Let me know if anything else is missing.

BhoumikPatidar (approved):

test/Common/standalone/CommandLine/Reproduce/RemapInputsFileCapture.test
```diff
@@ -0,0 +1,41 @@
...
+RUN: %clang %clangopts -c %p/Inputs/1.c -o %t1.replacement.o
+RUN: %clang %clangopts -c %p/Inputs/2.c -o %t1.2.o
+RUN: %echo "%t1.original.o=%t1.replacement.o" > %t1.remap.txt
+RUN: %echo "%t1.original.version=%p/Inputs/versionscript" >> %t1.remap.txt
+RUN: %echo "%t1.original.dynamic=%p/Inputs/list1" >> %t1.remap.txt
+RUN: %echo "%t1.original.extern=%p/Inputs/externlist" >> %t1.remap.txt
+RUN: %link %linkopts %t1.original.o %t1.2.o -o %t1.out --force-dynamic --version-script=%t1.original.version --dynamic-list=%t1.original.dynamic --extern-list=%t1.original.extern --remap-inputs-file=%t1.remap.txt --reproduce %t0.tar --dump-response-file %t1.response
+RUN: %filecheck %s --check-prefix=RESPONSE < %t1.response
+RUN: %mkdir %t1.extract
+RUN: %tar %gnutaropts -xf %t0.tar -C %t1.extract --strip-components=1
```
parth-07, line 23: @Steven6798  `-C` tar option currently fails on our Windows builders, can you please update the test so that it does not need `-C` option? (or fix the root cause of why it fails on Windows)

Steven6798: Ill fix this today.

PR #1702, opened 2026-08-13
Make handling of -plugin-opt=O2/3 more similar to lld

efriedma-quic:
In addition to setting the OptLevel, lld sets CGOptLevel and enables vectorization.  Make eld do the same.


quic-seaswara (approved):

apazos (commented):
@quic-seaswara, we need this fix in 23.x as well.

efriedma-quic:
https://github.com/qualcomm/eld/pull/1703

PR #1706, opened 2026-08-14
Use correct base branch for checking non-doc changes in PR CI

parth-07:
This commit fixes the pull-request CI workflow to use the correct base branch when checking for non-doc changes in pull-request CI.

Until now, if a pull-request was created for qualcomm/eld/release/23.x branch using parth/eld/SomeFeature branch, then the 'Check for non-doc changes' step in pull-request CI will take parth/eld/SomeFeature branch as the HEAD commit and parth/eld/release/23.x branch as the base. The base used here is wrong and should instead be
qualcomm/eld/release/23.x.

Using incorrect base has two key issues:

- If the repository of the branch using which PR is created does not have a branch named as the ${BASE_BRANCH} then the `git diff ...` command will error out and the workflow will fail.
- A more serious issue is that the PR author can update his/her ${REPO}/eld/${BASE_BRANCH} such that the 'check for non-doc changes' returns no diff and the workflow incorrectly passes without building/running tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,8 +72,10 @@ jobs:
         id: file-check
         working-directory: pr-${{ env.PR_NUMBER }}/llvm-project/llvm/tools/eld
         run: |
-          git fetch origin ${{ env.BASE_BRANCH_NAME }}
-          MERGE_BASE=$(git merge-base origin/${{ env.BASE_BRANCH_NAME }} HEAD)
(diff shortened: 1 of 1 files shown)


quic-seaswara (approved):

efriedma-quic:
Does this also need to land on main?

quic-seaswara:
@parth-07  is working on cherry picking this to main.

PR #1709, opened 2026-08-14
Remove `tar ... -C <dir>` usage from tests enabled for Windows

parth-07:
This commit modifies the tests which were enabled for Windows to not use `-C <dir>` tar option. `tar ... -C <dir>` command fails on Windows builders because the `...\Git\bin\tar.exe` does not correctly handle the Windows-style path with the `-C` option.

We should later debug exactly why tar is failing here and if we can use some other tar on Windows builders so that such errors are not encountered in the future.


quic-seaswara (approved):

parth-07:
With this change, `SysrootMarker.test` and `ReproduceSharedLibSoname.test` are not failing anymore. Verification: https://github.com/qualcomm/eld/actions/runs/31793909768/job/94746659506?pr=1709. The build is still failing because the below tests are getting timed out:

- Common/LTO/LTOSaveTempsNotWritable/LTOSaveTempsNotWritable.test
- lld/ELF/opt-level.ll

BhoumikPatidar (approved):

PR #1718, opened 2026-08-17
Add --trace=plugin=<name> scoped plugin tracing

tcurtiss:
Extend --trace=<category>[=<name>] (mirroring symbol=/section=/reloc=) to plugins, so an individual plugin's diagnostics can be enabled without tracing every plugin. GeneralOptions::tracePlugin() checks an exact-string/regex allowlist populated by the new plugin=<name> case, falling back to "trace everything" for the pre-existing bare --trace=plugin.

Plugin::isTraced() wraps this check and replaces every tracePlugins()-gated call site in Plugin.cpp, PluginManager.cpp, and GNULDBackend::RunPluginsAndProcessHelper with per-plugin gating via O->prolog().getPlugin()->isTraced().

(cherry picked from commit b4b7af1b3570e0ec1d4c44088aaf89ee4c9fea3e)

### DIFF
--- a/lib/Script/Plugin.cpp
+++ b/lib/Script/Plugin.cpp
@@ -49,6 +49,10 @@ Plugin::Plugin(plugin::Plugin::Type T, std::string LibraryName,
       ThisModule(Module), ThisConfig(Module.getConfig()),
       IsDefaultPlugin(DefaultPlugin) {}
 
+bool Plugin::isTraced() const {
+  return ThisConfig.options().tracePlugin(getPluginType());
+}
+
 std::string Plugin::resolvePath(const LinkerConfig &PConfig) {
   // Library already loaded!
   if (PluginLibraryHandle)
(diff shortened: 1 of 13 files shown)


lib/Script/Plugin.cpp
parth-07: We should perhaps use `Plugin::isTraced()` here as well.

parth-07: We should use `Plugin::isTraced()` here as well.

tcurtiss: Added the isTraced() to the plugin load/unload as noted above.
Added a plugin API isTraced() , so that plugins could also gate their tracing information for ALL entrypoints into the linker plugin (e.g. Init/Destroy)
Lastly, added unit tests for everything and ran clang-format on the changes to-date.

parth-07 (approved):

quic-seaswara (approved):

quic-areg (approved):

PR #1719, opened 2026-08-17
Main reassign example plugin

tcurtiss:
Add the ability to override symbol addresses within a section via ELD linker plugin API.  Example plugin included.


lib/LinkerWrapper/PluginADT.cpp
```diff
@@ -1020,6 +1020,20 @@ Expected<void> OutputSection::setPluginOverride(Plugin *P,
   return {};
 }
 
+Expected<void> OutputSection::setExcludeFromOverlapCheck(
+    const LinkerWrapper &LW) {
+  CHECK_LINK_STATE(LW, "CreatingSegments");
+  getOutputSection()->getSection()->setExcludedFromOverlapCheck();
```
parth-07, line 1026: Currently, this only affects the overlap checking of virtual addresses. The sections marked as excludedFromOverlapCheck are still checked for overlaps of file offsets and load memory addresses. What do you think about also skipping the overlap checks of file offsets and LMAs for sections marked as excludeFromOverlapCheck, or perhaps we can make the API name more strict such that it is clear from the name that it only skips VMA overlap checks?

lib/LinkerWrapper/PluginADT.cpp
```diff
@@ -1020,6 +1020,20 @@ Expected<void> OutputSection::setPluginOverride(Plugin *P,
   return {};
 }
 
+Expected<void> OutputSection::setExcludeFromOverlapCheck(
+    const LinkerWrapper &LW) {
+  CHECK_LINK_STATE(LW, "CreatingSegments");
```
parth-07, line 1025:
```suggestion
  CHECK_LINK_STATE(LW, "ActBeforeRuleMatching", "BeforeLayout", "ActBeforeSectionMerging"
    "CreatingSections", "ActBeforePerformingLayout", "CreatingSegments");
```

include/eld/PluginAPI/LinkerWrapper.h
```diff
@@ -571,6 +571,22 @@ class DLL_A_EXPORT LinkerWrapper {
...
+  /// Unlike resetSymbol(), this works on already-defined symbols.
+  ///
+  /// \param S Symbol whose address is to be set.
+  /// \param Addr Absolute value to assign to the symbol.
+  ///
+  /// \note This function must only be used in the \em AfterLayout link
+  /// state, i.e. after layout, garbage-collection, and relaxation have
+  /// converged, and before the symbol table and relocations are finalized
+  /// and emitted.
+  eld::Expected<void> setSymbolAddress(plugin::Symbol S, uint64_t Addr);
```
parth-07, line 588: This is a very interesting API. Eventually, we definitely need to record this API usage in the plugin activity log file. I will create a separate issue for the plugin activity log file improvement.

Why do we need to change the symbol binding to Absolute here?

parth-07: Additionally, is it okay to call this API on `local` symbols? This API will convert local symbols to absolute symbols, and eld emits all absolute symbols with `global` binding.

parth-07 (commented):
Thank you for improving the plugin framework. The key API added in this change, `setSymbolAddress`, seems very interesting. Do you want the plugin `Plugins/ReassignSectionAddresses/ReassignSectionAddresses.cpp` that is added in this change to be directly used by the users, or is it primarily meant to serve as an example for users where this API is useful?

lib/LinkerWrapper/PluginADT.cpp
tcurtiss: I intentionally kept the window of usage of this API narrow, so it could only be called in CreatingSegments state prior to AfterLayout.  This is similar to how setPluginOverride() was scoped, so I used that pattern for both setExcludeFromOverlapCheck() and setAllowedInNonLoadSegment().  I suppose allowing it before is 'safe' - but trying to avoid any possible usage error or misguided expectations.

include/eld/PluginAPI/LinkerWrapper.h
tcurtiss: To answer your question above - I included the sample plugin to show how it could be used, no need to merge that since I see you are keeping it lean with just the HelloWorld plugin example.

The setSymbolAddress can be called on any symbol, but a plugin can skip locals (or any other Sym.isXXX() types) before calling it.  Now that you mention it, I'll update my example plugin to skip locals out of correctness, even though the section contains no locals in my use case.

PR #1731, opened 2026-08-18
[ELD] Fix ARM OVERLAY VMA/LMA layout

### LINKED ISSUE #1724: [arm] Linux kernel link fails with "virtual address range overlaps" error
## Problem

The arm Linux kernel build fails when linked with ld.eld with the following error:
```
  LD      .tmp_vmlinux1
Note: Rule CONSTRUCTORS in output section .data does not have any sections specified
Warning: Unrecognized option `--pic-veneer'
Error: section .init.text virtual address range overlaps with .stubs
         >>> .init.text range is [0xC1C003C0, 0xC1CCED67]
         >>> .stubs range is [0xC1C01000, 0xC1C0134F]
Fatal: Linking had errors (.tmp_vmlinux1)
make[2]: *** [scripts/Makefile.vmlinux:72: vmlinux.unstripped] Error 1
make[1]: *** [/path/to/linux-7.1/Makefile:1338: vmlinux] Error 2
```
The build was also test with an implementation of --pic-veneer and the link fail with the same error. Also, this flag is also missing when building aarch64 and the link works, suggesting that this issue is not related to this flag.
The same kernel source, config, and toolchain links and boots successfully when using `ld.lld` instead of `ld.eld`.

## Environment / tool versions

- Kernel source: `linux-7.1`, `defconfig` (arm), unmodified upstream source tree.
- Host: `Linux 6.8.0-124-generic x86_64`.
- Compiler: `clang version 24.0.0git (https://github.com/llvm/llvm-project.git 89db6d6d7f542cee781f608c309634f1efb20dc3)`
- Linker under test: `ld.eld`, `--version` reports `GNU ld compatible linker - eld 24.0`; eld repo commit `c7cda1a5140d5e88e7895a72336541273372025b`.
- Control linker: `ld.lld`, same `llvm-project` build (`89db6d6d7f542cee781f608c309634f1efb20dc3`), same invocation flags.

Steven6798:
Handle OVERLAY members at their shared VMA, assign the first member's AT() LMA, and place subsequent members consecutively. Allow intentional virtual-address overlap between members of the same OVERLAY.

Add an ARM Linux kernel linker-script regression test covering section addresses and program-header VMAs/LMAs.

Fix: qualcomm#1724

### DIFF
--- a/lib/Target/CreateProgramHeaders.hpp
+++ b/lib/Target/CreateProgramHeaders.hpp
@@ -278,8 +278,42 @@ bool GNULDBackend::createProgramHdrs() {
     // Linker script overriding below.
     std::optional<uint64_t> scriptvma;
     bool doAlign = true;
+    OverlayDesc *overlay = (*out)->getOverlayDesc();
+    bool isOverlayMember = overlay != nullptr;
+    bool isFirstOverlayMember = isOverlayMember &&
+                                !overlay->members().empty() &&
+                                overlay->members().front() == *out;
+    OutputSectionEntry *previousOverlayMember = nullptr;
+    if (isOverlayMember && !isFirstOverlayMember) {
+      for (auto *Member : overlay->members()) {
+        if (Member == *out)
+          break;
+        previousOverlayMember = Member;
+      }
+    }
+
+    // OVERLAY members share one VMA, while their LMAs are consecutive. The
+    // member output-section descriptors intentionally have no VMA/LMA
+    // prologue; those values belong to the enclosing OverlayDesc.
+    if (isOverlayMember) {
+      if (overlay->hasStart()) {
+        overlay->start()->evaluateAndRaiseError();
+        scriptvma = overlay->start()->result();
+      } else if (isFirstOverlayMember) {
+        // With no explicit start, the overlay begins at the current dot.
+        scriptvma = dotSymbol->value();
+      } else {
+        // All members use the first member's VMA, including when that member
+        // is empty and therefore does not become the global previous section.
+        scriptvma = overlay->members().front()->getSection()->addr();
+      }
+      if (isCurAlloc)
+        dotSymbol->setValue(*scriptvma);
+      doAlign = false;
+    }
+
     // If the output section specified a VMA value.
-    if ((*out)->prolog().hasVMA()) {
+    if (!isOverlayMember && (*out)->prolog().hasVMA()) {
       (*out)->prolog().vma().evaluateAndRaiseError();
       // If the output section descriptor has an alignment specified
       // honor the alignment specified, the alignment would have been
@@ -345,7 +379,9 @@ bool GNULDBackend::createProgramHdrs() {
     // if this is the first section and the VMA was forced then
     // set PMA = VMA. For all sections following the first section
     // PMA will be calculated separately below
-    if ((*out)->prolog().hasLMA()) {
+    if (isOverlayMember) {
+      disconnect_lma_vma = true;
+    } else if ((*out)->prolog().hasLMA()) {
       useSetLMA = true;
       disconnect_lma_vma = true;
     }
@@ -460,7 +496,21 @@ bool GNULDBackend::createProgramHdrs() {
       cur->setAddr(vma);
       // Handle setting LMA and alignment of LMA
       // Explicitly align LMA to make the code easy to read
-      if (useSetLMA) {
+      if (isOverlayMember) {
+        // Only the first member consumes OVERLAY AT(...); later members are
+        // consecutive in load memory.
+        if (isFirstOverlayMember) {
+          if (overlay->hasLMA()) {
+            overlay->lma()->evaluateAndRaiseError();
+            pma = overlay->lma()->result();
+          } else {
+            pma = vma;
+          }
+        } else {
+          pma = previousOverlayMember->getSection()->pAddr() +
+                previousOverlayMember->getSection()->size();
+        }
+      } else if (useSetLMA) {
         (*out)->prolog().lma().evaluateAndRaiseError();
         pma = (*out)->prolog().lma().result();
       } else if (hasVMARegion || hasLMARegion) {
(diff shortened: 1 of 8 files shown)


lib/Target/CreateProgramHeaders.hpp
```diff
@@ -460,7 +479,20 @@ bool GNULDBackend::createProgramHdrs() {
...
+        // consecutive in load memory.
+        if (isFirstOverlayMember) {
+          if (overlay->hasLMA()) {
+            overlay->lma()->evaluateAndRaiseError();
+            pma = overlay->lma()->result();
+          } else {
+            pma = vma;
+          }
+        } else {
+          pma = prev->pAddr() + prev->size();
```
quic-areg: What if the first overlay member is empty?

test/ARM/standalone/linkerscript/OverlayArmKernelLayout/Inputs/arm-overlay.lds
```diff
@@ -0,0 +1,30 @@
...
+  . = 0xc1c00000;
+  .text : { *(.text) }
+  . = ALIGN(0x1000);
+
+  __vectors_lma = .;
+  OVERLAY 0xffff0000 : NOCROSSREFS AT(__vectors_lma)
+  {
+    .vectors { KEEP(*(.vectors)) }
+    .vectors.bhb.loop8 { KEEP(*(.vectors.bhb.loop8)) }
+    .vectors.bhb.bpiall { KEEP(*(.vectors.bhb.bpiall)) }
```
quic-areg, line 15: Does the test still pass if we reverse the order of the overlay sections?

lib/Target/CreateProgramHeaders.hpp
```diff
@@ -278,8 +278,25 @@ bool GNULDBackend::createProgramHdrs() {
...
+    OverlayDesc *overlay = (*out)->getOverlayDesc();
+    bool isOverlayMember = overlay != nullptr;
+    bool isFirstOverlayMember = isOverlayMember &&
+                                !overlay->members().empty() &&
+                                overlay->members().front() == *out;
+
+    // OVERLAY members share one VMA, while their LMAs are consecutive. The
+    // member output-section descriptors intentionally have no VMA/LMA
+    // prologue; those values belong to the enclosing OverlayDesc.
+    if (isOverlayMember && overlay->hasStart()) {
```
quic-areg: What do we do if the overlay has no explicit start? Do they still share VMAs?

test/ARM/standalone/linkerscript/OverlayArmKernelLayout/Inputs/arm-overlay.lds
Steven6798: I'll add a test for this. Thanks!

lib/Target/CreateProgramHeaders.hpp
Steven6798: The current implementation uses the global previous section for the next LMA, so this case may be incorrect. I'll look into this.

lib/Target/CreateProgramHeaders.hpp
Steven6798: The current code only forces a shared VMA when OVERLAY has an explicit start so I need to fix that. Thanks!

Steven6798:
@quic-areg I see that the zephyr check fails but I don't know what to make of the results.

Steven6798:
Workflows are failing due to reasons unrelated to this PR. I'll track these issues and re-run this PR after they are fixed.

parth-07 (changes requested):
@Steven6798 We also need to update `CreateScriptProgramHeaders` to handle correctly handle overlay when the linker script specifies PHDRS.

Steven6798:
> Steven Ramirez Rosa (@Steven6798) We also need to update `CreateScriptProgramHeaders` to handle correctly handle overlay when the linker script specifies PHDRS.

Thanks. I'll look into this.

test/ARM/standalone/linkerscript/OverlayArmKernelLayout/Inputs/overlay-reversed.lds
```diff
@@ -0,0 +1,16 @@
...
+  .text : { *(.text) }
+  . = ALIGN(0x1000);
+
+  OVERLAY 0xffff0000 : AT(0xc1c00000)
+  {
+    .vectors.bhb.bpiall { KEEP(*(.vectors.bhb.bpiall)) }
+    .vectors.bhb.loop8 { KEEP(*(.vectors.bhb.loop8)) }
+    .vectors { KEEP(*(.vectors)) }
+  }
+}
```
quic-areg, line 16: Can you place something after the overlay to make sure that the location counter is advanced correctly? Something like `after = .;`. From the issue desc, it looks like we are advancing the location counter after an overlay by the size of the last section rather than the largest section.

```sh
cat > 1.s << \!
.section .a, "ax", %progbits
  .space 0x20
.section .b, "ax", %progbits
  .space 0x40
.section .t, "ax", %progbits
.global _start
_start:
  .space 0x10
!

cat > script.t << \!
SECTIONS {
  . = 0x1000;
  .t : { *(.t) }
  OVERLAY 0x9000 : AT(0x8000) { .b { *(.b) } .a { *(.a) } }
  after = .;
}
!

clang -target arm-linux-gnueabi -c 1.s
ld.eld 1.o -T script.t -o A
ld.lld 1.o -T script.t -o B
llvm-readelf -s A B | grep after
```

lib/Target/CreateProgramHeaders.hpp
```diff
@@ -278,8 +278,42 @@ bool GNULDBackend::createProgramHdrs() {
...
+        previousOverlayMember = Member;
+      }
+    }
+
+    // OVERLAY members share one VMA, while their LMAs are consecutive. The
+    // member output-section descriptors intentionally have no VMA/LMA
+    // prologue; those values belong to the enclosing OverlayDesc.
+    if (isOverlayMember) {
+      if (overlay->hasStart()) {
+        overlay->start()->evaluateAndRaiseError();
```
quic-areg, line 300: Do we need to re-evaluate the start expression for each member?

Steven6798:
The initial scope of this PR was to address a failure in ARM Linux kernel build. The Overlay feature applies to more than just ARM and it looks like ELD is missing much more than I anticipated. We might need to break this feature into multiple follow-ups, address one thing at a time and test all architectures. This looks like a better approach than to focus everything in one PR focused just on ARM. Any feedback?

Steven6798:
> We might need to break this feature into multiple follow-ups, address one thing at a time and test all architectures. This looks like a better approach than to focus everything in one PR focused just on ARM.

Follow up: #1745

PR #1735, opened 2026-08-18
Allocate GOT/PLT slots into one shared section deterministically

### LINKED ISSUE #1834: fix non-deterministic GOT/PLT slot order

quic-areg:
Previously, every target created its own .got/.got.plt/.plt per input object
file and relied on section merging to stitch them back together. This PR instead
creates them once in a shared section and places every slot there, and makes the
slot order deterministic.

.rela.dyn and .rela.plt are created once alongside .got/.plt rather than per
input file.

Slots are now created during relocation scanning deterministically in input
order, matching lld. The scan is now serial. Slot creation within it was already
serialized on a mutex, so parallelism only decided which input reached the lock
first.

ifunc loops now iterate .plt instead of a DenseMap, which is not deterministic
even under a single-threaded link.

Fixes #1834, #1856

### DIFF
--- a/lib/Target/ELFDynamic.cpp
+++ b/lib/Target/ELFDynamic.cpp
@@ -16,6 +16,9 @@
 #include "eld/Core/Module.h"
 #include "eld/Diagnostics/DiagnosticPrinter.h"
 #include "eld/Fragment/DynStrFragment.h"
+#include "eld/Fragment/GOT.h"
+#include "eld/Object/OutputSectionEntry.h"
+#include "eld/Object/RuleContainer.h"
 #include "eld/Readers/ELFSection.h"
 #include "eld/Support/MsgHandling.h"
 #include "eld/SymbolResolver/LDSymbol.h"
@@ -341,14 +344,32 @@ void ELFDynamic::applyEntries(GNULDBackend &pBackend,
     applyOne(llvm::ELF::DT_STRSZ, DynStrSize);  // DT_STRSZ
   }
 
-  if (const ELFSection *GOTPLT = pBackend.getGOTPLT())
-    if (GOTPLT->size() != 0)
-      // DT_PLTGOT always points to the GOTPLT section. Glad that the
-      // linker treats .got.plt section as internal. DT_PLTGOT is needed
-      // only by lazy binding. Note that both ld and lld create
-      // DT_PLTGOT even with lazy binding and the image crashes without
-      // it on riscv qemu.
-      applyOne(llvm::ELF::DT_PLTGOT, GOTPLT->addr());
+  if (ELFSection *GOTPLT = pBackend.getGOTPLT())
+    if (GOTPLT->size() != 0) {
+      // DT_PLTGOT points at the reserved GOTPLT0 slot. The section's own
+      // fragment list is empty after merge and its addr field is stale, so find
+      // the slot by walking the output section.
+      uint64_t PLTGOTAddr = GOTPLT->addr();
+      if (OutputSectionEntry *Out = GOTPLT->getOutputSection())
+        for (RuleContainer *Rule : *Out) {
+          ELFSection *RS = Rule->getSection();
+          if (!RS)
+            continue;
+          const GOT *GotFrag = nullptr;
+          for (Fragment *F : RS->getFragmentList()) {
+            auto *G = llvm::dyn_cast<GOT>(F);
+            if (G && G->getType() == GOT::GOTPLT0) {
+              GotFrag = G;
+              break;
+            }
+          }
+          if (GotFrag) {
+            PLTGOTAddr = GotFrag->getAddr(m_Config.getDiagEngine());
+            break;
+          }
+        }
+      applyOne(llvm::ELF::DT_PLTGOT, PLTGOTAddr);
+    }
 
   pBackend.applyTargetDynamicEntries();
 
--- a/lib/Target/GNULDBackend.cpp
+++ b/lib/Target/GNULDBackend.cpp
@@ -1709,16 +1709,43 @@ uint64_t GNULDBackend::getSymbolValue(LDSymbol *pSymbol) const {
   return pSymbol->value();
 }
 
-ELFSection *GNULDBackend::getGOT() const {
-  return m_DynamicSectionHeadersInputFile->getGOT();
-}
+ELFSection *GNULDBackend::getGOT() const { return GOTSection; }
 
-ELFSection *GNULDBackend::getGOTPLT() const {
-  return m_DynamicSectionHeadersInputFile->getGOTPLT();
-}
+ELFSection *GNULDBackend::getGOTPLT() const { return GOTPLTSection; }
 
-ELFSection *GNULDBackend::getPLT() const {
-  return m_DynamicSectionHeadersInputFile->getPLT();
+ELFSection *GNULDBackend::getPLT() const { return PLTSection; }
+
+void GNULDBackend::initDynamicSections(ELFObjectFile &InputFile,
+                                       const DynamicSectionLayout &Layout) {
+  bool IsRela = Layout.RelType == llvm::ELF::SHT_RELA;
+  ELFSection &RelDyn = *m_Module.createInternalSection(
+      InputFile, LDFileFormat::DynamicRelocation,
+      IsRela ? ".rela.dyn" : ".rel.dyn", Layout.RelType, llvm::ELF::SHF_ALLOC,
+      Layout.RelAlign);
+  ELFSection &RelPLT = *m_Module.createInternalSection(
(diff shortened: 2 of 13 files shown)


quic-seaswara (commented):
A few questions before I do a full review.

From the PR description:
The scan is now serial. GOT slot creation within the scan was already serialized by a mutex, so the previous parallelism only affected which input acquired the lock first.

=> The GOT slot will still be created by whichever input reaches it first, which means the behavior remains non-deterministic.

Regarding the GOT/PLT design, they still allocate fragments on a per-GOT-slot and per-PLT-slot basis. What I was proposing instead is:

- A single GOT fragment that contains multiple GOT slots.

- A single PLT fragment that contains multiple PLT slots.

In this model, the individual slots would be private regions within their respective GOT or PLT fragment, rather than each slot being represented by a separate fragment.

This would eliminate fragment creation ordering as a source of non-determinism and provide a more natural representation of the GOT and PLT structures.

parth-07 (approved):
LGTM.

quic-areg:
> => The GOT slot will still be created by whichever input reaches it first, which means the behavior remains non-deterministic.

The first input requesting a slot does determine its index, but that input is deterministic for a fixed invocation of the linker. I.e. for the same link, slots will be created in the same order every time. Changing the command-line input order can change slot order, like it does in lld. The PR includes tests to verify that repeated links produce the same slots. 

> What I was proposing instead...

I want to keep this patch focused on fixing non-determinism. Removing per-slot fragments is a useful follow-up refactoring but it is not required to fix non-determinism.

lib/Target/ELFDynamic.cpp
```diff
@@ -341,14 +356,16 @@ void ELFDynamic::applyEntries(GNULDBackend &pBackend,
...
-      // it on riscv qemu.
-      applyOne(llvm::ELF::DT_PLTGOT, GOTPLT->addr());
+  if (ELFSection *GOTPLT = pBackend.getGOTPLT())
+    if (GOTPLT->size() != 0) {
+      // DT_PLTGOT points at the reserved GOTPLT0 slot. The section's own
+      // fragment list is empty after merge and its addr field is stale, so find
+      // the slot by walking the output section.
+      uint64_t PLTGOTAddr = GOTPLT->addr();
+      if (const GOT *G = findGOTPLT0(GOTPLT->getOutputSection()))
+        PLTGOTAddr = G->getAddr(m_Config.getDiagEngine());
```
quic-seaswara, line 366: Do we need to go through the fragment list here ? Can we just use the address of GOTPLT->getOutputELFSection()->addr() ?

lib/Target/GNULDBackend.cpp
quic-seaswara, line 1647: Do we need this special handling of DynamicSectionHeadersInputFile ? 

I feel this can be designed as an internal input file that the Module will create for all internal inputs.

We can remove these extra weight that we carry for every ELF object file.

```
ELFSection *GOT = nullptr;
ELFSection *GOTPLT = nullptr;
ELFSection *PLT = nullptr;
ELFSection *RelaDyn = nullptr;
ELFSection *RelaPLT = nullptr;

```

lib/Object/ObjectLinker.cpp
```diff
@@ -2229,54 +2229,19 @@ bool ObjectLinker::scanRelocations(bool IsPartialLink) {
...
-      auto &CopyRelocsRef = *CopyRelocs; // must dereference in the main thread
-      Pool->async([&] {
-        scanRelocationsHelper(Input, IsPartialLink, PluginVect, CopyRelocsRef);
-      });
-      AllCopyRelocs.push_back(std::move(CopyRelocs));
-    }
-    Pool->wait();
+  for (auto &Input : ThisModule->getObjectList()) {
+    auto CopyRelocs = std::make_unique<Relocator::CopyRelocs>();
+    scanRelocationsHelper(Input, IsPartialLink, PluginVect, *CopyRelocs);
```
quic-seaswara, line 2236: ScanRelocations is no longer multi-threaded? This can potentially impact link-time performance.

We would like ScanRelocations to remain multi-threaded and preserve the current parallelism.

I missed this aspect in my initial review comments.

quic-seaswara (changes requested):

lib/Object/ObjectLinker.cpp
parth-07: I don't think that we were getting multi-threaded benefits from scan relocations phase as the core component of ScanRelocations was guarded by a mutex. I believe that we should eventually move to a model where scan relocations is multi-threaded but in the parallel part it only stores decisions in a per-symbol bucket (so that there is no data-race), and then in the sequential part it actually allocates PLT/GOT slots. However, I believe that this design change is outside the scope of this PR.

quic-seaswara: I may be missing something, but why would we give up multithreading here? Even for purely static links, where no GOT slots need to be allocated by the linker, parallel processing should still provide meaningful link-time performance benefits.

My preference would be to address the non-determinism directly rather than remove the multithreading. If the current design exposes a non-deterministic behavior, it would be ideal to make the implementation deterministic while preserving the existing parallelism.

I also think the cleanup performed by this patch is valuable and should be kept. What I'm suggesting is that we separate the two concerns: retain the cleanup and refactoring improvements, but fix the non-determinism in a way that allows the code to remain multithreaded. That would let us preserve the performance benefits while still achieving deterministic behavior.

parth-07: In this case, @quic-seaswara @quic-areg  What do you think about the design that @bsaharsh implemented in https://github.com/qualcomm/eld/pull/1467. It keeps the multithreaded relocation scan phase and move the part that introduces non-determinism (and requires a mutex) into a separate follow-up sequential phase.

quic-seaswara: Yes, we can probably reuse the work implemented by @bsaharsh.

quic-areg: It might be worth measuring how much link time we give up by serializing ScanRelocations. ScanRelocations takes up <1% of link time and is largely (but not completely) currently dominated by a mutex. Serializing would increase link time on the order of seconds while some individual link stages can take on the order of minutes to complete. 

There also seems to be a TSan issue with the current approach https://github.com/qualcomm/eld/issues/1856

My original plan was to serialize ScanRelocations with a central GOT section to solve non-determinism (this PR) and then re-introduce parallelism, without a mutex in a follow-up patch. Saharsh's approach is similar to what I had in mind and roughly follows lld's approach:
1.  scan inputs in parallel and record deferred got/plt slot requests
2.  apply those requests deterministically in input order, serially 

I am in the processes of implementing this approach as we speak. It's looking good but I am not seeing the speedup I expected. 

I am okay with doing this up front, but my concern was that the combined patch would be easier to review and test if it were split, especially given the cost of serializing ScanRelocations is not that high. Please let me know your thoughts. Either way I am working on parallelism; let me know if it should go with this PR or the next one.

quic-seaswara: I am fine with introducing the parallelism in a follow up patch. 

If you are working with object files that do not have any GOT relocations, I would think it we will get a speed up.

quic-seaswara (approved):

quic-seaswara (changes requested):
There are other comments outside multithreading that needs to be cleared up.

lib/Target/ELFDynamic.cpp
quic-areg: Yes, fragment lookup is needed here. `GOTPLT->getOutputELFSection()->addr()` gives the output section base address which is not always the address of GOTPLT0. There is a test included in this patch for this case where GOTPLT0 has a nonzero offset.

lib/Target/GNULDBackend.cpp
quic-areg: I intentionally kept the old `ELFObjectFile` APIs in this patch to reduce the diff and limit the scope of the change. I intended to address this in a follow-up since there is a lot of API churn. I Agree that this could be a separate input file.

quic-seaswara: Let us raise issues for all the follow ups.

lib/Object/ObjectLinker.cpp
parth-07: I think parallelism should be in a separate PR. 

> It might be worth measuring how much link time we give up by serializing ScanRelocations. ScanRelocations takes up <1% of link time and is largely (but not completely) currently dominated by a mutex.

I agree that it is worth investigating whether the extra complexity of parallelizing ScanRelocations is actually backed up by the improvement in performance.

quic-seaswara (approved):

Steven6798 (approved):

PR #1740, opened 2026-08-19
[workflow] Remove artifacts after successful runs

### LINKED ISSUE #1739: [workflow] Remove artifacts after successful runs
Currently workflows that generate artifacts have a 1-day retention policy. This becomes a problem when the next runs fail for causes outside of ELD (like GitHub issues). We should remove the artifacts only after a successful run and we can have a 7-day retention policy in case the error keeps happening consistently.

Steven6798:
Artifacts are now removed after a succesful run. The retention policy has been changed to 7 days in case there are repeated failures.

Fix: qualcomm#1739


Steven6798:
Will address in a branch in eld instead of my fork.

PR #1741, opened 2026-08-20
[Reproduce] Fix thin archive member reproducer paths

### LINKED ISSUE #1737: --reproduce mishandles thin archive members with identical basenames
ELD’s --reproduce mishandles thin archive members with identical basenames. Different files such as a/foo.o and b/foo.o are stored under the same generated path, causing replay to lose or replace one member.

reproduce:
```
mkdir {a,b}

printf '%s\n' \
    'int only_a(void) { return 1; }' > a/foo.c
printf '%s\n' \
    'int only_b(void) { return 2; }' > b/foo.c
printf '%s\n' \
    'extern int only_a(void);' \
    'extern int only_b(void);' \
    'int main(void) { return only_a() + only_b(); }' > main.c
clang -c a/foo.c -o a/foo.o
clang -c b/foo.c -o b/foo.o
clang -c main.c -o main.o
ar rcT libsame.a a/foo.o b/foo.o

ld.eld main.o libsame.a -o same.out --reproduce=repro.tar

tar -xf repro.tar

cd reproduce.same.out 
bash -x response.txt
```

Result:

```
Error: multiple definition of symbol `only_a' in file libsame.a(a/foo.o) and libsame.a(b/foo.o)

### PR DESCRIPTION
**Pull request by rachitmeht**
Include the thin archive member name and child offset in the generated input identity. This prevents members with identical basenames from colliding in --reproduce archives and preserves both files during replay.

Resolves #1737

### DIFF
--- a/lib/Support/OutputTarWriter.cpp
+++ b/lib/Support/OutputTarWriter.cpp
@@ -74,10 +74,17 @@ void OutputTarWriter::addInputFile(const InputFile *File, bool IsLTOObject) {
 }
 
 std::string OutputTarWriter::getHashAndExtension(const Input *Ipt) const {
+  uint64_t InputHash = Ipt->getResolvedPathHash();
+  if (const auto *ArchiveMember = llvm::dyn_cast<ArchiveMemberInput>(Ipt))
+    // Thin archive members share the parent archive's resolved path. Include
+    // the member identity so members with the same basename cannot collide.
+    InputHash = llvm::hash_combine(InputHash, ArchiveMember->getMemberName(),
+                                   ArchiveMember->getChildOffset());
+
   // Returns filename passed to the linker along with the file hash.
   return std::string(
              llvm::sys::path::filename(Ipt->getInputFile()->getMappedPath())) +
-         "." + std::to_string(Ipt->getResolvedPathHash());
+         "." + std::to_string(InputHash);
 }
 
 /// Create mapping.ini file of input filepath to its sha2 hash + file extension
--- a/lib/Support/OutputTarWriter.cpp
+++ b/lib/Support/OutputTarWriter.cpp
@@ -74,7 +74,6 @@ void OutputTarWriter::addInputFile(const InputFile *File, bool IsLTOObject) {
 }
 
 std::string OutputTarWriter::getHashAndExtension(const Input *Ipt) const {
-  // Returns filename passed to the linker along with the file hash.
   return std::string(
              llvm::sys::path::filename(Ipt->getInputFile()->getMappedPath())) +
          "." + std::to_string(Ipt->getResolvedPathHash());
--- a/test/Common/standalone/CommandLine/Reproduce/ReproduceThinArchiveSameBasename.test
+++ b/test/Common/standalone/CommandLine/Reproduce/ReproduceThinArchiveSameBasename.test
@@ -1,0 +1,28 @@
+#---ReproduceThinArchiveSameBasename.test----------------- Executable -----------------#
+
+#BEGIN_COMMENT
+# Checks that --reproduce preserves thin-archive members with the same basename.
+#END_COMMENT
+#START_TEST
+RUN: %rm -rf %t.tmpdir
+RUN: %mkdir %t.tmpdir
+RUN: %mkdir %t.tmpdir/a
+RUN: %mkdir %t.tmpdir/b
+RUN: %clang %clangopts -c %p/Inputs/a/foo.c -o %t.tmpdir/a/foo.o
+RUN: %clang %clangopts -c %p/Inputs/b/foo.c -o %t.tmpdir/b/foo.o
+RUN: %clang %clangopts -c %p/Inputs/thin-main.c -o %t.tmpdir/main.o
+RUN: cd %t.tmpdir
+RUN: %ar cr %aropts --thin %t.tmpdir/libsame.a a/foo.o b/foo.o
+RUN: %link --no-threads %t.tmpdir/main.o %t.tmpdir/libsame.a -o %t.same.out \
+RUN:   --reproduce %t.repro.tar --dump-mapping-file %t.mapping.ini \
+RUN:   --dump-response-file %t.response
+RUN: %filecheck %s < %t.mapping.ini
+RUN: %mkdir %t.replay
+RUN: cd %t.replay
+RUN: %tar %gnutaropts -xf %t.repro.tar --strip-components=1
+RUN: %link --no-threads @response.txt -o %t.same.replay.out
+RUN: %diff %t.same.out %t.same.replay.out
+#END_TEST
+
+CHECK-DAG: Object/foo.o.{{[0-9]+}}
+CHECK-DAG: Object/foo.o.{{[0-9]+}}
--- a/test/Common/standalone/CommandLine/Reproduce/ReproduceThinArchiveSameBasename.test
+++ b/test/Common/standalone/CommandLine/Reproduce/ReproduceThinArchiveSameBasename.test
@@ -1,0 +1,27 @@
+#---ReproduceThinArchiveSameBasename.test----------------- Executable -----------------#
+
+#BEGIN_COMMENT
+# Checks that --reproduce preserves thin-archive members with the same basename.
+#END_COMMENT
+#START_TEST
+RUN: %rm -rf %t.tmpdir
+RUN: %mkdir %t.tmpdir
+RUN: %mkdir %t.tmpdir/a
+RUN: %mkdir %t.tmpdir/b
+RUN: %clang %clangopts -c %p/Inputs/a/foo.c -o %t.tmpdir/a/foo.o
+RUN: %clang %clangopts -c %p/Inputs/b/foo.c -o %t.tmpdir/b/foo.o
+RUN: %clang %clangopts -c %p/Inputs/thin-main.c -o %t.tmpdir/main.o
+RUN: cd %t.tmpdir
+RUN: %ar cr %aropts --thin %t.tmpdir/libsame.a a/foo.o b/foo.o
+RUN: %link --no-threads %t.tmpdir/main.o %t.tmpdir/libsame.a -o %t.same.out \
+RUN:   --reproduce %t.repro.tar --dump-mapping-file %t.mapping.ini
+RUN: %filecheck %s < %t.mapping.ini
+RUN: %mkdir %t.replay
+RUN: %tar %gnutaropts -xf %t.repro.tar -C %t.replay --strip-components=1
+RUN: cd %t.replay
+RUN: %python -c "import shlex, subprocess; args = shlex.split(open('response.txt').read())[1:]; args[args.index('-o') + 1] = r'%t.same.replay.out'; subprocess.check_call(shlex.split(r'%link') + args)"
+RUN: %diff %t.same.out %t.same.replay.out
+#END_TEST
+
+CHECK-DAG: Object/foo.o.{{[0-9]+}}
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by parth-07** [comment]
@rachitmeht Can you please add tests?

File: lib/Support/OutputTarWriter.cpp [hunk 1]
```diff
@@ -74,10 +74,17 @@ void OutputTarWriter::addInputFile(const InputFile *File, bool IsLTOObject) {
 }
 
 std::string OutputTarWriter::getHashAndExtension(const Input *Ipt) const {
+  uint64_t InputHash = Ipt->getResolvedPathHash();
+  if (const auto *ArchiveMember = llvm::dyn_cast<ArchiveMemberInput>(Ipt))
```
parth-07: Won't we only call `getHashAndExtension` on top-level inputs, and thus, it should never be called for archive members?

parth-07 (commented):
@rachitmeht With this fix, will the created reproducer work standalone?

test/Common/standalone/CommandLine/Reproduce/ReproduceThinArchiveSameBasename.test
```diff
@@ -0,0 +1,27 @@
+#UNSUPPORTED: windows, reproduce_fail
```
parth-07: Why is the test unsupported for windows? and we likely do not need reproduce_fail either here.

test/Common/standalone/CommandLine/Reproduce/ReproduceThinArchiveSameBasename.test
```diff
@@ -0,0 +1,27 @@
...
+RUN: %rm -rf %t.tmpdir
+RUN: %mkdir %t.tmpdir
+RUN: %mkdir %t.tmpdir/a
+RUN: %mkdir %t.tmpdir/b
+RUN: %clang %clangopts -c %p/Inputs/a/foo.c -o %t.tmpdir/a/foo.o
+RUN: %clang %clangopts -c %p/Inputs/b/foo.c -o %t.tmpdir/b/foo.o
+RUN: %clang %clangopts -c %p/Inputs/thin-main.c -o %t.tmpdir/main.o
+RUN: cd %t.tmpdir
+RUN: %ar cr %aropts --thin libsame.a a/foo.o b/foo.o
+RUN: %link --no-threads main.o libsame.a -o same.out --reproduce repro.tar --dump-mapping-file mapping.ini
```
parth-07: Can you prefix all output files using `%t`? It makes debugging easier as then we do not need additional efforts in finding the full path of the outputs.

lib/Support/OutputTarWriter.cpp
```diff
@@ -74,10 +74,17 @@ void OutputTarWriter::addInputFile(const InputFile *File, bool IsLTOObject) {
...
+    // Thin archive members share the parent archive's resolved path. Include
+    // the member identity so members with the same basename cannot collide.
+    InputHash = llvm::hash_combine(InputHash, ArchiveMember->getMemberName(),
```
parth-07: Would using the member name alone not be enough to make the hash different here?

parth-07 (changes requested):

test/Common/standalone/CommandLine/Reproduce/ReproduceThinArchiveSameBasename.test
rachitmeht: The windows restriction was there because I used bash -x to replay response.txt, I will change it.

lib/Support/OutputTarWriter.cpp
rachitmeht: I took reference for this from this lld implementation.  ELF/InputFiles.cpp

 // ThinLTO assumes that all MemoryBufferRefs given to it have a unique
  // name. If two archives define two members with the same name, this
  // causes a collision which result in only one of the objects being taken
  // into consideration at LTO time (which very likely causes undefined
  // symbols later in the link stage). So we append file offset to make
  // filename unique.

For a/foo.c and b/foo.c 
without offset it will be only  "libname.a + foo.c"

parth-07: The archive member names should be `a/foo.o` and `b/foo.o` instead of just `foo.o`.

rachitmeht: I read gnu ar docs, -q allows duplicate files to get appended. Maybe that's the reason lld implements with child offset for uniqueness.

parth-07: If we have duplicate file `a/foo.o` in a thin archive, then do we need to store them multiple times in the reproduce tarball with different hashes, or is just one file with the same hash enough?

Can you please check and report here whether the `outputTarWriter::getHashAndExtension` function is called for regular (non-thin) archive member?

rachitmeht: For an exact duplicate a/foo.o entry in a thin archive, one copied file and one mapping/hash are sufficient. I will change it to hash(Archive name + Member name) only.

getHashAndExtension() is not used for regular archive members. addInputFileToTar() skips archive members; only the regular archive itself is recorded. Thin members are added separately and use the archive-member hash logic.

parth-07:
@rachitmeht Can you reply/mark-resolve the comments which are completed?

test/Common/standalone/CommandLine/Reproduce/ReproduceThinArchiveSameBasename.test
```diff
@@ -0,0 +1,27 @@
...
+RUN: %clang %clangopts -c %p/Inputs/thin-main.c -o %t.tmpdir/main.o
+RUN: cd %t.tmpdir
+RUN: %ar cr %aropts --thin %t.tmpdir/libsame.a a/foo.o b/foo.o
+RUN: %link --no-threads %t.tmpdir/main.o %t.tmpdir/libsame.a -o %t.same.out \
+RUN:   --reproduce %t.repro.tar --dump-mapping-file %t.mapping.ini
+RUN: %filecheck %s < %t.mapping.ini
+RUN: %mkdir %t.replay
+RUN: %tar %gnutaropts -xf %t.repro.tar -C %t.replay --strip-components=1
+RUN: cd %t.replay
+RUN: %python -c "import shlex, subprocess; args = shlex.split(open('response.txt').read())[1:]; args[args.index('-o') + 1] = r'%t.same.replay.out'; subprocess.check_call(shlex.split(r'%link') + args)"
```
parth-07: Why do we have this inline python script here? What was wrong with replaying the response file using bash?

rachitmeht: > Why is the test unsupported for windows? and we likely do not need reproduce_fail either here.

> The windows restriction was there because I used bash -x to replay response.txt, I will change it. 

For changing the windows restrictions I changed it to python script

rachitmeht: Reproduce/Namespec.test this also uses python script for replay.

parth-07: Many tests use `%link response.txt` for replaying the reproducer link (for example: test/Common/standalone/CommandLine/Reproduce/SysrootMarker.test). Can you check if there is any benefit of using this long inline python script over the `%link response.txt` method?

rachitmeht: No, there is no benefit over `%link response.txt`, I will change it.

test/Common/standalone/CommandLine/Reproduce/ReproduceThinArchiveSameBasename.test
```diff
@@ -0,0 +1,28 @@
...
+RUN: %clang %clangopts -c %p/Inputs/b/foo.c -o %t.tmpdir/b/foo.o
+RUN: %clang %clangopts -c %p/Inputs/thin-main.c -o %t.tmpdir/main.o
+RUN: cd %t.tmpdir
+RUN: %ar cr %aropts --thin %t.tmpdir/libsame.a a/foo.o b/foo.o
+RUN: %link --no-threads %t.tmpdir/main.o %t.tmpdir/libsame.a -o %t.same.out \
+RUN:   --reproduce %t.repro.tar --dump-mapping-file %t.mapping.ini \
+RUN:   --dump-response-file %t.response
+RUN: %filecheck %s < %t.mapping.ini
+RUN: %mkdir %t.replay
+RUN: %tar %gnutaropts -xf %t.repro.tar -C %t.replay --strip-components=1
```
Steven6798: tar .... -C is not compatible with Windows. you can cd into %t.replay first that way you don't need -C

Steven6798 (changes requested):

lib/Support/OutputTarWriter.cpp
```diff
@@ -74,10 +74,16 @@ void OutputTarWriter::addInputFile(const InputFile *File, bool IsLTOObject) {
 }
 
 std::string OutputTarWriter::getHashAndExtension(const Input *Ipt) const {
+  uint64_t InputHash = Ipt->getResolvedPathHash();
```
parth-07: What do you think about correctly setting getResolvedPathHash for ArchiveMemberInput such that it is unique for different archive members? This way we would not need to do anything special for archive members in this function.

parth-07 (changes requested):

lib/Support/OutputTarWriter.cpp
rachitmeht: I changed resolvedPathHash() from hash(resolvedpath) --> hash(resolvedpath+membername) (For ArchiveMember only). It is working, but we are only changing hash but not resolvedpath itself. is it good?

rachitmeht: resolvedpath for an archive member only contains archive name currently.

parth-07: Can you tell me what are your thoughts on this? What do you think ideally should be the resolvedPath and Hash be of ArchiveMemberInput?

rachitmeht: Ideally for an ArchiveMember, We should have ResolvedPath = Archive path only, So,  Hash should also be hash(ResolvedPath) only. But, for thinArchive there can be ResolvedPath = complete Archive member path.

parth-07: Why should hash be hash(ResolvedPath)? What is the use of the hash of a InputFile/Input object?

rachitmeht: I mean't, Ideally, according to name semantically ResolvedPathHash should be Hash(ResolvedPath). Yes, every hash in the codebase is used for Mapping match purpose only. So, it makes sense to make it unique for archive members.

parth-07: We can always change `ResolvedPathHash` name to something like `InputHash` to make the new behaviour correct semantically.

lib/Support/OutputTarWriter.cpp
```diff
@@ -74,7 +74,6 @@ void OutputTarWriter::addInputFile(const InputFile *File, bool IsLTOObject) {
 }
 
 std::string OutputTarWriter::getHashAndExtension(const Input *Ipt) const {
-  // Returns filename passed to the linker along with the file hash.
```
parth-07, line 77: Why remove this comment?

rachitmeht: Yes, it should not be  removed, Thanks.

parth-07:
@rachitmeht Please rebase the PR and make clang-format builder happy?

PR #1751, opened 2026-08-22
Main setsymboladdress only (alternate to PR 1749)

tcurtiss:
Use this PR if the setExcludeFromOverlapCheck and setAllowedInNonLoadSegment plugin APIs in PR 1749 are not needed.

LinkerWrapper::setSymbolAddress(Symbol, Addr): detaches a defined symbol from its fragment and assigns it an absolute (SHN_ABS) value, callable during AfterLayout. Preserves original binding (e.g. global, local, weak)


lib/Core/Module.cpp
```diff
@@ -783,6 +783,18 @@ bool Module::resetSymbol(ResolveInfo *R, Fragment *F) {
   return true;
 }
 
+bool Module::setSymbolAddress(ResolveInfo *R, uint64_t Addr) {
+  if (!R || !R->outSymbol())
+    return false;
```
quic-seaswara, line 794: You might want  to return eld::Expected here.

lib/Core/Module.cpp
```diff
@@ -783,6 +783,18 @@ bool Module::resetSymbol(ResolveInfo *R, Fragment *F) {
...
+  // Clearing the fragment ref (rather than setBinding(Absolute)) is what
+  // makes the symbol emit with SHN_ABS; see GNULDBackend::getSymbolShndx.
+  // The symbol's original binding (e.g. Local) is preserved.
+  R->outSymbol()->setFragmentRef(FragmentRef::null());
+  R->setDesc(ResolveInfo::Define);
+  R->setValue(Addr, /*IsFinal=*/true);
```
quic-seaswara, line 800: This needs to be recorded in the map file, that a plugin changed this symbol to be absolute.

lib/LinkerWrapper/LinkerWrapper.cpp
```diff
@@ -842,6 +842,16 @@ eld::Expected<void> LinkerWrapper::resetSymbol(plugin::Symbol S, Chunk C) {
   return {};
 }
 
+eld::Expected<void> LinkerWrapper::setSymbolAddress(plugin::Symbol S,
+                                                    uint64_t Addr) {
+  CHECK_LINK_STATE(*this, "AfterLayout");
+  bool b = m_Module.setSymbolAddress(S.getSymbol(), Addr);
+  if (!b)
+    return std::make_unique<DiagnosticEntry>(DiagnosticEntry(
+        Diag::error_failed_to_set_symbol_address, {S.getName()}));
```
quic-seaswara: You might also want to show the origin of the symbol ?

quic-seaswara (changes requested):

lib/Core/Module.cpp
tcurtiss: That doesn't match the convention used in the rest of the file.  The top-level plugin API does use eld::Expected, though, as with the convention there.

lib/Core/Module.cpp
tcurtiss: Good point - let me figure out a clean way to wire that up in a concise way in the map file.

lib/LinkerWrapper/LinkerWrapper.cpp
tcurtiss: Agree - will do!

tcurtiss: Added decoratedPath() in latest commit, to pinpoint where the offending symbol was coming from

lib/Core/Module.cpp
tcurtiss: Added in latest commit, along with lit tests.  Fairly small code change, but had to thread a pluginOp thru to track the symbol addr modification by plugin all the way to map file generation.

test/Common/Plugin/SetSymbolAddress/SetSymbolAddress.cpp
```diff
@@ -0,0 +1,61 @@
...
+      TargetSymbol = "bar";
+  }
+
+  void processOutputSection(OutputSection O) override {}
+
+  Status Run(bool Trace) override {
+    if (CallInCreatingSections && getLinker()->isLinkStateCreatingSections())
+      return callSetSymbolAddress(0x1234);
+    if (!CallInCreatingSections && getLinker()->isLinkStateAfterLayout())
+      return callSetSymbolAddress(0x12345678);
```
quic-seaswara: I have a question, since this happens to be AfterLayout and if this address is out of bounds, linker wont be able to create a trampoline.

Can we add a test to make sure that is the behavior needed.

tcurtiss: Added a check limiting setSymbolAddress to data symbols, this would not be used on function symbols.  Also added a range-check to ensure the address is within the expected range of target architecture.

quic-seaswara (approved):

lib/LinkerWrapper/LinkerWrapper.cpp
```diff
@@ -842,6 +844,27 @@ eld::Expected<void> LinkerWrapper::resetSymbol(plugin::Symbol S, Chunk C) {
   return {};
 }
 
+eld::Expected<void> LinkerWrapper::setSymbolAddress(plugin::Symbol S,
+                                                    uint64_t Addr) {
+  CHECK_LINK_STATE(*this, "AfterLayout");
+  if (S.isFunction())
```
parth-07: We should also error out for `IFunc` symbols as well. IFunc is a GNU functionality, but it is widely supported and used because GNU libc heavily use this. As per GNU ABI, IFunc function symbols have a different symbol type than the ordinary symbols (in eld, IFunc symbols have the type `ResolveInfo::Type::IndirectFunc`)

include/eld/PluginAPI/LinkerWrapper.h
```diff
@@ -571,6 +571,33 @@ class DLL_A_EXPORT LinkerWrapper {
...
+  /// state, i.e. after layout, garbage-collection, and relaxation have
+  /// converged, and before the symbol table and relocations are finalized
+  /// and emitted.
+  ///
+  /// \note Reassigning a symbol's address this way does not change the
+  /// address range that the linker's virtual-address overlap check
+  /// associates with the symbol's original output section. If the new
+  /// address could fall inside another section's address range, declare
+  /// the symbol's output section with the \c INFO, \c COPY, \c DSECT, or
+  /// \c OVERLAY linker script section type. These types are equivalent in
```
parth-07, line 594: The user does not necessarily need to change the other output section's type to `INFO|COPY|DSECT` if the symbol's new address falls in this other output output section, right? The linker does not use symbol values during range overlap checks.

include/eld/PluginAPI/LinkerWrapper.h
```diff
@@ -571,6 +571,33 @@ class DLL_A_EXPORT LinkerWrapper {
...
+  /// ELD: each removes the section from the overlap check (and from
+  /// PT_LOAD segment placement) by clearing \c SHF_ALLOC on the section,
+  /// which is recommended whenever the section's addresses are bookkeeping
+  /// only and are not meant to be loaded at runtime.
```
parth-07: I do not think that users should trust non-allocatable section addresses even for bookkeeping. Non-alloc sections offsets can be used for bookkeeping.

lib/LinkerWrapper/LinkerWrapper.cpp
tcurtiss: Agree & done!

include/eld/PluginAPI/LinkerWrapper.h
tcurtiss: reworded to be a little more precise about the use case

include/eld/PluginAPI/LinkerWrapper.h
tcurtiss: clarified

parth-07 (approved):
Looks good to me!

PR #1752, opened 2026-08-22
fix: set correct state before writing output

### LINKED ISSUE #1728: eld incorrectly set the link state `ActBeforePerformingLayout` when it should instead set `ActBeforeWritingOutput`
We incorrectly set the link state `LinkState::ActBeforePerformingLayout` when we should set `LinkState::ActBeforeWritingOutput`.

In ELFObjectWriter.cpp, we have:

```bash
{
    PluginManager &PM = ThisModule.getPluginManager();
    ThisModule.setLinkState(LinkState::ActBeforePerformingLayout);
    if (!PM.callActBeforeWritingOutputHook()) {
      // Return generic error-code. Actual error is already reported!
      return make_error_code(std::errc::not_supported);
    }
  }
```

vansvan17:
## Summary

- set `LinkState::ActBeforeWritingOutput` before invoking the corresponding plugin hook
- strengthen the existing plugin test so it reports an error when the hook observes any other state

## Testing

- syntax-checked `ELFObjectWriter.cpp` with the generated external-LLVM compile command
- syntax-checked `ActBeforeWritingOutputPlugin.cpp` against LLVM 22.1.8
- external macOS build reaches C++ compilation but is blocked project-wide by the existing ELF `.comment` inline assembly being rejected by the Mach-O assembler

Fixes #1728

### DIFF
--- a/lib/Writers/ELFObjectWriter.cpp
+++ b/lib/Writers/ELFObjectWriter.cpp
@@ -203,7 +203,7 @@ ELFObjectWriter::writeObject(llvm::FileOutputBuffer &CurOutput) {
 
   {
     PluginManager &PM = ThisModule.getPluginManager();
-    ThisModule.setLinkState(LinkState::ActBeforePerformingLayout);
+    ThisModule.setLinkState(LinkState::ActBeforeWritingOutput);
     if (!PM.callActBeforeWritingOutputHook()) {
       // Return generic error-code. Actual error is already reported!
       return make_error_code(std::errc::not_supported);
--- a/lib/Core/Linker.cpp
+++ b/lib/Core/Linker.cpp
@@ -193,6 +193,7 @@ bool Linker::link() {
 
   {
     PluginManager &PM = ThisModule->getPluginManager();
+    ThisModule->setLinkState(LinkState::ActBeforePerformingLayout);
     if (!PM.callActBeforePerformingLayoutHook()) {
       // llvm::errs() << "callActBeforePerformingLayoutHook returning false!\n";
       return false;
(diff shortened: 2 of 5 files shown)


lib/Writers/ELFObjectWriter.cpp
```diff
@@ -203,7 +203,7 @@ ELFObjectWriter::writeObject(llvm::FileOutputBuffer &CurOutput) {
 
   {
     PluginManager &PM = ThisModule.getPluginManager();
-    ThisModule.setLinkState(LinkState::ActBeforePerformingLayout);
+    ThisModule.setLinkState(LinkState::ActBeforeWritingOutput);
```
parth-07, line 206: Can you also correctly call `setLinkState` before we call `PM.callActBeforePerformingLayout`?

vansvan17: Done in 61484ceb. I now set `ActBeforePerformingLayout` immediately before `callActBeforePerformingLayoutHook()`, and extended the existing hook test to assert the matching link state. Both modified files pass focused compiler checks.

parth-07 (approved):

lib/Writers/ELFObjectWriter.cpp
quic-seaswara, line 206: I would prefer this function to be called in Linker::emit()

quic-seaswara (changes requested):

lib/Writers/ELFObjectWriter.cpp
quic-seaswara: @parth-07  did you create an issue ?

parth-07: I have created an issue for this: https://github.com/qualcomm/eld/issues/1830

quic-seaswara (approved):

PR #1755, opened 2026-08-24
[GC] Hoist reached-list lookups

### LINKED ISSUE #1754: [GC] Avoid repeated reached-list map lookups per relocation
`GarbageCollection::setUpReachedSectionsAndSymbols()` performs `SectionReachedListMap` lookups inside the innermost relocation loop.

  For every relocation, it repeatedly calls:

  ```cpp
  MSectionReachedListMap.getReachedSymbolList(*ApplySect)
  MSectionReachedListMap.getReachedList(*ApplySect)
```

  ApplySect is unchanged while processing relocations for the same section, so the same DenseMap entries are looked up repeatedly.

rachitmeht:
Move reached symbol and section list lookups out of the per-relocation path.

On a Release ELD + Release Clang benchmark linking with
--gc-sections over 100 runs, Garbage Collection runtime improved from
0.4516s to 0.4365s.

Resolves #1754


parth-07 (commented):
Please add some performance improvement numbers in the commit message and the PR description.

quic-seaswara (approved):

quic-areg (approved):

PR #1759, opened 2026-08-24
Initialize sentinel rule hash in SectionMap constructor

### LINKED ISSUE #1758: SectionMap sentinel RuleContainer has no hash
The `SectionMap` constructor always creates a sentinel `OutputSectionEntry` with an empty name and pattern via `insert("","")`. The sentinel's `RuleHash` field stays as `std::nullopt`.  Any code that iterates all output sections and dereferences `getRuleHash()` will trigger an assertion failure.
```bash
optional<T>::operator* called on a disengaged value
```

BhoumikPatidar:
The `SectionMap` constructor creates an `OutputSectionEntry` with an empty name and pattern via `insert("","")`. This entry's `RuleContainer's` `RuleHash` field is never set and remains `std::nullopt`. Calling `RuleContainer::getRuleHash()` and dereferencing the result leads to an assertion failure.

Fix: After `insert("","")`, call `setRuleHash` on the returned `RuleContainer` with `llvm::hash_combine("","",0)` as this is an empty `RuleText`. This ensures every `RuleContainer` in the `SectionMap` has a valid hash.

Resolves #1758


parth-07 (changes requested):
Please add tests for this fix.

quic-seaswara (approved):

parth-07 (approved):
LGTM.

PR #1760, opened 2026-08-24
[ELD] Add initial OVERLAY VMA/LMA layout support

### LINKED ISSUE #1748: [linker-script] Failure in OVERLAY linker scripts with VMA/LMA
repro

```
#!/usr/bin/env bash
set -euo pipefail

cat > "inputs.s" <<'EOF'
.section .vectors, "ax", %progbits
  .space 0x20

.section .vectors.bhb.loop8, "ax", %progbits
  .space 0x20

.section .vectors.bhb.bpiall, "ax", %progbits
  .space 0x20

.section .stubs, "ax", %progbits
  .space 0x350

.section .init.text, "ax", %progbits
.global _start
_start:
  .space 0x1000
EOF

cat > "overlay.lds" <<'EOF'
ENTRY(_start)

SECTIONS
{
  . = 0xc1c00000;
  .text : { *(.text) }
  . = ALIGN(0x1000);

  __vectors_lma = .;
  OVERLAY 0xffff0000 : NOCROSSREFS AT(__vectors_lma)
  {
    .vectors { KEEP(*(.vectors)) }
    .vectors.bhb.loop8 { KEEP(*(.vectors.bhb.loop8)) }
    .vectors.bhb.bpiall { KEEP(*(.vectors.bhb.bpiall)) }
  }

  . = __vectors_lma + SIZEOF(.vectors) +
      SIZEOF(.vectors.bhb.loop8) + SIZEOF(.vectors.bhb.bpiall);

  __stubs_lma = .;
  .stubs ADDR(.vectors) + 0x1000 : AT(__stubs_lma)
  {
    *(.stubs)
  }

  . = __stubs_lma + SIZEOF(.stubs);
  . = ALIGN(8);
  .init.text : AT(ADDR(.init.text)) { *(.init.text) }
}
EOF

clang -c "inputs.s" -o "inputs.o"
ld.eld -T "overlay.lds" "inputs.o" -o "output.elf"
ld.lld -T "overlay.lds" "inputs.o" -o "output.elf"
```

ELD fails with the following error while LLD links successfully:
```
Error: section .init.text virtual address range overlaps with .stubs 
         >>> .init.text range is [0xC1C003B0, 0xC1C013AF] 
         >>> .stubs range is [0xC1C01000, 0xC1C0134F]
Fatal: Linking had errors (output.elf)
```

This pattern is used in the Linux ARM kernel

Steven6798:
Place OVERLAY members at a shared VMA and assign consecutive LMAs, including implicit overlay starts and overlays without AT(). Advance the location counter past the largest member and permit intentional VMA overlap between members of the same OVERLAY.

Fix: qualcomm#1748


Steven6798:
Note: This work will have follow ups as this does not complete the OVERLAY feature. It covers the basic scenarios seen in the new test added. This will serve as a good start and will unblock work like the arm Linux kernel.

test/Common/standalone/linkerscript/OverlaySharedVmaLma/OverlaySharedVmaLma.test
```diff
@@ -0,0 +1,70 @@
...
+#   9. Empty OVERLAY
+
+#END_COMMENT
+#START_TEST
+
+# OVERLAY with explicit start address
+RUN: %clang %clangopts -c %p/Inputs/input.s -o %t.o
+RUN: %link %linkopts %t.o -o %t.elf -T %p/Inputs/overlay.lds
+RUN: %readelf -S %t.elf | %filecheck %s -check-prefix=CHECK
+
```
quic-areg: can we check the `readelf -l` too for the `AT` cases?

Steven6798: I'll add this. Thanks!

Steven6798: This exposed a bug when dealing with multiple OVERLAY commands. The first member of the OVERLAY was not setting a new load-address range. That caused sections in different OVERLAY to be in the same program header, breaking the load and virtual addresses.

Thanks!!!

lib/Target/CreateProgramHeaders.hpp
```diff
@@ -278,6 +279,60 @@ bool GNULDBackend::createProgramHdrs() {
...
+
+    // Each OVERLAY block tracks its own largest member independently, so
+    // reset when starting a new overlay to support multiple OVERLAY blocks
+    // in the same linker script.
+    if (isFirstOverlayMember)
+      maxSize = 0;
+
+    // The first member of a new OVERLAY starts a new load-address range.
+    // Otherwise it could be merged with the preceding load segment.
+    if (isFirstOverlayMember && prevOut && prevOut->getOverlayDesc() != overlay)
```
quic-areg, line 300: What happens when the first overlay member is empty?

quic-areg (commented):
LGTM

lib/Target/CreateProgramHeaders.hpp
Steven6798: Nothing breaks because we still need that code. Most of the isFirstOverlayMember and isLastOverlayMember are used for things that should be done once to reduce unnecessary work. But since empty sections are being flagged as not wanted and they get dropped, we might be doing unnecessary work when processing these sections. This is a great follow-up to optimize the function when we know that a section will get dropped.

Thanks!!

quic-areg (approved):

lib/Script/ScriptFile.cpp
```diff
@@ -150,6 +150,12 @@ eld::Expected<void> ScriptFile::activate(Module &CurModule,
   // Resolve OVERLAY member names to OutputSectionEntry pointers so that layout
   // can discover overlay membership from the output-section entries.
   for (auto *O : OverlayDescs) {
+    // Overlay expressions are evaluated during layout; retain the script path
+    // for diagnostics produced at that point.
+    if (O->hasStart())
+      O->start()->setContext(getPath().str());
+    if (O->hasLMA())
+      O->lma()->setContext(getPath().str());
```
quic-seaswara, line 158: This has to be a seperate change.

lib/Target/CreateProgramHeaders.hpp
```diff
@@ -278,6 +279,60 @@ bool GNULDBackend::createProgramHdrs() {
...
+    bool isLastOverlayMember = isOverlayMember && !overlay->members().empty() &&
+                               overlay->members().back() == *out;
+    OutputSectionEntry *previousOverlayMember = nullptr;
+    if (isOverlayMember && !isFirstOverlayMember) {
+      for (auto *Member : overlay->members()) {
+        if (Member == *out)
+          break;
+        previousOverlayMember = Member;
+      }
+    }
```
quic-seaswara: This could be a function that OverlayDesc would contain.

lib/Target/CreateProgramHeaders.hpp
```diff
@@ -278,6 +279,60 @@ bool GNULDBackend::createProgramHdrs() {
...
+
+        scriptvma = overlay->start()->result();
+      } else if (isFirstOverlayMember) {
+        // With no explicit start, the overlay begins at the current dot.
+        scriptvma = dotSymbol->value();
+      } else {
+        // With no explicit start, all members use the first member's VMA,
+        // including when that member is empty and therefore does not become
+        // the global previous section.
+        scriptvma = overlay->members().front()->getSection()->addr();
```
quic-seaswara: overlay->getFirstSection() 

can we add this function ?

test/Common/standalone/linkerscript/OverlaySharedVmaLma/OverlaySharedVmaLma.test
```diff
@@ -0,0 +1,86 @@
...
+# Cases checked (a test can contain more than 1 test case):
+#   1. OVERLAY with explicit start address
+#   2. OVERLAY without explicit start address
+#   3. Next section after overlay land at `overlay_base + max(member_sizes)`
+#   4. Multiple OVERLAY
+#   5. OVERLAY with No explicit AT()
+#   6. Single member OVERLAY
+#   7. Multiple member OVERLAY
+#   8. OVERLAY with a zero-sized member
+#   9. Empty OVERLAY
```
quic-seaswara: Seperate tests, dont club all into one test.

test/Common/standalone/linkerscript/OverlaySharedVmaLma/OverlaySharedVmaLma.test
```diff
@@ -0,0 +1,86 @@
+#---OverlaySharedVmaLma.test------------------- Executable,LS------------------#
+#BEGIN_COMMENT
+# This test checks that overlay members share the same VMA.
+# All members should be placed at the overlay VMA, not sequentially.
+# Cases checked (a test can contain more than 1 test case):
+#   1. OVERLAY with explicit start address
+#   2. OVERLAY without explicit start address
+#   3. Next section after overlay land at `overlay_base + max(member_sizes)`
+#   4. Multiple OVERLAY
+#   5. OVERLAY with No explicit AT()
```
quic-seaswara: Need tests with linker map files. 

How does TLS work with overlays ?

quic-seaswara (changes requested):

lib/Target/CreateProgramHeaders.hpp
Steven6798: Will do.

lib/Target/CreateProgramHeaders.hpp
Steven6798: Will do.

test/Common/standalone/linkerscript/OverlaySharedVmaLma/OverlaySharedVmaLma.test
Steven6798: Sounds good.

lib/Script/ScriptFile.cpp
Steven6798: This is needed, otherwise "evaluateAndRaiseError" errors out.

test/Common/standalone/linkerscript/OverlaySharedVmaLma/OverlaySharedVmaLma.test
Steven6798: Will add linker map file test.

TLS outside OVERLAY is working just fine. Now inside OVERLAY is something that I have not look into. I can address this as a follow-up.

lib/Target/CreateProgramHeaders.hpp
```diff
@@ -278,6 +279,55 @@ bool GNULDBackend::createProgramHdrs() {
     // Linker script overriding below.
     std::optional<uint64_t> scriptvma;
     bool doAlign = true;
+
+    OverlayDesc *overlay = (*out)->getOverlayDesc();
+    bool isOverlayMember = overlay != nullptr;
+    bool isFirstOverlayMember = isOverlayMember &&
+                                !overlay->members().empty() &&
+                                overlay->members().front() == *out;
```
quic-seaswara: Let us add getFirstOverLayMember() and getLastOverLayMember()

lib/Target/CreateProgramHeaders.hpp
```diff
@@ -278,6 +279,55 @@ bool GNULDBackend::createProgramHdrs() {
...
+        // With no explicit start, all members use the first member's VMA,
+        // including when that member is empty and therefore does not become
+        // the global previous section.
+        scriptvma = overlay->getFirstSection()->addr();
+      }
+      if (isCurAlloc) {
+        dotSymbol->setValue(*scriptvma);
+        maxSize = std::max(maxSize, cur->size());
+      }
+      doAlign = false;
```
quic-seaswara, line 326: Why is doAlign set to false ? 

We should error out with PHDRS (createScriptProgramHdrs when doing OVERLAY's since it is not implemented). 

Please add a test with PHDRS, with the error reported by the linker.

quic-seaswara (changes requested):
Some nit picks, but overall good for the first support to get the kernel green.

lib/Target/CreateProgramHeaders.hpp
Steven6798: Sounds good!

lib/Target/CreateProgramHeaders.hpp
Steven6798: > Why is doAlign set to false?

Should we align if the VMA is set here and the next address outside the OVERLAY will be either set by the linker script (. =0x....) or set to *scriptvma + maxSize? Wouldn't alignment mess that up?

> Please add a test with PHDRS, with the error reported by the linker.

I'll add that test

quic-seaswara: Does the first overlay section get aligned to the maximum alignment of all the overlay members. Can you please check ?

Steven6798: I'll check

Steven6798: This is the current behavior:
- With an explicit OVERLAY start, the first member uses start exactly.
- Without an explicit start, the overlay uses the current dot exactly.

This matches what GNU ld does.

Steven6798: So far alignment must not be used. This is something that can be look into as the features keeps expanding. So far, I have had no use case where it requires alignment, but maybe as we combine OVERLAY with other stuff alignment might be needed.

quic-seaswara (approved):

PR #1763, opened 2026-08-25
[Reproduce] Support linker scripts resolved through -L

### LINKED ISSUE #1454: Reproduce does not work when linker script is found within a `-L` directory
Reproduce does not work when a linker script is found within a `-L` directory.

Reproducer:

```bash
#!/usr/bin/env bash

set -e

mkdir -p A/lib

cat >main.c <<\EOF
int foo() { return 1; }
int main() { return foo(); }
EOF

cat >script.t <<\EOF
EOF

clang -target x86_64-linux-gnu -o main.o main.c -c -ffunction-sections 
mv script.t A/lib

ld.eld -o main.eld.out main.o -L A/lib -T script.t --reproduce a.tar 
rm -rf reproduce && mkdir reproduce -p
tar -xvf a.tar -C reproduce --strip-components=1
cd reproduce
bash -x response.txt 
```

rachitmeht:
Capture linker scripts during early activation and rewrite -T paths to their mapped paths in reproduce response files. Add a regression test covering linker scripts found through a -L search directory.

Resolves #1454

### DIFF
--- a/lib/LinkerWrapper/GnuLdDriver.cpp
+++ b/lib/LinkerWrapper/GnuLdDriver.cpp
@@ -1680,6 +1680,7 @@ bool GnuLdDriver::processReproduceOption(
   if (!Config.options().getDumpResponse())
     os << getProgramName() << " ";
   size_t lastNamespecId = -1;
+  size_t lastScriptId = -1;
 
   auto zArgsRange = Args.filtered(T::dash_z);
   auto zArgIt = zArgsRange.begin();
@@ -1727,7 +1728,6 @@ bool GnuLdDriver::processReproduceOption(
     }
     case T::output_file:
     case T::Map:
-    case T::T:
     case T::R:
     case T::dynamic_list:
     case T::extern_list:
@@ -1735,6 +1735,28 @@ bool GnuLdDriver::processReproduceOption(
       os << arg->getSpelling() << ' ' << outputTar->rewritePath(arg->getValue())
          << ' ';
       break;
+    case T::T: {
+      bool rewritten = false;
+      for (size_t i = lastScriptId + 1; i < actions.size(); ++i) {
+        auto action = actions[i];
+        if (action->getInputActionKind() == eld::InputAction::Script) {
+          lastScriptId = i;
+          auto ipt = action->getInput();
+          if (ipt) {
+            std::string key = ipt->getInputFile()
+                                  ? ipt->getInputFile()->getMappedPath()
+                                  : ipt->getName();
+            os << arg->getSpelling() << ' ' << outputTar->rewritePath(key)
+               << ' ';
+            rewritten = true;
+          }
+          break;
+        }
+      }
+      if (!rewritten)
+        os << arg->getSpelling() << ' ' << arg->getValue() << ' ';
+      break;
+    }
     case T::remap_inputs:
       os << arg->getSpelling() << ' ' << arg->getValue() << ' ';
       break;
--- a/lib/Object/ObjectLinker.cpp
+++ b/lib/Object/ObjectLinker.cpp
@@ -248,6 +248,10 @@ bool ObjectLinker::readAndActivateLinkerScript(
       return true;
   }
 
+  // Capture scripts during early activation so the response file can rewrite
+  // -T paths that were resolved through a -L search directory.
+  addInputFileToTar(input, eld::MappingFile::LinkerScript);
+
   if (!parseLinkerScript(input))
     return false;
 
--- a/test/Common/standalone/CommandLine/Reproduce/ReproduceLinkerScriptSearchPath.test
+++ b/test/Common/standalone/CommandLine/Reproduce/ReproduceLinkerScriptSearchPath.test
@@ -1,0 +1,22 @@
+#---ReproduceLinkerScriptSearchPath.test----------------- Executable -----------------#
+
+#BEGIN_COMMENT
+# Checks that --reproduce rewrites a linker script found through a -L directory.
+#END_COMMENT
+#START_TEST
+RUN: %rm -rf %t.lib %t.reproduce %t.tar %t.out %t.replay.out
+RUN: %mkdir %t.lib
+RUN: %cp %p/Inputs/script1.t %t.lib/script.t
+RUN: %clang %clangopts -c %p/Inputs/1.c -o %t.main.o
+RUN: cd %t.lib
+RUN: %link --no-threads -L %t.lib -T script.t %t.main.o -o %t.out --reproduce %t.tar --dump-response-file %t.response
+RUN: %filecheck %s < %t.response
+RUN: %mkdir %t.reproduce
+RUN: %tar %gnutaropts -xf %t.tar -C %t.reproduce --strip-components=1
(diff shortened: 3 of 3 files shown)


lib/Object/ObjectLinker.cpp
```diff
@@ -248,6 +248,10 @@ bool ObjectLinker::readAndActivateLinkerScript(
       return true;
   }
 
+  // Capture scripts during early activation so the response file can rewrite
+  // -T paths that were resolved through a -L search directory.
+  addInputFileToTar(input, eld::MappingFile::LinkerScript);
```
parth-07, line 253: It is being called for both early activation and late activation, right?

Why can't the response file rewrite -T paths that were resolved through -L subdirectory if we capture the information later in full activation?

lib/LinkerWrapper/GnuLdDriver.cpp
```diff
@@ -1762,45 +1728,42 @@ bool GnuLdDriver::processReproduceOption(
...
-      break;
-    }
-    case T::R: {
-      auto path = getRewrittenActionInputPath(
-          eld::InputAction::JustSymbols, lastJustSymbolsId, arg->getValue());
-      if (!path)
-        return false;
-      os << arg->getSpelling() << ' ' << *path << ' ';
+      bool rewritten = false;
+      for (size_t i = lastScriptId + 1; i < actions.size(); ++i) {
```
parth-07, line 1740: Why do we need to determine the script's `InputFile` positionally? Why does `outputTar->rewritePath(...)` does not work here?

parth-07 (changes requested):

lib/LinkerWrapper/GnuLdDriver.cpp
rachitmeht: outputTar->rewritePath("script.t") does not find the mapping created for
"/work/A/lib/script.t" and falls back to "script.t"

 So, The ScriptAction lookup is needed to obtain the mapped resolved path.

lib/Object/ObjectLinker.cpp
rachitmeht: Yes, addInputFileToTar() may be called during both early and full activation. For a normal successful link, full activation usually happens and may capture the script later. The reason for keeping the early capture is to cover the broader --reproduce-on-fail case: any earlier input error, script error or interruption can prevent full activation from being reached.

lib/LinkerWrapper/GnuLdDriver.cpp
rachitmeht: Positionally determining InputFile will allow us to have args with same name but may resolve to different files. like 
 -L dir1 -T script.t -L dir2 -T script.t

lib/Object/ObjectLinker.cpp
parth-07: > The reason for keeping the early capture is to cover the broader --reproduce-on-fail case: any earlier input error, script error or interruption can prevent full activation from being reached.

Then can you fix the comments above `addInputFileToTar` here? The comments suggests that early activation is somehow required when script is found in `-L` subdirectory.

lib/LinkerWrapper/GnuLdDriver.cpp
parth-07: > outputTar->rewritePath("script.t") does not find the mapping created for
"/work/A/lib/script.t" and falls back to "script.t"

Yes, that is the core issue that needs to be fixed.

What is the benefit of determining the InputFile positionally and making the code and the flow more complex as opposed to fixing `outputTar->rewritePath(...)` to function correctly here?

> Positionally determining InputFile will allow us to have args with same name but may resolve to different files. like -L dir1 -T script.t -L dir2 -T script.t

Can you verify whether the 2 script.t here would resolve to different scripts if dir1 and dir2 both contains script.t?

rachitmeht: I verified, the 2 script.t will resolve to the first -L dir only. I was influenced by how we handle namespecs here.
I agree this approach doesn't fix `outputTar->rewritePath(...)`.
But if we fix that we must handle LinkerScript seperately in that function to give correct path. The current fix effectively does that. 
Will that be better design?

parth-07: >But if we fix that we must handle LinkerScript seperately in that function to give correct path. The current fix effectively does that.

Why would we need to handle LinkerScript separately in that function? Can you think of a solution that fixes this problem in a more general way?

rachitmeht: So, The issue will be for every script ( version script & LinkerScript ) which excecutes `ScriptAction::Activate() `
Because it changes the file name by 
```
    Path = Res->native();
  }
  setFileName(Path);
```
So, the MappedPath contains the wrong Name for scripts came through -L.

  But `processReproduceOption()` calls `rewritePath()` with the original argument spelling from the command line. Since `rewritePath()` only receives a string and does an exact lookup in
  `InputMap`, it cannot know that `script.t` should correspond to `/path/from/-L/script.t`.

  So we can either:
  have conditional handling of these scripts before calling `rewritePath()`, or 
  we can have a different Reproduce Specific map which will have the original file name. (we only need to store original map in `ScriptAction:: Activate()`)

test/Common/standalone/CommandLine/Reproduce/ReproduceLinkerScriptSearchPath.test
```diff
@@ -0,0 +1,22 @@
...
+#START_TEST
+RUN: %rm -rf %t.lib %t.reproduce %t.tar %t.out %t.replay.out
+RUN: %mkdir %t.lib
+RUN: %cp %p/Inputs/script1.t %t.lib/script.t
+RUN: %clang %clangopts -c %p/Inputs/1.c -o %t.main.o
+RUN: cd %t.lib
+RUN: %link --no-threads -L %t.lib -T script.t %t.main.o -o %t.out --reproduce %t.tar --dump-response-file %t.response
+RUN: %filecheck %s < %t.response
+RUN: %mkdir %t.reproduce
+RUN: %tar %gnutaropts -xf %t.tar -C %t.reproduce --strip-components=1
```
Steven6798, line 15: tar .... -C is not compatible with Windows. you can cd into %t.reproduce first that way you don't need -C

Steven6798 (changes requested):

test/Common/standalone/CommandLine/Reproduce/ReproduceLinkerScriptSearchPath.test
rachitmeht: Sure, I will change it.

lib/LinkerWrapper/GnuLdDriver.cpp
parth-07: @rachitmeht Based on the offline discussion, let's standardized the meaning of _name_ members in Input (`Name`, `OriginalFileName`, `ResolvedPath`) such that one of them can be effectively used for reproduce functionality.

PR #1765, opened 2026-08-25
Allow local versioned symbols

### LINKED ISSUE #1597: Local versioned symbol are not supported in regular object files
eld reports errors when it encounters local versioned symbols in regular object files, lld and bfd do not report any error here. lld and bfd sees local versioned symbols as literal symbols with '@' in them. We should do the same for compatibility with GNU.

parth-07:
This commit accepts local versioned symbols as plain local symbols. They are no longer rejected. There is no need to canonicalize, export, or assigned GNU symbol-version entries for local versioned symbols.

Resolves #1597


quic-seaswara (approved):

BhoumikPatidar (approved):

quic-areg (approved):
LGTM

PR #1777, opened 2026-08-27
[AdvancedLTO] Stop exporting plugin implementation class

svs-quic:
Remove DLL_A_EXPORT from the AdvancedLTO class and keep the implementation internal to the translation unit. The plugin loader only needs the exported C entry points provided by ELD_REGISTER_PLUGIN.

This fixes Windows builds where dllexport on the anonymous-namespace class was rejected, and also avoids MSVC exporting implicit copy/assignment members for the unique_ptr-owned module state.

### DIFF
--- a/Plugins/AdvancedLTO/AdvancedLTO.cpp
+++ b/Plugins/AdvancedLTO/AdvancedLTO.cpp
@@ -22,9 +22,9 @@
 #include <string>
 #include <unordered_map>
 
-namespace {
-
 using namespace eld::plugin;
+
+namespace {
 
 struct ModuleData {
   explicit ModuleData(BitcodeFile BCF) : BCFile(BCF) {}
@@ -33,7 +33,7 @@ struct ModuleData {
   std::unordered_map<std::string, Section> SectionsByName;
 };
 
-class DLL_A_EXPORT AdvancedLTO : public LinkerPlugin {
+class AdvancedLTO : public LinkerPlugin {
 public:
   AdvancedLTO() : LinkerPlugin("AdvancedLTO") {}
 


svs-quic:
This patch fixes a windows failure we were seeing on the CPULLVM builders. Verified that the failure is fixed using https://github.com/qualcomm/cpullvm-toolchain/pull/631. Given that it is a fix to the AdvancedLTO plugin, we only need it on the 23.x_lto_linker_scripts branch.

apazos (approved):

Plugins/AdvancedLTO/AdvancedLTO.cpp
```diff
@@ -238,6 +237,21 @@ class DLL_A_EXPORT AdvancedLTO : public LinkerPlugin {
   std::optional<BitcodeFile> LastBitcodeFile;
 };
 
+PluginBase *ThisPlugin = nullptr;
+
 } // namespace
 
-ELD_REGISTER_PLUGIN(AdvancedLTO)
+extern "C" {
```
parth-07: Why did we replace ELD_REGISTER_PLUGIN with this expansion here? ELD_REGISTER_PLUGIN should not cause any issues here I believe.

svs-quic: Thanks Parth. I was looking at what we were doing in our downstream implementation and tried to mimic that. I see that ELD_REGISTER_PLUGIN basically does what I've done and I've put it back and only retained the bare minimum changes needed. I've verified that the windows builder on CPULLVM passes using the same PR https://github.com/qualcomm/cpullvm-toolchain/pull/631

parth-07 (approved):
LGTM.

svs-quic:
I am going to go ahead and merge this since this branch is used only by CPULLVM and I have verified that things work there.

PR #1781, opened 2026-08-28
[CI] Check out LLVM release/23.x

svs-quic:
The CI tries to fetch the the llvm-project branch with the same name as the one in eld but in this case there is no release/23.x_lto_linkerscripts branch in the LLVM repo. This was causing the CI to fail early.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,6 +57,15 @@ jobs:
           ref: release/23.x
           path: pr-${{ env.PR_NUMBER }}/llvm-project
 
+      - name: Apply CPULLVM LTO linker script patch
+        working-directory: pr-${{ env.PR_NUMBER }}/llvm-project
+        run: |
+          curl -fsSL \
+            https://raw.githubusercontent.com/qualcomm/cpullvm-toolchain/release/qualcomm-software/23.x/qualcomm-software/patches/llvm-project/0004-LTO-with-linker-scripts.patch \
+            -o /tmp/lto-with-linker-scripts.patch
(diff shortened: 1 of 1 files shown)


svs-quic:
okay this wont work since the release/23.x_lto_linker_scripts has the ELD changes for lto w/linkerscripts that depends on the compiler changes that are currently only there on the CPULLVM 23.x branch. So even if we were to fetch the release/23.x branch on llvm-project the builder will fail compilation.

parth-07 (approved):

parth-07:
> okay this wont work since the release/23.x_lto_linker_scripts has the ELD changes for lto w/linkerscripts that depends on the compiler changes that are currently only there on the CPULLVM 23.x branch. So even if we were to fetch the release/23.x branch on llvm-project the builder will fail compilation.

To resolve this, should we use apply patch mechanism that the CPULLVM uses?

svs-quic:
> > okay this wont work since the release/23.x_lto_linker_scripts has the ELD changes for lto w/linkerscripts that depends on the compiler changes that are currently only there on the CPULLVM 23.x branch. So even if we were to fetch the release/23.x branch on llvm-project the builder will fail compilation.
> 
> To resolve this, should we use apply patch mechanism that the CPULLVM uses?

Thanks for the suggestion, Parth. I've tried to do that in the latest patch and the build seems to be passing.

quic-seaswara (approved):

efriedma-quic (approved):

quic-seaswara (approved):

PR #1782, opened 2026-08-28
Reject local-exec TLS relocations in shared libraries

### LINKED ISSUE #1780: ELD doesn't give proper error on encountering TPOFF relocation in shared library.
Eld only errors out on encountering TPOFF relocation in shared library because there is no `PT_TLS segment`. This is misleading and not a good error message. LLD properly reports that the relocation is unsupported for shared library. 

Reproducer:
```
# Assembly is needed to force exact local-exec TPOFF relocations in data.
cat > /tmp/tpoff32-shared.s <<\EOF
.weak weak_tls
.section .data
.align 4
result:
  .long 0xdeadbeef
  .reloc result, R_X86_64_TPOFF32, weak_tls
EOF

cat > /tmp/tpoff64-shared.s <<\EOF
.weak weak_tls
.section .data
.align 8
result:
  .quad 0xdeadbeefdeadbeef
  .reloc result, R_X86_64_TPOFF64, weak_tls
EOF
clang -target x86_64-linux-gnu -c /tmp/tpoff32-shared.s -o /tmp/tpoff32-shared.o
clang -target x86_64-linux-gnu -c /tmp/tpoff64-shared.s -o /tmp/tpoff64-shared.o

LDs=(ld.eld ld.lld ld.bfd)
SFs=(eld lld bfd)

for obj in tpoff64-shared; do
  for i in "${!SFs[@]}"; do
    ${LDs[$i]} -m elf_x86_64 -shared /tmp/$obj.o -o /tmp/$obj.${SFs[$i]}.so
  done
done
```

Result:
```
Error: PT_TLS segment not found, required by the image
Error: Relocation error when applying relocation `R_X86_64_TPOFF64' for symbol `weak_tls' referred from /tmp/tpoff64-shared.o[.data] symbol defined in /tmp/tpoff64-shared.o
Fatal: Linking had errors (tpoff64-shared.eld.so)
ld.lld: error: relocation R_X86_64_TPOFF64 against weak_tls cannot be used with -shared
>>> defined in /tmp/tpoff64-shared.o
>>> referenced by /tmp/tpoff64-shared.o:(.data+0x0)
ld.bfd: BFD (GNU Binutils for Ubuntu) 2.38 assertion fail ../../bfd/elf64-x86-64.c:4074

### PR DESCRIPTION
**Pull request by rachitmeht**
Classify target-specific local-exec TLS relocations and report an error when they are used with -shared. Add regression tests for all supported targets.

Resolves #1780

### DIFF
--- a/include/eld/Diagnostics/DiagRelocations.inc
+++ b/include/eld/Diagnostics/DiagRelocations.inc
@@ -82,3 +82,5 @@ DIAG(warn_unsupported_reloc_for_ifunc, DiagnosticEngine::Warning,
      "Unsupported relocation %0 used for STT_GNU_IFUNC symbol '%1'")
 DIAG(warn_invalid_reloc_for_ifunc, DiagnosticEngine::Warning,
      "Invalid relocation %0 used for STT_GNU_IFUNC symbol '%1'")
+DIAG(tls_le_relocation_in_shared, DiagnosticEngine::Fatal,
+     "relocation %0 against %1 referred from %2 cannot be used with -shared")
--- a/lib/Target/Relocator.cpp
+++ b/lib/Target/Relocator.cpp
@@ -292,6 +292,8 @@ bool Relocator::reportNonPICRelocation(const Relocation &reloc) const {
 }
 
 bool Relocator::checkPICRelocSupported(const Relocation &reloc) const {
+  if (!isTLSRelocTypeSupported(reloc))
+    return false;
   if (!config().isCodeIndep())
     return true;
   if (isPICRelocTypeSupported(reloc))
@@ -320,6 +322,26 @@ bool Relocator::checkDynamicRelocAllowed(const Relocation &reloc,
   return true;
 }
 
+bool Relocator::isTLSRelocTypeSupported(const Relocation &reloc) const {
+  if (!config().options().hasShared())
+    return true;
+
+  if (!isTLSLocalExecReloc(reloc))
+    return true;
+
+  FragmentRef *Ref = reloc.targetRef();
+  std::string Location = reloc.getSourcePath(config().options());
+  if (Ref && Ref->frag() && Ref->frag()->getOwningSection())
+    Location = Ref->frag()->getOwningSection()->getLocation(Ref->offset(),
+                                                            config().options());
+  ResolveInfo *sym = reloc.symInfo();
+  config().raise(Diag::tls_le_relocation_in_shared)
+      << getName(reloc.type()) << (sym ? getSymbolName(sym) : "<unknown>")
+      << Location;
+
+  m_Module.setFailure(true);
+  return false;
+}
 bool Relocator::doDeMangle() const {
   return m_Config.options().shouldDemangle();
 }
--- a/include/eld/Target/Relocator.h
+++ b/include/eld/Target/Relocator.h
@@ -138,6 +138,12 @@ public:
   bool checkDynamicRelocAllowed(const Relocation &reloc,
                                 const ELFSection &section, bool isAbs) const;
 
+  bool isTLSRelocTypeSupported(const Relocation &reloc) const;
+
+  virtual bool isTLSLocalExecReloc(const Relocation &reloc) const {
+    return false;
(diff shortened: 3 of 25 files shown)

### CODE REVIEW

File: include/eld/Diagnostics/DiagRelocations.inc [hunk 1]
```diff
@@ -81,4 +83,4 @@ DIAG(verbose_remapped_internal_reloc, DiagnosticEngine::Verbose,
 DIAG(warn_unsupported_reloc_for_ifunc, DiagnosticEngine::Warning,
      "Unsupported relocation %0 used for STT_GNU_IFUNC symbol '%1'")
 DIAG(warn_invalid_reloc_for_ifunc, DiagnosticEngine::Warning,
-     "Invalid relocation %0 used for STT_GNU_IFUNC symbol '%1'")
\ No newline at end of file
+     "Invalid relocation %0 used for STT_GNU_IFUNC symbol '%1'")
```
quic-seaswara, line 84: need to show input object file in the error.

quic-seaswara (changes requested):

quic-areg (approved):

include/eld/Diagnostics/DiagRelocations.inc
parth-07: @quic-seaswara Do you mean in the newly added diagnostic `tls_le_relocation_in_shared`?. This diagnostic `warn_invalid_reloc_for_ifunc` is not getting modified in this patch.

parth-07: @rachitmeht We need both the input object file and the section name (if possible) in the `tls_le_relocation_in_shared` diagnostic. Also, please prefix the diagnostic name with the diagnostic severity.

parth-07 (changes requested):
Can we add a test in `Common/standalone` instead of adding the test in each target subdirectory?

parth-07:
@rachitmeht Can you please reply to the comments or mark them as resolved after resolving them?

lib/Target/Relocator.cpp
```diff
@@ -320,6 +322,22 @@ bool Relocator::checkDynamicRelocAllowed(const Relocation &reloc,
   return true;
 }
 
+bool Relocator::checkTLSLocalExecReloc(const Relocation &reloc) const {
```
parth-07: The function name does not make it clear what it is checking. Can you please improve the function name here?

lib/Target/Relocator.cpp
```diff
@@ -320,6 +322,22 @@ bool Relocator::checkDynamicRelocAllowed(const Relocation &reloc,
...
+  if (!config().options().hasShared())
+    return true;
+
+  if (!isTLSLocalExecReloc(reloc))
+    return true;
+
+  ResolveInfo *sym = reloc.symInfo();
+  config().raise(Diag::tls_le_relocation_in_shared)
```
parth-07, line 338: We are only printing the section name (`.text`) instead of section + offset (`.text + 0x3`). Can you please update here to print section with the offset information?

parth-07 (changes requested):

lib/Target/Relocator.cpp
rachitmeht: Sure

lib/Target/Relocator.cpp
rachitmeht: Yes

rachitmeht:
@parth-07 


> Can we add a test in `Common/standalone` instead of adding the test in each target subdirectory?

Should I add tests for all the Relocation Types in a single test in Common/standalone

parth-07:
> Can we add a test in Common/standalone instead of adding the test in each target subdirectory?

> > Should I add tests for all the Relocation Types in a single test in Common/standalone

As much as possible, yes. For the cases where you need to use assembly test inputs, those can be in target-specific test directoreis.

parth-07 (approved):
LGTM.

parth-07:
@rachitmeht Can you please rebase the PR and make clang-format builder happy?

PR #1785, opened 2026-08-28
[workflow] Implement thread sanitizer nightly workflow

### LINKED ISSUE #1784: [workflow] Create thread sanitizer nightly workflow
Create a nightly workflow similar to sanitize.yml that uses the threads sanitizer option.

Steven6798:
Add a nightly workflow similar to sanitize.yml that use the threads sanitizer option.

Fix: qualcomm#1784


quic-seaswara:
we should probably have check-llvm-unit to make sure that LLVM is green before checking eld.

Steven6798:
> we should probably have check-llvm-unit to make sure that LLVM is green before checking eld.

Both llvm and eld have failures and the workflows fails only when eld fails. I think we can still enable the workflow even if it is currently red. This will help us test all the fixes. I can disable the nightly schedule and only let the manual trigger for now. What do you think?

quic-seaswara (approved):

quic-shaksuma (approved):

PR #1788, opened 2026-08-28
Add qemu_x86_64 to Zephyr CI

BhoumikPatidar:
Adds `x86_64` as a sixth board. It builds with GCC from the Zephyr SDK.

Adds toolchain input to `zephyr-run.yml`, defaulting to `cpullvm` for existing backends and a `qemu_x86_64` backend using GCC from the Zephyr SDK. Adds a step applying KEEP rules to Zephyr's linker scripts for the C++ exception tables, to be dropped once they land upstream

Also adds `--retry-failed` for flaky tests and `--personal-access-token` to avoid GitHub API rate limits.

### DIFF
--- a/.github/workflows/zephyr-run.yml
+++ b/.github/workflows/zephyr-run.yml
@@ -15,6 +15,11 @@ on:
         description: "Arch label, used for ccache keys and build-status records"
         required: true
         type: string
+      toolchain:
+        description: cpullvm or zephyr-sdk
+        required: false
+        type: string
+        default: "cpullvm"
       extra_args:
         description: "Extra Kconfig passed to west build (-D) and twister (-x); may be empty"
         required: false
@@ -69,9 +74,9 @@ jobs:
         uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 #v6.1.0
         with:
           path: ${{ env.CCACHE_DIR }}
-          key: ccache-${{ inputs.arch }}-cpullvm${{ env.CPULLVM_VERSION }}-${{ github.run_id }}
+          key: ccache-${{ inputs.arch }}-${{ inputs.toolchain }}-${{ github.run_id }}
           restore-keys: |
-            ccache-${{ inputs.arch }}-cpullvm${{ env.CPULLVM_VERSION }}-
+            ccache-${{ inputs.arch }}-${{ inputs.toolchain }}-
 
       - name: ccache stats (before build)
         run: ccache -s -v || ccache -s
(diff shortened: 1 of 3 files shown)


quic-seaswara (approved):

.github/workflows/zephyr-run.yml
```diff
@@ -225,9 +276,10 @@ jobs:
             --timestamps \
             -j 3 \
             -v \
-            -x=ZEPHYR_TOOLCHAIN_VARIANT=host/llvm \
-            -x=CONFIG_LLVM_USE_ELD=y \
-            -x=CONFIG_COMPILER_RT_RTLIB=y \
+            --retry-failed 2 \
+            --retry-interval 10 \
```
quic-areg, line 280: Should we quarantine these tests instead? My concern is that an intermittent linker failure could still produce a green result and we know about it. Runtime might also be a concern currently since the builds are close to the 6h gh limit.

quic-areg (commented):
How does this work without upstream eld build support with GCC?

quic-seaswara:
> How does this work without upstream eld build support with GCC?

are you referring to using clang instead of gcc ?

.github/workflows/zephyr-run.yml
quic-seaswara: How long does this take ? 

I think the retry is needed even with bfd.

.github/workflows/zephyr-run.yml
```diff
@@ -129,6 +152,27 @@ jobs:
       - name: Set safe.directory
         run: git config --global --add safe.directory "$GITHUB_WORKSPACE/zephyr"
 
+      # When a linker script is present, retention is the script's responsibility. 
+      # KEEP is the mechanism for sections that must survive garbage collection.
+      # Will be dropped once the change is merged in Zephyr upstream.
+      - name: Add linker-script KEEP rules for C++ exception tables
```
parth-07, line 158: Can you please add an FAQ entry for this as a follow-up patch?

BhoumikPatidar:
> How does this work without upstream eld build support with GCC?

This is via a workaround currently.
I am passing `-B<dir>` and setting `GNULD_LINKER` to `ld path` and `GNULD_LINKER_IS_BFD=OFF`

Zephyr has `FindGnuLd.cmake` which forces `-fuse-ld=bfd`. But by pre-setting `GNULD_LINKER` we can avoid this probe.

Consider this run on a small subset of twister tests on my fork to confirm eld by explicitly printing the `.comment` section - 
https://github.com/BhoumikPatidar/eld/actions/runs/33375683913/job/99436559130

.github/workflows/zephyr-run.yml
BhoumikPatidar: `--retry-failed` retries tests failures, not build errors. So the case is limited to intermittent runtime failures.
On why retry is needed: Across multiple runs locally as well as on github runners some tests fail randomly at runtime. This behavior was observed for both `bfd` and `eld` as linkers.

BhoumikPatidar: for x86_64 it takes around 2.5 hours. But riscv does take over 5 hours.

quic-areg (approved):
lgtm

PR #1791, opened 2026-08-31
[Driver] Group never-read options as ignored; alias --library-path to -L

### LINKED ISSUE #1186: Dead driver options
Several options are declared in the linker driver TableGen files but never referenced anywhere in the linker source. Passing them on the command line is silently accepted but does nothing. If any of these are intentionally ignored for compatibility reasons, they should go in `grp_compatorignoredopts`. 

This is misleading because the flags appear in `--help` output, giving the appearance that they are functional. 

Affected options:

This was checked by grep-ing each option's td identifier against eld source. 

Options supported by GNU ld:

```
--allow-shlib-undefined
--no-allow-shlib-undefined
--add-needed
--no-add-needed
--copy-dt-needed-entries
--no-copy-dt-needed-entries
--fix-cortex-a8
--no-fix-cortex-a8
--library-path[=<path>]
--Ttext-segment=<address>
--rpath-link[=<path>]
--hash-size=<size>
-g                            # gold 
```

We might be making misleading GNU compatibility claims with these options. 

ELD-specific:

```
--map-section=<section>
--align-segments            
--use-shlib-undefines
```

saket0187:
This PR tries to fix #1186 

This PR moves linker driver options that are declared in TableGen but never read by the linker into `grp_compatorignoredopts` so `--help` stops presenting them as functional, and aliases `--library-path` to -L since silently discarding a library search path is a bug rather than an intentional no-op.

### DIFF
--- a/include/eld/Driver/ARMLinkerOptions.td
+++ b/include/eld/Driver/ARMLinkerOptions.td
@@ -26,12 +26,6 @@ def use_mov_veneer
       Group<grp_arm_linker>;
 def grp_arm_linker_only : OptionGroup<"opts">,
                           HelpText<"ARM Linker Option ONLY">;
-def fix_cortex_a8 : Flag<["--"], "fix-cortex-a8">,
-                    HelpText<"Fix the Cortex A8 bug">,
-                    Group<grp_arm_linker_only>;
-def no_fix_cortex_a8 : Flag<["--"], "no-fix-cortex-a8">,
-                       HelpText<"Dont Fix the Cortex A8 bug">,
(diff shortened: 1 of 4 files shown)


quic-seaswara (approved):

quic-seaswara (approved):

parth-07 (approved):

quic-seaswara (changes requested):
Can you please expand the commit message ?

saket0187:
> Can you please expand the commit message ?

Expanded the commit message. Please let me know if this works.

quic-seaswara (approved):

PR #1794, opened 2026-08-31
Remove duplicate LLVM globals and run llvm_shutdown from LW

quic-areg:
LLVMSupport was linked into both the ld.eld executable and the LinkerWrapper shared library. This created duplicate LLVM globals that are supposed to be cleaned up by llvm_shutdown, but the shutdown call only came from the executable, not the library.  On windows, this could lead to a race condition between thread creation and process exit.

This patch moves the linker entry point to LinkerWrapper and ensures that we only have one copy of LLVM globals, all in the library rather than the executable.

### DIFF
--- a/lib/LinkerWrapper/Driver.cpp
+++ b/lib/LinkerWrapper/Driver.cpp
@@ -18,11 +18,56 @@
 #include "eld/Support/TargetRegistry.h"
 #include "eld/Support/TargetSelect.h"
 #include "eld/Target/TargetMachine.h"
+#include "llvm/ADT/STLExtras.h"
 #include "llvm/ADT/StringSwitch.h"
 #include "llvm/MC/TargetRegistry.h"
+#include "llvm/Support/ManagedStatic.h"
 #include "llvm/Support/Path.h"
+#include "llvm/Support/PrettyStackTrace.h"
 #include "llvm/Support/Process.h"
+#include "llvm/Support/Signals.h"
+#include "llvm/Support/StringSaver.h"
 #include <optional>
+
+// If a command line option starts with "@", the driver reads its suffix as a
+// file, parse its contents as a list of command line options, and insert them
+// at the original @file position. If file cannot be read, @file is not expanded
+// and left unmodified. @file can appear in a response file, so it's a recursive
+// process.
+static llvm::ArrayRef<const char *>
+maybeExpandResponseFiles(llvm::ArrayRef<const char *> Args,
+                         llvm::BumpPtrAllocator &Alloc) {
+  // Expand response files.
+  llvm::SmallVector<const char *, 256> SmallVec;
+  for (const char *Arg : Args)
+    SmallVec.push_back(Arg);
+  llvm::StringSaver Saver(Alloc);
+  llvm::cl::ExpandResponseFiles(Saver, llvm::cl::TokenizeGNUCommandLine,
+                                SmallVec);
+
+  // Pack the results to a C-array and return it.
+  const char **Copy = Alloc.Allocate<const char *>(SmallVec.size() + 1);
+  std::copy(SmallVec.begin(), SmallVec.end(), Copy);
+  Copy[SmallVec.size()] = nullptr;
+  return llvm::ArrayRef(Copy, SmallVec.size() + 1);
+}
+
+int Driver::main(int Argc, const char **Argv) {
+  // Standard set up, so program fails gracefully.
+  llvm::BumpPtrAllocator Alloc;
+  llvm::sys::PrintStackTraceOnErrorSignal(Argv[0]);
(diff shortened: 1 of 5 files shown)


quic-seaswara (approved):

Steven6798 (approved):

test/Common/standalone/DriverNoLLVMStatics/driver-no-llvm-statics.test
```diff
@@ -0,0 +1,9 @@
+## UNSUPPORTED: windows
```
parth-07, line 1: Why is it unsupported for Windows?

lib/LinkerWrapper/Driver.cpp
```diff
@@ -18,12 +18,69 @@
...
+}
+
+int Driver::main(int Argc, const char **Argv) {
+  // Standard set up, so program fails gracefully.
+  llvm::BumpPtrAllocator Alloc;
+  llvm::sys::PrintStackTraceOnErrorSignal(Argv[0]);
+  llvm::PrettyStackTraceProgram StackPrinter(Argc, Argv);
+  llvm::llvm_shutdown_obj Shutdown;
+  // This is declared after Shutdown so that it is destroyed first.
+  llvm::ScopedFatalErrorHandler FatalHandler(shutdownOnFatalError);
```
parth-07: Do we strictly need this for resolving the race-condition issue?

lib/LinkerWrapper/Driver.cpp
```diff
@@ -18,12 +18,69 @@
...
+  const char **Copy = Alloc.Allocate<const char *>(SmallVec.size() + 1);
+  std::copy(SmallVec.begin(), SmallVec.end(), Copy);
+  Copy[SmallVec.size()] = nullptr;
+  return llvm::ArrayRef(Copy, SmallVec.size() + 1);
+}
+
+/// report_fatal_error() calls exit(1) (or abort()) directly, so it
+/// cleans up nothing. This cleanup routine is called before the exit.
+static void shutdownOnFatalError(void *, const char *Reason, bool) {
+  llvm::errs() << "Fatal: " << Reason << "\n";
```
parth-07: I am not sure if we can always rely on `llvm::errs()` and so on for fatal errors. The cause of an LLVM fatal error might be stream failure...

test/Common/standalone/DriverNoLLVMStatics/driver-no-llvm-statics.test
quic-areg: `%eld` on windows does not have a symbol table

lib/LinkerWrapper/Driver.cpp
quic-areg: Yes, I think so; we call `llvm::report_fatal_error` in a few places, and that exits directly without cleanup. This is a simple RAII object for registering the handler where we call `shutdown` before exit. LLD has something similar: 

```
  // Dealloc/destroy ManagedStatic variables before calling _exit().
  // In an LTO build, allows us to get the output of -time-passes.
  // Ensures that the thread pool for the parallel algorithms is stopped to
  // avoid intermittent crashes on Windows when exiting.
  if (!CrashRecoveryContext::GetCurrent())
    llvm_shutdown();
```
note the windows comment

lib/LinkerWrapper/Driver.cpp
quic-areg: Thanks for catching this: I updated the PR to use what LLVM does in its handler.

lib/LinkerWrapper/Driver.cpp
quic-seaswara: It appears that lld uses CrashRecoveryContext

https://llvm.org/doxygen/classllvm_1_1CrashRecoveryContext.html

This might be useful to look into. 

Thoughts ?

quic-areg: > Do we strictly need this for resolving the race-condition issue?

After looking at this again, I think the fatal error handler is not needed for this fix, and my original comparison to lld was incorrect. Calling shutdown from the fatal handler may actually be unsafe since the handler runs synchronously on the thread that reported the error. In that case we should let the process terminate without attempting static cleanup.  This is what lld actually does; what I cited before comes from the ordinary exit path, not fatal errors.

> It appears that lld uses CrashRecoveryContext

It looks like lld uses it to return control to the caller when it is used as a library. I don't know if it's applicable to this patch. 

I am going to remove the fatal handler and keep this patch focused on the duplicate LLVM statics and normal shutdown path.

quic-seaswara (approved):

quic-seaswara:
@quic-areg : Please cherry-pick this to release/23.x branch

FYI @apazos

PR #1799, opened 2026-09-01
Scope the libc++ quarantine to the cpullvm boards

BhoumikPatidar:
The `cpp.libcxx.libcxx.picolibc / libraries.hash_map.cxx.djb2` entry had no platforms filter, so it also applied to `qemu_x86_64`, which uses GCC from the Zephyr SDK rather than cpullvm. The quarantined reason is a cpullvm libc++abi built without exceptions, which does not apply there.
`libraries.hash_map.cxx.djb2` passes on qemu_x86_64 and `cpp.libcxx.libcxx.picolibc` filters out via `TOOLCHAIN_HAS_LIBCXX`.

### DIFF
--- a/.github/zephyr-quarantine.yml
+++ b/.github/zephyr-quarantine.yml
@@ -47,9 +47,18 @@
 - scenarios:
   - cpp.libcxx.libcxx.picolibc
   - libraries.hash_map.cxx.djb2
+  platforms:
+  - qemu_riscv32/.*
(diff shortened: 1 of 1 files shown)


quic-seaswara (approved):

quic-seaswara:
@quic-areg pls review

quic-areg (approved):

PR #1801, opened 2026-09-01
[command-line] --discard-locals now discard ".L" local temporaries

### LINKED ISSUE #1800: [command-line] --discard-locals not discarding ".L" local temporaries
The --discard-locals command not discarding ".L" prefixed local temporaries. This can be seen during the ARM linux kernel build, causing the following failure:

```
  LD      arch/arm/boot/compressed/vmlinux
following symbols must have non local/private scope:
.L_user_stack_end
make[3]: *** [arch/arm/boot/compressed/Makefile:153: arch/arm/boot/compressed/vmlinux] Error 1
make[3]: *** Deleting file 'arch/arm/boot/compressed/vmlinux'
make[2]: *** [arch/arm/boot/Makefile:57: arch/arm/boot/compressed/vmlinux] Error 2
make[1]: *** [arch/arm/Makefile:300: zImage] Error 2
```

Steven6798:
The --discard-locals command now discards ".L" prefixed local temporaries.

Fix: qualcomm#1800

### DIFF
--- a/lib/Target/GNULDBackend.cpp
+++ b/lib/Target/GNULDBackend.cpp
@@ -1191,12 +1191,15 @@ void GNULDBackend::sizeSymTab() {
     return;
 
   bool isStripLocal = (S == GeneralOptions::StripLocals);
+  bool isStripTemporaries = (S == GeneralOptions::StripTemporaries);
 
   size_t strtab = 1;
   std::vector<ResolveInfo *> &RVect = m_Module.getSymbols();
   uint64_t NumSymbols = 0;
   for (auto &Sym : RVect) {
     if ((isStripLocal && NumSymbols && Sym->isLocal()) ||
+        (isStripTemporaries && Sym->isLocal() &&
+         llvm::StringRef(Sym->name()).starts_with(".L")) ||
         m_SymbolsToRemove.count(Sym)) {
       continue;
     }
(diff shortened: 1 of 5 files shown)


Steven6798:
This fix makes the ARM Linux kernel no longer fail.

quic-shaksuma (approved):

quic-areg (approved):

lib/Target/GNULDBackend.cpp
```diff
@@ -1409,7 +1413,10 @@ GNULDBackend::emitRegNamePools(llvm::FileOutputBuffer &pOutput) {
   for (auto &S : symbols) {
     m_pSymIndexMap[S->outSymbol()] = symIdx;
     // Null symbol is already emitted.
-    if ((isStripLocal && S->isLocal()) || m_SymbolsToRemove.count(S)) {
+    if ((isStripLocal && S->isLocal()) ||
+        (isStripTemporaries && S->isLocal() &&
+         llvm::StringRef(S->name()).starts_with(".L")) ||
+        m_SymbolsToRemove.count(S)) {
```
quic-seaswara, line 1419: What happens if you have --emit-relocs ?

You need to test for partial link to make sure they are not stripped away. 

If --keep-labels is true, we should remove .L symbols.

quic-seaswara (changes requested):

lib/Target/GNULDBackend.cpp
Steven6798: I'll look into this. If symbols are not preserved when --emit-relocs is present, then doesn't that also affect --discard-all?

quic-seaswara: Yes, looks like there is a bug (or) we may need to verify this.

Steven6798: Everything is looking good. I'll add the updated tests.

quic-seaswara (approved):

PR #1819, opened 2026-09-03
refactor: rename 'checkUnsignedRange' to 'reportUnsignedOverflow'

### LINKED ISSUE #1738: `Relocator::checkUnsignedRange` does not check unsigned range
`Reloctor::checkUnsignedRange` simply reports the overflow error instead of checking and reporting if overflowed. We should either rename the function or actually check the range within in.

tecnolgd:
What's Changed

- Renamed `checkUnsignedRange` to `reportUnsignedOverflow` across `Relocator.h`, `Relocator.cpp` and all `Target/` backends to accurately describe that the function reports an overflow error instead of performing range validation check.

Closes #1738

### DIFF
--- a/include/eld/Target/Relocator.h
+++ b/include/eld/Target/Relocator.h
@@ -182,8 +182,8 @@ protected:
 Relocator::Result checkSignedRange(Relocation &Rel, Relocator &R, int64_t Value,
                                    unsigned Bits);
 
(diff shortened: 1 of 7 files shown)


quic-areg (approved):
LGTM

quic-seaswara (approved):
Thanks for the patch!

PR #1823, opened 2026-09-04
[AArch64] Add --relax/--no-relax flag for TLS relaxations

### LINKED ISSUE #1793: [AArch64] --relax/--no-relax flag not supported
Currently ELD does not support --relax/--no-relax flags for AArch64.

Related: #1658

deepakshirkem:
By default (`--no-relax`), new TLS relaxations for global symbols  are disabled, preserving existing behavior. Pass `--relax` to   enable IE->LE relaxation for global symbols in static executables.

Fixes #1793


parth-07 (commented):
Thank you for pushing this fix so quickly!

Now most of the targets support --relax/--no-relax functionality. Can you please refactor the handling of --relax/--no-relax option in such a way that we don't need target-specific handling of --relax/--no-relax?

deepakshirkem:
@parth-07, I have unified --relax/--no-relax for x86 and AArch64  into a single flag in GnuLinkerOptions.td parsed by GnuLdDriver.

RISCV still uses its own BRiscvRelax flag because it has many target-specific relax variants (ToC, Xqci, GOT, Tbljal). 
Unifying RISCV requires more careful analysis and testing. I will address RISCV unification as a follow-up PR.

lib/LinkerWrapper/GnuLdDriver.cpp
```diff
@@ -381,6 +381,10 @@ bool GnuLdDriver::processOptions(llvm::opt::InputArgList &Args) {
   // -pie
   Config.options().setPIE(Args.hasFlag(T::pie, T::no_pie, false));
 
+  // --relax/--no-relax: enable/disable relaxations (enabled by default)
+  Config.options().setRelax(
+      Args.hasFlag(T::relax, T::no_relax, /*default=*/true));
```
parth-07: @quic-seaswara Should we enable the relaxations by default, or keep them disabled for now (until we are very sure that all the relaxations are robust)?

quic-seaswara: We might want to consider this enabled as default after all the testing is complete.

parth-07: @deepakshirkem Can you please make relaxations as disabled by default (except RISCV, as for that it was already enabled by default)?

deepakshirkem: @parth-07  Updated the default to disabled (false) for all targets except RISCV.

parth-07 (approved):
LGTM.

deepakshirkem:
@quic-seaswara Ping.

deepakshirkem:
@quic-seaswara Ping :))

PR #1838, opened 2026-09-09
[LinkerScript] Update hashing mechanism to use MD5 for file contents

### LINKED ISSUE #1188: We should use MD5 hash instead of xx3 hash for `plugin::LinkerScript::getHash()`
MD5 hash is far better than xx3 hash for ensuring that two file contents are the same. xx3 hash is designed for speed instead of being collision-resistant. MD5 is much better for our use case because it has much lower collision rate. Hence, we should use it (or some other equally good hash strategy) for `plugin::LinkerScript::getHash()`.

malto101:
`LinkerScript::getHash()` previously hashed only script file paths, so
  identical scripts at different times/machines produced stable hashes, but two
  logically different scripts at the same path could collide, and content
  changes without a path change went undetected. This matters for consumers of
  the hash such as the `--flto-options=cache` ThinLTO cache key and the
  `plugin::LinkerScript::getHash()` plugin API.
  
  - `LinkerScript::getHash()` now returns a combined hash: SHA1 of each script
    path mixed with an MD5 of that file's contents (including INCLUDE files),
    instead of hashing paths alone.
    
This resolves #1188


test/Common/Plugin/GetLinkerScriptHash/GetLinkerScriptHash.test
```diff
@@ -0,0 +1,32 @@
+#---GetLinkerScriptHash.test----------------------- Executable,LS --------------------#
```
quic-seaswara, line 2: add appropriate REQUIRES since it uses a variety of tools available on linux to build.

test/Common/Plugin/GetLinkerScriptHash/GetLinkerScriptHash.test
```diff
@@ -0,0 +1,32 @@
...
+
+## INCLUDE file contents participate in the hash.
+RUN: mkdir -p %t.incdir
+RUN: cp %p/Inputs/script-with-include.t %t.incdir/script.t
+RUN: cp %p/Inputs/extra.t %t.incdir/extra.t
+RUN: %link %linkopts %t1.o -T %t.incdir/script.t -L %t.incdir -o %t.inc1.out --plugin-config %p/Inputs/plugin.config --flto-options=cache=%t.cache > %t.hashInc1 2>&1
+RUN: cp %p/Inputs/extra-other.t %t.incdir/extra.t
+RUN: %link %linkopts %t1.o -T %t.incdir/script.t -L %t.incdir -o %t.inc2.out --plugin-config %p/Inputs/plugin.config --flto-options=cache=%t.cache > %t.hashInc2 2>&1
+RUN: not diff %t.hashInc1 %t.hashInc2
+#HASH: linker script hash: {{[0-9A-F]{40}}}
```
quic-seaswara: Can we print the linker script hash in the map file ? 

You can then use the map file to figure out if the hashes are same. 

Does any change to comments in the linker script affect the hash as well ? Just checking.

quic-areg (approved):
LGTM

quic-seaswara (changes requested):

test/Common/Plugin/GetLinkerScriptHash/GetLinkerScriptHash.test
malto101: > Does any change to comments in the linker script affect the hash as well

i would expect it to change, as we reading the raw file content, let me verify it once

Edit: just verified, it does so

test/Common/Plugin/GetLinkerScriptHash/GetLinkerScriptHash.test
malto101: Done

test/Common/Plugin/GetLinkerScriptHash/GetLinkerScriptHash.test
```diff
@@ -0,0 +1,41 @@
...
+RUN: not diff %t.hashA %t.hashB
+RUN: %grep "Linker script hash:" %t.diff1.map > %t.mapHashA
+RUN: %grep "Linker script hash:" %t.diff2.map > %t.mapHashB
+RUN: not diff %t.mapHashA %t.mapHashB
+
+## INCLUDE file contents participate in the hash.
+RUN: mkdir -p %t.incdir
+RUN: cp %p/Inputs/script.WithInclude.t %t.incdir/script.t
+RUN: cp %p/Inputs/extra.t %t.incdir/extra.t
+RUN: %link %linkopts %t1.o -T %t.incdir/script.t -L %t.incdir -o %t.inc1.out --plugin-config %p/Inputs/plugin.config --flto-options=cache=%t.cache > %t.hashInc1 2>&1
```
quic-seaswara, line 35: Everything in the test has a variable from lit.cfg, use %not, %cp %mkdir etc.

quic-seaswara (approved):
Just a nit pick, but I am approving this patch.

test/Common/Plugin/GetLinkerScriptHash/GetLinkerScriptHash.test
malto101: had slipped my mind, 
updated the same

test/Common/Plugin/GetLinkerScriptHash/GetLinkerScriptHash.test
quic-seaswara: Will this pass on windows / mac ?

quic-seaswara:
@parth-07  pls review

test/Common/Plugin/GetLinkerScriptHash/GetLinkerScriptHash.test
malto101: it should be work as long as it has unix shell access.
Edit: but I see that windows is using ps here, give me some time to verify

lib/Core/LinkerScript.cpp
```diff
@@ -169,38 +170,40 @@ std::string LinkerScript::getHash() {
...
 
+static void mixInMD5(llvm::SHA1 &Hasher, llvm::StringRef Contents) {
+  llvm::MD5::MD5Result Digest =
+      llvm::MD5::hash(llvm::arrayRefFromStringRef(Contents));
+  Hasher.update(llvm::ArrayRef<uint8_t>(Digest.data(), Digest.size()));
+}
+
 /// Compute a hash for the given linker script and add it to the combined hash
 /// if it's a file. Otherwise, add the given text to the hash.
 void LinkerScript::addToHash(llvm::StringRef FilenameOrText) {
```
parth-07, line 181: Not exactly a new issue, but something that we should keep track of. `addToHash` function is not reproduce-functionality friendly. It takes in a file name and reads it to compute its hash. However, we typically pass it the original file path, instead of the filename in the reproduce tarball. Hence, it computes the hash by reading the original file path, which may or may not exist, and maybe stale. 

For example:

```bash
#!/usr/bin/env bash
set -e

mkdir -p tmp

cat > tmp/1.c <<\EOF
void _start(void) {}
EOF

cat > tmp/original.t <<\EOF
SECTIONS {
  .foo : { *(.text) }
}
EOF

cat > tmp/hashed.t <<\EOF
SECTIONS {
  .foo : { *(.text) }
}
EOF

cat > tmp/other.t <<\EOF
SECTIONS {
  .bar : { *(.text) }
}
EOF

cat > tmp/plugin.config <<\EOF
GlobalPlugins:
  - Type : LinkerPlugin
    Name : GetLinkerScriptHash
    Library: GetLinkerScriptHash
EOF

cat > tmp/mapping.ini <<\EOF
[Hash]
tmp/original.t=tmp/hashed.t
EOF

clang -target riscv64 -c tmp/1.c -o tmp/1.o

ld.eld tmp/1.o -T tmp/hashed.t -o tmp/a.eld.out --plugin-config tmp/plugin.config \
  --flto-options=cache=tmp/cache --mapping-file=tmp/mapping.ini \
  -Map tmp/a.map > tmp/a.log

cp tmp/other.t tmp/hashed.t

ld.eld tmp/1.o -T tmp/hashed.t -o tmp/b.eld.out --plugin-config tmp/plugin.config \
  --flto-options=cache=tmp/cache --mapping-file=tmp/mapping.ini \
  -Map tmp/b.map > tmp/b.log 

grep "linker script hash" tmp/a.log tmp/b.log
grep "Linker script hash" tmp/a.map tmp/b.map
llvm-readelf -S tmp/a.eld.out | grep '\.foo\|\.bar'
llvm-readelf -S tmp/b.eld.out | grep '\.foo\|\.bar'
```

In this example, the linker script has changed, however, the hash remains the same:

```
+ grep 'linker script hash' tmp/a.log tmp/b.log
tmp/a.log:linker script hash: 2EB839B99656D8EF2AC6118EDEB4A0989B1FEC93
tmp/b.log:linker script hash: 2EB839B99656D8EF2AC6118EDEB4A0989B1FEC93
+ grep 'Linker script hash' tmp/a.map tmp/b.map
tmp/a.map:Linker script hash: 2EB839B99656D8EF2AC6118EDEB4A0989B1FEC93
tmp/b.map:Linker script hash: 2EB839B99656D8EF2AC6118EDEB4A0989B1FEC93
+ llvm-readelf -S tmp/a.eld.out
+ grep '\.foo\|\.bar'
  [ 1] .foo              PROGBITS        0000000000000000 001000 000010 00  AX  0   0  2
+ llvm-readelf -S tmp/b.eld.out
+ grep '\.foo\|\.bar'
  [ 1] .bar  
```

parth-07: Perhaps this issue should be addressed separately.

parth-07 (approved):
Looks good to me.

lib/Core/LinkerScript.cpp
quic-seaswara: @parth-07 can you open a issue for this ? @malto101 do you want to work on that ?

malto101: @quic-seaswara yes

malto101:
tested on mac and linux
works as intended

tests on windows is pending
will let you know by tom EOD

lib/Core/LinkerScript.cpp
parth-07: I have created an issue for this: #1859.

parth-07:
Hi @malto101,

Were you able to verify this patch on Windows?

malto101:
Hello @parth-07 @quic-seaswara 
the feature itself works
but the tests fails because the test extracts just the hash line from each .map file using %grep, 

(currently used in window's grep substitution) findstr "Linker script hash:" (without /C:) treats the quoted phrase as separate OR'd words — "Linker" OR "script" OR "hash:" — not a literal substring. So it pulls in every other line containing any of those words too

malto101:
to handle windows test too,
i have raised #1875 , this will solve the problem,
we can merge the above and then merge this

malto101:
Hello @quic-seaswara @parth-07 
the windows test is passing when i moved from grep to filecheck,
we can merge it

quic-seaswara (approved):

PR #1840, opened 2026-09-10
Support TLS offsets across multiple PT_TLS segments for x86_64

BhoumikPatidar:
Fixes `x86-64` TLS offset computation for images containing multiple `PT_TLS` segments.

The `x86-64` backend assumed that an output image contained exactly one `PT_TLS` segment and calculated the thread-pointer offset from that segment alone. Linker scripts can generate multiple non-empty TLS segments with padding between them, causing TLS relocations to receive incorrect offsets.

The backend now considers all non-empty `PT_TLS` segments, finds the lowest and highest TLS address, uses the maximum TLS alignment and computes the thread-pointer offset from the complete aligned TLS span.

This fixes failing test cases in `picolibc test suite` that exercise the case of multiple `PT_TLS` segments.

### DIFF
--- a/lib/Target/X86/x86_64Relocator.cpp
+++ b/lib/Target/X86/x86_64Relocator.cpp
@@ -14,6 +14,8 @@
 #include "llvm/ADT/Twine.h"
 #include "llvm/BinaryFormat/ELF.h"
 
+#include <algorithm>
+#include <limits>
 using namespace eld;
 
 //===--------------------------------------------------------------------===//
@@ -570,19 +572,35 @@ Relocator::Result VerifyRelocAsNeededHelper(
 void x86_64Relocator::computeTLSOffsets() {
   std::vector<ELFSegment *> tlsSegments =
       getTarget().elfSegmentTable().getSegments(llvm::ELF::PT_TLS);
-
-  if (tlsSegments.empty()) {
+  if (tlsSegments.empty())
     return;
+
+  // The x86-64 TPOFF value is relative to the thread pointer.  With TLS
+  // Variant 2, the thread pointer is placed after the complete TLS image,
+  // rounded up to the required alignment.  A linker script may produce more
+  // than one PT_TLS segment, so use the complete span rather than one
+  // segment's memsz.  Empty PT_TLS segments do not contribute to the image.
+  bool hasNonEmptySegment = false;
+  uint64_t lo = std::numeric_limits<uint64_t>::max();
+  uint64_t hi = 0;
+  uint64_t alignment = 1;
+  for (ELFSegment *Segment : tlsSegments) {
+    if (Segment->memsz() == 0)
+      continue;
+
+    const uint64_t segmentEnd = Segment->vaddr() + Segment->memsz();
+    lo = std::min(lo, Segment->vaddr());
+    hi = std::max(hi, segmentEnd);
+    alignment = std::max(alignment, Segment->align());
+    hasNonEmptySegment = true;
   }
 
-  ASSERT(tlsSegments.size() == 1,
-         "Multiple TLS segments not supported in x86_64 backend");
+  if (!hasNonEmptySegment)
+    return;
 
-  ELFSegment *tlsSegment = tlsSegments[0];
-  uint64_t templateSize = tlsSegment->memsz();
-  uint64_t alignment = tlsSegment->align();
-  templateSize = llvm::alignTo(templateSize, alignment);
-  GNULDBackend::setTLSTemplateSize(templateSize);
+  const uint64_t alignedEnd = llvm::alignTo(hi, alignment);
+  const uint64_t threadPointerOffset = alignedEnd - lo;
+  GNULDBackend::setTLSTemplateSize(threadPointerOffset);
 }
 
 template <typename T>
--- a/test/x86_64/linux/TLSLocalExecMultipleSegments/Inputs/1.c
+++ b/test/x86_64/linux/TLSLocalExecMultipleSegments/Inputs/1.c
@@ -1,0 +1,5 @@
+__thread int u;
+__attribute__((aligned(256))) __thread int v;
+__thread int a = 12;
+__attribute__((aligned(128))) __thread int b = 13;
+int foo(void) { return a; }
--- a/test/x86_64/linux/TLSLocalExecMultipleSegments/Inputs/script.t
+++ b/test/x86_64/linux/TLSLocalExecMultipleSegments/Inputs/script.t
@@ -1,0 +1,12 @@
+PHDRS {
+  CODE PT_LOAD;
+  DATA PT_LOAD;
+  T1 PT_TLS;
+  T2 PT_TLS;
+}
+
+SECTIONS {
+  .text (0x1000) : { *(.text*) } :CODE
+  .tdata (0x4000) : { *(.tdata*) } :DATA :T1
+  .tbss : ALIGN(0x400) { *(.tbss*) } :DATA :T2
+}
--- a/test/x86_64/linux/TLSLocalExecMultipleSegments/TLSLocalExecMultipleSegments.test
+++ b/test/x86_64/linux/TLSLocalExecMultipleSegments/TLSLocalExecMultipleSegments.test
@@ -1,0 +1,15 @@
+#---TLSLocalExecMultipleSegments.test--------------------------- Executable,LS ---------------#
+#BEGIN_COMMENT
+# This checks x86-64 local-exec TLS relocation computation when the image has
+# multiple PT_TLS segments and the TLS segments have alignment padding.
+#END_COMMENT
+#START_TEST
+RUN: %clang %clangopts -c %p/Inputs/1.c -o %t.o
+RUN: %link %linkopts -o %t.out %t.o -T %p/Inputs/script.t
+RUN: %readelf -l %t.out | %filecheck %s --check-prefix=READELF
+RUN: %objdump -d %t.out | %filecheck %s --check-prefix=OBJDUMP
+#END_TEST
+
+READELF: TLS {{.*}} RW 0x400
+READELF: TLS {{.*}} RW 0x400
+OBJDUMP: leaq -0x800


lib/Target/X86/x86_64Relocator.cpp
```diff
@@ -570,19 +572,38 @@ Relocator::Result VerifyRelocAsNeededHelper(
...
+  // segment's memsz.  Empty PT_TLS segments do not contribute to the image.
+  bool hasNonEmptySegment = false;
+  uint64_t lo = std::numeric_limits<uint64_t>::max();
+  uint64_t hi = 0;
+  uint64_t alignment = 1;
+  for (ELFSegment *Segment : tlsSegments) {
+    if (Segment->memsz() == 0)
+      continue;
+
+    const uint64_t segmentAlignment = std::max<uint64_t>(1, Segment->align());
```
parth-07: nitpic: why do we need `std::max(...)` here? Also, if `segmentAlign` is only used at line 596 (`alignment = ....`), then can we simply use `Segment->align()` at line 596?

lib/Target/X86/x86_64Relocator.cpp
```diff
@@ -570,19 +572,38 @@ Relocator::Result VerifyRelocAsNeededHelper(
...
+    const uint64_t segmentEnd = Segment->vaddr() + Segment->memsz();
+    lo = std::min(lo, Segment->vaddr());
+    hi = std::max(hi, segmentEnd);
```
parth-07, line 593: Will this work if the TLS segments are non-contiguous? Is it even valid to have non-contiguous TLS segments? 

Also, is the behavior of multiple TLS segments specified by the ABI or does each runtime/loader is free to choose the behaviour here?

parth-07 (commented):
This is a really interesting case. Thank you for analyzing/debugging it so quickly!

lib/Target/X86/x86_64Relocator.cpp
quic-seaswara: TLS segments need to be contiguous. Linker will produce an error.

lib/Target/X86/x86_64Relocator.cpp
```diff
@@ -570,19 +572,38 @@ Relocator::Result VerifyRelocAsNeededHelper(
...
+    alignment = std::max(alignment, segmentAlignment);
+    hasNonEmptySegment = true;
```
quic-seaswara, line 595: Why check for empty segments ?

lib/Target/X86/x86_64Relocator.cpp
```diff
@@ -570,19 +572,38 @@ Relocator::Result VerifyRelocAsNeededHelper(
...
+  }
+
+  if (!hasNonEmptySegment) {
+    return;
+  }
```
quic-seaswara: nit : remove extra braces ?

lib/Target/X86/x86_64Relocator.cpp
```diff
@@ -570,19 +572,38 @@ Relocator::Result VerifyRelocAsNeededHelper(
...
 
-  ELFSegment *tlsSegment = tlsSegments[0];
-  uint64_t templateSize = tlsSegment->memsz();
-  uint64_t alignment = tlsSegment->align();
-  templateSize = llvm::alignTo(templateSize, alignment);
-  GNULDBackend::setTLSTemplateSize(templateSize);
+  const uint64_t alignedEnd = llvm::alignTo(hi, alignment);
+  const uint64_t threadPointerOffset = alignedEnd - lo;
+  GNULDBackend::setTLSTemplateSize(threadPointerOffset);
```
quic-seaswara, line 603: Why does this not affect other targets ?

quic-seaswara (changes requested):
Can we accompany this with a build and run test ?

lib/Target/X86/x86_64Relocator.cpp
BhoumikPatidar: Thanks, addressed these, its not required.
Had added that because for p_align both 0 and 1 mean no alignment, but as we initialize alignment with 1, that case is not an issue

lib/Target/X86/x86_64Relocator.cpp
BhoumikPatidar: Consider the case when the linker script has two TLS segments for .tdata and .tbss but the program has no .tdata.
This leads to the vaddr set to 0 for that segment. That makes `lo` incorrectly set to 0.
This was observed in a test case in picolibc test suite - `constructor-skip`

lib/Target/X86/x86_64Relocator.cpp
BhoumikPatidar: `computTLSOffsets()` is virtual, so this implementation called only for x86_64Relocator.

quic-areg (approved):
LGTM

BhoumikPatidar:
> Can we accompany this with a build and run test ?

The glibc loader does not handle multiple TLS segments. The last TLS segment acts as the only TLS segment.

parth-07 (changes requested):
We also need to update `GNULDBackend::finalizeTLSSymbol` to use the right TLS segment when computing TLS symbol values. Wrong TLS segment can be picked when the TLS segment order in the PHDRS is different from the TLS segments order in the ascending virtual addresses.

Reproducer:

```bash
#!/usr/bin/env bash
set -e

cat >tmp/pr1840_tls_order.c <<\EOF
__attribute__((section(".tdata.low"))) __thread int low = 1;
__attribute__((section(".tdata.high"), aligned(256))) __thread int high = 2;
int foo(void) { return low; }
EOF

cat >tmp/pr1840_tls_order.t <<\EOF
PHDRS {
  CODE PT_LOAD;
  DATA PT_LOAD;
  T1 PT_TLS;
  T2 PT_TLS;
}

SECTIONS {
  .text 0x1000 : { *(.text*) } :CODE
  .low 0x4000 : { *(.tdata.low*) } :DATA :T2
  .high 0x5000 : { *(.tdata.high*) } :DATA :T1
}
EOF

clang -target x86_64-linux-gnu -c tmp/pr1840_tls_order.c -o tmp/pr1840_tls_order.o

for ld in ld.eld ld.bfd; do
  $ld -m elf_x86_64 -o tmp/pr1840_tls_order.$ld.out tmp/pr1840_tls_order.o -T tmp/pr1840_tls_order.t
  llvm-objdump -d tmp/pr1840_tls_order.$ld.out
done
```

Output:

```bash
+ ld.eld -m elf_x86_64 -o tmp/pr1840_tls_order.ld.eld.out tmp/pr1840_tls_order.o -T tmp/pr1840_tls_order.t
+ llvm-objdump -d tmp/pr1840_tls_order.ld.eld.out

tmp/pr1840_tls_order.ld.eld.out:	file format elf64-x86-64

Disassembly of section .text:

0000000000001000 <foo>:
    1000: 55                           	pushq	%rbp
    1001: 48 89 e5                     	movq	%rsp, %rbp
    1004: 64 48 8b 04 25 00 00 00 00   	movq	%fs:0x0, %rax
    100d: 48 8d 80 00 df ff ff         	leaq	-0x2100(%rax), %rax
    1014: 8b 00                        	movl	(%rax), %eax
    1016: 5d                           	popq	%rbp
    1017: c3                           	retq
+ for ld in ld.eld ld.bfd
+ ld.bfd -m elf_x86_64 -o tmp/pr1840_tls_order.ld.bfd.out tmp/pr1840_tls_order.o -T tmp/pr1840_tls_order.t
+ llvm-objdump -d tmp/pr1840_tls_order.ld.bfd.out

tmp/pr1840_tls_order.ld.bfd.out:	file format elf64-x86-64

Disassembly of section .text:

0000000000001000 <foo>:
    1000: 55                           	pushq	%rbp
    1001: 48 89 e5                     	movq	%rsp, %rbp
    1004: 64 48 8b 04 25 00 00 00 00   	movq	%fs:0x0, %rax
    100d: 48 8d 80 00 ef ff ff         	leaq	-0x1100(%rax), %rax
    1014: 8b 00                        	movl	(%rax), %eax
    1016: 5d                           	popq	%rbp
```

eld has incorrectly resolved `leaq	-0x2100(%rax), %rax` instruction. The value `-0x2100` is incorrect because of wrong TLS segment used to compute the value of `low` symbol.

BhoumikPatidar:
> We also need to update `GNULDBackend::finalizeTLSSymbol` to use the right TLS segment when computing TLS symbol values. Wrong TLS segment can be picked when the TLS segment order in the PHDRS is different from the TLS segments order in the ascending virtual addresses.
> 
> Reproducer:
> 
> ```shell
> #!/usr/bin/env bash
> set -e
> 
> cat >tmp/pr1840_tls_order.c <<\EOF
> __attribute__((section(".tdata.low"))) __thread int low = 1;
> __attribute__((section(".tdata.high"), aligned(256))) __thread int high = 2;
> int foo(void) { return low; }
> EOF
> 
> cat >tmp/pr1840_tls_order.t <<\EOF
> PHDRS {
>   CODE PT_LOAD;
>   DATA PT_LOAD;
>   T1 PT_TLS;
>   T2 PT_TLS;
> }
> 
> SECTIONS {
>   .text 0x1000 : { *(.text*) } :CODE
>   .low 0x4000 : { *(.tdata.low*) } :DATA :T2
>   .high 0x5000 : { *(.tdata.high*) } :DATA :T1
> }
> EOF
> 
> clang -target x86_64-linux-gnu -c tmp/pr1840_tls_order.c -o tmp/pr1840_tls_order.o
> 
> for ld in ld.eld ld.bfd; do
>   $ld -m elf_x86_64 -o tmp/pr1840_tls_order.$ld.out tmp/pr1840_tls_order.o -T tmp/pr1840_tls_order.t
>   llvm-objdump -d tmp/pr1840_tls_order.$ld.out
> done
> ```
> 
> Output:
> 
> ```shell
> + ld.eld -m elf_x86_64 -o tmp/pr1840_tls_order.ld.eld.out tmp/pr1840_tls_order.o -T tmp/pr1840_tls_order.t
> + llvm-objdump -d tmp/pr1840_tls_order.ld.eld.out
> 
> tmp/pr1840_tls_order.ld.eld.out:	file format elf64-x86-64
> 
> Disassembly of section .text:
> 
> 0000000000001000 <foo>:
>     1000: 55                           	pushq	%rbp
>     1001: 48 89 e5                     	movq	%rsp, %rbp
>     1004: 64 48 8b 04 25 00 00 00 00   	movq	%fs:0x0, %rax
>     100d: 48 8d 80 00 df ff ff         	leaq	-0x2100(%rax), %rax
>     1014: 8b 00                        	movl	(%rax), %eax
>     1016: 5d                           	popq	%rbp
>     1017: c3                           	retq
> + for ld in ld.eld ld.bfd
> + ld.bfd -m elf_x86_64 -o tmp/pr1840_tls_order.ld.bfd.out tmp/pr1840_tls_order.o -T tmp/pr1840_tls_order.t
> + llvm-objdump -d tmp/pr1840_tls_order.ld.bfd.out
> 
> tmp/pr1840_tls_order.ld.bfd.out:	file format elf64-x86-64
> 
> Disassembly of section .text:
> 
> 0000000000001000 <foo>:
>     1000: 55                           	pushq	%rbp
>     1001: 48 89 e5                     	movq	%rsp, %rbp
>     1004: 64 48 8b 04 25 00 00 00 00   	movq	%fs:0x0, %rax
>     100d: 48 8d 80 00 ef ff ff         	leaq	-0x1100(%rax), %rax
>     1014: 8b 00                        	movl	(%rax), %eax
>     1016: 5d                           	popq	%rbp
> ```
> 
> eld has incorrectly resolved `leaq -0x2100(%rax), %rax` instruction. The value `-0x2100` is incorrect because of wrong TLS segment used to compute the value of `low` symbol.

I have created #1892 to track this case.

parth-07 (approved):

quic-seaswara (approved):

PR #1857, opened 2026-09-15
[AArch64] Avoid creating unnecessary .got.plt section

deepakshirkem:
ELD was unnecessarily creating the `.got.plt` section when only `.got` is required. The `.got.plt` section should only be created when PLT entries are needed.

This change adds the reserved GOT0 entry to `.got` and avoids creating `.got.plt` when it is not required.

Part of #1030


lib/Target/AArch64/AArch64GOT.cpp
```diff
@@ -11,6 +11,21 @@
...
+  if (!R)
+    return G;
+
+  Relocation *Rel = Relocation::Create(
+      llvm::ELF::R_AARCH64_ABS64, 64, make<FragmentRef>(*G, 0), 0);
+  Rel->setSymInfo(R);
+  O->addRelocation(Rel);
+
+  return G;
+}
```
quic-seaswara: AArch64 does not reserve GOT0 like RISC-V ABI mentions, AFAIK.

What does `lld` do ?

deepakshirkem: @quic-seaswara I checked LLD and the AArch64 SysV ABI. AArch64 does not appear to require a reserved GOT0 entry like RISC-V.

LLD also does not seem to create a reserved GOT0 unconditionally for AArch64. I think the fix here should instead avoid creating `.got.plt` when only regular GOT entries are needed, without adding GOT0.

Would you suggest following LLD's behavior here?

deepakshirkem:
@quic-seaswara The updated approach no longer adds a reserved GOT0 for AArch64. Regular GOT references now stay in .got, while GOTPLT0 is created only when the PLT path explicitly requests it, which matches the behavior I observed with LLD.
Please take another look at this approach and run the CI?

deepakshirkem:
The CI failures were caused by tests that still expected the old AArch64 behavior where `.got.plt` was created unconditionally.

With the new approach, regular GOT references use only `.got`, and `.got.plt` is created only when the PLT path actually requires it. Because of that, some section indices, GOT addresses, and instruction offsets changed.

I updated the affected AArch64 tests to match the new layout, and marked the common `InitializeGOTPLT0ToDynamic` test unsupported for AArch64 since that test assumes the old GOTPLT0/SOS-loader behavior.

@parth-07 @quic-seaswara @quic-areg

deepakshirkem:
@quic-areg Ping ::)

PR #1858, opened 2026-09-15
Restrict executable ELF inputs to --just-symbols and drop unneeded reads

quic-seaswara:
An ELF executable can only be used via --just-symbols, which only needs each symbol's name, type, and value.


test/Common/standalone/InvalidObject/invalidobject.test
```diff
@@ -1,9 +1,13 @@
 UNSUPPORTED: riscv32, riscv64, x86
+#---invalidobject.test--------------------------- Executable -----------------#
+#BEGIN_COMMENT
 # This tests that the linker does not error when there is an executable
```
parth-07: The test description needs to be updated as the linker now errors out when there is an executable.

quic-areg (approved):

test/Common/standalone/InvalidObject/invalidobject.test
quic-seaswara: Done

PR #1863, opened 2026-09-16
[Zephyr] Fix Zephyr nightly ordering and quarantine known failures

quic-areg:
When zephyr nightly starts before nightly CI finishes, the workflow will fail because it depends on nighltly's toolchain artifact. Now, the zephyr nightly workflow is automatically triggered after nightly, similar to how PR CI automatically triggers the zephyr PR workflow.

Quarantine the known ARM EXIDX overflow and the RV32 XIP board overflow, and disable fatal warnings for the LLVM TLS test to make the workflow green again.


BhoumikPatidar (approved):

.github/zephyr-quarantine.yml
```diff
@@ -152,6 +152,21 @@
...
+  comment: >-
+    Same EXIDX merging issue as bluetooth.init.test_20. See issue #225.
+
+- scenarios:
+  - integration\.kernel
+  platforms:
+  - qemu_riscv32_xip/.*
+  comment: >-
+    The test's noinit section is 0x4020 bytes, but the board has 0x4000 bytes
+    of RAM. GCC also overflows by the same amount as ELD.
```
quic-seaswara, line 168: Fix sounds fine but we might want to watch if the community ends up fixing this, is there any way to mark this as XFAIL ? This willl allow if the test was fixed in zephyr upstream.

quic-seaswara (approved):

.github/zephyr-quarantine.yml
quic-areg: There isn't an `XFAIL` afaik but twister does take a `--quarantine-verify` that runs only the quarantined tests. We can add a check to the nightly workflow to make sure the quarantine list is still valid in a follow-up.

Steven6798 (approved):

PR #1868, opened 2026-09-16
[workflow] Update nightly check to use the latest LLVM release

### LINKED ISSUE #1867: [workflow] Update Linux nightly to use the latest LLVM release.
Currently, the Linux nightly workflow uses main. The main branch is not stable, and the latest release should be used instead.

Steven6798:
The workflow now uses LLVM 23.x instead of main.

Fix: #1867


.github/workflows/nightly-test.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     if: github.repository == 'qualcomm/eld'
     runs-on: ubuntu-22.04
     env:
-      BASE_BRANCH_NAME: ${{ github.event.pull_request.base.ref }}
+      BASE_BRANCH_NAME: ${{ github.event.pull_request.base.ref || 'release/23.x' }}
```
quic-seaswara, line 17: Why is nightly-test using release/23.x ? 

`eld` needs to always build with LLVM latest.

When building linux kernel use LLVM 23.x and overlay nightly `eld`

quic-seaswara (changes requested):

.github/workflows/nightly-test.yml
Steven6798: Ohh i see. I misunderstood. Thanks for the info.

Steven6798:
I create the branch in ELD instead of my fork to test the Linux nightly workflow.

PR #1875, opened 2026-09-17
[test] Use real grep on Windows instead of findstr for %grep

malto101:
findstr does not support literal multi-word substring matching the way %grep-using RUN lines expect: findstr "a b" matches lines containing ANY of the words "a" or "b" (word-OR semantics), not the literal substring "a b". This caused spurious failures such as Common/Plugin/GetLinkerScriptHash/GetLinkerScriptHash.test, where an unrelated line containing "script" and other individual words was matched, corrupting the diff comparison even though the actual feature output was correct.

Git for Windows ships a real grep.exe, and lit.cfg already resolves rm/mkdir the same way via which(), so resolve grep consistently instead of hardcoding findstr.


quic-areg:
when would we need to use grep instead of FileCheck?

quic-seaswara:
> when would we need to use grep instead of FileCheck?

Let us work on a follow up to remove usages of grep and use FileCheck.

malto101:
> Let us work on a follow up to remove usages of grep and use FileCheck.
closing this in favor for that

closing this PR in favour for above

PR #1878, opened 2026-09-17
Copy runtime libraries for LLVM test suite

BhoumikPatidar:
Remove the test executable RPATH flags and istead copy the nightly libc++ runtime libraries into `/usr/local/lib` before running the LLVM test suite.


parth-07 (commented):
What is the motivation behind this change?

.github/workflows/nightly-llvm-test-suite.yml
```diff
@@ -49,15 +49,24 @@ jobs:
...
 
+      - name: Copy runtime libraries
+        run: |
+          ELD_BIN=$(which ld.eld)
+          LIB_DIR="$(dirname "$ELD_BIN")/../lib"
+          sudo cp -L \
+            "$LIB_DIR/libc++.so.1" \
+            "$LIB_DIR/libc++abi.so.1" \
+            "$LIB_DIR/x86_64-unknown-linux-gnu/libunwind.so.1" \
+            /usr/local/lib/
```
quic-seaswara, line 60: Why are we copying this to /usr/local/lib ?

quic-seaswara (changes requested):

.github/workflows/nightly-llvm-test-suite.yml
BhoumikPatidar: Should we copy to toolchain's lib/ instead and have LD_LIBRARY_PATH pointing at it since it's not on default search path?

BhoumikPatidar:
> What is the motivation behind this change?

Mainly for good practice, avoiding using RPATH for resolving runtime libs

parth-07 (approved):

.github/workflows/nightly-llvm-test-suite.yml
quic-seaswara: linker already has RPATH set to $ORIGIN/../lib right ? So if they are in the lib directory it should work ?

BhoumikPatidar: ld.eld can use $ORIGIN/../lib. But the test suite binaries are in build-llvm-test-suite. They cant discover the libraries as such

quic-seaswara: sorry I did not know this. I am fine with using the -rpath flag. This is not portable to self hosted machines if we decide to go that way. 

please abandon this PR.

quic-seaswara (changes requested):

PR #1882, opened 2026-09-17
Harden workflows and resolve static-analysis findings

### LINKED ISSUE #1881: [workflow] Address Run QC Preflight Checks failures
Version 2.0.11 of reusable-qcom-preflight-checks-orchestrator.yml is throwing errors. Fix them and suppress false positives.

Example of a run: https://github.com/qualcomm/eld/actions/runs/35246458822/job/105287791960

Steven6798:
- Replace curl-pipe-shell installs with download-then-execute
- Validate dynamic SQLite identifiers and use safe YAML loading
- Route GitHub values through environment variables
- Replace raw buffer allocation with std::vector
- Document and suppress verified C++ analyzer false positives

Fix: qualcomm#1881


include/eld/PluginAPI/PluginADT.h
```diff
@@ -1448,6 +1488,11 @@ class DLL_A_EXPORT InputSymbol final {
 
 /// LinkerConfig houses functions for plugins to access linker configurations
 /// and options.
+// This is a type declaration; its constructors do not return values.
+// This is a type declaration; its constructors do not return values.
```
quic-seaswara, line 1492: Sorry for jumping on this, there is just repeated comments in the source code.

include/eld/PluginAPI/PluginADT.h
```diff
@@ -1448,6 +1488,11 @@ class DLL_A_EXPORT InputSymbol final {
...
+// clang-format off
+// nosemgrep: reusable-workflows.semgrep_rules.cxx.funcret.gen.non-void-function-no-return
```
quic-seaswara, line 1494: Do you need this ?

quic-seaswara (changes requested):

include/eld/PluginAPI/PluginADT.h
Steven6798: Yes, I saw that. I have it fixed locally.

include/eld/PluginAPI/PluginADT.h
Steven6798: Yes.

Steven6798:
Workflow was rolled back to a previous version. This fix is no longer needed at the moment.

PR #1889, opened 2026-09-17
[NFC] Dynamic section inputs cleanup

quic-areg:
Dynamic section header inputs are now owned by Module's internal input list instead of keeping a separate ELFObjectFile.


lib/Core/Linker.cpp
```diff
@@ -467,14 +467,17 @@ bool Linker::resolve() {
...
-    // Add dynamic section header inputs to the front of the input list.
-    ThisModule->getObjectList().insert(
-        ThisModule->getObjectList().begin(),
-        Backend->getDynamicSectionHeadersInputFile());
+    // Dynamic section headers must precede real inputs so that GOT/PLT header
+    // fragments retain their position in merged output sections.
+    InputFile *DynamicSectionHeaders = ThisModule->getInternalInput(
+        Module::InternalInputType::DynamicSectionHeaders);
+    ThisModule->getObjectList().insert(ThisModule->getObjectList().begin(),
+                                       DynamicSectionHeaders);
```
quic-seaswara: There is only one GOT/PLT/GOTPLT section, so should be ok even if it as the end.

quic-areg: Done

quic-seaswara (approved):

Steven6798 (approved):

PR #1913, opened 2026-09-19
[ARM] Implement R_ARM_THM_PC12 relocation

deepakshirkem:
Implement the `apply` function for `R_ARM_THM_PC12` using the formula `((S + A) | T) - Pa`, following LLD behavior, where:

* `Pa = P & ~3` — the relocation place aligned down to a 4-byte boundary
* The implicit addend `A` is extracted from the Thumb-2 LDR literal instruction using the `U` bit and `imm12`
* The Thumb bit `T` participates in the relocation calculation before being ignored in the final displacement
* Positive and negative offsets are supported, with the sign encoded in the `U` bit
* The absolute displacement must fit in the 12-bit `imm12` field, giving a range of `[-4095, 4095]`

The relocated value is encoded back into the Thumb-2 instruction by updating the `U` bit in the upper halfword and `imm12` in the lower halfword.


quic-seaswara (approved):

Steven6798:
@deepakshirkem Can you look into the cases in the following LLD test? arm-thumb2-ldrlit.s

Do we have tests for them?

deepakshirkem:
@Steven6798 Added the test. Thank you!

Steven6798 (approved):

